PR #146, opened 2018-08-24
Update erlexec from 1.7.5 to 1.9.3

robinclart:
This means that everything now works on the latest version of erlang
and elixir.


kasima:
Thanks @robinclart! Confirmed this now works with the latest in brew:

- elixir 1.7.2
- erlang 21.0.6

PR #159, opened 2018-09-05
Feature/omg 233 watcher tests clean up

purbanow:


apps/omg_watcher/lib/web/views/transaction.ex
```diff
@@ -22,15 +22,18 @@ defmodule OMG.Watcher.Web.View.Transaction do
...
-      "newowner1",
-      "newowner2",
-      "sig1",
-      "sig2",
-      "spender1",
-      "spender2"
+    transaction
+    |> Map.delete(:__meta__)
+    |> Map.from_struct()
+    |> Serializer.Response.encode16([
```
pnowosie: Since it's transaction's view lets keep these keys in constant list `@keys_to_encode`

apps/omg_watcher/priv/static/swagger.json
```diff
@@ -1,109 +1,43 @@
 {
   "swagger": "2.0",
   "paths": {
-    "/transactions": {
```
pnowosie: Let's ignore this file from git-tracking

PR #161, opened 2018-09-05
chore: update credo and turn off deprecated checks

pgebal:


apps/omg_api/lib/block_queue/core.ex
```diff
@@ -25,6 +25,7 @@ defmodule OMG.API.BlockQueue.Core do
   alias OMG.API.BlockQueue
   alias OMG.API.BlockQueue.Core
   alias OMG.API.BlockQueue.GasPriceAdjustmentStrategyParams, as: GasPriceParams
+  alias OMG.Eth.BlockSubmission
```
purbanow: do you need this alias?

PR #166, opened 2018-09-07
feat: handle indexed fields in root contract events

pgebal:


apps/omg_eth/lib/eth/root_chain.ex
```diff
@@ -236,24 +234,38 @@ defmodule OMG.Eth.RootChain do
...
+
     deposit_blknum
   end
 
-  @spec filter_receipt_events([%{topics: [binary], data: binary()}], binary) :: [tuple]
-  def filter_receipt_events(receipt_logs, signature) do
+  defp decode_deposit(log) do
+    indexed_keys = [:owner, :blknum]
+    indexed_keys_types = [:address, {:uint, 256}]
+    non_indexed_signature = "Deposit(address,uint256)"
```
paulperegud: I think skipping signature and not using parse_event/2 would be a better, less confusing way of dealing with this. I.e. there is no such thing as "Deposit(address,uint256)". Admitting that parse_event/2 is limited and should be rewritten later is probably a more readable solution.

apps/omg_eth/lib/eth/root_chain.ex
```diff
@@ -236,24 +234,38 @@ defmodule OMG.Eth.RootChain do
...
+    non_indexed_keys = [:currency, :amount]
+    non_indexed_key_types = [:address, {:uint, 256}]
+    indexed_keys = [:owner, :utxo_pos]
+    indexed_keys_types = [:address, {:uint, 256}]
+
+    parse_events_with_indexed_fields(log, {non_indexed_keys, non_indexed_key_types}, {indexed_keys, indexed_keys_types})
+  end
+
+  @spec filter_receipt_events([%{topics: [binary], data: binary()}], binary, (map() -> map())) :: [map()]
+  def filter_receipt_events(receipt_logs, signature, decode_log) do
```
pdobacz: I think I'd have a slight preference towards running `Enum.map(&decode_deposit/1)` in the function that calls this, instead of passing a function parameter, to keep simple.

Also I think this can be `defp` now, perhaps?

PR #167, opened 2018-09-10
Feature/quick fix slow sync

pdobacz:
Composed out of various fixes that make the demos feasible again:

fix: proper Utxo decoding in exiting when called from exit validator  …
fix: new default polling intervals for `:dev` environment to fix demo
fix: adjust demo 02 to new Watcher's API
fix: allow depositer to start from last synced eth height + naming

### DIFF
--- a/apps/omg_api/lib/state/core.ex
+++ b/apps/omg_api/lib/state/core.ex
@@ -26,7 +26,7 @@ defmodule OMG.API.State.Core do
 
   @maximum_block_size 65_536
 
-  defstruct [:height, :last_deposit_height, :utxos, pending_txs: [], tx_index: 0]
+  defstruct [:height, :last_deposit_child_blknum, :utxos, pending_txs: [], tx_index: 0]
 
   alias OMG.API.Block
   alias OMG.API.Crypto
@@ -37,7 +37,7 @@ defmodule OMG.API.State.Core do
 
   @type t() :: %__MODULE__{
           height: non_neg_integer(),
-          last_deposit_height: non_neg_integer(),
+          last_deposit_child_blknum: non_neg_integer(),
           utxos: utxos,
           pending_txs: list(Transaction.Recovered.t()),
           tx_index: non_neg_integer()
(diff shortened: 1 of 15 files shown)


apps/omg_api/lib/state/core.ex
```diff
@@ -388,38 +386,38 @@ defmodule OMG.API.State.Core do
...
+      |> Enum.filter(&utxo_exists?(&1, state))
 
     event_triggers =
       exiting_utxos
       |> Enum.map(fn %{owner: owner, utxo_pos: utxo_pos} ->
         %{exit: %{owner: owner, utxo_pos: Utxo.Position.decode(utxo_pos)}}
       end)
 
-    state =
+    new_utxos_in_state =
```
pgebal: That name is misleading. Those are not new utxos. Those are utxos that survived exits.

apps/omg_watcher/lib/exit_validator/validator.ex
```diff
@@ -36,7 +36,6 @@ defmodule OMG.Watcher.ExitValidator.Validator do
   end
 
   defp exists?(utxo_exit) do
-    Utxo.position(blknum, txindex, oindex) = Utxo.Position.decode(utxo_exit.utxo_pos)
-    OMG.API.State.utxo_exists?(%{blknum: blknum, txindex: txindex, oindex: oindex})
+    OMG.API.State.utxo_exists?(utxo_exit)
```
pgebal: Since it's a one liner I would move that State call to line 27.

apps/omg_watcher/lib/application.ex
```diff
@@ -38,12 +38,12 @@ defmodule OMG.Watcher.Application do
...
         [
           %{
-            synced_height_update_key: :last_depositer_block_height,
+            synced_height_update_key: :last_depositer_eth_height,
             service_name: :depositer,
             block_finality_margin: block_finality_margin,
             get_events_callback: &OMG.Eth.RootChain.get_deposits/2,
             process_events_callback: &OMG.API.State.deposit/1,
-            get_last_synced_height_callback: &OMG.Eth.RootChain.get_root_deployment_height/0
+            get_last_synced_height_callback: &OMG.DB.last_depositer_eth_height/0
```
pgebal: To not get sync issues, we need to remember to initialize that to root_deployment height, or take maximum of root deployment height and last depositer eth height.

PR #169, opened 2018-09-10
fix: figure out getter synced eth height based on persisted child blknum

pdobacz:
it used to rely on the persisted getter eth height which could remain un-updated
to quickly fix changed that to take persisted child block number into account


apps/omg_watcher/lib/block_getter/core.ex
```diff
@@ -397,4 +397,25 @@ defmodule OMG.Watcher.BlockGetter.Core do
   defp zero_fee_for(%Transaction.Recovered{signed_tx: %Transaction.Signed{raw_tx: %Transaction{cur12: cur12}}}, fee_map) do
     Map.put(fee_map, cur12, 0)
   end
+
+  @doc """
+  Given a persited `synced_height` and actual child block number State reports to figure out the exact
```
pgebal: Something is wrong with this sentence.

PR #175, opened 2018-09-14
OMG-221 - refactor: make omg_jsonrpc not dependent on omg_api

pdobacz:
co-authored with @pgebal during refactor Friday

closes OMG-221

### DIFF
--- a/apps/omg_jsonrpc/lib/server.ex
+++ b/apps/omg_jsonrpc/lib/server.ex
@@ -14,13 +14,16 @@
 
 defmodule OMG.JSONRPC.Server.Handler do
   @moduledoc """
-  Exposes OMG.API via jsonrpc 2.0 over HTTP. It leverages the generic OMG.JSONRPC.Exposer convenience module
+  Exposes an API via jsonrpc 2.0 over HTTP. It leverages the generic OMG.JSONRPC.Exposer convenience module
 
   Only handles the integration with the JSONRPC2 package
   """
   use JSONRPC2.Server.Handler
 
+  # Running :omg_jsonrpc requires setting this environmental variable
+  @api_module Application.fetch_env!(:omg_jsonrpc, :api_module)
+
   def handle_request(method, params) do
-    OMG.JSONRPC.Exposer.handle_request_on_api(method, params, OMG.API)
+    OMG.JSONRPC.Exposer.handle_request_on_api(method, params, @api_module)
   end
 end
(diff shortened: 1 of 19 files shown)


apps/omg_jsonrpc/lib/server.ex
```diff
@@ -14,13 +14,16 @@
...
   @moduledoc """
-  Exposes OMG.API via jsonrpc 2.0 over HTTP. It leverages the generic OMG.JSONRPC.Exposer convenience module
+  Exposes an API via jsonrpc 2.0 over HTTP. It leverages the generic OMG.JSONRPC.Exposer convenience module
 
   Only handles the integration with the JSONRPC2 package
   """
   use JSONRPC2.Server.Handler
 
+  # Running :omg_jsonrpc requires setting this environmental variable
+  @api_module Application.fetch_env!(:omg_jsonrpc, :api_module)
```
paulperegud: I believe this one is broken. I.e. Application.fetch_env is supposed to work in runtime, not in compile time. This effectively reads value from files that exist on dev machine during compilation (mix.exs or config.exs) and will ignore whatever way we will use to provide configuration to packaged release.

paulperegud: You can test it by running `iex -S mix run --no-start` and changing the value of env variable via :application.set_env

apps/omg_performance/lib/performance.ex
```diff
@@ -158,6 +158,7 @@ defmodule OMG.Performance do
 
     Application.put_env(:omg_eth, :contract_addr, contract_addr)
 
+    # FIXME
```
paulperegud: Somehow this line does not trigger credo, while it should.

apps/omg_jsonrpc/lib/server.ex
pdobacz: actually, we gave it a moment of thought and realised, we want it to be a compile-time constant in fact. Does this make it easier to live with?

apps/omg_performance/lib/performance.ex
pdobacz: apparently `# FIXME a` works, I'll file an issue to `credo`

apps/omg_jsonrpc/lib/server.ex
paulperegud: Yeah, this works fine. Commenting that such usage is intended would be good to make reading simpler.

PR #177, opened 2018-09-17
OMG-234 Wallet provider quickly synces up with long child chain of almost empty blocks

pgebal:


paulperegud:
@pgebal I will need some help with testing it. Code so far looks fine.

apps/omg_api/lib/root_chain_coordinator.ex
```diff
@@ -53,17 +55,18 @@ defmodule OMG.API.RootchainCoordinator do
   end
 
   def handle_call({:check_in, synced_height, service_name}, {pid, _}, state) do
-    {:ok, state} = Core.check_in(state, pid, synced_height, service_name)
+    {:ok, state, services_to_sync} = Core.check_in(state, pid, synced_height, service_name)
+    Enum.each(services_to_sync, fn pid -> send(pid, :sync) end)
```
paulperegud: It seems that for every ethereum height, service can get from 1 to N-1 :sync messages, where N is number of services.

apps/omg_api/lib/root_chain_coordinator/core.ex
```diff
@@ -11,46 +11,60 @@
...
-  def init(allowed_services, rootchain_height) do
-    %__MODULE__{allowed_services: allowed_services, rootchain_height: rootchain_height}
+  def init(allowed_services, root_chain_height) do
+    %__MODULE__{allowed_services: allowed_services, root_chain_height: root_chain_height}
   end
 
   @doc """
   Updates Ethereum height on which a service is synchronized.
+  Returns list of pids of services to synchronize on next Ethereum height.
+  List is not empty only when service that checks in is the last synchronizing on a given hight.
```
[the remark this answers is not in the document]
pgebal: Agree on typo. The second part: the comment is right. That list is not empty only for the last service. Let's talk about that in person.

apps/omg_api/lib/root_chain_coordinator.ex
pgebal: That's not the case.

apps/omg_api/lib/root_chain_coordinator/service.ex
```diff
@@ -11,10 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-defmodule OMG.API.RootchainCoordinator.Service do
+defmodule OMG.API.RootChainCoordinator.Service do
   @moduledoc """
-  Represents a service that is coordinated by rootchain coordinator.
-  Such a service is expected to get rootchain height by calling `RootchainCoordinator.get_height()` function
+  Represents a service that is coordinated by root chain coordinator.
+  Such a service is expected to get root chain height by calling `RootChainCoordinator.get_height()` function
```
paulperegud: Mentioning that service will get `:sync` message is worthwhile.

PR #179, opened 2018-09-18
Add a test case covering lack of outputs and value set to zero

paulperegud:
This is a behavior emergent from our implementation. I'm not sure if it is a good idea to cover it with a unit test. Maybe we should document it instead.

### DIFF
--- a/apps/omg_api/test/state/core_test.exs
+++ b/apps/omg_api/test/state/core_test.exs
@@ -629,6 +629,46 @@ defmodule OMG.API.State.CoreTest do
     end
   end
 
+  @tag fixtures: [:alice, :state_empty]
+  test "Output can have a zero value; can't be used as input though", %{alice: alice, state_empty: state} do
+    fee = %{eth() => 2}
+
+    state
+    |> Test.do_deposit(alice, %{amount: 10, currency: eth(), blknum: 1})
+    |> (&Core.exec(Test.create_recovered([{1, 0, 0, alice}], eth(), [{alice, 8}, {alice, 0}]), fee, &1)).()
+    |> success?
+    |> (&Core.exec(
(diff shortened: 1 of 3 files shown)


paulperegud:
After talk with @pdobacz we've decided that this behavior is OK and it should be covered with tests, so we know how system behaves.

apps/omg_api/test/state/core_test.exs
```diff
@@ -629,6 +629,46 @@ defmodule OMG.API.State.CoreTest do
...
+  @tag fixtures: [:alice, :state_empty]
+  test "Output can have a zero value; can't be used as input though", %{alice: alice, state_empty: state} do
+    fee = %{eth() => 2}
+
+    state
+    |> Test.do_deposit(alice, %{amount: 10, currency: eth(), blknum: 1})
+    |> (&Core.exec(Test.create_recovered([{1, 0, 0, alice}], eth(), [{alice, 8}, {alice, 0}]), fee, &1)).()
+    |> success?
+    |> (&Core.exec(
+          Test.create_recovered([{2, 0, 0, alice}, {2, 1, 0, alice}], eth(), [{alice, 5}, {alice, 1}]),
```
pnowosie: I think it should be `{2, 0, 1 ...` not `{2, 1, 0, ...}` 
...and including existing utxo `{2, 0, 0}` make it harder to reason about - please use just zero-output to show the behavior.

paulperegud: Good catch. Fixed and simplified.

paulperegud:
While adding one more test I've discovered that I've made a mistake in specifying utxo_pos of zero value input. It's `{1000, 0, 1}`, not `{2, 0, 1}` :man_facepalming: 

Rechecked everything. Merging.

PR #180, opened 2018-09-18
OMG-232 fix: avoid nonce too low problems, where there aren't ones really

pdobacz:
now the nonce too low (and any other) response from geth will be processed by BlockQueue.Core. True problems with nonce too low, i.e. when it wasn't the submission that got mined are detected and make the BlockQueue crash

Also added some tests and fixes to allow BlockQueue to restart in some circumstances


pdobacz:
leaving fix&merge up to the reviewer, merge after https://github.com/omisego/elixir-omg/pull/181 is merged as this branch/PR has 181 merged in and includes resolving conflicts around that

PR #182, opened 2018-09-20
chore: use ex_doc

pgebal:


apps/omg_jsonrpc/lib/jsonrpc.ex
```diff
@@ -14,7 +14,7 @@
 
 defmodule OMG.JSONRPC do
   @moduledoc """
-  A JSONRPC 2.0 gateway to an API - automatically exposed via `ExposeSpec`
+  A JSONRPC 2.0 gateway to an API - automatically exposed via `OMG.JSONRPC.ExposeSpec`
   ## Usage
```
pnowosie: This `h2` header looks strange in the modules list... maybe we can show it only in module's page?

pgebal: done

PR #190, opened 2018-09-27
Feature/omg 241 watcherdb aware of deposits only

### LINKED ISSUE #157: Watcher not showing available UTXOs for deposits to Plasma contract
After following the steps within `demo_01` to depositing ETH on Plasma, UTXOs are not showing as available upon queries on the watcher until the initial transaction is made on the Plasma chain

pnowosie:
Fixed issue that deposited funds to child chain were not available as utxo set.

Closes #157

### DIFF
--- a/apps/omg_watcher/lib/application.ex
+++ b/apps/omg_watcher/lib/application.ex
@@ -112,7 +112,7 @@ defmodule OMG.Watcher.Application do
 
   defp deposit_events_callback(deposits) do
     :ok = OMG.API.State.deposit(deposits)
-    OMG.Watcher.DB.EthEventDB.insert_deposits(deposits)
+    _ = OMG.Watcher.DB.EthEventDB.insert_deposits(deposits)
     :ok
   end
 end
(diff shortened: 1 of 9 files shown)


apps/omg_watcher/test/integration/watcher_api_test.exs
```diff
@@ -0,0 +1,97 @@
...
+      "amount" => 10,
+      "blknum" => token_deposit_blknum,
+      "txindex" => 0,
+      "oindex" => 0,
+      "currency" => token_addr,
+      "txbytes" => nil
+    }
+
+    # utxo from deposit should be available
+    assert [eth_deposit, token_deposit] == IntegrationTest.get_utxos(alice)
```
pgebal: We can skip code in the following lines.

apps/omg_watcher/lib/application.ex
```diff
@@ -109,4 +109,10 @@ defmodule OMG.Watcher.Application do
         :child_chain_exit
     end
   end
+
+  defp deposit_events_callback(deposits) do
+    :ok = OMG.API.State.deposit(deposits)
+    OMG.Watcher.DB.EthEventDB.insert_deposits(deposits)
```
pgebal: Why don't we match that with :ok?

pnowosie: I've added `:ok` match on each particular insert - so it's either all pass or fail on first error

PR #196, opened 2018-10-01
refactor: rename variables and functions in block getter module

pgebal:


apps/omg_watcher/lib/block_getter/core.ex
```diff
@@ -34,27 +34,27 @@ defmodule OMG.Watcher.BlockGetter.Core do
...
   ]
 
   @type t() :: %__MODULE__{
-          synced_height_update_blocks: MapSet.t(),
+          height_sync_blknums: MapSet.t(),
           synced_height: pos_integer(),
-          last_consumed_block: non_neg_integer,
-          started_block_number: non_neg_integer,
+          last_applied_block: non_neg_integer,
+          start_download_blknum: non_neg_integer,
```
paulperegud: A comment would be useful here: // highest blknum that is being downloaded

Issue #188, opened 2018-09-25
Watcher returns 500 Internal Server Error from /utxos if given an invalid address


kevsul:
If I pass an invalid address format to the watcher e.g. `http://localhost:4000/utxos?address=0` the error response is `500 Internal Server Error`.

I would have expected `400 Bad Request`

I went to check if swagger validated the address param, but I can't find the swagger.json file. The omg_watcher/config.exs has 
```  
  swagger_files: %{
    "priv/static/swagger.json" => [router: OMG.Watcher.Web.Router]
  }
```
that file doesn't seem to exist in the repo, but calling `http://localhost:4000/api/swagger/swagger.json` _does_ return the swagger doc, so I'm confused 😕

pnowosie:
Hi @kevsul thank for reporting response mismatch - we will add better input validation.

Regarding swagger - swagger.json is autogenerated during build and therefore not stored in repo.

PR #200, opened 2018-10-04
Feature/omg 259 wallet provider user provided input validation

purbanow:
Example:
http://localhost:4000/utxos?address=0
respond:
400
{"result":"error","data":{"description":"Provided address is invalid.","code":"address:bad_address_encoding"}}

Closes: #188


pnowosie:
Since a lot API changes ahead we will propose another approach therefore.

PR #201, opened 2018-10-08
Docs: signedTxBytes is not a part of RLP encoded signed transaction

paulperegud:
Fixing error in docs: signedTxBytes is not a part of RLP encoded signed transaction - it can always be recreated, if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ A developer instance of `geth` runs Ethereum locally and prefunds an account.
 However, when `geth` terminates, the state of the Ethereum network is lost.
 
 ```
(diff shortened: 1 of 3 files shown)


pdobacz:
@paulperegud @Pongch I think a larger portion of the document can be dropped: `rlp_encoded` and `signedTxBytes` are the same thing now, I think, also point `2.4` seems kinda redundant.

Pongch:
@pdobacz point `2.4` is redundant for the sake of being consistent with other points. ie. each point explicitly show the input and output of the function-- so as to be easier to follow for the reader

PR #208, opened 2018-10-12
fix(watcher): do not download more blocks if last applied block is lagging

pgebal:


apps/omg_watcher/lib/block_getter/core.ex
```diff
@@ -215,7 +222,7 @@ defmodule OMG.Watcher.BlockGetter.Core do
     potential_next_block_numbers =
       first_block_number
       |> Stream.iterate(&(&1 + block_interval))
-      |> Stream.take_while(&(&1 < next_child))
+      |> Stream.take_while(&(&1 < next_child and &1 <= last_applied_block + maximum_last_applied_block_lag))
```
pthomalla: instead of this condition condition in Enum.take

apps/omg_watcher/lib/block_getter/core.ex
```diff
@@ -33,17 +35,32 @@ defmodule OMG.Watcher.BlockGetter.Core do
           }
   end
 
+  defmodule Config do
```
pthomalla: block_interval fits Config

PR #210, opened 2018-10-15
Fix/omg 277 15sec block perf testing

pdobacz:


pdobacz:
Thanks to @rodiazet for contributing part of this PR

apps/omg_api/test/support/integration/deposit_helper.ex
```diff
@@ -25,10 +25,14 @@ defmodule OMG.API.Integration.DepositHelper do
   def deposit_to_child_chain(to, value, token \\ @eth)
 
   def deposit_to_child_chain(to, value, @eth) do
-    {:ok, receipt} = Eth.RootChain.deposit(value, to) |> Eth.DevHelpers.transact_sync!()
+    {:ok, %{"blockNumber" => deposit_eth_height} = receipt} =
```
pik694: Shouldn't we refactor this function and the latter one (version for ERC20 token)? Besides the preparation in the latter one, they are basically the same. They just call different Root Chain functions, but checks are the same.

apps/omg_api/test/integration/happy_path_test.exs
```diff
@@ -96,4 +96,14 @@ defmodule OMG.API.Integration.HappyPathTest do
 
     assert {:error, {_, "Internal error", "utxo_not_found"}} = Client.call(:submit, %{transaction: token_tx})
   end
+
+  defp wait_for_current_child_block(blknum, timeout \\ 10_000, contract \\ nil) do
```
pik694: this looks quite similar to `wait_for_root_chain_block` from `DepositHelper`, shouldn't they be in the same place ? and

PR #212, opened 2018-10-16
Add dockerfile

jbunce:
Add Dockerfile to this branch

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,53 @@
+FROM ubuntu:16.04
+
+MAINTAINER Jake Bunce <jake@omise.co>
+
+ARG USER=plasma
+ARG GROUP=plasma
+ARG UID=1000
+ARG GID=1000
+ARG HOME=/home/jbunce
+RUN groupadd --gid     && useradd --uid 501 --gid  --shell /bin/bash --create-home --home /Users/jbunce jbunce
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+
+RUN groupadd --gid "${GID}" "${USER}" && \
+    useradd \
+      --uid ${UID} \
+      --gid ${GID} \
+      --create-home \
+      --shell /bin/bash \
+      ${USER}
+
+ARG BUILD_PACKAGES="build-essential autoconf libtool libgmp3-dev libssl-dev wget gettext"
```
paulperegud: libgmp3-dev -> libgmp-dev. We need to update the readme.

Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+
+USER plasma
+
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8 
+
+WORKDIR /home/plasma/elixir-omg/
+
+RUN sudo -H pip3 install --upgrade pip \
+  && sudo -H -n ln -s /usr/bin/python3 python \
```
paulperegud: This one is a bit strange. Maybe update-alternatives instead?

Dockerfile
jbunce: ack

Dockerfile
jbunce: I'll look into that

PR #213, opened 2018-10-16
Change DevHelpers passphrase

paulperegud:
Geth does not allow empty passphrase on the testnet. Change it to non-empty one.

Temporary change. Do not merge into dev / master! Normally, use your own script to deploy contract and configure chain on the testnet.

@jbunce


jbunce:
LGTM. Let's chat about the deployment process because devs won't be able to execute their own scripts. This works for now.

PR #218, opened 2018-10-18
fix: accelarate start OMG.API.State

pthomalla:


apps/omg_db/lib/db.ex
```diff
@@ -36,7 +36,11 @@ defmodule OMG.DB do
   end
 
   def utxos(server_name \\ @server_name) do
-    GenServer.call(server_name, {:utxos})
+    # in this time database shout read 3 millions utxos
```
pdobacz: - `should`
 - `max_time_of_read_ms`

apps/omg_db/lib/leveldb_server.ex
```diff
@@ -64,12 +64,11 @@ defmodule OMG.DB.LevelDBServer do
   end
 
   def handle_call({:utxos}, _from, %__MODULE__{db_ref: db_ref} = state) do
-    keys_stream = Exleveldb.stream(db_ref, :keys_only)
-
     result =
-      keys_stream
+      db_ref
+      |> Exleveldb.stream()
```
pdobacz: but this looks like is going through all the values of all the blocks in the database.  I'm afraid this will end up much slower, once blocks accumulate (or might even make the child chain non-restartable - mind that the Watcher isn't writing blocks here (yet), so this is why you got your speed up - the restart of the child chain might be worse off now).

I think we cannot do this like that. Either there is another (simple) way, or we need to double the timeout instead (logging some `info` about what's happening and that it might take a while is necessary then)

pthomalla: I test it on child chain and there is speed up on Start.
I'm afraid that ```Exleveldb.stream(rd_ref, :keys_only)```  reds also value from DB but then cut it out.

pdobacz: OK @pthomalla I have finally been able to check this. It seems to not impede child chain server start, I can see the same speedup like you do.

Let's go forward with merging this, but please, keep the large timeout and the `this might take a while` `Logger.info` message that you'll find in the `release` branch.

pdobacz: OK @pthomalla I have finally been able to check this. It seems to not impede child chain server start, I can see the same speedup like you do.

Let's go forward with merging this, but please, keep the large timeout and the `this might take a while` `Logger.info` message that you'll find in the `release` branch.

pdobacz: OK @pthomalla I have finally been able to check this. It seems to not impede child chain server start, I can see the same speedup like you do.

Let's go forward with merging this, but please, keep the large timeout and the `this might take a while` `Logger.info` message that you'll find in the `release` branch.

pdobacz: OK @pthomalla I have finally been able to double check this. You're right, it seems to not impede child chain server start on my setup, I can see the same speedup like you do.

Let's go forward with merging this, but please, keep the large timeout and the `this might take a while` `Logger.info` message that you'll find in the `release` branch.

PR #219, opened 2018-10-18
Update contracts

paulperegud:
Switch to new, updated version of contracts.

Changes relevant to elixir-omg code are:
* confirmation signatures are turned off properly (without breaking other security guarantees)
* changed `depositFrom` API
* lots of bug fixes


paulperegud:
Merged manually because GH is falling apart.

PR #222, opened 2018-10-19
OMG-285 - wallet provider - gets N most recent transactions, optionally filtering by recipient

purbanow:


kevsul:
This is probably outside the scope of this PD, but why is the filter by recipient? 
I mean, for a given address, when is it useful only to get the transactions where that address is the recipient? 
I think getting both sent and received transactions makes more sense (e.g. a wallet or block explorer would want to show both types of transaction)

jarindr:
> This is probably outside the scope of this PD, but why is the filter by recipient?
> I mean, for a given address, when is it useful only to get the transactions where that address is the recipient?
> I think getting both sent and received transactions makes more sense (e.g. a wallet or block explorer would want to show both types of transaction)

actually it should be both send and received, maybe miscommunicated.

kasima:
Seems like a miscommunication here. Let's change to filter by `address` and return both sent and received.

purbanow:
Moved to #228 PR

PR #226, opened 2018-10-19
make watcher db resistant to replaying blocks

pnowosie:


apps/omg_watcher/lib/block_getter.ex
```diff
@@ -31,9 +31,9 @@ defmodule OMG.Watcher.BlockGetter do
   @spec download_block(pos_integer()) ::
           {:ok, Block.t() | Core.PotentialWithholding.t()} | {:error, Core.block_error(), binary(), pos_integer()}
   def download_block(requested_number) do
-    {:ok, {requested_hash, _time}} = Eth.RootChain.get_child_chain(requested_number)
+    {:ok, {requested_hash, timestamp}} = Eth.RootChain.get_child_chain(requested_number)
     response = OMG.JSONRPC.Client.call(:get_block, %{hash: requested_hash})
-    Core.validate_download_response(response, requested_hash, requested_number, :os.system_time(:millisecond))
+    Core.validate_download_response(response, requested_hash, requested_number, timestamp)
```
pdobacz: whoa, that `system_time` was there for a reason - detecting block wihholding (see other clauses of the function). If docs/comments are insufficient, please add some more, so that no one else misinterprets that

PR #227, opened 2018-10-19
Prevent event listeners from processing some events twice on restart

paulperegud:
By writing to db `current_ethereum_height - maturity_block` and reading value from db as if it was `current_ethereum_height` listener sets itself up for processing of the same events twice. This patch fixes it.

### DIFF
--- a/apps/omg_api/lib/ethereum_event_listener/core.ex
+++ b/apps/omg_api/lib/ethereum_event_listener/core.ex
@@ -61,7 +61,7 @@ defmodule OMG.API.EthereumEventListener.Core do
         next_event_height_lower_bound: next_event_height_upper_bound + 1
     }
 
-    db_updates = [{:put, update_key, next_event_height_upper_bound}]
+    db_updates = [{:put, update_key, next_sync_height}]
 
     {:get_events, {next_event_height_lower_bound, next_event_height_upper_bound}, new_state, db_updates}
   end
--- a/apps/omg_api/lib/ethereum_event_listener/core.ex
+++ b/apps/omg_api/lib/ethereum_event_listener/core.ex
@@ -33,6 +33,16 @@ defmodule OMG.API.EthereumEventListener.Core do
           block_finality_margin: non_neg_integer()
         }
 
+  def init(update_key, service_name, last_synced_ethereum_height, block_finality_margin) do
+    %__MODULE__{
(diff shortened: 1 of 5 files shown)


apps/omg_api/lib/ethereum_event_listener/core.ex
```diff
@@ -61,7 +61,7 @@ defmodule OMG.API.EthereumEventListener.Core do
         next_event_height_lower_bound: next_event_height_upper_bound + 1
     }
 
-    db_updates = [{:put, update_key, next_event_height_upper_bound}]
+    db_updates = [{:put, update_key, next_sync_height}]
```
pgebal: When we set last height to next_sync_height, update it in db and process crashes, then after restart we can get
next_event_height_lower_bound = next_sync_height. Then next_event_height_upper_bound < next_event_height_lower_bound. What happens then? Even if that works and Exthereumex does not crash, in another round we process the same events twice.

paulperegud: Can EEL at the moment of writing to DB have events processed at the height < ethereum_height - finality_margin?

apps/omg_api/lib/ethereum_event_listener/core.ex
```diff
@@ -33,6 +33,16 @@ defmodule OMG.API.EthereumEventListener.Core do
           block_finality_margin: non_neg_integer()
         }
 
+  def init(update_key, service_name, last_synced_ethereum_height, block_finality_margin) do
+    %__MODULE__{
+      synced_height_update_key: update_key,
+      next_event_height_lower_bound: max(last_synced_ethereum_height - block_finality_margin + 1, 0),
+      synced_height: last_synced_ethereum_height,
+      service_name: service_name,
```
pthomalla: service_name is unused in Core,
in event listener it is incorrect to reach the implementation core state

apps/omg_api/lib/ethereum_event_listener/core.ex
paulperegud: Fixed, added test emulating service restart

apps/omg_api/lib/ethereum_event_listener/core.ex
paulperegud: This is a fix branch. Will add TODO.

PR #228, opened 2018-10-22
Feature/omg 285 wallet provider gets n most recent transactions, optionally filtering by address

purbanow:


apps/omg_watcher/lib/web/controllers/transaction.ex
```diff
@@ -36,6 +39,24 @@ defmodule OMG.Watcher.Web.Controller.Transaction do
     |> respond(conn)
   end
 
+  @doc """
+  Retrieves a list of transactions
+  """
+  def get_transactions(conn, params) do
+    address = Map.get(params, "address", nil)
```
pnowosie:
```suggestion
    address = Map.get(params, "address")
```

apps/omg_watcher/lib/web/views/transaction.ex
```diff
@@ -23,6 +23,23 @@ defmodule OMG.Watcher.Web.View.Transaction do
...
+    |> Enum.map(&serialize(&1))
+    |> Serializer.Response.serialize(:success)
+  end
+
+  def render("transaction_encode.json", %{transaction: transaction}) do
+    OMG.API.State.Transaction.encode(transaction)
+    |> Serializer.Response.serialize(:success)
+  end
+
+  defp serialize(transaction) do
```
pnowosie: `serialize` for the name seems improper here... and `render` is by convention taken... hmm
`render_transaction` :question:

apps/omg_watcher/test/web/controllers/transaction_test.exs
```diff
@@ -111,6 +111,42 @@ defmodule OMG.Watcher.Web.Controller.TransactionTest do
...
+
+      %{"data" => txs, "result" => "success"} =
+        TestHelper.rest_call(:get, "/transactions?address=#{address_encoded}&limit=#{limit}")
+
+      assert [
+               %{"txblknum" => 3000, "txindex" => 1},
+               %{"txblknum" => 3000, "txindex" => 0}
+             ] == Enum.map(txs, fn tx -> %{"txblknum" => tx["txblknum"], "txindex" => tx["txindex"]} end)
+
+      assert [true, true] == for(tx <- txs, do: tx["newowner1"] == address or tx["newowner2"] == address)
```
pnowosie: Remove. Since we exactly know add the appropriate `newowner? => ^address` in above assertion

pnowosie: Please add more test
- [ ] shows tx that contain requested address as the `sender` and not `recipient`
- [ ] tx contains requested address as both sender & recipient is listed once
- [ ] tx without inputs filtered by recipient
- [ ] tx without outputs (`amount = 0`) filtered by sender
- [ ] queries on empty DB (both with/out limit, address) returns empty array
Please try to be explicit in assertions, e.g. expected is recipient 1 to have specified address

apps/omg_watcher/lib/web/views/transaction.ex
purbanow: I think `serialize ` is a proper name because the function is responsible for preparing the `transaction` object like `Serializer.Response.serialize`is preparing.

apps/omg_watcher/test/web/controllers/transaction_test.exs
```diff
@@ -111,6 +113,174 @@ defmodule OMG.Watcher.Web.Controller.TransactionTest do
...
 
+  #  TODO Remove below skip tags after `OMG-245 - Run tests on (real) postgres database ` .
+  #  There are problems between Ecto and SQlite
+  describe "Controller.TransactionTest - transactions" do
+    @tag :skip
+    @tag fixtures: [:alice, :bob, :phoenix_ecto_sandbox]
+    test "endpoint returns tx that contain requested address as the sender and not recipient", %{
+      alice: alice,
+      bob: bob,
+      phoenix_ecto_sandbox: phoenix_ecto_sandbox
```
pnowosie: This `phoenix_ecto_sandbox: phoenix_ecto_sandbox` isn't used - pls remove from all tests

PR #229, opened 2018-10-22
Enable recovery after coordinator crash

paulperegud:


apps/omg_watcher/lib/application.ex
```diff
@@ -30,7 +30,8 @@ defmodule OMG.Watcher.Application do
       # Start workers
       {OMG.API.State, []},
       {OMG.Watcher.Eventer, []},
-      {OMG.API.RootChainCoordinator, MapSet.new([:depositer, :fast_validator, :slow_validator, :block_getter])},
+      {OMG.API.RootChainCoordinator,
+       MapSet.new([:depositer, :fast_validator, :slow_validator, OMG.Watcher.BlockGetter])},
```
pgebal: Why do we treat BlockGetter differently?

paulperegud: All of those are atoms, as they should be. It is just in that in case of BlockGetter, there is a 1-1 correspondence between module and named service. In case of EthereumEventListeners - it is three of them per one module. So the name of the module cant be reused.

pdobacz: a side note: I'm in favor of `OMG.Watcher.BlockGetter` and also thinking of taking a similar route with the `:`-styled ones (to have them have their modules). I see it as a chore to cleanup the `watcher/application.ex` module in a foreseeable future.

PR #230, opened 2018-10-22
Feature/omg 279 timeout of state exec in block application

pik694:
Fixes bug that resulted in `BlockGetter` and `State` inconsistent state whenever `BlockGetter` failed and stoped abnormally.


apps/omg_watcher/lib/block_getter/o_m_g/watcher/block_getter/supervisor.ex
```diff
@@ -0,0 +1,42 @@
+# Copyright 2018 OmiseGO Pte Ltd
```
pdobacz: sth's wrong with the path, should be `lib/block_getter/supervisor.ex` (`OMG.Watcher` part is "consumed" by `apps/omg_watcher` according to our conventions

pik694: sorry, it it caused by the fact that I did create this file via module creator and did not realize that it was created in a strange directory

PR #235, opened 2018-10-24
feature: add timestamp and eth_height to transaction endpoint

pik694:
Transaction endpoint now returns data from block the transaction is included in. The additional info is `eth_height` and `timestamp`

### DIFF
--- a/apps/omg_watcher/lib/db/transaction.ex
+++ b/apps/omg_watcher/lib/db/transaction.ex
@@ -49,16 +49,10 @@ defmodule OMG.Watcher.DB.Transaction do
     belongs_to(:block, DB.Block, foreign_key: :blknum, references: :blknum, type: :integer)
   end
 
-  def get(hash, preload_block \\ false) do
-    transaction =
-      __MODULE__
-      |> Repo.get(hash)
-
-    if preload_block do
-      transaction |> Repo.preload(:block)
-    else
-      transaction
-    end
+  def get(hash) do
+    __MODULE__
+    |> Repo.get(hash)
+    |> Repo.preload(:block)
   end
 
   def get_last(limit) do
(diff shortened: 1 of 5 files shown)


apps/omg_watcher/lib/db/transaction.ex
```diff
@@ -49,9 +49,16 @@ defmodule OMG.Watcher.DB.Transaction do
...
 
-  def get(hash) do
-    __MODULE__
-    |> Repo.get(hash)
+  def get(hash, preload_block \\ false) do
+    transaction =
+      __MODULE__
+      |> Repo.get(hash)
+
+    if preload_block do
```
pnowosie:
```suggestion
    REMOVE: ~~if preload_block do
```

apps/omg_watcher/lib/db/transaction.ex
```diff
@@ -49,9 +49,16 @@ defmodule OMG.Watcher.DB.Transaction do
     belongs_to(:block, DB.Block, foreign_key: :blknum, references: :blknum, type: :integer)
   end
 
-  def get(hash) do
-    __MODULE__
-    |> Repo.get(hash)
+  def get(hash, preload_block \\ false) do
+    transaction =
+      __MODULE__
+      |> Repo.get(hash)
```
pnowosie: Let's load block with tx - it's not much
```suggestion
      |> Repo.get(hash)
      |> Repo.preload(:block)
```

apps/omg_watcher/lib/db/transaction.ex
```diff
@@ -78,6 +86,7 @@ defmodule OMG.Watcher.DB.Transaction do
       )
 
     Repo.all(query)
```
pnowosie: Let's preload with query
```suggestion
    Repo.all(query, preload: :block)
```

pnowosie:
Can you please fix [this](https://github.com/omisego/elixir-omg/blob/c2d18478a3be9bdb88ff58a89d996a15558ecf72/apps/omg_watcher/lib/web/controllers/transaction.ex#L203)
id(:path, ~~:integer~~string, "Id of the transaction", required: true)

PR #236, opened 2018-10-24
fix: improve LevelDB init - more verbose and fool proof

pdobacz:
- LevelDBServer.init won't ever create a leveldb instance
- a database-not-empty error isn't silenced out
- one shouldn't use `mix do` in elixir 1.6 for databases reset (updated README)

Thanks to @kevsul for reporting and tracking

### DIFF
--- a/apps/omg_db/lib/db.ex
+++ b/apps/omg_db/lib/db.ex
@@ -80,29 +80,30 @@ defmodule OMG.DB do
     GenServer.call(server_name, :last_exiter_eth_height)
   end
 
-  def init do
+  def init(server_name \\ @server_name) do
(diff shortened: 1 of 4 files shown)


apps/omg_db/lib/db.ex
```diff
@@ -80,29 +80,30 @@ defmodule OMG.DB do
     GenServer.call(server_name, :last_exiter_eth_height)
   end
 
-  def init do
+  def init(server_name \\ @server_name) do
```
pik694: Not rather `server_module`?

pdobacz: it's both by coincidence, so decided to be consistent (in other functions it is the server's name that makes more sense, actually)

PR #238, opened 2018-10-24
fix: turn on transactions endpoint tests

pgebal:


apps/omg_watcher/lib/db/transaction.ex
```diff
@@ -69,15 +69,17 @@ defmodule OMG.Watcher.DB.Transaction do
     query =
       from(
         tx in __MODULE__,
-        distinct: tx.txhash,
         left_join: output in assoc(tx, :outputs),
         left_join: input in assoc(tx, :inputs),
         where: output.owner == ^address or input.owner == ^address,
-        order_by: [desc: tx.blknum, desc: tx.txindex],
-        limit: ^limit
+        order_by: [desc: tx.txindex]
```
pdobacz: is the removal of `desc: tx.blknum` here intentional?

apps/omg_watcher/lib/db/transaction.ex
```diff
@@ -69,15 +69,17 @@ defmodule OMG.Watcher.DB.Transaction do
...
       )
 
-    Repo.all(query)
+    # TODO: use DISTINCT and LIMIT instead, sqlite_ecto does not support DISTINCT on multiple columns
+    query
+    |> Repo.all()
+    |> Enum.dedup_by(fn %{txhash: txhash} -> txhash end)
```
pdobacz: 1/ `Stream.` instead of `Enum.` to leverage the next `take` and lazily load the result of the query perhaps (not sure)?
2/ `.uniq_by` not `.dedup_by`, the latter only removes adjacent duplicates [see](https://hexdocs.pm/elixir/Enum.html#dedup_by/2)

apps/omg_watcher/lib/db/transaction.ex
pgebal: no, that's a bug

apps/omg_watcher/lib/db/transaction.ex
pgebal: Don't know about the 1/ but we want dedup_by, not uniq_by. We don't want to filter out transactions that are not uniq.  Transaction that has given address in both input and output would be filtered out when we use uniq_by.

pnowosie: To get lazy load we need [Repo.stream(queryable, opts)](https://hexdocs.pm/ecto/Ecto.Repo.html#c:stream/2) instead encapsulated in DB transaction (as docs stated)

However, I opted for not to merge yet

PR #252, opened 2018-10-28
Fix/stabilize getter state

pdobacz:
see each of the respective commits for a more comprehensible view. Possibly we can revert one of these commits if unsure


apps/omg_watcher/lib/block_getter/core.ex
```diff
@@ -558,6 +566,13 @@ defmodule OMG.Watcher.BlockGetter.Core do
         end
 
       nil ->
+        _ =
+          Logger.warn(fn ->
```
pgebal: I don't think that should be a warning. I think we can have 3 ethereum blocks without any plasma block published on them.

pdobacz: this isn't the case. We're looking up a particular plasma block number in submissions - `nil` case is when you know a particular plasma block, but you can't find it's exact submission (because we're limiting ourselves to 2000 eth blocks surrounding the persisted `synced_height`). In such case, everything might work fine (if `synced_height` is correct), but might not, and then the warning can lead one to where the problem is.

apps/omg_watcher/test/block_getter/core_test.exs
```diff
@@ -366,29 +368,217 @@ defmodule OMG.Watcher.BlockGetter.CoreTest do
...
+             Core.get_blocks_to_apply(state, [], next_synced_height + 1)
+  end
+
+  test "long running applying block scenario" do
+    # this test replicates a long running scenario, with various inputs from the root chain coordinator
+    # We're testing if we're applying blocks and height updates correctly
+
+    # child block submissions on the root chain, by eth_height
+    submissions = %{
+      57 => [%{blknum: 0000, eth_height: 57}],
```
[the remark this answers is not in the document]
pdobacz: LOL, probably 0_000 is more in line with coding style. Thx :)

PR #256, opened 2018-10-30
fix: allow margin in reading submissions on BlockGetter start

pdobacz:
a subset of #252 which we decided to include in the release

### DIFF
--- a/apps/omg_watcher/test/block_getter/core_test.exs
+++ b/apps/omg_watcher/test/block_getter/core_test.exs
@@ -317,6 +317,7 @@ defmodule OMG.Watcher.BlockGetter.CoreTest do
     |> assert_check([1_000])
   end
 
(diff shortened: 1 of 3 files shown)


apps/omg_watcher/test/block_getter/core_test.exs
```diff
@@ -317,6 +317,7 @@ defmodule OMG.Watcher.BlockGetter.CoreTest do
     |> assert_check([1_000])
   end
 
+  @tag :capture_log
```
paulperegud: is this a leftover?

pdobacz: no, intended, one case issues a `warn` log

paulperegud: ok

PR #257, opened 2018-10-30
OMG-262 - Improve Watcher integration test - waiting on Eth events

purbanow:


apps/omg_watcher/test/integration/watcher_api_test.exs
```diff
@@ -130,11 +128,9 @@ defmodule OMG.Watcher.Integration.WatcherApiTest do
         alice.addr
       )
 
-    {:ok, %{"status" => "0x1"}} = Eth.WaitFor.eth_receipt(txhash2, @timeout)
+    IntegrationTest.wait_until_block_getter_fetches_block_after_current_child_block(@timeout)
 
-    # exiting spends UTXO on child chain
-    # wait until the exit is recognized and attempt to spend the exited utxo
-    Process.sleep(4_000)
+    {:ok, %{"status" => "0x1"}} = Eth.WaitFor.eth_receipt(txhash2, @timeout)
```
pgebal: I think order of those lines should be switched.

apps/omg_watcher/test/support/integration/test_helper.ex
```diff
@@ -41,6 +41,14 @@ defmodule OMG.Watcher.Integration.TestHelper do
     utxos
   end
 
+  def wait_until_block_getter_fetches_block_after_current_child_block(timeout) do
```
pgebal: Make this name less specific. In some uses of this function we wait for exitor.

purbanow: I changed methods to:
`wait_for_current_block_fetch` - this function is more general so I didn't change for more specific name
`wait_for_block_fetch`

PR #259, opened 2018-11-04
chore: remove usages of `Supervisor.Spec`

pik694:
As of `Elixir 1.5` module `Supervisor.Spec` has been marked as deprecated, so the changes in this PR remove usages of `Supervisor.Spec` and use the new notation instead.

### DIFF
--- a/apps/omg_watcher/lib/application.ex
+++ b/apps/omg_watcher/lib/application.ex
@@ -57,7 +57,7 @@ defmodule OMG.Watcher.Application do
     children = [
       # Start the Ecto repository
       %{
-        id: :omg_watcher_repo,
+        id: OMG.Watcher.DB.Repo,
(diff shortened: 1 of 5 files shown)


apps/omg_watcher/lib/application.ex
```diff
@@ -50,63 +50,74 @@ defmodule OMG.Watcher.Application do
...
-
     # Define workers and child supervisors to be supervised
     block_finality_margin = Application.get_env(:omg_api, :ethereum_event_block_finality_margin)
     margin_slow_validator = Application.get_env(:omg_watcher, :margin_slow_validator)
 
     children = [
       # Start the Ecto repository
-      supervisor(OMG.Watcher.DB.Repo, []),
+      %{
+        id: :omg_watcher_repo,
```
paulperegud: One more place where lowercase atom ID is used.

PR #263, opened 2018-11-07
Feature/deferred config 

purbanow:
mix.lock:
add deferred config 

docs/source_consumption_log.md:
add deferred config  license
delete SQLite licenses


docs/source_consumption_log.md
```diff
@@ -24,11 +24,11 @@ cowlib 1.0.2            -> ISC
 credo 0.9.3             -> MIT
 db_connection 1.1.3     -> Apache 2.0
 decimal 1.5.0           -> Apache 2.0
+deferred_config 0.1.1   -> MIT
 dialyxir 0.5.1          -> Apache 2.0
 ecto 2.2.10             -> Apache 2.0
 eleveldb 2.2.20         -> Apache 2.0
 erlexec 1.7.5           -> BSD
-esqlite                 -> Apache 2
```
pik694: If you're deleting this dependency, shouldn't you remove it from `mix.lock` file as well?

docs/source_consumption_log.md
```diff
@@ -60,7 +60,6 @@ ranch 1.3.2             -> ISC
 poolboy 1.5.1           -> Apache 2.0
 sbroker                 -> Apache 2
 socket 0.3.13           -> WTFPL
-sqlitex                 -> MIT
```
pik694: Same as above

pik694: there is another dep which is probably redundant: `sqlite_ecto2`

PR #264, opened 2018-11-08
Unified API proposal

kevsul:


docs/unified_api.md
```diff
@@ -0,0 +1,105 @@
...
+/transaction.create
+/transaction.submit
+/block.get
+/status
+```
+
+## Architecture
+Suggestion from Robin is to have a sort of ReverseProxy app that sits in front of the other services and routes the calls to the appropriate service e.g. `/api/transaction.submit` goes to the ChildChain and `/api/account.get_utxos` goes to the Watcher.
+
+I think this can work with HTTP calls, although I'm not sure about WebSockets...
```
jarindr: @kevsul this is the websocket implementation guideline https://github.com/omisego/ewallet/blob/master/docs/guides/ewallet_api_websockets.md

ewallet is using a customized Elixir Phoenix websocket socket base for authentication btw.

kevsul: Thanks @jarindr 
The issue is about the ReverseProxy app though. 
With HTTP the proxy can easily route on the endpoint and redirect to the appropriate app. 
But with WebSockets I imagine that the proxy would have to parse each message to get the action, then route on that. It's probably not all that different to HTTP, but just seems more cumbersome.
I also don't know how WebSocket events would work if there's a proxy server in the middle.

Tagging @robinclart here, because I'm not sure exactly what he had in mind with the proxy server - I may be misunderstanding the intention.

jarindr: oh, it is under architecture 🤣 my bad. Anyway about reverse proxy websocket is possible and not that different from normal http calls since the handshake happen with http upgrade header anyway.
for ref we can use: https://www.nginx.com/blog/websocket-nginx/ as guideline for websocket proxy.
- About reverse proxy, are we talking about devop level or app level (nginx or phoenix)

pnowosie: Thank you @kevsul, I like proposition very much, especially simplicity and not fixing to the HTTP.
Great job :exclamation:

docs/unified_api.md
```diff
@@ -0,0 +1,105 @@
...
+    }
+  }
+}
+```
+## OmiseGo Plasma API
+An incomplete list of endpoints to demonstrate the format:
+```
+/account.get_balance
+/account.get_utxos
+/account.get_transactions
```
pnowosie: I think better explain the purpose will do `transaction.all` as it returns last transactions (possibly filtered by several criteria)

pnowosie: I think better explain the purpose will do transaction.all as it returns last transactions (possibly filtered by several criteria)
```suggestion
/transaction.all
```

docs/unified_api.md
```diff
@@ -0,0 +1,105 @@
...
+/transaction.get
+/transaction.create
+/transaction.submit
+/block.get
+/status
+```
+
+## Architecture
+Suggestion from Robin is to have a sort of ReverseProxy app that sits in front of the other services and routes the calls to the appropriate service e.g. `/api/transaction.submit` goes to the ChildChain and `/api/account.get_utxos` goes to the Watcher.
+
```
pnowosie: Regarding `Reverse Proxy` I think we have a requirement that Watcher should prevent it's user to submit transaction in certain conditions. Therefore this would be handled by the Watcher's endpoint which decides to pass the tx submition to the plasma chain or to warn user about the reason.

docs/unified_api.md
```diff
@@ -0,0 +1,105 @@
...
+## OmiseGo Plasma API
+An incomplete list of endpoints to demonstrate the format:
+```
+/account.get_balance
+/account.get_utxos
+/account.get_transactions
+/utxo.get_exit_data
+/utxo.get_challenge_data
+/transaction.get
+/transaction.create
```
pnowosie: I might be biased but `/transaction.create` means `create` as in `CRUD` - so inserting tx into DB.
I'm thinking of better word here to express the purpose as we're basically helping the user to encode the transaction that is ready to be sign & submit... hmm
 * `/transaction.prepare`
 * `/transaction.build`
 * `/transaction.smth_more_to_the_point` :arrow_left:

pnowosie: Anyway, best would be probably remove the endpoint and prepare `encoding library` which encapsulates all contract weirdness and can be reused in child chain, watchers and wallets

docs/unified_api.md
kevsul: We could probably leave in both. 
`/transaction.all` is more generic (i.e. it doesn't have to be filtered by address)
But `/account.get_transactions` feels more natural to get the account's transactions

docs/unified_api.md
kevsul: Yeah, I think we already agreed to remove this and let the client libs handle it.

docs/unified_api.md
jarindr: more note: if the reverse proxy is on the app level, on @omisego/wallet we already have an elixir app doing that https://github.com/omisego/ewallet/tree/master/apps/url_dispatcher which can be reuse easily in this case. cc. @T-Dnzt

docs/unified_api.md
kevsul: Does that mean that the ChildChain submit endpoint should not be called by the user ever? Not sure how that would be enforced.
Or is it just for convenience - so the Watcher can do some validation of the transaction and not submit an invalid one?

docs/unified_api.md
```diff
@@ -0,0 +1,105 @@
...
+   - On error, response contains response.error
+ - Watcher API:
+   - **Must** have 'Content-Type': 'application/json' header
+   - On error, response contains response.result === 'error'
+
+Also on the roadmap is the **Informational API Service** that will provide non-critical convenience APIs.
+
+All three APIs should behave consistently.
+  
+## Proposal
```
whoisjeremylam: Not sure if it's appropriate in this document - an option to be able to set CORS (for which components)?

docs/unified_api.md
```diff
@@ -0,0 +1,105 @@
...
+The eWallet already has a well defined API, using HTTP-RPC (rather than REST).
+ - [eWallet Admin API](https://ewallet.staging.omisego.io/api/admin/docs.ui)
+ - [eWallet Client API](https://ewallet.staging.omisego.io/api/client/docs.ui)
+
+We can follow the same model and ensure consistency across all OmiseGo services.
```
pnowosie:
```suggestion
We can follow the same model and ensure consistency across all OmiseGO services.
```

docs/unified_api.md
pdobacz: >the ChildChain submit endpoint should not be called by the user ever?
if the particular user is running a Watcher, probably no, it should not. It's not enforceable, but is a reasonable default that might not be easy to weasel out of.

But I suspect there will be watcher-less users, that have low-stakes accounts and don't care or are fine with some amount of trust. They can either speak to `submit` directly or speak to a 3P api, which speaks to a Watcher they trust, which in turn speaks to `submit`. 

> just for convenience

it is  not just for convenience. A particular transaction can be unsafe to send, depending on child chain conditions. Also, come MoreVP, the Watcher will need to track outgoing in-flight transactions.

pnowosie: In my understanding everyone can freely submit theirs transactions directly to the CC.
...and (pls correct me if I'm wrong) there will be Watcher extra-security feature :point_up: which prevents (or warns) the user about submitting tx in certain conditions. (Maybe I went to far about how it'll implemented)

docs/unified_api.md
pnowosie: Besides the benefits CORS could bring... Enabling CORS for all paths will be easy and should not affect this design. 

However if we would like to select the paths that are enabled for CORS - this approach has a drawback proposing all calls to be made with POST.  For instance it will be easier to configure enabling CORS just for GET request than to switch it by the path.

kevsul:
I've added to the list of API endpoints and events for each service. 
The details such as params, return types, etc. will be done elsewhere - this is just an attempt to get a complete list.
Please take a look and let me know if there's anything missing (the `events` section certainly needs work).

PR #265, opened 2018-11-08
Feature/omg 306 spend on finalization

pdobacz:
This is a first of series of PRs. I need to punt some tidies (as enumerated in numerous) and fixes to let this through and become a building block for MoreVP


README.md
```diff
@@ -462,11 +462,17 @@ Events:
...
+These should be treated as a prompt to mass exit immediately.
+
 Events:
 
 **invalid_block**
 
-Event informing about that particular block is invalid.
+Event informing about that particular block is invalid
+
+**unchallenged_exit**
```
purbanow: 1. Don't you think this event will be thrown too often(lots of false alarms)? 

I think there will be more valid/proper exits than invalid ones so from users perspective there is no point to challenged valid exit if the proof doesn't exist. 

Am I missing something?

apps/omg_api/lib/application.ex
```diff
@@ -53,7 +53,7 @@ defmodule OMG.API.Application do
              %{
                synced_height_update_key: :last_exiter_eth_height,
                service_name: :exiter,
-               block_finality_margin: block_finality_margin,
+               block_finality_margin: 0,
```
purbanow: Why 0? There will be no block_finality margin?

docs/exit_validation.md
```diff
@@ -1,48 +1,74 @@
...
+    * if `false` and there is less than `sla_margin` time till finalization -> `:unchallenged_exit`
+4. Spend utxos in `State` on exit finalization or challenging
+
+### Mandatory automatic challenging with a strategy + spend prohibition
+
+The above notifications aren't enough to guarantee soundness of the chain.
+
+By the power of defaults we should:
+  - force the user running the Watcher to keep a (modestly) funded Ethereum account (`challenger`), that will be used to challenge **automatically** all of the invalid exits seen
+  - do the automatic challenging according to some randomized strategy, that will ensure that the challenges won't burn each other
```
purbanow: So the `watcher` will use/keep user private key to make those automatic challenges to the root chain contract

docs/exit_validation.md
```diff
@@ -1,48 +1,74 @@
...
+  - prohibit spending if the chain is approaching invalid exit finalization
+  - prohibit spending also when there are valid exits, but the `challenger` address is underfunded
+
+To ensure the above works, challenges shouldn't involve a need to place large bonds nor should be required to be done from addresses holding funds.
```
purbanow: So who is going to pay for those automatic challenges? How will it work?

apps/omg_watcher/lib/eventer/event.ex
```diff
@@ -93,13 +93,34 @@ defmodule OMG.Watcher.Eventer.Event do
...
+            eth_height: pos_integer()
+          }
+  end
+
+  defmodule UnchallengedExit do
+    @moduledoc """
+    Notifies about an invalid exit, that is dangerously approaching finalization, without being challneged
+
+    It is a prompt to exit
+    """
```
purbanow: If you added `It is a prompt to exit` to `UnchallengedExit` maybe it is worth to add this to other `byzantine_bad_chain` events(invalid_block etc) to be consistent with what is in README.md

apps/omg_watcher/test/integration/invalid_exit_test.exs
```diff
@@ -65,19 +65,20 @@ defmodule OMG.Watcher.Integration.InvalidExitTest do
...
         amount: 10,
         currency: @eth,
         owner: alice.addr,
-        utxo_pos: utxo_pos
+        utxo_pos: utxo_pos,
+        eth_height: eth_height
       })
 
-    assert_push("invalid_exit", ^invalid_exit_event)
+    assert_push("invalid_exit", ^invalid_exit_event, 5_000)
```
purbanow: why only here it required to change the timeout to 5_000 for `assert_push`?

apps/omg_api/test/exit_processor/core_test.exs
```diff
@@ -0,0 +1,238 @@
...
+
+  alias OMG.API.Crypto
+  alias OMG.API.ExitProcessor.Core
+  alias OMG.API.State
+  alias OMG.API.Utxo
+  alias OMG.Watcher.Eventer.Event
+
+  require Utxo
+
+  @eth Crypto.zero_address()
```
purbanow: In the whole codebase, we always declare `@eth Crypto.zero_address()` maybe it is worth to make chore for cleaning this up?

purbanow:
The test coverage decrease from 82 to 79.9(https://coveralls.io/builds/19988437). Maybe you can have a quick look on this

README.md
pdobacz: > active exit having gone too long without being challenged

so this only is raised if the unchallenged exit approaches finality (which is very bad). This isn't about "currently isn't challenged" but "hasn't been challenged".

ah I see what's missing, the word "invalid", thx good catch

apps/omg_api/lib/application.ex
pdobacz: yes, for the child chain, for exits we want to remove them from UTXO set immediately, regardless of reorgs etc. I'll make an explicit comment.

Not sure how this margin managed to slip through in fact.

docs/exit_validation.md
pdobacz: TBD, but the idea is to have them paid from that `challenger` account which holds a non-critical amount of eth, just for challenging - hence allowing the Watcher (or other automatic service) to operate with control over `challenger`'s private key.

In terms of "pay", as opposed to "provide ether for gas", the challenger should be reimbursed from the exit bond being slashed. It isn't yet fully clear how to do this though.

apps/omg_watcher/lib/eventer/event.ex
pdobacz: it is there, line 465, to cover the whole section

apps/omg_watcher/test/integration/invalid_exit_test.exs
pdobacz: only here we don't know how long to wait exactly, so the custom timeout is useful to not wait anything extra. In other cases, we're waiting on other things (a particular child block validation, `BlockGetter` stopping) so the timeout is not necessary.

apps/omg_api/test/exit_processor/core_test.exs
pdobacz: I think this is good. `Crypto.zero_address()` is the common function for those apps which depend on `omg_api`. `@eth` is just syntax sugar to shorten stuff and make readable. We have little option to have a 4-char `@eth` which is defined only once and not make a mess.

Also note that `@eth` isn't used in "productioney" code (only tests, perf-test, devhelpers etc), which is good and should remain like this probably. And only code that would otherwise use `Crypto.zero_address` a lot should make this shortcut.

pdobacz:
@purbanow the reason for the coveralls drop is `ExitProcessor` being in `omg_api`, when it belongs (and is documented as such) in `omg_watcher`. Thanks!

PR #267, opened 2018-11-13
Add rinkeby manual

pik694:
Contains changes corresponding to running child chain on a testnet and steps required to deploy the contract manually.


README.md
```diff
@@ -146,17 +161,66 @@ Note that you'll need to pass the configuration file each time you run `mix` wit
...
+authority_raw = from_hex(authority)
+
+# fund the authority
+value = :math.pow(10, 18)
+        |> round
+        |> to_hex()
+{:ok, _} = Ethereumex.HttpClient.eth_send_transaction(%{from: faucet, to: authority, value: value})
+
+#deploy the contract
+{:ok, _} = deploy_result = Eth.RootChain.create_new("./", authority_raw)
```
pdobacz: instead of this assignment, you should just `...create_new(...) |> deploy_sync!()`, _et voila_

README.md
```diff
@@ -146,17 +161,66 @@ Note that you'll need to pass the configuration file each time you run `mix` wit
...
+import Eth.Encoding
+
+{:ok, _} = Application.ensure_all_started(:ethereumex)
+
+{:ok, accounts} = Ethereumex.HttpClient.eth_accounts()
+faucet = accounts
+         |> Enum.at(0) # assuming that this is an unlocked faucet
+
+# create and unlock authority account
+password = "secret_password"
```
pdobacz: unused

README.md
```diff
@@ -146,17 +161,66 @@ Note that you'll need to pass the configuration file each time you run `mix` wit
...
+And then in `IEx`
+```elixir
+alias OMG.Eth.DevHelpers
+alias OMG.Eth
+
+import Eth.Encoding
+
+{:ok, _} = Application.ensure_all_started(:ethereumex)
+
+{:ok, accounts} = Ethereumex.HttpClient.eth_accounts()
```
pdobacz: `{:ok, [faucet | _]} = ...` looks better and is more robust, but still has a ton of assumptions.

Regardless of the finer-grained comments, I think this script duplicates `prepeare_env!` a lot, while it could be made a rather straightforward extension of that. I think it should take less default assumptions and take some arguments (like the `faucet` address, that is only assumed unlocked and funded, instead of assuming it is the first acc known to `geth`).

Then, in this readme here, you just explain the usage of the modified `prepare_env!()` script + make a note about the necessity of funding the faucet (of course, this is sth like "personal faucet", that the user maintains, uses and reuses for (possibly) other testnet needs.

Lastly, mention that the `authority` account needs refunding from time to time (but this should rather be done using `geth attach` or `MEW` or sth

README.md
```diff
@@ -103,16 +103,31 @@ If you don't already have access to a developer instance of `geth`, follow the [
...
 Alternatively, a persistent developer instance that does not lose state can be started with the following command:
-```
+```bash
 geth --dev --dev.period 1 --rpc --rpcapi personal,web3,eth,net  --rpcaddr 0.0.0.0 --datadir ~/.geth
 ```
 
+##### Connecting to a testnet
+
+Another alternative might be running the whole setup on some official testnet, ex. `rinkeby`.
+It is recommended to provide the `datadir` parameter, otherwise the default value will be used.
```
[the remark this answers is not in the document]
pik694: I thought that keys would mix, but rinkeby has indeed its own keystorage.

README.md
```diff
@@ -103,16 +103,26 @@ If you don't already have access to a developer instance of `geth`, follow the [
...
 ```
 
 ##### Persistent developer `geth` instance
 Alternatively, a persistent developer instance that does not lose state can be started with the following command:
-```
+```bash
 geth --dev --dev.period 1 --rpc --rpcapi personal,web3,eth,net  --rpcaddr 0.0.0.0 --datadir ~/.geth
 ```
 
+#### Connecting to a testnet
```
pdobacz: for consistency with the later section: `to a non-dev chain`

PR #268, opened 2018-11-13
Feature/omg 306 spend on finalization2

pdobacz:
https://github.com/omisego/elixir-omg/pull/265 continued with critical tidies and fixes


apps/omg_api/lib/state/core.ex
```diff
@@ -395,26 +395,28 @@ defmodule OMG.API.State.Core do
...
+          {:ok, {[exit_event], [db_update], {list(Utxo.Position.t()), list(Utxo.Position.t())}}, new_state :: t()}
   def exit_utxos([%{utxo_pos: _} | _] = exit_infos, %Core{} = state) do
     exit_infos
     |> Enum.map(&Utxo.Position.decode(&1.utxo_pos))
     |> exit_utxos(state)
   end
 
   def exit_utxos(exiting_utxos, %Core{utxos: utxos} = state) do
-    exiting_utxos = Enum.filter(exiting_utxos, &utxo_exists?(&1, state))
+    _ = if exiting_utxos != [], do: Logger.info(fn -> "Recognized exits #{inspect(exiting_utxos)}" end)
```
pthomalla: if statement should be inside function

apps/omg_api/lib/state.ex
```diff
@@ -209,18 +207,14 @@ defmodule OMG.API.State do
...
 
+    # persistence is required to be here, since propagating the block onwards requires restartability including the
+    # new block
     :ok = DB.multi_update(db_updates)
 
     ### casts, note these are no-ops if given processes are turned off
     FreshBlocks.push(block)
-    BlockQueue.enqueue_block(block.hash, block.number)
-    # enrich the event triggers with the ethereum height supplied
-    event_triggers
```
pthomalla: child chain don't need events?

apps/omg_api/lib/state.ex
```diff
@@ -209,18 +207,14 @@ defmodule OMG.API.State do
         } ms"
       end)
 
+    # persistence is required to be here, since propagating the block onwards requires restartability including the
+    # new block
     :ok = DB.multi_update(db_updates)
```
pthomalla: inconsistency a bit in when to complete the data

apps/omg_api/lib/state.ex
```diff
@@ -176,30 +167,37 @@ defmodule OMG.API.State do
...
-
-  defp do_form_block(state, eth_height \\ nil) do
     {:ok, child_block_interval} = Eth.RootChain.get_child_block_interval()
 
-    {core_form_block_duration, {:ok, {%Block{number: blknum} = block, event_triggers, db_updates}, new_state}} =
+    _ = Logger.debug(fn -> "Forming new block..." end)
+
+    {core_form_block_duration,
+     {:ok, {%Block{number: blknum, hash: blkhash} = block, _event_triggers, db_updates}, new_state}} =
       :timer.tc(fn -> Core.form_block(child_block_interval, state) end)
```
pthomalla: why :close_block don't print duration but :form_block print?

apps/omg_api/lib/state.ex
```diff
@@ -176,30 +167,37 @@ defmodule OMG.API.State do
...
   """
   def handle_call({:close_block, eth_height}, _from, state) do
-    {duration, {result, new_state}} = :timer.tc(fn -> do_form_block(state, eth_height) end)
-    _ = Logger.debug(fn -> "Closing block done in #{inspect(round(duration / 1000))} ms" end)
-    {:reply, result, new_state}
+    {:ok, child_block_interval} = Eth.RootChain.get_child_block_interval()
+
+    {:ok, {_block, event_triggers, db_updates}, new_state} = Core.form_block(child_block_interval, state)
+
+    # enrich the event triggers with the ethereum height supplied
```
pthomalla: move line down, which this comment

apps/omg_api/test/state/core_test.exs
```diff
@@ -522,7 +522,8 @@ defmodule OMG.API.State.CoreTest do
             {[
                %{exit: %{owner: ^expected_owner, utxo_pos: ^utxo_pos_exit_1}},
                %{exit: %{owner: ^expected_owner, utxo_pos: ^utxo_pos_exit_2}}
-             ], [{:delete, :utxo, {@child_block_interval, 0, 0}}, {:delete, :utxo, {@child_block_interval, 0, 1}}]},
+             ], [{:delete, :utxo, {@child_block_interval, 0, 0}}, {:delete, :utxo, {@child_block_interval, 0, 1}}],
+             {[_, _], []}},
```
pthomalla: why ```[_,_]``` isn't ```^utxo_pos_exits```?

apps/omg_eth/lib/eth.ex
```diff
@@ -35,9 +35,8 @@ defmodule OMG.Eth do
 
   def get_ethereum_height do
     case Ethereumex.HttpClient.eth_block_number() do
-      {:ok, "0x" <> height_hex} ->
-        {height, ""} = Integer.parse(height_hex, 16)
-        {:ok, height}
+      {:ok, height_hex} ->
+        {:ok, int_from_hex(height_hex)}
```
pthomalla: a bit confusing, using import, ```Encoding.int_from_hex``` it is more direct

apps/omg_eth/lib/eth.ex
```diff
@@ -133,11 +132,12 @@ defmodule OMG.Eth do
     signature |> ExthCrypto.Hash.hash(ExthCrypto.Hash.kec()) |> to_hex()
   end
 
-  defp filter_not_removed(logs) do
-    logs |> Enum.filter(&(not Map.get(&1, "removed", true)))
+  defp filter_not_removed(logs, want_removed: want_removed) do
```
pthomalla: the function name now does not match, 
maybe use ```:maps.filter``` directly in get_ethereum_events?

apps/omg_api/lib/application.ex
```diff
@@ -51,12 +51,15 @@ defmodule OMG.API.Application do
...
-               synced_height_update_key: :last_exiter_eth_height,
-               service_name: :exiter,
                # 0, because we want the child chain to make UTXOs spent immediately after exit starts
                block_finality_margin: 0,
+               synced_height_update_key: :last_exiter_eth_height,
+               service_name: :exiter,
                get_events_callback: &OMG.Eth.RootChain.get_exits/2,
-               process_events_callback: &OMG.API.State.exit_utxos/1,
+               process_events_callback: fn exits ->
+                 {status, db_updates, _validities} = OMG.API.State.exit_utxos(exits)
```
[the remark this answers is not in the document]
pdobacz: it is not what `EthereumEventListener` expects. Here we force `exit_utxos` to behave like a proper `process_events_callback`

apps/omg_api/lib/root_chain_coordinator/core.ex
```diff
@@ -39,9 +39,9 @@ defmodule OMG.API.RootChainCoordinator.Core do
   `allowed_services` - set of names of services that are being synchronized
   `root_chain_height` - current root chain height
   """
-  @spec init(MapSet.t(), non_neg_integer()) :: t()
+  @spec init([atom], non_neg_integer()) :: t()
```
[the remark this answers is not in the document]
pdobacz: yeah, I find the `[...]` notation a bit confusing too

apps/omg_api/lib/state/core.ex
pdobacz: no, it should be outside, otherwise we'd get an empty log line

apps/omg_api/lib/state.ex
pdobacz: no, it doesn't

apps/omg_api/lib/state.ex
pdobacz: for `form_block` (child chain) it is the only logs that would print on doing a new block. Also there are calculations that actually matter for the child chain (hashing).

In `close_block` (watcher) case these logs and time isn't relevant in my opinion

apps/omg_eth/lib/eth.ex
pdobacz: there's a lot of occurrences in this file and `Encoding` is an `Eth` internal tool. There would be a lot of `Encoding...`. Then, by consistency, we'd flood `Eth.RootChain` which these `Encoding` too.

In general I prefer explicit, but in this situation I think no prefix is more readable

apps/omg_eth/lib/eth.ex
pdobacz: renamed function + used `match?` for more readable implementation.

I'd rather avoid erlangey calls if possible

pdobacz: renamed function + used `match?` for more readable implementation.

I'd rather avoid erlangey calls if possible

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -80,155 +81,182 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
-  end
-
-  @tag fixtures: [:empty_state, :alice, :state_alice_deposit]
-  test "can work with State to determine valid exits and finalize them", %{
+  @tag fixtures: [:empty_state, :alice, :state_empty, :events]
+  test "handles invalid exit finalization - doesn't forget and activates", %{
     empty_state: processor_state,
     alice: %{addr: alice},
-    state_alice_deposit: state
+    state_empty: state,
```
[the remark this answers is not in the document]
pdobacz: yeah, just realized that too and was planning to change that, thx for the catch :)

apps/omg_api/lib/application.ex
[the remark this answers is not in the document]
pthomalla: Yes, you are right

apps/omg_api/lib/state.ex
pthomalla: :close_block and :form_block they are very similar, I thought about separating the common part.

apps/omg_eth/lib/eth/encoding.ex
```diff
@@ -39,4 +39,15 @@ defmodule OMG.Eth.Encoding do
   # credo:disable-for-next-line Credo.Check.Consistency.SpaceAroundOperators
   @spec from_hex(<<_::16, _::_*8>>) :: binary
   def from_hex("0x" <> encoded), do: Base.decode16!(encoded, case: :lower)
+
+  @doc """
+  Decodes to an integer, see `to_hex`
+  """
+  # because https://github.com/rrrene/credo/issues/583, we need to:
+  # credo:disable-for-next-line Credo.Check.Consistency.SpaceAroundOperators
+  @spec int_from_hex(<<_::16, _::_*8>>) :: non_neg_integer
```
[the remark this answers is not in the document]
pdobacz: to be consistent with `from_hex/1`, that would require a big change. Is there a strong reason to change I'm missing?

PR #272, opened 2018-11-20
Feature/omg 297 chain operator doesn't submit empty blocks

purbanow:


apps/omg_api/lib/block_queue.ex
```diff
@@ -138,16 +139,27 @@ defmodule OMG.API.BlockQueue do
       {:noreply, state1}
     end
 
-    def handle_cast({:enqueue_block, block_hash, block_number}, %Core{} = state) do
-      state2 = Core.enqueue_block(state, block_hash, block_number)
+    def handle_cast({:enqueue_block, %Block{transactions: [], number: block_number, hash: block_hash}}, %Core{} = state) do
+      state1 = Core.enqueue_block(state)
 
       _ =
         Logger.info(fn ->
```
pthomalla: unnecessary log

apps/omg_api/lib/block_queue/core.ex
```diff
@@ -126,6 +126,10 @@ defmodule OMG.API.BlockQueue.Core do
     end
   end
 
+  def enqueue_block(state) do
```
pthomalla: unnecessary function, handle_cast should send what the got to Core

apps/omg_api/lib/state/core.ex
```diff
@@ -290,45 +290,49 @@ defmodule OMG.API.State.Core do
...
-      [db_updates_new_utxos, db_updates_spent_utxos, db_updates_block, db_updates_top_block_number]
-      |> Enum.concat()
-
-    new_state = %Core{
-      state
-      | tx_index: 0,
-        height: height + child_block_interval,
-        pending_txs: []
-    }
+    if Enum.empty?(txs) do
```
pthomalla: add form_block with match empty pending_txs, 
you will not have this big indentation

apps/omg_api/lib/fresh_blocks.ex
```diff
@@ -56,9 +56,18 @@ defmodule OMG.API.FreshBlocks do
     {:reply, result, state}
   end
 
-  def handle_cast({:push, block}, state) do
+  def handle_cast({:push, %Block{transactions: [], number: block_number, hash: block_hash}}, state) do
```
pthomalla: instead getting Block with an empty transactions, it should get :not_form or something else

docs/exit_validation.md
```diff
@@ -52,7 +52,8 @@ Causes to emit an `:invalid_finalization` event
     * if `true` -> noop,
     * if `false` -> emit `:invalid_exit` event  which leads to challenge
     * if `false` and there is less than `sla_margin` time till finalization -> `:unchallenged_exit`
-4. Spend utxos in `State` on exit finalization or challenging
+4. Spend utxos in `State` on exit finalization or challenging by call of `ExitProcessor.check_validity` method
```
pdobacz: this is not correct I think. Weren't we supposed instead to add a point here and say how validity checks are done, when and by what?

apps/omg_watcher/config/config.exs
```diff
@@ -13,7 +13,8 @@ config :omg_watcher,
   maximum_block_withholding_time_ms: 10_000,
   block_getter_height_sync_interval_ms: 2_000,
   maximum_number_of_unapplied_blocks: 50,
-  eth_exit_finality_margin: 12
+  eth_exit_finality_margin: 12,
+  exit_processor_validation_interval_ms: 5_000
```
pdobacz: add a shorter version of this interval to `config/dev` and even a shorter one to `config/test`

docs/exit_validation.md
purbanow:
```
....
4. Spend utxos in `State` on exit finalization or challenging
....
6. `ExitProcessor` check validation of exits is call periodically by itself with `exit_processor_validation_interval_ms`
interval and by `BlockGetter` every time when applying new block. 
```

apps/omg_api/lib/state/core.ex
purbanow: Thanks !

apps/omg_api/lib/block_queue/core.ex
purbanow:
```
 def handle_cast({:enqueue_block, %Block{transactions: []}}, %Core{} = state) do
      state1 = Core.enqueue_block(state)
      {:noreply, state1}
    end

    def handle_cast({:enqueue_block, %Block{number: block_number, hash: block_hash}}, %Core{} = state) do
      state1 = Core.enqueue_block(state, block_hash, block_number)

      _ =
        Logger.info(fn ->
          "Enqueuing block num '#{inspect(block_number)}', hash '#{inspect(Base.encode16(block_hash))}'"
        end)

      submit_blocks(state1)
      {:noreply, state1}
    end
```

The extra handle_cast is to prevent to run following code:
```
 _ =
        Logger.info(fn ->
          "Enqueuing block num '#{inspect(block_number)}', hash '#{inspect(Base.encode16(block_hash))}'"
        end)

      submit_blocks(state1)
```

apps/omg_api/lib/fresh_blocks.ex
pdobacz: 1. it's still misleading, because you return an empty block, but you don't persist it. If the name `form_block` is the problem, then perhaps `maybe_form_block`? to indicate the optionality. But I think `form_block` as a signal sent to an actor is still fine - `State` replies "no I won't" and that's it. The optionality is quite evident from the return type (list of Blocks, which might be empty) and a `@doc` can clarify any doubts further. Note, that with your approach you have `enqueue_block` and `push` which don't do what the verb says, so same problem. 
2. Good catch here, it might be the case that `wait_for_enqueue` is a leftover/redundant check :(. Let's check this

pdobacz: due to problems with handling of empty blocks in the Watcher (`close_block`), let's consider reverting to your original idea with a pre-check in `BlockQueue`

apps/omg_api/lib/block_queue.ex
```diff
@@ -118,16 +120,18 @@ defmodule OMG.API.BlockQueue do
...
 
     @doc """
-    Checks the status of both Ethereum root chain and the top mined child block number to decide what to do
+    Checks the status of the Ethereum root chain, the top mined child block number
+    and status of State to decide what to do
     """
     def handle_info(:check_ethereum_status, %Core{} = state) do
       {:ok, height} = Eth.get_ethereum_height()
       {:ok, mined_blknum} = Eth.RootChain.get_mined_child_block()
+      {_, is_empty_block} = OMG.API.State.get_status()
```
pthomalla: maybe shorter function is_empty_block or is_beginning?

apps/omg_api/test/block_queue/core_test.exs
```diff
@@ -210,135 +219,141 @@ defmodule OMG.API.BlockQueue.CoreTest do
...
-               |> enqueue_block("2", 2 * @child_block_interval)
+               |> enqueue_block("1", @child_block_interval, 0)
+               |> enqueue_block("2", 2 * @child_block_interval, 0)
                |> get_blocks_to_submit()
     end
 
-    test "Block generation is driven by Ethereum height" do
-      assert {:dont_form_block, queue} =
+    test "Block generation is driven by last enqueued block Ethereum height and if block is empty or not" do
+      %Core{minimal_enqueue_block_gap: minimal_enqueue_block_gap, parent_height: parent_height} = empty()
```
pthomalla: empty should be a fixture

PR #275, opened 2018-11-22
[elixir-omg] Modify Dockerfile to use static binary release of solc

jbunce:
`solc` 0.5.0 breaks compatibility with our smart contacts. This PR pins `solc` to 0.4.23 to fix builds

https://github.com/omisego/plasma-contracts/issues/41

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,8 +26,7 @@ RUN apt-get update \
   sudo \
   git \
   python3-pip \
-  python3-dev \
-  solc 
+  python3-dev 
 
 RUN wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb \
   && dpkg -i erlang-solutions_1.0_all.deb \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -52,6 +51,11 @@ ENV LANG=C.UTF-8
 
 WORKDIR /home/plasma/elixir-omg/
 
+RUN wget https://github.com/ethereum/solidity/releases/download/v0.4.23/solc-static-linux \
```
pdobacz: Jenkins builds use 0.4.24:
https://github.com/omisego/elixir-omg/blob/develop/Jenkinsfile
https://hub.docker.com/r/omisegoimages/blockchain-base/~/dockerfile/

alternatively we could consistently do 0.4.25, which seems to be the recent 0.4

jbunce: Version bumped to 0.4.25

Issue #266, opened 2018-11-09
Watcher API responses redefinition


pnowosie:
I'd like to start a discussion to redefine data WatcherAPI returns to make it align better for new database structure.

**Proposed changes**
 * Added Plasma's `Block` definition
 * Changed `Transaction` definition for 4in/4out and reuse other definitions
 * New definition of `TransactionItem` - list of transaction items with less data
 * Changed `Utxo` definition - added `owner` and removed `txbytes`

You can take a look at swagger definition changes in [commit](https://github.com/omisego/elixir-omg/commit/b672df4537c10f29418a6c2c4cb0f694e91c029d) or review it in [SwaggerUI](https://petstore.swagger.io/?url=https%3A%2F%2Fs3-eu-west-1.amazonaws.com%2Fwatcher-api%2Fswagger.json#/)

jarindr:
I love this, some question before we go further:
- why do we choose to rename `transaction` to `transactionItem`

pnowosie:
Yeah - sorry it's a little hard to think of it from git-diff. 
So `Transaction` is there as well and contains mostly the same data but nicely structured.
`TransactionItem` is a new structure for items of the list of transactions (e.g.: `/transactions?address=`) and is limited to most important data.

kevsul:
I understand that you want to reduce the amount of data returned for each transaction in the list, but I wonder if the reduced data is useful at all?

I can think of 2 reasons you would want to get a list of txs for an account: a BlockExplorer and a Wallet. In both cases you would want to show the value (and currency) of the tx, and whether it's incoming or outgoing
e.g. [0x742d35cc6634c0532925a3b844bc454e4438f44e](https://etherscan.io/address/0x742d35cc6634c0532925a3b844bc454e4438f44e)

It's a lot of traffic if the client has to make separate call for each transaction to get the missing details...

You _could_ try to calculate the value in the watcher and add that to TransactionItem. But I'd be inclined to just send the entire Transaction and let the client do it.

pnowosie:
I can agree to the point that whether transaction list is filtered by particular account address, at least direction currency and amount of value (sent/received) by address is the interesting part.
However in more general case with *n-inputs* and *k-outputs* this is probably a waste of throughput as we couldn't present these data on the list.

kevsul:
Ok, so if the address is passed in, then the server can calculate `value`, `currency` and `direction` and include them in the returned data. `value` is the amount sent or received by the address.

[...]

PR #277, opened 2018-11-22
Feature/api specs proposal

kevsul:


docs/api_specs/04_watcher_websocket_specs.md
```diff
@@ -0,0 +1,139 @@
...
+--------- | ------- | -----------
+amount | Integer | 
+currency | HEX-encoded string | 
+owner | HEX-encoded string | 
+utxo_pos | Integer |
+eth_height | Integer |
+
+
+
+## Other planned events
```
pdobacz: Here I think we can take on an approach, where we only notify, when there's a clear action to be taken. 

Perhaps we don't want to just forward all the events from Ethereum. We of course may decide that we do want that, but to keep consistent with the standard exits, we should remove (or at least separate visibly/deprioritize):

`inflight_exit_started`, `inflight_exit_challenged`, `piggyback_to...`, `fees_exited`,  `new_deposit`, `exit_started`, `exit_challenged`, `exit_success`.

Next thing is - if we want these - which of these are topics, which are events - it depends on the exact way these will be subscribed to. My first guess - when subscribing I want to track a particular address, in the spirit of `transfer:{address}` - maybe more generic, like `activity:{address}`

Next: I'd distinguish the critical events, that should become prompts to do something (for now the plan is that we'll only prompt, not act autonomously, but that's also debated), probably all filled under Topic `byzantine`:

`invalid_in_flight_exit` - which prompts to provide a known competitor
`invalid_piggyback` - which prompts to provide a challenge. Optionally can distinguish between input/output
`invalid_in_flight_challenge` - which prompts to provide a proof of canonicity
`in_flight_exit_required` - which prompts to start an IFE from a local transaction that is jeopardized by a byzantine child
`piggyback_required` - which prompts to piggyback on an IFE. Optionally can distinguish between input/output
`invalid_fee_exit` - might as well blend into `unchallenged_exit` as the consequences are the same.

docs/api_specs/04_watcher_websocket_specs.md
```diff
@@ -0,0 +1,139 @@
...
+in no particular order or severity
+To be clarified...
+
+ * inflight_exit_started
+ * inflight_exit_challenged
+ * piggyback_to_input
+ * piggyback_to_output
+ * invalid_fee_exit
+ * fees_exited
+ * fees
```
pdobacz: I understand this is "fees paid"-like event.

There's an option to include fees summary in `new_block` (favorite).

But it's hard to reason about fee events in general, we should think about the architecture of that, considering what we know at this point.

Also we might want to include fee information in `address_spent/received` events and in the "rich" transaction data from informational API

docs/api_specs/04_watcher_websocket_specs.md
```diff
@@ -0,0 +1,139 @@
...
+ * new_block
+ * new_transaction
+ * new_deposit
+ * deposit_spendable
+ * exit_started
+ * exit_challenged
+ * exit_success
```
pdobacz: `exit_finalized` regardless of the fate of this

docs/api_specs/04_watcher_websocket_specs.md
```diff
@@ -0,0 +1,139 @@
...
+
+ * inflight_exit_started
+ * inflight_exit_challenged
+ * piggyback_to_input
+ * piggyback_to_output
+ * invalid_fee_exit
+ * fees_exited
+ * fees
+ * new_block
+ * new_transaction
```
pdobacz: that's super verbose - wouldn't a natural flow be "be informed about a new block -> look up all txs from that block"?

docs/api_specs/04_watcher_websocket_specs.md
```diff
@@ -0,0 +1,139 @@
...
+ * new_deposit
+ * deposit_spendable
```
pdobacz: there's a long discussion about this in PT (OMG-177), but there's some pivot into this being just waiting for `eth_height_of_deposit + some_margin_child_chain_promised` - so it can fall into the "DIY" category of ethereum events

docs/api_specs/05_informational_api_specs.md
```diff
@@ -0,0 +1,388 @@
...
+
+### JSON Body
+
+Key | Type | Description
+--------- | ------- | -----------
+address | Hex encoded string | Address of the funds owner
+
+
+
+## Account - Get Utxos
```
pdobacz: I think currently it falls under `security critical`. Otherwise the user doesn't know what to spend or exit in case of a byzantine condition.

Honestly, figuring out how to draw the line here evades for a long time :/.

docs/api_specs/02_operator_api_specs.md
```diff
@@ -0,0 +1,86 @@
+# Child chain
```
pdobacz: we might want to add `settle` which will take a transaction and RC-exchange settlement proof, to start designing around that.

Then we'll probably mirror that in `Watcher`, just like we're doing with `submit`

docs/api_specs/04_watcher_websocket_specs.md
```diff
@@ -0,0 +1,139 @@
...
+> Both event types have the same structure.
+Key | Type | Description
+--------- | ------- | -----------
+child_blknum | Integer | 
+child_txindex | Integer | 
+child_block_hash | HEX-encoded string |
+submited_at_ethheight | integer |
+tx | object | Structure of signed transaction
+
+Blocks are validated by the Watcher after a short (not-easily-configurable) finality margin. By consequence, above events will be emitted no earlier than that finality margin. In case extra finality is required for high-stakes transactions, the client is free to wait any number of Ethereum blocks (confirmations) on top of submitted_at_ethheight.
```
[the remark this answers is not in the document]
pnowosie: Yes, it's copied from README... here is probably out of context... we don't need it here

docs/api_specs/05_informational_api_specs.md
kevsul: Oof, this is a tough one!
You're right that the user will need to get their utxos to be able to exit.
But tracking utxos is going to affect the performance of the Watcher...

Maybe it could be in both places - the Informational API (need a better name for this) can always track and store utxos in its DB, so they can be quickly retrieved by e.g. a Wallet.

And Watcher doesn't maintain a list of utxos anywhere, but can provide a account.get_utxos call that can dynamically look them up. It's computationally more expensive, but should only happen in unusual circumstances.

pnowosie: I think `looking for utxo only in unusual cases`  is OK. Would it make sense to not serve this functionality in "Security" & "Info API" under the same name? It will do the same, but the first one should make it clear `don't call me too often`.

Any way we'd need to store owner address in OMG.DB, computing it from signature for every lookup will be too much.

docs/api_specs/04_watcher_websocket_specs.md
pdobacz: EDIT: `in_flight_exit_required` might be heavily "nice-to-have". The flow can be, that on any event that indicates a byzantine chain, user already knows that all in flight transactions should be IFEd. Scratching that from the above list. As part of containing the byzantine condition, the user will ask the Watcher for such txs.

docs/api_specs/05_informational_api_specs.md
pdobacz: yeah - agree on the dual nature of this endpoint - same API, different implementation and performance. In the `security critical` world, the UTXO set will just be brute-force scanned (order-of-magnitude millions of entries max - as determined by exit-ability implications of the protocol).

Mind however 

>we'd need to store owner address in OMG.DB

it is stored - every UTXO's value is `{amount, currency, owner}`, `owner` being the address.
When you have the UTXO positions necessary, you can grab the blocks, txbytes and calculate proof from that.

pnowosie:
As a reminder from #280 we need to keep specs markdowns here
 - [ ] index.html.md
 - [ ] includes/*.md - without the leading numbers in names

docs/api_specs/includes/watcher_websocket_specs.md
```diff
@@ -0,0 +1,214 @@
...
+  },
+  "join_ref": null,
+  "event": "address_received"
+}
+```
+
+
+## Topic `childchain`
+
+### Events `new_block`
```
purbanow: Typo, `Events` -> `Event`

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,314 @@
...
+
+<aside class="warning">
+The most critical function of the Watcher is to monitor the ChildChain and report dishonest activity. 
+The user must call the `/status` endpoint periodically to check. Any situation that requires the user to either exit or challenge an invalid exit will be included in the `byzantine_events` field.
+</aside>
+
+### Byzantine events
+All of the following events indicate a byzantine chain and that the user should either exit or challenge.
+
+#### `invalid_exit`
```
kevsul: Is this really a 'byzantine' event?
My understanding of 'byzantine' is that the chain has been compromised somehow (most likely due to a dishonest operator) and everyone should exit. 
An `invalid_exit` though could just be a dishonest user, and as long as it's challenged the chain itself is fine.

pdobacz: I'd say yes. The actor (user) is acting byzantine in the sense "not following protocol"/"arbitrarily".

I think it falls in the broad category of "byzantine" behavior, whereas what you seem to have in mind is "byzantine child chain". I tried to follow this differentiation or otherwise be clear about what's going on, but might have missed some spots.

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,314 @@
...
+    "success": true,
+    "data": {
+        "last_validated_child_block_number": 10000,
+        "last_mined_child_block_timestamp": 1535031020,
+        "last_mined_child_block_number": 11000,
+        "eth_syncing": true,
+        "byzantine_events":
+        [
+            {
+                "event": "invalid_exit",
```
pdobacz: ok, so moving this approach forward, we should have a similar construction for other conditions? Like (following status quo here, we can structure differently, if you think it's better)
`:invalid_block`, `:block_withholding`, `:unchallenged_exit`, `:invalid_piggyback`, `submit_competitor` (`invalid_ife`? `invalid_canonicality`? split to `:canonical_ife` + `:noncanonical_ife`, where the title would reflect what's byzantine?), `:piggyback_required` (`:piggyback_possible` depending on how we serve the user) , (optional) `tx_requires_ife`

(I'm making the names up now, based on the MoreVP diagram, sorry for any inconsistency).

CC @purbanow

docs/api_specs/includes/watcher_api_specs.md
kevsul: Got it, thanks 👍

docs/api_specs/includes/watcher_api_specs.md
kevsul: Right.
I'm not going to add any of these now though - the docs should be updated as these features are implemented

pdobacz: Yes, LGTM, but let's keep challenging this design and refer to MoreVP specs to be sure there aren't simpler/cleaner alternatives. To not fall in the trap of overspeccing (I tend to do that a lot ;) ).

When we settle on on that, I can update the [diagram](https://docs.google.com/drawings/d/1UaAMZTJBbikTM0eFSbNM7ZVgi1HZrLTfwYCUvHGUaW0/edit) to be more in sync with the names and division. But for now let's keep as is, conceptually it is the same.

With `noncanonical_ife` and `invalid_ife_challenge` being the same event, I was hoping they could be merged, but now, [here](https://github.com/omisego/plasma-contracts/blob/elixir_omg_morevp_compat/contracts/RootChain.sol#L526) and [here](https://github.com/omisego/plasma-contracts/blob/elixir_omg_morevp_compat/contracts/RootChain.sol#L579) I see there are differences in ways to respond are big enough to separate.

pdobacz: There's a fundamental question though, that I keep bumping into, so I'll drop it here.

Separation/merging between "telling an ACTION is required" and "giving all info to perform it".

1/ Currently we're doing the latter - push event (or `status/`) returns a "reference" to an ACTION (`:invalid_exit` with a respective utxo position to challenge) and then the user asks for details (proofs etc) on another endpoint `/utxo.get_challenge_data utxo_pos`
2/ there are performance issues - getting the full information is likely to be heavier than just the prompt, especially when we go in the "security critical Watcher" mode
3/ OTOH, API-wise, the thinnest API is when you just drop everything on the user in `status/` and not have `/utxo.get_challenge_data utxo_pos`
4/ OTOH^2, to be consistent, we'd need to move the `/utxo.get_challenge_data utxo_pos` which is additional cleanup work

So 2 vs 3 vs 4 needs to be considered. I'm slightly in favor of not removing `/utxo.get_challenge_data utxo_pos` and adding its MoreVP counterparts, but I might be wrong.

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -312,3 +312,110 @@ A block containing an invalid tx has been processed [are there other reasons a b
...
+<aside class="warning"> Not Implemented Yet.</aside> 
+
+Event details:
+Key | Type | Description
+--------- | ------- | -----------
+inflight_blocknum | integer | Block number of the in-flight transaction
+inflight_tx_index | integer | Index of the in-flight transaction
+
+#### `piggyback_available`
+> A piggyback_available event
```
purbanow: 1. Maybe `available_piggyback` to be consistent with `invalid_piggyback`.
2. We must not forget to specify endpoint which will be providing necessary information to make piggyback.

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -312,3 +312,110 @@ A block containing an invalid tx has been processed [are there other reasons a b
...
+
+Event details:
+Key | Type | Description
+--------- | ------- | -----------
+inflight_blocknum | integer | Block number of the in-flight transaction
+inflight_tx_index | integer | Index of the in-flight transaction
+inflight_io_index | integer | Index of the input or output that can be exitted.
+
+#### `invalid_piggyback`
+> A invalid_piggyback event
```
purbanow: 1. We must not forget to specify endpoint which will be providing necessary information to make a challenge of piggyback.

docs/api_specs/includes/watcher_api_specs.md
kevsul: I lean towards separating "reporting of the state" from "giving all info necessary to respond to it".
I think that there will be multiple polls to `/status`, but the user will only need to get the necessary info to perform the action once, so including all the info each time seems unnecessarily heavyweight.

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -417,4 +417,275 @@ inflight_tx_index | integer | Index of the in-flight transaction
...
+Attribute | Type | Description
+--------- | ------- | -----------
+txhash | Hex encoded string | The hash of the in-flight transaction
+
+
+
+## Inflight Exit - Get Piggyback Data
+
+```shell
+curl http://localhost:4000/inflight_exit.get_piggyback_data -d '{"txhash": "bdf562c24ace032176e27621073df58ce1c6f65de3b5932343b70ba03c72132d"}'
```
purbanow: This endpoint requires some modifications because is expecting an only txhash but returns:
a) `inflight_txbytes` which is ok 
b) `inflight_io_index` which index?

Potential solutions:
1. return only `inflight_txbytes`
2. return `inflight_txbytes` + list of indexes

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,691 @@
...
+
+```json
+{
+    "event": "piggyback_available",
+    "details": {
+        "inflight_blocknum"  : 10000,
+        "inflight_tx_index"  : 12,
+        "inflight_io_index"  : 1
+    }
+}
```
purbanow: Let's assume there is `InFlight transaction` with 4 inputs and 4 outputs then we're going to emit 8 events, right? Maybe it will be better to emit one `piggyback_available` event per `InFlifht transacion`

purbanow: We must design this event together with `inflight_exit.get_piggyback_data` endpoint to be elegant from a user flow perspective.

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,691 @@
...
+inflight_tx_index | integer | Index of the in-flight transaction
+
+#### `piggyback_available`
+> A piggyback_available event
+
+```json
+{
+    "event": "piggyback_available",
+    "details": {
+        "inflight_blocknum"  : 10000,
```
pdobacz: Not sure if not premature but I have a comment here: 

In order to search for piggybacks that target my address (following our today's conversation and assuming "no filtering" path is picked), this address must be delivered here. 

Probably to be able to get the "piggyback information", caller of `status/` would also need to know whether that is an input or output to piggyback on. There is a number of ways one could architect that, the fundamental thing is that the flow can be followed through by the user.

docs/api_specs/includes/watcher_api_specs.md
kevsul: Ok, makes sense. 
(they're not going to be 'events' anymore, they'll be entries in a list)

So, if I understand correctly:
- there's an entry for every in-flight exit with just {blocknum, txindex}
- the user needs to check if they own any inputs or outputs in the tx
- if they do, they need to get the piggyback data for the tx
- then they need to call RootChain.piggybackInFlightExit() for each input or output that they own

docs/api_specs/includes/watcher_api_specs.md
kevsul: Ok, so the data returned is just the tx bytes, and is the same for any input or output, right?
In that case, index isn't necessary.

docs/api_specs/includes/watcher_api_specs.md
kevsul: Does the caller need to know whether it's an input or output?
In Rootchain.sol implies you only need the inflight tx bytes and the index:
```
function piggybackInFlightExit(
        bytes _inFlightTx,
        uint8 _outputIndex
    )
```
i.e. the "piggyback information" appears to be the same whether it's an input or an output.

pdobacz: ah right, it is because the input vs output is encoded in `_outputIndex`, 0-3 means input, 4-7 means output, assuming 4in 4out tx.

It all depends on what information do you want to serve to the user, I'm not saying explicit input/output flag it's required, it's just worth thinking through and syncing with the "piggyback information" endpoint.

There's however the "only mine" step in the MoreVP flow, where you don't piggyback an input, if you're the only input owner, but maybe we can treat it as optimization? again, just prompting to think through.

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,691 @@
...
+{
+    "event": "invalid_block",
+    "details": {
+        "blockhash"  : "DB32876CC6F26E96B9291682F3AF4A04C2AA2269747839F14F1A8C529CF90225",
+        "blocknum"  : 10000,
+    }
+}
+```
+
+A block containing an invalid tx has been processed [are there other reasons a block can be invalid?]. User should exit.
```
pdobacz: re the question here: yes
 - invalid merkle root
 - totally corrupt, malformed data, that can't be decoded

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,691 @@
...
+        "inflight_tx_index"  : 12,
+        "inflight_input_index"  : 1,
+        "competing_blocknum"  : 10000,
+        "competing_tx_index"  : 10,
+        "competing_input_index"  : 0,
+    }
+}
+```
+
+An non-canonical in-flight exit has been started. It should be challenged.
```
pdobacz: 1/ `non-canonical transaction in-flight exit` or `in-flight exit of a non-canonical...` is more precise. Ah I see we settled on `noncanonical_ife` as the name, I guess the abbreviation there is ok, but in the description a full, precise wording is probably clearer.
2/ the non-canonicity can be caused by a transaction, that isn't included in a block, so the tripplet `"competing_..."` won't cover all cases. The most visual scenario is when two IFEs for competing txs are started, but none is included in a block.

The part that is related to the competing tx being included should probably be optional, but pls double check this with the protocol specs

And it also digs into the issue what to serve in here and what to accept in the respective "information returning" endpoint - one that gets called to take action on Ethereum

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,691 @@
...
+inflight_blocknum | integer | Block number of the in-flight transaction
+inflight_tx_index | integer | Index of the in-flight transaction
+inflight_io_index | integer | Index of the input or output that can be exitted.
+
+#### `invalid_piggyback`
+> A invalid_piggyback event
+
+```json
+{
+    "event": "invalid_piggyback",
```
pdobacz: same issue here - the tx to challenge with might not be an included one

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,691 @@
...
+inflight_io_index | integer | Index of the input or output of the in-flight transaction
+spending_blocknum | integer | Block number of the spending transaction
+spending_tx_index | integer | Index of the spending transaction
+spending_input_index | integer | Index of the spent input iofn the spending transaction
+
+
+## Inflight Exit - Get Exit Data
+
+```shell
+curl http://localhost:4000/inflight_exit.get_data -d '{"txhash": "bdf562c24ace032176e27621073df58ce1c6f65de3b5932343b70ba03c72132d"}'
```
pdobacz: input here is `txbytes` not `txhash`, c.f. OMG-345

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,691 @@
...
+```json
+{
+    "version": "1",
+    "success": true,
+    "data": {
+        "last_validated_child_block_number": 10000,
+        "last_mined_child_block_timestamp": 1535031020,
+        "last_mined_child_block_number": 11000,
+        "eth_syncing": true,
+        "byzantine_events":
```
pdobacz: we can also plan ahead to have space for the functionality of Watcher to track outgoing (in-flight) transactions (that could later be IFEd, if there's need). They also could be served somewhere here (not sure though), c.f. OMG-309

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,691 @@
...
+> The above command returns JSON document:
+
+```json
+{
+    "version": "1",
+    "success": true,
+    "data": {
+        "txbytes": "F847010180808080940000...",
+        "sigs": "7C29FB8327F60BBFC62...",
+        "input_txs" : [
```
pdobacz: this is more elegant, but the convention is to abide by whatever the contract requires (we can also consider dropping that convention), which would be two equal-sized lists of `input_txs` and `input_proofs`

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,691 @@
...
+Attribute | Type | Description
+--------- | ------- | -----------
+txhash | Hex encoded string | The hash of the in-flight transaction
+
+
+
+## Inflight Exit - Get Challenge Data
+
+```shell
+curl http://localhost:4000/inflight_exit.get_challenge_data -d '{"txhash": "bdf562c24ace032176e27621073df58ce1c6f65de3b5932343b70ba03c72132d"}'
```
pdobacz: similar here - how does this correlate with what we're getting from `status`? how would the user's flow be satisfied?

Hm.. in general I think it could be worthwhile to write out all of the `ACTION:...` entries from the MoreVP handling flows, in detail. E.g.:

1/ user pings `status/`
2/ user takes some data out of the response
3/ optionally user makes some decision based on that data
4/ user calls some endpoint using some piece of that data
5/ user turns the data from that response into an ethereum tx and calls the contract

That would allow us to double check whether all the actions are possible

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,691 @@
...
+```json
+{
+    "version": "1",
+    "success": true,
+    "data": {
+        "inflight_txbytes": "F847010180808080940000...",
+        "inflight_input_index": 1,
+        "competing_txbytes": "F317010180808080940000...",
+        "competing_input_index": 1,
+        "competing_txid": 2600003920012,
```
pdobacz: also note that `competing_input_index` might be 0, which means that no proof is there and the challenge uses a non-included transaction (or a transaction which inclusion is unknown)

docs/api_specs/includes/watcher_api_specs.md
purbanow: Yes, it's the same so an index is not necessary

docs/api_specs/includes/watcher_api_specs.md
purbanow: Exactly like that :+1:

README.md
```diff
@@ -559,3 +559,67 @@ To run a development `iex` REPL with all code loaded:
 ```bash
 iex -S mix run --no-start
 ```
+
+# Working with API Spec's
```
pnowosie: We have described it better in the README of `gh-pages` branch. Probably here we should just point to that README

docs/api_specs/includes/watcher_api_specs.md
kevsul: I've added this in inflight_exit_scenarios.md, which has led me to a few changes in the approach

1. /status returns a list of `byzantine_events` and also a list of `inflight_txs` and `inflight_exits`. I think separating them is cleaner and it avoids duplicating information in each event.
2. Events are related to the transactions by `txbytes`.  This is maybe a too heavy and could be done on `txhash` instead. But `txbytes` is used almost everywhere.
3. I'm assuming the Watcher can track piggybacks. For example it will report if an output is available to be piggybacked. If that output does get piggybacked then it won't report it as available anymore. Not sure if this is actually feasible?

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,650 @@
...
+> A piggyback_available event
+
+```json
+{
+    "event": "piggyback_available",
+    "details": {
+        "txbytes": "F3170101C0940000...",
+        "available_outputs" : [0, 1],
+        "available_inputs" : [0, 1]
+    }
```
purbanow: I think possible values for:
`available_outputs` 4, 5, 6,7 
`available_inputs`: 0, 1, 2, 3
Maybe it is worth to  mention somewhere here?

pdobacz: note that the `outputs` being numbered "on top" of `inputs` (4,5,6,7) should probably be seen as leakage of contract implementation details.

docs/api_specs/includes/watcher_api_specs.md
pdobacz: @pik694 @purbanow @paulperegud please, take a look at the bit `txbytes` vs `txhash` in Kevin's docs vs how you're solving on your side

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,650 @@
...
+}
+```
+
+An in-flight exit has been started and can be piggybacked. If all inputs are owned by the same address, then `available_inputs` will not be present.
+<aside class="warning"> Not Implemented Yet.</aside>
```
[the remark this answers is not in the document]
kevsul: This is to address the "only mine" situation. If all the inputs belong to the same owner, then there's no need to piggyback.
The intention here is that if this is the case, the user won't get a `piggyback_available`.
If we always always return all possible inputs, then the user will get the `piggyback_available` event, but then need to check if they really need to piggyback or not.

Either way is valid of course - this way leaves that logic in the Watcher as opposed to the User.

docs/api_specs/includes/watcher_api_specs.md
kevsul: Yeah, I would prefer to keep them 0-based in the API, and not expose the contract implementation details.

docs/inflight_exit_scenarios.md
```diff
@@ -0,0 +1,409 @@
...
+        "byzantine_events": [
+            {
+                "event": "block_withholding",
+                "details": {
+                    "blockhash"  : "DB32876CC6F...",
+                    "blocknum"  : 10000,
+                }
+            }
+        ],
+        "inflight_txs": [
```
pdobacz: might be worth pointing out that this is not in our immediate "features to implement" set. Until then, user only receives the `block_withholding` part and needs to know `txbytes` of in-flight txs themselves

docs/inflight_exit_scenarios.md
```diff
@@ -0,0 +1,409 @@
...
+RootChain.startInFlightExit(
+    response.data.txbytes, 
+    response.data.input_txs, 
+    response.data.input_proofs, 
+    response.data.sigs, 
+    {"value": inFlightExitBond}
+)
+```
+
+#### 3. Piggyback his output
```
pdobacz: the user sort of knows their "transaction" and their "exposure" in that transaction, but wouldn't it be more "by the books" to first wait on an appropriate information about piggybacking opportunity, ask for piggyback input and piggyback? 

Except for document clarity it is also reasonable, since then the user is certain that the exit start has mined on ethereum

docs/inflight_exit_scenarios.md
```diff
@@ -0,0 +1,409 @@
...
+                "piggybacked_outputs" : [0]
+            }
+        ]
+    }
+}
+ ```
+
+Because Bob has already started an exit for the transaction - there is a `piggyback_available` event indicating that `output[1]` can be piggybacked. This output is to Alice's address. Note that as Alice is the sole owner of the inputs, she does not need to piggyback any input.
+
+#### 1. Piggyback her output on the IFE
```
pdobacz: again, aren't we hitting the API to get Piggyback info? (either let's document that step, or maybe drop an unnecessary endpoint, leaving only `status`? not sure...)

docs/api_specs/includes/watcher_api_specs.md
purbanow: So the flow will be the following:
1.  A User gets `piggyback_available` event with "available_outputs" : [0, 1] 
2.  A User calls `get_piggyback_data`  
3.  A user wants to make piggyback for output 1
4.  A user must manually add 4 to output 1.
5.  A user calls root chain.`piggybackInFlightExit` method with `_inFlightTx` + `_outputIndex` = 5

Am I missing something ? 

note:
```
 function piggybackInFlightExit(
        bytes _inFlightTx,
        uint8 _outputIndex
    )
```

docs/inflight_exit_scenarios.md
```diff
@@ -0,0 +1,409 @@
...
+
+response:
+```json
+{
+    "version": "1",
+    "success": true,
+    "data": {
+        "inflight_txbytes": "F847010180808080940000...",
+        "inflight_input_index": 1,
+        "spending_txbytes": "F317010180808080940000...",
```
pdobacz: I think it is swapped with `inflight_txbytes`. Same comment to next section

docs/inflight_exit_scenarios.md
kevsul: How will the user know which transactions are in-flight?
The sender may know that they've submitted a transaction that hasn't been put into a block, but the receiver probably won't.

docs/api_specs/includes/watcher_api_specs.md
```diff
@@ -0,0 +1,650 @@
...
+Attribute | Type | Description
+--------- | ------- | -----------
+txbytes | Hex encoded string | The in-flight transaction that the event relates to
+
+#### `piggyback_available`
+> A piggyback_available event
+
+```json
+{
+    "event": "piggyback_available",
```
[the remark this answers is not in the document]
kevsul: The input and output addresses are already in the `inflight_txs`, so the idea is that the user can look them up there, instead of having them duplicated here.

docs/inflight_exit_scenarios.md
kevsul: Yeah, that endpoint has been dropped - the info required is minimal (`txbytes` and `outputIndex`) and it's already in the event.

docs/inflight_exit_scenarios.md
kevsul: You're right - correct behaviour would be to wait for the `piggyback_available` event.

docs/api_specs/includes/watcher_api_specs.md
kevsul: Correct, except for `2.` isn't necessary anymore because all the info necessary to piggyback is already in the event.

I know that manual +4 is ugly, but that's how the contract works. We could try to hide that from the user by automatically doing the +4 for them, but I feel it makes the API more brittle.
0-based arrays feel more neutral somehow.

I'm open to being persuaded otherwise though...

docs/inflight_exit_scenarios.md
pdobacz: oh, but it's all about the sender, sender has initiative here. I've recently realized, that if the receiver starts the IFE, then the receiver is exposed to being slashed/grieved, [cf.](https://github.com/omisego/elixir-omg/blob/master/docs/morevp.md#honest-receiver-should-not-start-in-flight-exits)

docs/api_specs/includes/watcher_api_specs.md
pdobacz: I would pitch doing `0-based and abstractly correct` in `status/events` and `contract-friendly` in the return of "get-piggyback-information" endpoint (to adhere by the conventions), but isn't this the endpoint that we just dubbed redundant and removed :joy: ? (Sorry if I'm lost, I'm not fully following the thread, feel free to disregard comment).

docs/inflight_exit_scenarios.md
kevsul: Yeah, agree with that scenario, but I don't see how the sender has an incentive to start an in-flight exit.

Ok, if there are multiple inputs from different address, then one of the senders _should_ start the in-flight, because then they have a chance of exiting that input if one of the _other_ senders is dishonest. A pretty rare edge case, IMO

However there is only one sender, they have nothing to gain by starting the exit.
If the sender is dishonest they can just start an MVP exit of the UTXO they spent in the in-flight transaction. Sure, this can be challenged by the receiver of the in-flight tx, but the not if the receiver doesn't know about it.
I guess someone else could challenge and the sender will lose their `standardExitBond`

But if the sender is honest, then they've 'spent' those tokens and it's not their problem if the receiver can't exit them. Unless they _want_ the receiver to exit in return for goods or something. Or maybe they have a significant amount of change in the transaction, then of course they will want to exit.

Hmm, maybe I've convinced myself :)

In any case, not implementing this yet affects some of the other api decisions and user flows e.g. I had assumed that the trigger to start an exit would be to see the user's address in the list.
But if we're not including this list it puts a lot more work on the user i.e. every time a user sends a transaction, they should store the details of it in case the chain goes byzantine and the transaction is in-flight. Are we comfortable with requiring wallets / apps to be responsible for this?

pdobacz: if the tx is withheld or included in an invalid position and the sender doesn't open an IFE didn't send anything (e.g. goods shouldn't be dispatched). It should be seen as a failed transaction and sender is to blame here.

Oh wait, I just realized, that the sender cannot fully prove validity and "exitability" of the funds they send to the receiver :(. An inclusion proof isn't enough. So it seems I might be wrong about the "receiver shouldn't IFE" part or at least we need some additional protocol, which puts the full burden of staring IFEs on one of the parties, so that disputes can be resolved somehow.

My recent update to `morevp.md` might also be too optimistic then :(.

This way or another, in the first iteration, it is the user who should know their respective IFE's due. So in case the receiver needs that, they should receive the prompt out-of-band and track the IF-txs themselves. (note that the plan is to start handling that for the user in the immediate future)

docs/api_specs/includes/watcher_api_specs.md
kevsul: Yes, I removed 'get-piggyback-information' because the only info you need are `txbytes` and the (controversial) index.

docs/api_specs/includes/informational_api_specs.md
```diff
@@ -0,0 +1,389 @@
...
+            "timestamp": 1540455586,
+            "hash": "0017372421F9A92BEDB7163310918E623557AB5310BEFC14E67212B660C33BEC",
+            "eth_height": 97425,
+            "blknum": 68290001
+        }
+    ]
+}
+```
+
+Gets a block with the given id
```
pdobacz: copy-paste-o?

docs/api_specs/includes/watcher_websocket_specs.md
```diff
@@ -0,0 +1,114 @@
...
+timestamp | integer |
+
+
+
+# RootChain - Events
+The RootChain contract raises certain events on Ethereum
+<aside class="warning">TODO We may want to forward these events through the watcher</aside> 
+<aside class="warning">TODO Most of these events are not currently implemented in the RootChain contract</aside> 
+
+ * new_deposit
```
pdobacz: requires grooming wrt. `status/` now returning a lot of these

docs/inflight_exit_scenarios.md
```diff
@@ -0,0 +1,415 @@
...
+ ```
+
+ Alice sees that her in-flight exit is in progress and she can now piggyback her ouput. Note that as Alice is the sole owner of the inputs, she does not need to piggyback any input.
+
+#### 4. Piggyback the output
+The second argument is `5` because she is piggybacking the first output.
+```
+RootChain.piggybackInFlightExit(
+    response.data.txbytes,
+    5,
```
pdobacz: I fear that the `+4`  part will bite us. Maybe the solution to consider is to take an analogous path to `challengeInFlightExitOutputSpent` vs its `input` version.

We'd then just have `RootChain.piggybackInFlightExitOutput`/`-Input`, which would do the `+4`, and the `4` of that is kept in contract's settings?  CC @paulperegud 

Then the data from `status/` would be usable "as-is" in the contract

pnowosie:
Closes #266 
I'm happy with the resulted API redefinition.

docs/api_specs/includes/informational_api_specs.md
```diff
@@ -0,0 +1,389 @@
+# Watcher - Informational API Service
+
+API specification of the Watcher's Informational Service
+
+
+## Account - Get Balance
+
+```shell
+curl http://localhost:4000/account.get_balance -d '{"address": "b3256026863eb6ae5b06fa396ab09069784ea8ea"}'
```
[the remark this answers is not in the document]
kevsul: Cool. Note that the '-X POST' isn't strictly necessary because curl is smart enough to recognise that it has to be a POST if you're send data. But leaving it in makes it more explicit so I'll put it back everywhere.

kasima:
giphy 1

PR #278, opened 2018-11-23
feat: support 4 in / 4 out transactions

pgebal:


apps/omg_api/lib/state/core.ex
```diff
@@ -168,98 +163,111 @@ defmodule OMG.API.State.Core do
...
+         utxos,
+         %Transaction.Recovered{
+           signed_tx: %Transaction.Signed{raw_tx: raw_tx}
+         } = recovered_tx
        ) do
-    with :ok <- utxo_not_from_the_future_block?(state, blknum) do
-      check_utxo_and_extract_amount(state, Utxo.position(blknum, txindex, oindex), spender, spent_cur)
+    inputs = Transaction.get_inputs(raw_tx)
+
+    with {:ok, input_utxos} <- get_input_utxos(utxos, [], inputs),
```
pdobacz: use `Enum.reduce` instead of rolling your own recurrence?

apps/omg_api/lib/state/core.ex
```diff
@@ -168,98 +163,111 @@ defmodule OMG.API.State.Core do
...
+        end
+        |> Enum.all?()
+
+      if amounts_add_up, do: :ok, else: {:error, :amounts_do_not_add_up}
+    end
   end
 
-  defp same_currency?(utxo_currency, spent_currency) do
-    if utxo_currency == spent_currency, do: :ok, else: {:error, :incorrect_currency}
+  defp same_currencies?(input_amounts, output_amounts) do
```
pdobacz: probably not necessary and would hurt in later developments - what if someone adds an input in a currency solely to cover a tx fee?

Maybe it's just safer to do `amounts_do_not_add_up` in situations when the output currencies aren't covered by inputs.

apps/omg_api/lib/state/core.ex
```diff
@@ -168,98 +163,111 @@ defmodule OMG.API.State.Core do
...
          %Core{height: height, tx_index: tx_index, utxos: utxos} = state,
-         %Transaction{
-           blknum1: blknum1,
-           txindex1: txindex1,
-           oindex1: oindex1,
-           blknum2: blknum2,
-           txindex2: txindex2,
-           oindex2: oindex2
-         } = tx
+         %Transaction{inputs: inputs} = tx
```
pdobacz: use public API.

In addition - maybe `Transaction.get_inputs` should return `Utxo.Position` to not have to handle maps outside `Transaction`.

apps/omg_api/lib/state/core.ex
```diff
@@ -168,98 +163,111 @@ defmodule OMG.API.State.Core do
...
-      state
-      | utxos:
-          utxos
-          |> Map.delete(Utxo.position(blknum1, txindex1, oindex1))
-          |> Map.delete(Utxo.position(blknum2, txindex2, oindex2))
-          |> Map.merge(new_utxos_map)
-    }
+    utxos =
+      inputs
+      |> Enum.reduce(utxos, fn %{blknum: blknum, txindex: txindex, oindex: oindex}, utxos ->
```
pdobacz: as soon as you get rid of the map-representation of utxo-pos you can do `Map.drop`

apps/omg_api/lib/state/core.ex
```diff
@@ -308,7 +317,9 @@ defmodule OMG.API.State.Core do
     db_updates_spent_utxos =
       txs
       |> Enum.flat_map(fn %Transaction.Recovered{signed_tx: %Transaction.Signed{raw_tx: tx}} ->
-        [Utxo.position(tx.blknum1, tx.txindex1, tx.oindex1), Utxo.position(tx.blknum2, tx.txindex2, tx.oindex2)]
+        tx
+        |> Transaction.get_inputs()
+        |> Enum.map(fn %{blknum: blknum, txindex: txindex, oindex: oindex} -> Utxo.position(blknum, txindex, oindex) end)
```
pdobacz: yeah, `Utxo.Position` should be returned `Transaction` functions, and here in `State.Core` we never use anyting else for utxo position

apps/omg_api/lib/state/transaction/recovered.ex
```diff
@@ -24,31 +24,52 @@ defmodule OMG.API.State.Transaction.Recovered do
...
+              spenders: spenders,
               signed_tx: signed_tx
             }}
   end
 
-  defp get_spender(_hash_no_spenders, @empty_signature), do: {:ok, nil}
-  defp get_spender(hash_no_spenders, sig), do: Crypto.recover_address(hash_no_spenders, sig)
+  defp get_spenders(_hash_no_spenders, acc, []), do: {:ok, Enum.reverse(acc)}
+
+  defp get_spenders(hash_no_spenders, acc, [@empty_signature | sigs]) do
```
pdobacz: `Enum.filter(sigs, &not_empty_sig/1) |> Enum.map(&get_spenders/1)`

apps/omg_api/lib/state/transaction/recovered.ex
```diff
@@ -24,31 +24,52 @@ defmodule OMG.API.State.Transaction.Recovered do
...
+  end
+
+  @doc """
+  Checks if everyone of the given spenders signed the transaction
+  """
+  @spec all_spenders_authorized?(t(), list()) :: :ok
+  def all_spenders_authorized?(%__MODULE__{spenders: spenders}, inputs_spenders) do
+    spenders = MapSet.new(spenders)
+    inputs_spenders = MapSet.new(inputs_spenders)
+    same_spenders = MapSet.subset?(spenders, inputs_spenders) and MapSet.subset?(inputs_spenders, spenders)
```
pdobacz: looks like only one of the conditions is required - that all spenders authorized. Condition that there are no "extra" spenders is superfluous.

apps/omg_api/test/api/core_test.exs
```diff
@@ -64,7 +58,7 @@ defmodule OMG.API.CoreTest do
 
   @tag fixtures: [:alice, :bob]
   test "encoded transaction is corrupt", %{alice: alice, bob: bob} do
-    encoded_signed_tx = TestHelper.create_encoded([{1, 2, 3, alice}, {2, 3, 4, bob}], eth(), [{alice, 7}])
+    encoded_signed_tx = TestHelper.create_encoded([alice, bob], [{1, 2, 3}, {2, 3, 4}], [{alice, eth(), 7}])
```
pdobacz: I'm against this change in `create_encoded` - one of its purposes was to be able to pass the owner to sign together with the input utxo_position. It also looks like a loooot of changes... why?

It is also super-hard to check the changes in the tests now.

apps/omg_api/test/api/core_test.exs
```diff
@@ -99,68 +93,37 @@ defmodule OMG.API.CoreTest do
...
-    missing4 =
-      %Transaction.Signed{partial_signed_tx2 | sig2: @empty_signature}
-      |> Transaction.Signed.encode()
-
-    assert {:error, :signature_missing_for_input} == Core.recover_tx(missing1)
-    assert {:error, :signature_missing_for_input} == Core.recover_tx(missing2)
-    assert {:error, :signature_missing_for_input} == Core.recover_tx(missing3)
-    assert {:error, :signature_missing_for_input} == Core.recover_tx(missing4)
+  @tag fixtures: [:alice]
+  test "transaction is not allowed to have input and empty sigs", %{alice: alice} do
```
pdobacz: this test should be reverted to follow the old flow - create a valid tx and then take the sigs away. No need for the new API of `create_signed`

apps/omg_api/test/state/core_test.exs
```diff
@@ -56,24 +60,24 @@ defmodule OMG.API.State.CoreTest do
...
 
   @tag fixtures: [:alice, :bob, :state_empty]
-  test "when spending inputs must have the same currency", %{alice: alice, bob: bob, state_empty: state} do
+  test "input currencies do not match output currencies", %{alice: alice, bob: bob, state_empty: state} do
     state
     |> Test.do_deposit(alice, %{amount: 10, currency: eth(), blknum: 1})
     |> Test.do_deposit(alice, %{amount: 0, currency: not_eth(), blknum: 2})
     |> (&Core.exec(
-          Test.create_recovered([{1, 0, 0, alice}, {2, 0, 0, alice}], eth(), [{bob, 7}, {alice, 3}]),
+          Test.create_recovered([alice], [{1, 0, 0}, {2, 0, 0}], [{bob, eth(), 7}, {alice, eth(), 3}]),
```
pdobacz: please, let's try reverting this change in `create_recovered` API and then reapproach this review.

I'd even opt for having a backwards-compatibility version with `create_recovered/signed(inputs_with_owners, single_currency, outputs)`, just for single-currency transactions, to prevent this storm of changes here.

skipping the review of these test for now

apps/omg_api/test/integration/happy_path_test.exs
```diff
@@ -36,16 +36,16 @@ defmodule OMG.API.Integration.HappyPathTest do
     token: token,
     alice_deposits: {deposit_blknum, token_deposit_blknum}
   } do
-    raw_tx = Transaction.new([{deposit_blknum, 0, 0}], eth(), [{bob.addr, 7}, {alice.addr, 3}])
+    raw_tx = Transaction.new([{deposit_blknum, 0, 0}], [{bob.addr, eth(), 7}, {alice.addr, eth(), 3}])
```
pdobacz: for the record - the update of `Transaction.new`'s API is OK, this is far less frequently invoked and is not a test helper like `create_encoded` et al.

apps/omg_api/test/state/prop_test.exs
```diff
@@ -156,7 +156,7 @@ defmodule OMG.API.State.PropTest do
     end
   end
 
-  @tag capture_log: true
+  # @tag capture_log: true
```
pdobacz: ?

apps/omg_api/test/state/prop_test.exs
```diff
@@ -146,8 +146,8 @@ defmodule OMG.API.State.PropTest do
         (result == :ok)
         |> when_fail(
           (fn ->
-             Logger.info("History: #{inspect(history)}")
-             Logger.error("Result: #{inspect(result)}")
+             Logger.error("History: #{inspect(history, pretty: true)}")
```
pdobacz: leftovers?

apps/omg_watcher/lib/web/controllers/transaction.ex
```diff
@@ -92,19 +90,47 @@ defmodule OMG.Watcher.Web.Controller.Transaction do
...
-      |> Enum.map(fn %{} = output ->
-        output = output |> Enum.into(%{}, fn {k, v} -> {String.to_existing_atom(k), v} end)
-        %{output | owner: OMG.API.Crypto.decode_address!(output.owner)}
-      end)
-    }
+      |> Enum.map(fn %{"currency" => currency} -> currency end)
+      |> Enum.dedup()
+      |> Enum.count()
+
+    cond do
```
pdobacz: these checks and processing look a bit out of place here, why?

apps/omg_watcher/test/challenger/core_test.exs
```diff
@@ -33,23 +33,10 @@ defmodule OMG.Watcher.Challenger.CoreTest do
...
-        cur12: <<0::160>>,
-        newowner1: <<1::160>>,
-        amount1: amount1,
-        newowner2: <<0::160>>,
-        amount2: amount2
-      },
-      sig1: <<0::520>>,
-      sig2: <<0::520>>
-    }
+    raw_tx =
```
pdobacz: hm, why do we use this here? why aren't the existing `creates` sufficient?

apps/omg_watcher/test/eventer/core_test.exs
```diff
@@ -28,9 +28,9 @@ defmodule OMG.Watcher.Eventer.CoreTest do
   @tag fixtures: [:alice, :bob]
```
pdobacz: skipping the review of these test until the test helper thing is resolved

apps/omg_api/lib/state/core.ex
pgebal: There is one use, when we filter inputs by block number only. Matching by Utxo.position is cumbersome, since you need it to write it with placeholders. I know that Paweł T. also has a use similar use case.

pdobacz: can you point me to that?

apps/omg_api/test/api/core_test.exs
pgebal: I don't like that either, but the thing is that now, we don't care about owner of a particular input utxo. We just care if each input owner signed the transaction. 
For example that:
```
TestHelper.create_encoded([{1, 2, 3, alice}, {2, 3, 4, bob}], [{alice, eth(), 7}])
```
creates a transaction that is valid even if bob is the owner of {1,2,3} and alice is the owner of {2,3,4}.

pdobacz: so, this is exactly why the test helper is here - someone who writes a test doesn't care about how signatures are created/verified - alice spends - she signs, bob spends - he signs. 

I still think it is better to not have this huge change in the test because of that. All the changes should be hidden within the test helper. You can for example `dedup` the signatures there, or you can support providing inputs without the `, alice}`, as a backwards-compatible extension of the helper.

apps/omg_api/lib/state/core.ex
whoisjeremylam: Sorry to jump into the conversation.

When we first launch, it's likely we will only accept a very restricted list of currencies for fees (maybe only ETH to begin with). So it's very likely that if a user is sending an ERC20 token, that they will need to add an input to pay the fee.

apps/omg_api/lib/state/transaction/recovered.ex
pgebal: We need to reduce to get a proper error message. I think this rec function is ok, but I can change that to reduce if you think that's more readable.

pdobacz: ah ok right. I wish there was an `Enum.` function to handle that in a one-liner.

alternative is

```
result = Enum.filter(sigs, &not_empty_sig/1) |> Enum.map(&get_spenders/1)

case Enum.find(result, check if error) do
  nil -> result
  result -> result
end
```
or similar. I have some preference for avoiding recurrence.

pgebal: ok

apps/omg_watcher/lib/web/controllers/transaction.ex
pgebal: Since deeper in the code we allow for multiple currencies I think that's the right place to put those validation. If we would not validate that here, we would break the API of this endpoint.

pdobacz: `State.Transaction.create_from_utxos` seems like the right place, and that comes before `parse...`, so why not there?

pdobacz: wait, for example: `:currency_mixing_not_possible` is actually thrown there so this is a duplicate check, so is `at_least_one_input_required`. Am I missing something?

pgebal: You are right. My mistake.

pgebal: It turned out I actually need to validate number of inputs right there in the controller. I need to get a currency. In endpoint API currency is tied to inputs. I need to extract it from inputs and put into outputs.

apps/omg_api/lib/state/core.ex
```diff
@@ -168,98 +164,105 @@ defmodule OMG.API.State.Core do
...
+  defp amounts_add_up?(input_amounts, output_amounts, fees) do
+    outputs_covered =
+      for {output_currency, output_amount} <- Map.to_list(output_amounts) do
+        fee = Map.get(fees, output_currency, 0)
+        input_amount = Map.get(input_amounts, output_currency, 0)
+        input_amount >= fee + output_amount
+      end
+      |> Enum.all?()
+
+    fees_covered =
```
pdobacz: ah I see the point with fees better now. This is a bit that's a little underdefined on the requirements end.

In this implementation, the fee must be paid for every currency that pops up in input or output.

That's probably not what we intend - most likely we want at least one currency of input to cover the fee. This way the user chooses which currency to pay with.

But we can amend it (and document!) in a sequel PR to not delay this. Please loop in @kasima and/or @whoisjeremylam to confirm on the desired behavior.

apps/omg_api/lib/state/transaction.ex
```diff
@@ -67,53 +62,51 @@ defmodule OMG.API.State.Transaction do
...
-      inputs
+      (input_utxos ++ outputs)
       |> Enum.map(& &1.currency)
-      |> Enum.uniq()
+      |> Enum.dedup()
 
-    if match?([_], currencies),
-      do: {:ok, currencies |> hd()},
-      else: {:error, :currency_mixing_not_possible}
+    # NOTE we support one currency
```
pdobacz: do we still want this check?

apps/omg_api/test/api/core_test.exs
```diff
@@ -30,28 +30,27 @@ defmodule OMG.API.CoreTest do
...
-          [input1, input2] |> Enum.map(fn {blknum, txindex, oindex, _} -> {blknum, txindex, oindex} end),
-          eth(),
-          [{alice, 7}, {bob, 3}] |> Enum.map(fn {newowner, amount} -> {newowner.addr, amount} end)
-        )
+        inputs
+        |> Enum.map(fn {blknum, txindex, oindex, _} -> {blknum, txindex, oindex} end)
+        |> Transaction.new([{alice.addr, eth(), 7}, {bob.addr, eth(), 3}])
 
-      encoded_signed_tx = TestHelper.create_encoded([input1, input2], eth(), [{alice, 7}, {bob, 3}])
+      encoded_signed_tx = TestHelper.create_encoded(inputs, [{alice, eth(), 7}, {bob, eth(), 3}])
```
pdobacz: what about https://github.com/omisego/elixir-omg/pull/278#discussion_r236172001 ? that was the main issue previously - we want to not have all this diff in tests.

apps/omg_api/lib/state/transaction.ex
pgebal: I kept it to not break convenience API.

pdobacz: ah ok, because despite the flexible tx support, we still have single-currency transactions, gotcha

apps/omg_api/test/api/core_test.exs
pgebal: Sorry, I applied it only to spenders.

apps/omg_api/lib/state/core.ex
pdobacz: this still doesn't behave as expected I think. 1/ it requires all fees to be paid up for all inputs 2/ it additionally has a `fees_covered` condition which is a list that's treated as a bool, which checks additionally, if there any excess inputs also cover the fee.

Are there `Core` tests that cover these behaviors so we can see how it works?

apps/omg_api/test/state/core_test.exs
```diff
@@ -748,6 +775,54 @@ defmodule OMG.API.State.CoreTest do
...
+    |> (&Core.exec(
+          Test.create_recovered([{1, 0, 0, alice}, {2, 0, 0, alice}], [{bob, eth(), 1}, {bob, not_eth(), 2}]),
+          zero_fees_map(),
+          &1
+        )).()
+    |> success?
+  end
+
+  @tag fixtures: [:alice, :bob, :state_empty]
+  test "respects fees for transactions with mixed currencies", %{
```
pdobacz: ah, there's the fee-test, so the behavior is not like the suggested (fee is covered in at least one currency).

Make sure that you test also situations like when fee is covered in `eth()` but `not_eth()` leg of the tx doesn't even have enough to cover the output, to ensure that fee handling doesn't surprisingly break the "basic" validity.

apps/omg_api/test/state/prop_test.exs
```diff
@@ -156,6 +156,7 @@ defmodule OMG.API.State.PropTest do
     end
   end
 
+  @tag skip: true
```
pdobacz: briefly explain why and mention OMG-351

PR #279, opened 2018-11-23
[elixir-omg] Add support for zero touch Childchain Linux deployments

jbunce:
Adds support for zero touch deployments of the Childchain on Linux. Mac deployments are possible but you need to hack the Ethereum client before building the container via:

```
sed -i 's/localhost:8545/docker.for.mac.localhost:8545/g' ./apps/omg_performance/lib/performance.ex
sed -i 's/localhost:8545/docker.for.mac.localhost:8545/g' ./deps/ethereumex/config/test.exs
sed -i 's/localhost:8545/docker.for.mac.localhost:8545/g' apps/omg_eth/config/config.exs
```

For Linux:

```
docker build -t plasma . && docker run -ti -p9656:9656 -e PLASMA_SERVICE='CHILDCHAIN' -e PLATFORM='MAC' plasma
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,4 +76,4 @@ RUN apt-get purge -y
 
 USER plasma
 
-ENTRYPOINT ["/bin/bash"]
+ENTRYPOINT ["./launcher.py"]
--- a/contracts/requirements.txt
+++ b/contracts/requirements.txt
@@ -1,3 +1,5 @@
+gitpython==2.1.11
 py-solc==3.1.0
 py-solc-simple==0.0.10
+requests==2.20.1
 semantic-version==2.6.0
--- a/deploy_and_populate.sh
+++ b/deploy_and_populate.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+mix run --no-start -e  '
+   contents = OMG.Eth.DevHelpers.prepare_env!() |> OMG.Eth.DevHelpers.create_conf_file()
+   "~/config.exs" |> Path.expand() |> File.write!(contents)
+ '
--- a/launcher.py
+++ b/launcher.py
@@ -1,0 +1,171 @@
+#!/usr/bin/python3
+##!/usr/local/bin/python3
+import json
+import logging
+import os
+import subprocess
+import sys
+
+import git
+import requests
+
+
+class ChildchainLauncher:
+    ''' ChildchainLauncher: A module to launch a Childchain service
+    '''
+    def __init__(self, git_commit_hash: str, platform: str):
+        self.git_commit_hash = git_commit_hash
+        self.platform = platform
+
+    def start(self):
+        ''' Start the launch process for the Childchain service
+        '''
+        logging.info('Service type to launch is Plasma Childchain')
+        logging.info(
+            'Starting launch process for build {}'.format(self.git_commit_hash)
+        )
+        self.ethereum_client = check_ethereum_client(self.platform)
+        logging.info('Ethereum client is {}'.format(self.ethereum_client))
+        if self.compile_application() is False:
+            logging.critical('Could not compile application. Exiting.')
+            sys.exit(1)
+        if self.deploy_plasma_contract() is False:
+            logging.critical('Contract not deployed. Exiting.')
+            sys.exit(1)
+        if self.initialise_childchain_database() is False:
+            logging.critical('Could not initialise database. Exiting.')
+            sys.exit(1)
+        self.start_childchain_service()
+
+    def compile_application(self) -> bool:
+        ''' Execute a mix compile
+        '''
+        result = subprocess.run(['mix', 'compile'], stdout=subprocess.DEVNULL)
+        if result.returncode == 0:
+            logging.info('Elixir mix compile successful')
+            return True
+        return False
+
+    def deploy_plasma_contract(self) -> bool:
+        ''' Deploy the smart contract and populate the ~/config.exs file
+        '''
+        result = subprocess.run(
+            ['./deploy_and_populate.sh'], stdout=subprocess.PIPE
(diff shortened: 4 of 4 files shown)


jbunce:
```
2018-11-23 12:33:42,531 INFO    :Starting Launcher
2018-11-23 12:33:42,545 INFO    :Service type to launch is Plasma Childchain
2018-11-23 12:33:42,546 INFO    :Starting launch process for build 752b7c8e96b0ad11e9e71dac378e4b3272e67962
2018-11-23 12:33:42,563 INFO    :Connected to the Ethereum client
2018-11-23 12:33:42,564 INFO    :Ethereum client is b'{"jsonrpc":"2.0","id":67,"result":"Geth/v1.8.14-stable/darwin-amd64/go1.10.3"}\n'
2018-11-23 12:33:47,585 INFO    :Elixir mix compile successful
2018-11-23 12:33:59,166 INFO    :Smart contract deployed and authority address populated
2018-11-23 12:34:03,683 INFO    :Childchain database initialised
b'Interactive Elixir (1.6.6) - press Ctrl+C to exit (type h() ENTER for help)'
```

Dockerfile
```diff
@@ -76,4 +76,4 @@ RUN apt-get purge -y
 
 USER plasma
```
pdobacz: I think we should not use `plasma` through out this code. "plasma" is a design framework.

Terms which should be used is:
 - child chain - for the name of the blocks/ledger/data
 - child chain server - for the application "running" the child chain 
 - operator - the entity that's running the child chain server - which is what probably fits in this particular spot.

There are a couple of occurrences of "plasma" throughout this PR

deploy_and_populate.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
pdobacz: I'd like 2 comments put in this file here:
 - why is this different from other `mix` incantations which went inline into `launcher.py` - unless the reason I'm missing is obvious
 - a loud remark that `DevHelpers` - and this by consequence - is not intended for production (a `# TODO` ideally)

launcher.py
```diff
@@ -0,0 +1,173 @@
+#!/usr/bin/python3
```
pdobacz: I'd like some brief description of the intention and scope of this script. What it adds on top of the `mix` commands it runs

Also make reference to it and explain in `README.md` - an indication when should I use this and when should I rather use Vagrant and when should I just run the `mix` commands in shell would be useful.

If the sole intended client of `launcher.py` is the `Dockerfile`, I'd rather have all these files in a subdir `docker` or sth.

launcher.py
```diff
@@ -0,0 +1,173 @@
...
+            'Could not initialise the database. Error: {}'.format(
+                result.stdout
+            )
+        )
+        return False
+
+    def start_childchain_service(self):
+        ''' Start the childchain service
+        '''
+        os.chdir(os.getcwd() + '/apps/omg_api')
```
pdobacz: FYI, the `cd` is probably going away soon (https://github.com/omisego/elixir-omg/pull/274). We will be updating the `README.md` and let you know.

launcher.py
```diff
@@ -0,0 +1,173 @@
...
+        process = subprocess.Popen(
+            [
+                "/usr/local/bin/iex ", "-S ", "mix", "run", "--config",
```
pdobacz: pls be aware that this starts the REPL (`iex`)

contracts/requirements.txt
```diff
@@ -1,3 +1,5 @@
+gitpython==2.1.11
```
pdobacz: this is `requirements.txt` to use for contract compiling. Not sure if these should be mixed.

Dockerfile
jbunce: That wasn't introduced via this PR.

That being said, I'm happy to modify it in separate change if you feel it should be changed. I think you've slightly misunderstood though, this is for the user of the running container so that we don't run the application as root. I don't want to follow that naming convention for creating Linux system users. It would result in three separate Dockerfiles for running the same code, which is very inefficient.

I propose elixir-user, or similar.

launcher.py
jbunce: It adds nothing to the `mix` commands. It executes them to enable start up of the service. At the moment we cannot start the chain manually, nor can we recover from a failed chain. This is a work in progress, and as such the `README` should not be updated. I'm checking in smaller PRs so that you don't have a PR from hell to review.

I'm not sure on the benefit of moving `launcher.py` to a subdir. Additionally, it will require more code to be written to prevent path problems when trying to execute code.

launcher.py
jbunce: Thanks, good to hear.

launcher.py
jbunce: From https://github.com/omisego/elixir-omg#setting-up-a-child-chain-server-a-developer-environment, `iex -S mix run --config ~/config.exs` is the command to start the chain?

Dockerfile
pdobacz: `elixir-user` is good of course, I'm just after cleaning up the "plasma-terminology" here, so anything that doesn't touch this is good.

I'd clean up now however, since more `plasma` occurrences are introduced in this PR.

And sorry about late noticing, I know this's been around since before this PR.

launcher.py
pdobacz: if it adds nothing, except a convenience to just do everything in one `Dockerfile` line, then maybe we should consider this temporary and look towards substituting this with `Mix.Tasks` which we have somewhere in plans. Unless python specifically will provide you with some tools not available from Mix.

the subdir benefit would be hiding that and marking it is used specifically by the docker setup. But if that's cumbersome then ok.

PR #280, opened 2018-11-23
Add Slate to repo

kevsul:
Proposal for including all the Slate code so that we can build the API documentation directly from here.

Note that due to Slate's `includes` directory structure, we'll have to move the 0*_specs.md files into an  `includes` directory on `develop` as well.


pnowosie:
I think I got it. So Slate generator files will be sitting in `docs/api_specs` in `gh-pages` branch.
Therefore we no longer need the numbers preceding specs file names because we control the order in the index file.

...and maybe we don't need index & includes/*.md files here but only in develop branch to avoid possible merge conflicts. Then we will be commiting just generated html here.

What do you think?

kevsul:
I think the issue is on which branch do we want to run `bundle exec middleman build ...` to generate the html files. 
If we want to do it on the `gh-pages` branch, then the `*.md` files have to be on that branch. I don't think we'll get merge conflicts though - those files should never change on the `gh-pages` branch, so it should always be a fast-forward merge out from `develop`.

The other option is to keep everything on `develop` (including the Slate lib and src files) and build there and then merge the generated html files out to `gh-pages`, but I don't think that will work because the generated html files will be local files on the `develop` branch, but not on `gh-pages`

Can you think of a better way?

pnowosie:
So let me rephrase, **I think we are aligned with most concerns here, and I like this proposition**.
What I'm proposing is:
 * keep `index.html.md` and `includes\*.md` files only on `develop` - they won't be changed in `gh-pages` so anyway they belong to `develop` and we are free to rename them and so... (probably someone will commit these files to `gh-pages` soon by mistake... :laughing:)

Slate generator code and generated htmls will be (as proposed) on `gh-pages` and that's the right place where generation process will be run. 
**I like** to have just specs markdowns on `develop` to being changed along with the feature which changes the API itself.
**I don't like** to have these markdowns on `gh-pages` just to make clear our intention that we only generating specs from development branch. IMHO the fact their are required by generation process but their are missing there make this intention clear :confused: ...and we are free to add/rename/delete do whatever to these files not worried about the `gh-pages`

PR #283, opened 2018-11-27
OMG-308 docs: add the whiteboarded diagram + description

pdobacz:
note the Diagram linked inside the md file

### DIFF
--- a/docs/exit_validation.md
+++ b/docs/exit_validation.md
@@ -8,6 +8,24 @@ NOTE:
 * `SLA` Service Level Agreement
 * `sla_margin` margin of service lever agreement validation (in Ethereum blocks)
 
+## Notes on the Child Chain Server
+
+This document focuses on the Watcher.
+
+For completeness we give a quick run-down of the rules followed by the Child Chain Server, in terms of processing exits sent on the root chain contract.
+
+1. Child Chain operator's objective is to pro-actively minimize the risk of chain becoming invalid.
+The risk is that this can happen if any exit gets finalized despite being invalid.
+2. To satisfy this objective:
+    - the child chain server listens to every `ExitStarted` event and immediately "spends" the exited utxo, preventing exit invalidation
+    - the child chain server listens to every `InFlightExitStarted` event and immediately "spends" the exiting tx's **inputs**
+    - the child chain server listens to every `InFlightExitPiggybacked` event (on outputs) and immediately "spends" the piggybacked outputs - as long as the IFEing tx has been included in the chain and the output exists
+
+There are scenarios, when a race condition/reorg on the root chain might make the Child Chain Server block spending of a particular UTXO **too late**, regardless of the immediacy mentioned above.
+This is OK, as long as the delay doesn't exceed a predetermined `sla_margin`.
+
+## Standard Exits
+
 ### The purpose of having exit validation as a separate service is to:
 1. proactively prevent user from losing money
     - prohibit user from spending on a dangerous chain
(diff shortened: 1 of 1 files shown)


docs/exit_validation.md
```diff
@@ -87,3 +105,88 @@ All the general rules will apply in the MoreVP world.
...
+
+With MoreVP, we need to handle another type of exit game, which is the in-flight exit game, as specced out [here](docs/morevp.md).
+
+In terms of handling within the Watcher, similar principles will apply:
+  - we gather and keep in `OMG.Watcher.ExitProcessor`'s persistent state the current state of in-flight txs and exits
+  - we periodically check the validity of this state, emitting events and allowing for actions as necessary
+  - we touch `OMG.API.State` only when the in-flight exit finalizes
+  - if some invalid IFE or piggyback runs unchallenged for too long, it should be a prompt to exit (ala `:unchallenged_exit` above)
+
+**[Diagram](https://docs.google.com/drawings/d/1UaAMZTJBbikTM0eFSbNM7ZVgi1HZrLTfwYCUvHGUaW0/edit?usp=sharing)** illustrates the flows described below (with the addition of handling on the Child Chain Server side for comparison).
```
paulperegud: Minor problem with "spend all inputs" block on ch-ch side.
tx1 = [[A, B],[C]], honest spend by A, not included somehow. IFE by B.
tx2 = [[B],[D]], double spend by B. IFE by B.
Now A can be only piggybacked, it can't be spend on plasma chain anymore because of "spend all inputs".

paulperegud: There is an attack vector which opens with "outputs mine" block in the watcher side of the diagram. Watcher also needs to check if inputs are in correct block which it has access to. Only in such case outputs can be exited at all.

If watcher does not perform such check, operator can use it as a part of his own DoS of Ethereum. By creating IFE [[input_from_withhold_block],[[A1],[A2],[A3],[A4]]] operator can provoke four Alices to do a piggyback, amplifying it's own gas investment, creating a DDoS out of his own DoS.

docs/exit_validation.md
```diff
@@ -87,3 +105,88 @@ All the general rules will apply in the MoreVP world.
...
+🕐 Periodically (same as with standard exits), `ExitProcessor` should:
+  - find invalid piggybacks and do something
+  - find non-canonical transactions that are seen as canonical in the root chain contract
+  - find canonical transactions that are seen as non-canonical in the root chain contract
+
+#### Checking if I should piggyback
+
+Any IFE started might be one such that the user should piggyback onto.
+
+This happens when an input or an output of the in-flight transaction submitted mentions the user as owner.
```
paulperegud: As above - outputs are only worth piggybacking if inputs are in non-withheld correct block.

docs/exit_validation.md
paulperegud: There is an attack vector which opens with "outputs mine" block in the watcher side of the diagram. Watcher also needs to check if inputs are in correct block which it has access to. Only in such case outputs can be exited at all.

If watcher does not perform such check, operator can use it as a part of his own DoS of Ethereum. By creating IFE [[input_from_withhold_block],[[A1],[A2],[A3],[A4]]] operator can provoke four Alices to do a piggyback, amplifying it's own gas investment, creating a DDoS out of his own DoS.

pdobacz: but A has had the intention to spend, so ch-ch is (in a way) justified to drop the `A` from spendable. `B` is attacking `A` in this situation, and will get punished with at least one `exit_bond`.

not sure how sever it is. An alternative which would potentially heal this is when `spend all inputs` happens only after the Phase 1 passes, and the tx is deemed canonical. But that might open another can of worms.

docs/exit_validation.md
pdobacz: I'm not sure I follow.

Inputs come with inclusion proofs on the root chain contract, so even if they were withheld, they become "unwithheld" at the moment of `startInFlightExit`.

And even if they ended up being in an incorrect block, the drill should still be the same - IFE is justified for those interested, and they should all piggyback + a mass exit independently.

docs/exit_validation.md
pdobacz: pasting a richer description by @paulperegud: (let's continue here)

> 
> There is an attack vector which opens with "outputs mine" block in the watcher side of the diagram. Watcher also needs to check if inputs are in correct block which it has access to. Only in such case outputs can be exited at all.
> 
> If watcher does not perform such check, operator can use it as a part of his own DoS of Ethereum. By creating IFE [[input_from_withhold_block],[[A1],[A2],[A3],[A4]]] operator can provoke four Alices to do a piggyback, amplifying it's own gas investment, creating a DDoS out of his own DoS.

PR #285, opened 2018-12-03
Update exit_validation.md

kevsul:
Fix some typos and change wording to (I think) clarify some points.

### DIFF
--- a/docs/exit_validation.md
+++ b/docs/exit_validation.md
@@ -108,7 +108,7 @@ Absence of challenges within some period (like `sla_margin`) must result in clie
 
 ## In-flight exits
 
-With MoreVP, we need to handle another type of exit game, which is the in-flight exit game, as specced out [here](docs/morevp.md).
+With MoreVP, we need to handle another type of exit game, which is the in-flight exit game, as specced out [here](morevp.md).
 
 In terms of handling within the Watcher, similar principles will apply:
   - we gather and keep in `OMG.Watcher.ExitProcessor`'s persistent state the current state of in-flight txs and exits
--- a/docs/exit_validation.md
+++ b/docs/exit_validation.md
@@ -14,52 +14,52 @@ This document focuses on the Watcher.
 
 For completeness we give a quick run-down of the rules followed by the Child Chain Server, in terms of processing exits sent on the root chain contract.
 
-1. Child Chain operator's objective is to pro-actively minimize the risk of chain becoming invalid.
-The risk is that this can happen if any exit gets finalized despite being invalid.
+1. The Child Chain operator's objective is to pro-actively minimize the risk of chain becoming invalid.
+The Child Chain will become invalid if any invalid exit gets finalized.
 2. To satisfy this objective:
-    - the child chain server listens to every `ExitStarted` event and immediately "spends" the exited utxo, preventing exit invalidation
-    - the child chain server listens to every `InFlightExitStarted` event and immediately "spends" the exiting tx's **inputs**
-    - the child chain server listens to every `InFlightExitPiggybacked` event (on outputs) and immediately "spends" the piggybacked outputs - as long as the IFEing tx has been included in the chain and the output exists
+    - the Child Chain server listens to every `ExitStarted` Root Chain event and immediately "spends" the exited utxo. This blocks the user from spending that utxo, thus preventing exit invalidation.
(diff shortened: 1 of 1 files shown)


docs/exit_validation.md
```diff
@@ -14,52 +14,52 @@ This document focuses on the Watcher.
...
-1. The exit is processed after the tx, so it is known to be invalid from the start. Should emit an `:invalid_exit` event.
-2. The exit is processed before the tx, but the tx is within `sla_margin` period, so the child chain is valid.
+1. The exit is processed *after* the tx that invalidates it, so it is known to be invalid from the start. Should emit an `:invalid_exit` event.
+2. The exit is processed *before* the tx that invalidates it and the tx is within `sla_margin` period, so the child chain is still valid.
 Causes an `:invalid_exit` event.
-3. The exit is processed before the tx, but tx is after `sla_margin` period, so the child chain is byzantine.
+3. The exit is processed *before* the tx that invalidates it, but the tx occurs *after* `sla_margin` period, so the child chain is byzantine.
 Causes to emit an `:unchallenged_exit` event.
-3. An exit is unchallenged within `sla_margin` period, so the child chain is jeopardized.
+3. An invalid exit is still unchallenged after `sla_margin` period, so the child chain is jeopardized.
```
kevsul: This is the opposite of what was there originally (i.e. the exit is unchallenged before the `sla_margin` is up). Have I misunderstood it?

docs/exit_validation.md
```diff
@@ -14,52 +14,52 @@ This document focuses on the Watcher.
...
+    - because our child chain implementation will never get close to this happening, so it is a symptom of a (subtle) hack attempt on the child chain server.
     By "subtle" we mean that it is a hack that counts on getting away with exit invalidation being submitted too late, rather than on just dropping a huge invalid UTXO and attempting to exit from that
-3. If an exit is invalid and unchallenged within a short period from it's finalization it must be challenged **AND** watcher must not allow spending and depositing
-    - because otherwise, we are "tightly" following the minimum safety requirements
-    - because otherwise, unchallenged exits may break the chain without honest holders having time to escape.
-    This is the case when an exit, a spend invalidating it and exit of that coincide and are unchallenged.
-4. (optional) If a finalization of an invalid exit occurs, watcher must prompt an exit without guarantees of recovery of funds
+3. If an exit is invalid and remains unchallenged within a short period of time from its finalization, it must be challenged **AND** watcher must not allow spending and depositing, because:
```
kevsul: Question: How can the Watcher prevent deposits? I didn't think the Watcher was involved with deposits at all.

docs/exit_validation.md
pdobacz: ah right, so this document follows the "end goal" approach, where exits become "unchallenged" when they enter a time period calculated based on finalization (`finalization - sla_margin`). So "still unchallenged within `sla_margin` from finalization" maybe?

But mind that implementation-wise, we're technically reaching this goal by calculating based on initiation (`startExit + sla_margin`). This is almost equivalent, and we have a future note for ourselves to amend that.

Regardless, I think the doc should state the former (`finalization - sla_margin`) consistently, If the new wording reads differently then this needs to be reverted

docs/exit_validation.md
pdobacz: it will be. All ethereum-txs making the deposits should be routed through the Watcher which will not allow that if it is unsafe, see OMG-119

docs/exit_validation.md
kevsul: Ok, makes sense but I think in some places it reads as if the sla_margin is counted ahead from when the exit is started.

I'll revise it again, but to be clear:
`sla_margin` is a number of ethereum blocks prior to finalization.

For example, if finalization is 10 blocks and sla_margin is 3 blocks, then it is possible to challenge an invalid exit for up to 7 blocks after it has started, but after that it becomes 'unchallenged' and the user should exit.

Correct?

pdobacz: Yes, this is the intended behavior, currently implemented the converse way around.

Just be careful about a related thing from your comment:
It is still possible to challenge an invalid exit after `unchallenged`. Challenge must come before finalization. (Ah I see I worded that in a confusing way: `finalization` and `scheduled finalization time` could have been differentiated - Challenge must come before finalization, but we substract `sla_margin` from `scheduled finalization time (SFT)` c.f. Tesuji design doc).

It is just that after (whichever way around) some threshold moment had passed, exit is prompted, regardless of challenges etc.

PR #289, opened 2018-12-04
feat: be more verbose about skipping form_block and its reasons

pdobacz:
Small improvement in `:debug` logging to tell us why the child chain would refuse to submit blocks.

### DIFF
--- a/apps/omg_api/lib/block_queue/core.ex
+++ b/apps/omg_api/lib/block_queue/core.ex
@@ -333,7 +333,24 @@ defmodule OMG.API.BlockQueue.Core do
          },
          is_empty_block
        ) do
-    parent_height - last_enqueued_block_at_height > minimal_enqueue_block_gap and !wait_for_enqueue and !is_empty_block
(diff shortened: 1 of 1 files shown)


apps/omg_api/lib/block_queue/core.ex
```diff
@@ -333,7 +333,23 @@ defmodule OMG.API.BlockQueue.Core do
          },
          is_empty_block
        ) do
-    parent_height - last_enqueued_block_at_height > minimal_enqueue_block_gap and !wait_for_enqueue and !is_empty_block
+    it_is_time = parent_height - last_enqueued_block_at_height > minimal_enqueue_block_gap
+    should_form_block = it_is_time and !wait_for_enqueue and !is_empty_block
+
+    if !should_form_block do
+      log_data = %{
```
pnowosie: Is this `log_data` used for logging only ❓ If so please move it inside `Logger.debug` f-ction. We will avoid memory allocation in higher log levels

PR #292, opened 2018-12-05
OMG-346 docs: add known issue of retries/timeouts to morevp.md

pdobacz:
also expands and tidies the alicebobbings somewhat

### DIFF
--- a/docs/morevp.md
+++ b/docs/morevp.md
@@ -363,6 +363,20 @@ The following scenario demonstrates this attack:
 Mallory has therefore caused Bob to lose `exit bond`, even though Bob was acting honestly.
 We want to mitigate the impact of this attack as much as possible so that this does not prevent users from receiving funds.
 
+#### Honest transaction retries attack
+
+Retrying a transaction that has failed for a trivial reason is not safe under MoreVP.
+
+Scenario is:
+1. Honest Alice creates/signs/submits a transaction `tx1`
+2. This fails, either tacitly or loudly - `tx1` doesn't get included in a block
+3. Alice is forced to in-flight exit, even if she just made a trivial mistake (e.g. incorrect fee)
+4. If instead Alice retries with amended `tx2`, then she opens an attack on her funds:
+    - if the child chain is nice, `tx2` will get included in a valid, non-withheld block, all is good
+    - if the child chain decides to go rogue, Alice is left defenseless, because she double-spent her input, i.e. she can't in-flight exit neither `tx1` nor `tx2` anymore
+
+See [Timeouts section](docs/morevp.md#Timeouts) for discussion on one possible mitigation.
+However, due to uncertainty of timeouts in MoreVP, other mitigations for the retry problem might be necessary.
 
 ##### Mitigations for Honest Exit Bond Slashing
 
@@ -386,11 +400,14 @@ Our system requires only a single node be properly incentivized to challenge, an
 Modeling the “correct” size of the exit bond is an ongoing area of research.
 
 
-###### Timeouts
+##### Timeouts
 
-We can add timeouts to each transaction (“must be included in the chain by block X”) to reduce number of vulnerable transactions at any point in time.
+We can add timeouts to each transaction (“must be included in the chain by block X”) to
+ - reduce number of transactions vulnerable to [**Honest Exit Bond Slashing**](docs/morevp.md#Honest-Exit-Bond-Slashing) point in time.
+ - alleviate [**Honest transaction retries attack**](docs/morevp.md#Honest-transaction-retries-attack), allowing Alice to just wait the timeout and retry
 This will probably also be necessary from a user experience point of view, as we don’t want users to accidentally sign a double-spend simply because the first transaction hasn’t been processed yet.
 
+**TODO** At this point, it is uncertain how the timeouts scheme would modify MoreVP and whether it's feasible at all.
 
 ## Appendix
 
--- a/docs/morevp.md
+++ b/docs/morevp.md
@@ -342,6 +342,23 @@ Bob’s exit will have priority of position of `UTXO1`.
 11. Bob receives the value of `UTXO2` first, Operator receives the value of `UTXO4` second (ideally contract is empty by this point).
 All bonds are refunded.
 
+### Honest receiver should not start in-flight exits
+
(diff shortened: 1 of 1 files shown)


docs/morevp.md
```diff
@@ -342,6 +341,43 @@ Bob’s exit will have priority of position of `UTXO1`.
 11. Bob receives the value of `UTXO2` first, Operator receives the value of `UTXO4` second (ideally contract is empty by this point).
 All bonds are refunded.
 
+### Operator tries to steal funds from a multi-input in-flight transaction.
+
+1. Alice spends `UTXO1a`, Malory spends `UTXO1m` in `TX1` to Bob, creating `UTXO2`.
+2. `TX1` is in-flight.
+3. Operator creates invalid deposit, creating `UTXO3`.
+4. Operator spends `UTXO3` in `TX3`, creating `UTXO4`.
+5. Operator starts an exit referencing `TX3` and places `exit bond`.
```
pnowosie: Regarding `in-flight` exit - operator needs to prove inclusion of each input in the chain. Have you had in mind that `UTXO3` is an output created in transaction spending mentioned invalid deposit (not a deposit itself). I think this is some mental shortcut used here - as operator could not prove inclusion of `UTXO3` therefore his exit is not accepted (:question:)

pnowosie: Oh and the previous case shares the same drawback

docs/morevp.md
```diff
@@ -351,18 +387,32 @@ It’s possible for an honest user to start an exit and have their exit bond sla
 This can occur if one of the inputs to a transaction is malicious and signs a second transaction spending the same input.
 
 The following scenario demonstrates this attack:
-1. Mallory spends `UTXO1` in `TX1` to Bob, creating `UTXO2`.
+1. Mallory spends `UTXO1m` and Alice spends `UTXO1a` in `TX1` to Bob, creating `UTXO2`.
 2. `TX1` is in-flight.
 3. Operator begins withholding blocks while `TX1` is still in-flight.
-4. Bob starts an exit referencing `TX1` and places `exit bond`.
-5. Mallory spends `UTXO1` in `TX2`.
+4. Alice starts an exit referencing `TX1` and places `exit bond`.
```
pnowosie: It's not relevant to the case, but we should mention here that `Alice` also piggybacks to her input and puts `pg-bond`. This bond will be refunded to her - but I think it makes the scenario more clear and logical... Here `Alice` has to piggyback to the input in `period 1` otherwise it will be lost, so the `Mallory` can wait as long as she piggybacks and then reveals the competitor.

docs/morevp.md
```diff
@@ -351,18 +387,32 @@ It’s possible for an honest user to start an exit and have their exit bond sla
...
+Mallory has therefore caused Alice to lose `exit bond`, even though Alice was acting honestly.
 We want to mitigate the impact of this attack as much as possible so that this does not prevent users from receiving funds.
 
+#### Honest transaction retries attack
+
+Retrying a transaction that has failed for a trivial reason is not safe under MoreVP.
+
+Scenario is:
+1. Honest Alice creates/signs/submits a transaction `tx1`
+2. This fails, either tacitly or loudly - `tx1` doesn't get included in a block
```
pnowosie: `either tacitly or loudly` - it might be just me but don't get it - can you rephrase please.

docs/morevp.md
pdobacz: I think by "invalid deposit", a transaction that looks like this is meant:
```
inputs: []
outputs: [{operator, token, bazilion} <~~~~UTXO3]
```
it is included by the operator in a child chain block and its inclusion can be proven.

Implementation-wise it is not a deposit, but it protocol-wise it is - a transaction with no inputs.

As a consequence, `TX3` is a "valid" transaction, from point of view of IFE

docs/morevp.md
pdobacz: it means either an error response will be returned from the child chain server or just silence. In both situations Alice is in trouble. I'll clarify

docs/morevp.md
```diff
@@ -391,10 +391,11 @@ The following scenario demonstrates this attack:
...
 3. Operator begins withholding blocks while `TX1` is still in-flight.
 4. Alice starts an exit referencing `TX1` and places `exit bond`.
+4. Alice piggybacks onto the exit and places `piggyback bond`.
 5. Mallory spends `UTXO1m` in `TX2`.
 6. In period 1 of the exit for `TX1`, Mallory challenges the canonicity of `TX1` by revealing `TX2`.
 7. No one is able to respond to the challenge in period 2, so `TX1` is determined to be non-canonical.
-8. After period 2, Mallory receives Alice's `exit bond`, no one exits any UTXOs.
+8. After period 2, Mallory receives Alice's `exit bond`, Alice receives `UTXO1a` and `piggyback bond`.
 
 Mallory has therefore caused Alice to lose `exit bond`, even though Alice was acting honestly.
```
pnowosie: Can we add a comment that the Mallory is real looser here because she basically burned her `UTXO1m`... This will help to highlight that this attack comes at cost.

docs/morevp.md
pnowosie: > I think by "invalid deposit", a transaction `inputs: []`

I believe it's still not possible to start `in-flight` exit from `TX3` because it contradicts the `exitability` property then:

> transaction can be called “exitable” if the transaction is correctly formed (e.g. more input value than output value,

docs/morevp.md
pdobacz: Hm, you're right, with a few caveats:
1/ this is conditional on the fact, that none of the double-spent inputs can be piggybacked and exited. I think this piece of the protocol is not to depend on and there's a slight chance that letting exactly one of such piggybacks out will be useful elsewhere.
2/ Mallory could craft `TX1` so that her loss is miniscule - as long as this doesn't make Alice suspicious, e.g. she could split her input to `TX1` into two portions, e.g. 1 wei + 10 eth and double spend only the 1 wei

I'll make a note of this

docs/morevp.md
pnowosie: Oh, no we're talking about the same thing:
`[] - tx2 -> UTXO3 - tx3 -> UTXO4`
I got it now... :smiley_cat:

PR #294, opened 2018-12-06
[elixir-omg] Add Sentry support for omg_api

jbunce:
As title


jbunce:
Tested locally that the childchain starts up without the Sentry DSN ENV populated. Good to merge.

PR #299, opened 2018-12-10
[elixir-omg] Add Watcher credentials via an ENV

jbunce:


apps/omg_watcher/config/dev.exs
```diff
@@ -45,7 +45,12 @@ config :omg_watcher, OMG.Watcher.DB.Repo,
   timeout: 60_000,
   connect_timeout: 60_000,
   # DATABASE_URL format is following `postgres://{user_name}:{password}@{host:port}/{database_name}`
-  url: {:system, "DATABASE_URL", "postgres://omisego_dev:omisego_dev@localhost/omisego_dev"},
+  url: {:system, "DATABASE_URL",
```
pdobacz: In general: 
```
"postgres://#{System.get_env(\"WATCHER_DATABASE_USER\")}:#{System.get_env(\"etc etc\")}...."
```

maybe you can save some escaping using `"""` for multiline preformatted string.

Maybe defining `"WATCHER_DATABASE_USER"` strings in variables could also help here.

But specifically - it seems like you're having two redundant ways of specifying `url` in `ENV` variables. Why not just use `DATABASE_URL` and have the literal `"postgres://..."` remain the default?

Finally, maybe you should be using `prod.exs` here?

jbunce:
Abandoning.

PR #300, opened 2018-12-10
[elixir-omg] Add Sentry support to other modules

jbunce:
Adds Sentry support to Watcher, Database and RPC modules


mix.exs
```diff
@@ -56,6 +56,14 @@ defmodule OMG.Umbrella.MixProject do
     ]
   end
 
+def application do
```
pdobacz: oh, maybe I misspoke. You probably don't want/need this bit, as far as I understand umbrella apps/`mix.exs` this is ignored.

I think what you can do, is add the `{:sentry....}` entry into `deps` in this (top level) `mix.exs`, instead of for every subapp (eg. instead of [here](https://github.com/omisego/elixir-omg/blob/335f640e69075ddc377d03ae3f876919781df81b/apps/omg_api/mix.exs)). If so, then note that currently, the entry in `deps` in `omg_api/mix.exs`.

If you have a way to test whether all is working as intended, can you try removing all `sentry` entries from `extra_applications: [:sentry...`, and see if it still works? See [here](https://www.amberbit.com/blog/2017/9/22/elixir-applications-vs-extra_applications-guide/) for an explanation of what `extra_applications` is and why `logger` is always there.

PR #313, opened 2018-12-16
[elixir-omg/docker-compose] Auto stand services up locally

jbunce:
Stands up a Childchain, Watcher, Geth, ContractExchanger, Postgres for local development.


jbunce:
Start via `docker-compose up --build`

PR #322, opened 2018-12-17
[elixir-omg] Fix lint error

jbunce:
Fixes `apps/omg_api/lib/application.ex:71: Expression produces a value of type {'error',_} | {'ok',pid()}, but this value is unmatched`


jbunce:
Abandoning.

PR #328, opened 2018-12-19
OMG-355 feat: use PriorityQueueLib for cheaper deployments

pdobacz:
gas cost of the largest deployment (`RootChain.sol`) is now ~5M, giving us ~20% buffer to hack more interesting stuff into the contract. 

This uses a dedicated branch on https://github.com/omisego/plasma-contracts, as PR over there is pending.


apps/omg_eth/mix.exs
```diff
@@ -54,15 +54,22 @@ defmodule OMG.Eth.MixProject do
...
 
     contract_paths =
       ["RootChain.sol", "MintableToken.sol"]
-      |> Enum.map(&Path.join(contracts_dir, &1))
       |> Enum.join(" ")
 
-    output_path = Path.join(mixfile_path, "../../_build/contracts")
-    "solc #{contract_paths} --overwrite --abi --bin --optimize --optimize-runs 1 -o #{output_path}"
+    output_path = Path.join([mixfile_path, "../..", "_build/contracts"])
+    compilation_path = Path.join([mixfile_path, "../..", contracts_dir])
```
pik694: I would also consider renaming this into sth like `sources_dir`

pik694: I'm just being fussy here

apps/omg_eth/test/support/librarian.ex
```diff
@@ -0,0 +1,83 @@
...
+  alias OMG.Eth
+
+  import Eth.Encoding
+
+  @tx_defaults Eth.Defaults.tx_defaults()
+
+  @gas_contract_libs 1_180_000
+
+  @doc """
+  Provides linked bytecode for a particular contract. All required libs are hardcoded inside
```
pik694: consider adding in docs that libraries are deployed within this call

apps/omg_eth/mix.exs
pdobacz: ok so,

1/ `compilation_path` is what it says it is - a path from where it is invoking the compilation. Actually it needs to be like that, because otherwise solc goes crazy about the library placeholders (it uses the relative paths from `compilation_path` so it may run out of space... don't ask why)
2/ I'll leave a comment on the above to clarify
3/ `contract_paths` - again this is what it is just the paths are trivial :). I'll think about naming better or commenting

pik694: I mentioned it as it may be a little confusing

PR #329, opened 2018-12-20
fix: prevent child chain allowing spends from exiting deposit

pdobacz:
Child chain's `:exiter`'s finality margin cannot be zero - it must be greater than its `deposit finality margin`. Otherwise there's an option of child chain letting a UTXO be spent that had been exited immediately after the deposit (because deposit's wait and must wait). Such exit processed by `:exiter` would be a noop.

So `:exiter` must wait at least whatever `:depositor` is waiting, to always come after it.

### DIFF
--- a/apps/omg_api/lib/application.ex
+++ b/apps/omg_api/lib/application.ex
@@ -51,8 +51,8 @@ defmodule OMG.API.Application do
           {OMG.API.EthereumEventListener, :start_link,
            [
              %{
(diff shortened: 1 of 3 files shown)


pdobacz:
(:x: is because of circleCI, Jenkins :heavy_check_mark: [here](https://jenkins.omisego.io/job/omisego/job/elixir-omg/job/PR-329/) )

PR #331, opened 2018-12-20
fix cleanbuild by making call to OMG.API dynamic

pdobacz:
when we did @api_module.call_sth that would cause compiler to complain, in case @api_module wasn't compiled yet (e.g. on CircleCI)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -67,7 +67,7 @@ podTemplate(
 
         stage('Cleanbuild') {
             withEnv(["MIX_ENV=test", "DATABASE_URL=${DATABASE_URL}"]) {
-                sh("mix do compile --warnings-as-errors --force, test --exclude test")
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -67,7 +67,7 @@ podTemplate(
 
         stage('Cleanbuild') {
             withEnv(["MIX_ENV=test", "DATABASE_URL=${DATABASE_URL}"]) {
-                sh("mix do compile --warnings-as-errors --force, test --exclude test")
+                sh("mix do clean, compile --warnings-as-errors --force, test --exclude test")
```
pnowosie: [out of curiosity]: I though it fetched code on clear instance and build from scratch, or this `clean` does smth which helps here ❓

pdobacz: code is already built in this moment, since `Unit test` built it in `MIX_ENV=test` env. The other order of execution on CircleCI made the issue hit there, but not here

Issue #332, opened 2018-12-23
Cannot create an authority address


myvidster:
Getting error when attempting to create an authority address.

mix run --no-start -e  '
   contents = OMG.Eth.DevHelpers.prepare_env!() |> OMG.Eth.DevHelpers.create_conf_file()
   "~/config.exs" |> Path.expand() |> File.write!(contents)
 '

** (MatchError) no match of right hand side value: {:ok, %{"blockHash" => "0x9c6fddc0789f756fa796bc3119fe411e49150410e1557d4da16e4eccf4597a82", "blockNumber" => "0xcbb", "contractAddress" => "0xf13d901d39c263cac2a63dc5c2d0c0c9cd6f93f1", "cumulativeGasUsed" => "0x3fc820", "from" => "0x43f9925f9e7775e974524953cbc28df7c6876b9c", "gasUsed" => "0x3fc820", "logs" => [], "logsBloom" => "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000", "status" => "0x0", "to" => nil, "transactionHash" => "0xf6751a93fed5c4ae6a85786df5d7f68f52424bb45cc23226f01e877d07bc9f95", "transactionIndex" => "0x0"}}
    test/support/dev_helpers.ex:98: OMG.Eth.DevHelpers.transact_sync!/1
    test/support/dev_helpers.ex:107: OMG.Eth.DevHelpers.deploy_sync!/1
    test/support/dev_helpers.ex:49: OMG.Eth.DevHelpers.prepare_env!/1
    (stdlib) erl_eval.erl:680: :erl_eval.do_apply/6
    (stdlib) erl_eval.erl:888: :erl_eval.expr_list/6
    (stdlib) erl_eval.erl:411: :erl_eval.expr/5
    (stdlib) erl_eval.erl:449: :erl_eval.expr/5
    (stdlib) erl_eval.erl:126: :erl_eval.exprs/5
    (elixir) lib/code.ex:232: Code.eval_string/3
    (elixir) lib/enum.ex:765: Enum."-each/2-lists^foreach/1-0-"/2
[...]

PR #333, opened 2018-12-24
[docker-compose] Fix Postgres path for Linux users

jbunce:


jbunce:
Should fix https://github.com/omisego/elixir-omg/issues/332

PR #334, opened 2018-12-24
[elixir-omg] Fix Plasma contract commit

jbunce:
Match commit in https://github.com/omisego/plasma-contracts/pull/38


pdobacz:
thanks @jbunce! Maybe let's make it point to `master` since `priority-queue-as-library` got just merged

PR #335, opened 2019-01-03
[elixir-omg] Get RPC address from env

jbunce:


apps/omg_rpc/config/dev.exs
```diff
@@ -8,7 +8,7 @@ use Mix.Config
 # with webpack to recompile .js and .css sources.
 config :omg_rpc, OMG.RPC.Web.Endpoint,
   http: [port: 9656],
-  url: [host: "localhost", port: 9656],
+  url: [host: System.get_env("RPC_ADDRESS"), port: 9656],
```
paulperegud: I propose a more descriptive name: `"RPC_ADDRESS" -> "CHILDCHAIN_RPC_ADDRESS"`

It should default to `"localhost"` if nothing is set in environment.

jbunce:
Abandoning

PR #337, opened 2019-01-03
Omg 328 add spend set to omgdb

pnowosie:
Proof of concept to show we can base security critical Watcher on OMG.DB

### DIFF
--- a/apps/omg_watcher/lib/challenger.ex
+++ b/apps/omg_watcher/lib/challenger.ex
@@ -17,6 +17,7 @@ defmodule OMG.Watcher.Challenger do
   Manages challenges of exits
   """
 
+  alias OMG.API.State.Transaction
   alias OMG.API.Utxo
   require Utxo
   alias OMG.Watcher.Challenger.Challenge
@@ -35,9 +36,11 @@ defmodule OMG.Watcher.Challenger do
   Returns challenge for an exit
   """
   @spec create_challenge(Utxo.Position.t()) :: {:ok, Challenge.t()} | {:error, :invalid_challenge_of_exit}
-  def create_challenge(utxo_exit) do
-    with {:ok, challenging_tx} <- DB.Transaction.get_transaction_challenging_utxo(utxo_exit) do
-      {:ok, Core.create_challenge(challenging_tx, utxo_exit)}
+  def create_challenge(Utxo.position(blknum, txindex, oindex) = utxo_exit) do
+    with {:ok, spending_blknum} <- OMG.DB.spent_blknum({blknum, txindex, oindex}),
+         {:ok, hashes} <- OMG.DB.block_hashes([blknum, spending_blknum]),
+         {:ok, [creating_block, spending_block]} <- OMG.DB.blocks(hashes) do
+      {:ok, Core.create_challenge(creating_block, spending_block, utxo_exit)}
     else
       {:error, :utxo_not_spent} -> {:error, :invalid_challenge_of_exit}
     end
--- a/apps/omg_watcher/lib/challenger.ex
+++ b/apps/omg_watcher/lib/challenger.ex
@@ -17,12 +17,10 @@ defmodule OMG.Watcher.Challenger do
   Manages challenges of exits
   """
 
-  alias OMG.API.State.Transaction
   alias OMG.API.Utxo
   require Utxo
   alias OMG.Watcher.Challenger.Challenge
   alias OMG.Watcher.Challenger.Core
-  alias OMG.Watcher.DB
 
   @doc """
   Challenges exit.
--- a/apps/omg_watcher/lib/challenger.ex
+++ b/apps/omg_watcher/lib/challenger.ex
@@ -35,7 +35,8 @@ defmodule OMG.Watcher.Challenger do
   """
   @spec create_challenge(Utxo.Position.t()) :: {:ok, Challenge.t()} | {:error, :invalid_challenge_of_exit}
   def create_challenge(Utxo.position(blknum, txindex, oindex) = utxo_exit) do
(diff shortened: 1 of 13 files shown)


apps/omg_watcher/test/web/controllers/challenge_test.exs
```diff
@@ -27,6 +27,7 @@ defmodule OMG.Watcher.Web.Controller.ChallengeTest do
 
   @eth Crypto.zero_address()
 
+  @tag skip: true
```
pnowosie: To make this test pass I'd need to prepare test data and push it to OMG.DB.
I decided to skip them as this is PoC and there is a [integration test](https://github.com/omisego/elixir-omg/blob/8c2fbb2c3d28a1e19c3d28253d9c2b0dd7d20cf7/apps/omg_watcher/test/integration/invalid_exit_test.exs#L86) which covers challeging

apps/omg_watcher/lib/challenger/core.ex
```diff
@@ -17,49 +17,83 @@ defmodule OMG.Watcher.Challenger.Core do
...
+    |> Map.fetch!(:owner)
+  end
+
+  # finds transaction in given block and input index spending given utxo
+  @spec get_spending_transaction_with_index(Block.t(), Utxo.Position.t()) ::
+          {Transaction.Signed.t(), non_neg_integer()} | false
+  defp get_spending_transaction_with_index(%Block{transactions: txs}, utxo_pos) do
+    txs
+    |> Enum.map(&Transaction.Signed.decode/1)
+    |> Enum.find_value(fn {:ok, %Transaction.Signed{raw_tx: tx} = tx_signed} ->
```
pdobacz: provide a brief comment to this bit (`find_value` call), explain what it does

apps/omg_watcher/test/web/controllers/challenge_test.exs
```diff
@@ -51,6 +52,7 @@ defmodule OMG.Watcher.Web.Controller.ChallengeTest do
     } = TestHelper.success?("utxo.get_challenge_data", %{"utxo_pos" => utxo_pos})
   end
 
+  @tag skip: true
```
pdobacz: you should probably recreate this test (to be precise - test the same behavior) in the "new approach tests".

Ah this requires you to tweak the `Challenger.create_challenge` a bit, along the lines of:
```
with spending_blknum_response <- OMG.DB.spent_blknum({blknum, txindex, oindex}),
     {:ok, spending_blknum} <- Core.ensure_is_challengable(spending_blknum_response), # other possible return here is {:error, :some_appropriate_error_ideally_one_that_we_used_to_throw}
....
```

apps/omg_watcher/lib/challenger.ex
```diff
@@ -35,9 +34,11 @@ defmodule OMG.Watcher.Challenger do
   Returns challenge for an exit
   """
   @spec create_challenge(Utxo.Position.t()) :: {:ok, Challenge.t()} | {:error, :invalid_challenge_of_exit}
-  def create_challenge(utxo_exit) do
-    with {:ok, challenging_tx} <- DB.Transaction.get_transaction_challenging_utxo(utxo_exit) do
-      {:ok, Core.create_challenge(challenging_tx, utxo_exit)}
+  def create_challenge(Utxo.position(blknum, txindex, oindex) = utxo_exit) do
```
pdobacz: you might want to remove `DB.Transaction.get_transaction_challenging_utxo` for completeness (and the stuff only that called/used)

apps/omg_watcher/lib/challenger/core.ex
```diff
@@ -17,49 +17,83 @@ defmodule OMG.Watcher.Challenger.Core do
...
     }
   end
 
+  @spec get_creating_transaction(Block.t(), Utxo.Position.t()) :: Transaction.Signed.t()
+  defp get_creating_transaction(
+         %Block{
+           transactions: txs,
+           number: blknum
+         },
+         Utxo.position(blknum, txindex, _oindex)
```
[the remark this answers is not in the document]
pnowosie: Surprisingly this is opposite - `^` gives a compilation error. No pin - matches.
https://gist.github.com/pnowosie/f455967eb71cac929b5efd64e7de9dee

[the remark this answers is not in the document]
pdobacz: ah wait, you're right, one doesn't need to pin in here. They must be the same to match. Sorrrrryyy

apps/omg_watcher/lib/challenger/core.ex
```diff
@@ -17,49 +17,83 @@ defmodule OMG.Watcher.Challenger.Core do
...
+       ) do
+    {:ok, signed_tx} =
+      txs
+      |> Enum.fetch!(txindex)
```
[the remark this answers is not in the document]
pnowosie: I've changed db query to find out if we get the blknum (where utxo was spent) or :not_found. 
Once we get blknum then all cases that we couldn't find expected data (block hash, block, tx in block...) seems corrupted state to me.

apps/omg_watcher/test/web/controllers/challenge_test.exs
pnowosie: I added test to [challenger_core](https://github.com/omisego/elixir-omg/blob/79146aaa846b62d5eb88dc53e5f316aed3526116/apps/omg_watcher/test/challenger/core_test.exs#L91)

pnowosie:
Agreed to merge this. But before:
- [ ]  Add TODOs when filling OMG.DB spend collection and in security-critical Watcher functions 
- [ ] Update PT task ##161355645
- [ ] run simple perftest before and after

PR #352, opened 2019-01-10
Omg 300 chain operator minimally conforms to more vp

pthomalla:


apps/omg_api/lib/state/core.ex
```diff
@@ -248,30 +251,34 @@ defmodule OMG.API.State.Core do
...
-         %Transaction{} = tx
+         %Transaction.Recovered{} = tx
        ) do
     new_utxos_map =
       tx
       |> non_zero_utxos_from(height, tx_index)
       |> Map.new()
 
-    inputs = Transaction.get_inputs(tx)
+    inputs = Transaction.get_inputs(tx.signed_tx.raw_tx)
```
pdobacz: prefer pattern matching to get `raw_tx` out

apps/omg_api/lib/state/core.ex
```diff
@@ -248,30 +251,34 @@ defmodule OMG.API.State.Core do
...
+         %Transaction.Recovered{signed_tx: %Transaction.Signed{raw_tx: %Transaction{} = tx}, signed_tx_hash: hash},
+         height,
+         tx_index
+       ) do
     outputs = Transaction.get_outputs(tx)
 
     for {%{owner: owner, currency: currency, amount: amount}, oindex} <- Enum.with_index(outputs) do
-      {Utxo.position(height, tx_index, oindex), %Utxo{owner: owner, currency: currency, amount: amount}}
+      {Utxo.position(height, tx_index, oindex),
+       %Utxo{owner: owner, currency: currency, amount: amount, signed_tx_hash: hash}}
```
pdobacz: `signed_tx_hash` field should have a better name... `creating_txhash` perhaps?

apps/omg_api/lib/state/core.ex
```diff
@@ -428,6 +435,67 @@ defmodule OMG.API.State.Core do
     {:ok, {event_triggers, db_updates, validities}, new_state}
   end
 
+  defp sigs_chope(<<>>), do: []
```
pdobacz: spelling + wouldn't this fit better in `Transaction.Signed`? you even have a `@signature_length` defined there

apps/omg_api/lib/state/core.ex
```diff
@@ -428,6 +435,67 @@ defmodule OMG.API.State.Core do
...
+    {:ok, tx_recover} = Transaction.Recovered.recover_from(%Transaction.Signed{raw_tx: tx, sigs: sigs_chope(sigs)})
+
+    {inputs, _invalid} =
+      tx_recover.signed_tx.raw_tx |> Transaction.get_inputs() |> Enum.split_with(&utxo_exists?(&1, state))
+
+    db_updates = to_delete_db_update(inputs)
+    new_state = %{state | utxos: Map.drop(utxos, inputs)}
+    event_trigger = %{in_flight_exits: %{utxo_pos: inputs}}
+
+    {{db_updates, event_trigger}, new_state}
```
pdobacz: avoid the switch of the order of args/returns list, eg. put `event_triggers` consistently first

apps/omg_api/lib/state/core.ex
```diff
@@ -428,6 +435,67 @@ defmodule OMG.API.State.Core do
...
+  defp sigs_chope(<<sig::bytes-size(65), rest::binary>>), do: [sig | sigs_chope(rest)]
+
+  @spec in_flight_exits(in_flight_txs :: [in_flight_exit()], state :: t()) ::
+          {:ok, {[exit_event], [db_update]}, new_state :: t()}
+  def in_flight_exits(in_flight_txs, state) do
+    {db_updates_list_and_events, new_state} =
+      in_flight_txs
+      |> Enum.map_reduce(state, &in_flight_exit/2)
+
+    {event_triggers, db_updates} =
```
pdobacz: Taking my comment on function naming and `exit_utxos` further, it looks like you should DRY these a lot, so that after you get a list of UTXO positions to work with, all 3 "versions" (for standard exits, for in flight exit inputs, for piggybacks) use the same DRY code.

apps/omg_api/lib/state/transaction.ex
```diff
@@ -199,6 +199,15 @@ defmodule OMG.API.State.Transaction do
   defp parse_address(<<_::160>> = address_bytes), do: {:ok, address_bytes}
   defp parse_address(_), do: {:error, :malformed_address}
 
+  def decode(signed_tx_bytes) do
```
pdobacz: I presume the input here isn't `signed_tx_bytes` but rather `raw_tx_bytes`.

Also please try to make naming and flow consistent with the version for `Transaction.Signed`. One imporant thing is `from_rlp` can't be included in the `try-rescue` block, see `Transaction.Signed.decode` and friends

apps/omg_api/test/state/core_test.exs
```diff
@@ -486,21 +487,21 @@ defmodule OMG.API.State.CoreTest do
...
   test "depositing produces db updates, that don't leak to next block", %{
-    alice: alice,
+    alice: %{addr: aplice_addr} = alice,
     state_empty: state
   } do
     assert {:ok, {_, [utxo_update, height_update]}, state} =
              Core.deposit([%{owner: alice.addr, currency: eth(), amount: 10, blknum: 1}], state)
 
-    assert utxo_update == {:put, :utxo, {{1, 0, 0}, %{owner: alice.addr, currency: eth(), amount: 10}}}
+    assert {:put, :utxo, {{1, 0, 0}, %{owner: ^aplice_addr, currency: @eth, amount: 10}}} = utxo_update
```
pdobacz: typo...

apps/omg_api/test/state/core_test.exs
```diff
@@ -25,6 +25,7 @@ defmodule OMG.API.State.CoreTest do
 
   require Utxo
 
+  @eth Crypto.zero_address()
```
pdobacz: what about additions to the `State.CoreTest` for the new functionality?

apps/omg_eth/lib/eth/root_chain.ex
```diff
@@ -74,6 +75,17 @@ defmodule OMG.Eth.RootChain do
...
 
+  def piggyback_in_flight_exit(in_flightTx, outputIndex, from, contract \\ nil, opts \\ []) do
+    defaults =
+      @tx_defaults
+      |> Keyword.put(:gas, 1_000_000)
+      |> Keyword.put(:value, @piggyback_bond)
+
+    opts = defaults |> Keyword.merge(opts)
+    contract = contract || from_hex(Application.get_env(:omg_eth, :contract_addr))
+    Eth.contract_transact(from, contract, "piggybackInFlightExit(bytes, uint8)", [in_flightTx, outputIndex], opts)
```
pdobacz: `piggybackInFlightExit(bytes, uint8)` is suspicious, normally the signatures don't contain spaces

apps/omg_eth/lib/eth/root_chain.ex
```diff
@@ -228,6 +248,35 @@ defmodule OMG.Eth.RootChain do
...
 
+  defp get_inflight_data(hash) do
+    {:ok, eth_tx} = Ethereumex.HttpClient.eth_get_transaction_by_hash(hash)
+    <<"0x", encode_inflight::binary>> = eth_tx["input"]
+
+    ABI.decode(
+      %ABI.FunctionSelector{
+        function: "startInFlightExit",
+        types: [:bytes, :bytes, :bytes, :bytes],
+        method_id: <<132, 97, 33, 149>>
```
pdobacz: there is a function which will calculate this `method_id` based on the signature, pls use it here instead of hardcoding

apps/omg_eth/lib/eth/root_chain.ex
```diff
@@ -228,6 +248,35 @@ defmodule OMG.Eth.RootChain do
          do: {:ok, Enum.map(logs, &decode_exit_started/1)}
   end
 
+  defp get_inflight_data(hash) do
+    {:ok, eth_tx} = Ethereumex.HttpClient.eth_get_transaction_by_hash(hash)
+    <<"0x", encode_inflight::binary>> = eth_tx["input"]
```
pdobacz: instead of this pattern match and later `Base.decode` use `Eth.Encoding.from_hex()`

apps/omg_watcher/lib/application.ex
```diff
@@ -71,7 +73,71 @@ defmodule OMG.Watcher.Application do
           depositor: %{sync_mode: :sync_with_coordinator},
           exit_processor: %{sync_mode: :sync_with_root_chain},
           exit_finalizer: %{sync_mode: :sync_with_coordinator},
-          exit_challenger: %{sync_mode: :sync_with_root_chain}
+          exit_challenger: %{sync_mode: :sync_with_root_chain},
```
pdobacz: modifications to the Watcher weren't supposed to be in this PR, why? (not only this is done elsewhere, but also appears incorrect - in flight exits are handled by `ExitProcessor` for the Watcher).

In fact, all of the modifications to `omg_watcher` look like they can and should be undone, except the tests and `libsecp...`

mix.exs
```diff
@@ -36,7 +36,6 @@ defmodule OMG.Umbrella.MixProject do
       # NOTE: we're overriding for the sake of `omg_api` mix.exs deps. Otherwise the override is ignored
       # TODO: removing the override is advised, but it gives undefined symbol errors, see
       #       https://github.com/exthereum/exth_crypto/issues/8#issuecomment-416227176
-      {:libsecp256k1, "~> 0.1.4", compile: "${HOME}/.mix/rebar compile", override: true},
```
pdobacz: there's a comment relevant to that line just above. If the override is unnecessary, and you're removing it, the dangling comment is horrible, please be careful

apps/omg_watcher/test/integration/in_flight_test.exs
```diff
@@ -0,0 +1,100 @@
+# Copyright 2018 OmiseGO Pte Ltd
```
pdobacz: this file duplicates the integration tests already under `inflight_exit_test.exs`, please merge.

Also I'd envision this as a one joint integration test, which would poke all the integrations - including that one, being checking if the child chain correctly spent the utxos.

Not 100% sure, but for now this is my vote, as otherwise the integration tests will get uncomfortably long to run.

CC @paulperegud

apps/omg_watcher/test/integration/in_flight_test.exs
```diff
@@ -0,0 +1,100 @@
...
+      )
+      |> Eth.DevHelpers.transact_sync!()
+
+    Eth.DevHelpers.wait_for_root_chain_block(eth_height + 10)
+
+    {:ok, %{"status" => "0x1", "blockNumber" => eth_height}} =
+      Eth.RootChain.piggyback_in_flight_exit(in_flight_exit_info["in_flight_tx"], 4, alice.addr)
+      |> Eth.DevHelpers.transact_sync!()
+
+    Eth.DevHelpers.wait_for_root_chain_block(eth_height + 10)
```
pdobacz: where does the `+ 10` come from? I suspect this can be parametrized using some config variable, something ala [this](https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher/test/integration/block_getter_test.exs#L130).

2 occurrences

apps/omg_watcher/test/integration/in_flight_test.exs
```diff
@@ -0,0 +1,100 @@
...
+      Eth.RootChain.in_flight_exit(
+        in_flight_exit_info["in_flight_tx"],
+        in_flight_exit_info["input_txs"],
+        in_flight_exit_info["input_txs_inclusion_proofs"],
+        in_flight_exit_info["in_flight_tx_sigs"],
+        alice.addr
+      )
+      |> Eth.DevHelpers.transact_sync!()
+
+    Eth.DevHelpers.wait_for_root_chain_block(eth_height + 10)
```
pdobacz: shouldn't be necessary, one should be able to just fire both `Eth.Rootchain` txs one right after the other.

apps/omg_watcher/lib/db/transaction.ex
```diff
@@ -98,6 +98,10 @@ defmodule OMG.Watcher.DB.Transaction do
   defp query_get_by_blknum(base, nil), do: base
   defp query_get_by_blknum(base, blknum), do: base |> from(where: [blknum: ^blknum])
 
+  def get_all do
```
pdobacz: undo?

apps/omg_watcher/lib/db/txoutput.ex
```diff
@@ -25,7 +25,6 @@ defmodule OMG.Watcher.DB.TxOutput do
   alias OMG.Watcher.DB.Repo
 
   require Utxo
-
```
pdobacz: undo for tidyness?

apps/omg_api/lib/application.ex
```diff
@@ -51,6 +53,42 @@ defmodule OMG.API.Application do
...
+        id: :piggyback,
+        start: {
+          OMG.API.EthereumEventListener,
+          :start_link,
+          [
+            %{
+              # TODO check if synced_height_update_key is appropriate
+              synced_height_update_key: :last_exiter_eth_height,
+              service_name: :piggyback,
+              block_finality_margin: deposit_finality_margin,
```
pdobacz: `block_finality_margin` is wrong in both additions. You must ensure `inflights` and `piggybacks` get processed after deposits (see `exiter` below). Dry the "`+ 1`" bit into a variable `exiters_finality_margin` and don't remember to move the comment `we need to be just one block after...` accordingly!

apps/omg_api/lib/application.ex
```diff
@@ -51,6 +53,42 @@ defmodule OMG.API.Application do
...
            ]}
       },
+      %{
+        id: :in_flight_exit,
+        start: {
+          OMG.API.EthereumEventListener,
+          :start_link,
+          [
+            %{
+              # TODO check if synced_height_update_key is appropriate
```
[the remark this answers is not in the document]
pdobacz: :point_up: still needs attention "Also get_last_synced_height_callback below should reflect that"

apps/omg_eth/lib/eth/root_chain.ex
```diff
@@ -228,6 +248,36 @@ defmodule OMG.Eth.RootChain do
...
+  Returns InFlightExit from a range of blocks.
+  """
+  def get_in_flight_exit_starts(block_from, block_to, contract \\ nil) do
+    contract = contract || from_hex(Application.get_env(:omg_eth, :contract_addr))
+    signature = "InFlightExitStarted(address,bytes32)"
+
+    with {:ok, logs} <- Eth.get_ethereum_events(block_from, block_to, signature, contract) do
+      {:ok,
+       Enum.map(logs, fn log ->
+         get_inflight_data(log["transactionHash"])
```
pdobacz: not sure, but it looks like the result of `get_in_flight_exit_starts` differs a lot from the other "get_logs" kinds of functions. I think its worth making it consistent, i.e. 
1/ return a map with the event fields
2/ I think also `eth_height` is getting included everywhere, we'll need it anyway, so it's worth to be consistent
3/ include the `get_inflight_data` data as a special field to this map

This will also have the benefit of a more sensible type [here](https://github.com/omisego/elixir-omg/pull/352/files#diff-a19f0c837d4bfde56a9db59968310a30R47) and a more readable pattern match [here](https://github.com/omisego/elixir-omg/pull/352/files#diff-a19f0c837d4bfde56a9db59968310a30R425)

apps/omg_api/test/state/core_test.exs
```diff
@@ -598,6 +600,79 @@ defmodule OMG.API.State.CoreTest do
...
+
+    state =
+      state
+      |> (&Core.exec(tx, zero_fees_map(), &1)).()
+      |> success?
+
+    expected_owner = alice.addr
+    utxo_pos_exits_in_flight = [[Transaction.encode(raw_tx), [], [], []]]
+    utxo_pos_exits_piggyback = [%{txhash: tx_hash, output_index: 4}]
+    expected_position = {:utxo_position, @child_block_interval, 0, 0}
```
pdobacz: use `Utxo.position(...)` (several occurrences)

apps/omg_watcher/test/integration/inflight_exit_test.exs
```diff
@@ -71,5 +73,46 @@ defmodule OMG.Watcher.Integration.WatcherApiTest do
...
+      Eth.RootChain.in_flight_exit(
+        in_flight_exit_info["in_flight_tx"],
+        in_flight_exit_info["input_txs"],
+        in_flight_exit_info["input_txs_inclusion_proofs"],
+        in_flight_exit_info["in_flight_tx_sigs"],
+        alice.addr
+      )
+      |> Eth.DevHelpers.transact_sync!()
+
+    Eth.DevHelpers.wait_for_root_chain_block(eth_height + 10)
```
pdobacz: still this `+ 10` here, should be parametrized, see https://github.com/omisego/elixir-omg/pull/352#discussion_r247893668

also the other comment seems unhandled: https://github.com/omisego/elixir-omg/pull/352#discussion_r247894084

apps/omg_api/lib/application.ex
```diff
@@ -23,6 +23,7 @@ defmodule OMG.API.Application do
 
   def start(_type, _args) do
     deposit_finality_margin = Application.fetch_env!(:omg_api, :deposit_finality_margin)
+    exiters_finality_margin = Application.fetch_env!(:omg_api, :exiters_finality_margin)
```
pdobacz: I see your point, but this is dangerous. Note that you've missed one spot where you should add the new `exiters_finality_margin` config variable (in the root `config/config.exs`, where you'll find `deposit_finality_margin`. 

After you add that, I think the best way to make it safe is to make a runtime check **here** which raises an error if `deposit_finality` <= `exiters_finality`

PR #353, opened 2019-01-10
fix: make decoding transaction fail when signatures do not have a proper length

pgebal:


apps/omg_api/lib/state/transaction/signed.ex
```diff
@@ -49,7 +49,7 @@ defmodule OMG.API.State.Transaction.Signed do
   end
 
   defp reconstruct_tx([sigs | raw_tx_rlp], signed_tx_bytes) do
-    with true <- Enum.all?(sigs, &signature_length?/1),
+    with :ok <- if(Enum.all?(sigs, &signature_length?/1), do: :ok, else: {:error, :bad_signature_length}),
```
paulperegud: Why not `else` corresponding to `with`? Else would return `{:error, :bad_signature_length}`. It would allow you skip this additional `if`.

PR #359, opened 2019-01-11
OMG-359_switch_from_push_system_to_pull_system_for_byzantine_events_2

purbanow:


apps/omg_watcher/lib/api/status.ex
```diff
@@ -51,7 +50,7 @@ defmodule OMG.Watcher.API.Status do
         last_mined_child_block_number: last_mined_child_block_number,
         last_mined_child_block_timestamp: last_mined_child_block_timestamp,
         eth_syncing: Eth.Geth.syncing?(),
-        byzantine_events: prepare_events(events_processor ++ events_blockgetter)
+        byzantine_events: events_processor ++ events_blockgetter
```
pgebal: It's not something about your changes, but since we are in this file, two things can be fixed:
1. You can be more specific in defining type t(), byzantine_events is of type list(OMG.Watcher.Event.t())
2. events_blockgetter -> events_block_getter

apps/omg_watcher/lib/web/views/status.ex
```diff
@@ -19,10 +19,25 @@ defmodule OMG.Watcher.Web.View.Status do
...
+  alias OMG.Watcher.Event
   alias OMG.Watcher.Web.Serializers
 
   def render("status.json", %{status: status}) do
     status
+    |> prepare_byzantine_events()
     |> Serializers.Response.serialize(:success)
   end
+
+  defp prepare_byzantine_events(%{byzantine_events: byzantine_events} = status) do
```
pgebal: I would rather name it format_byzantine_events

apps/omg_watcher/lib/event.ex
```diff
@@ -129,25 +129,8 @@ defmodule OMG.Watcher.Event do
...
-  def add_event_name_field(%UnchallengedExit{} = event) do
-    put_event_name_value(event, UnchallengedExit.name())
-  end
-
-  defp put_event_name_value(event, value) do
-    event
-    |> Map.from_struct()
-    |> Map.put(:event_name, value)
-  end
+  def get_event_name(%InvalidBlock{}), do: InvalidBlock.name()
```
pgebal: Again, since we are here, please change typo in UnchallengedExit doc, challneged -> challenged

PR #365, opened 2019-01-14
fix: Add Sentry to Watcher's Phoenix endpoint

jbunce:
Exceptions from the Watcher are not getting caught by Sentry. Phoenix applications require this statement to be added. https://hexdocs.pm/sentry/Sentry.Phoenix.Endpoint.html

### DIFF
--- a/apps/omg_watcher/lib/web/sockets/endpoint.ex
+++ b/apps/omg_watcher/lib/web/sockets/endpoint.ex
@@ -15,6 +15,7 @@
 defmodule OMG.Watcher.Web.Endpoint do
   use Phoenix.Endpoint, otp_app: :omg_watcher
   use Appsignal.Phoenix
+  use Sentry.Phoenix.Endpoint
 
   socket("/socket", OMG.Watcher.Web.Socket)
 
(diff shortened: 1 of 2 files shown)


unnawut:
The build for this PR is returning this dialyzer warning:

```
apps/omg_watcher/lib/web/sockets/endpoint.ex:15: Expression produces a value of type 'error' | 'excluded' | 'ignored' | 'unsampled' | {'ok',binary() | pid() | #{'__struct__':='Elixir.Task', 'owner':=_, 'pid':=_, 'ref':=_}}, but this value is unmatched
```

It's from sentry library not matching the return [after calling `Sentry.capture_exception/2`](https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/phoenix_endpoint.ex#L39). So it should be safe enough to add this warning to our `dialyzer.ignore-warnings`.

pdobacz:
@jbunce 

regarding @unnawut 's comment (Thank you O :) ), if you add the ignore as suggested, please add also a comment `# Sentry related ignores` or similar (following the current convention in `dialyzer.ignore-warnings`.)

PR #371, opened 2019-01-21
test: bring back capture_log to decrease red in test and assert

pdobacz:
lets have the two-fold check of byzantine event reporting, where applicable
The message delivered to humans is as important as one delivered to machines


apps/omg_watcher/test/integration/block_getter_test.exs
```diff
@@ -38,6 +38,8 @@ defmodule OMG.Watcher.Integration.BlockGetterTest do
   alias OMG.Watcher.Web.Channel
   alias OMG.Watcher.Web.Serializers.Response
 
+  import ExUnit.CaptureLog
```
pnowosie:
```suggestion
  import ExUnit.CaptureLog :D
```

PR #369, opened 2019-01-17
Omg 312 Challenging non-canonical IFEs using block txs

pdobacz:
polished version of scrapped https://github.com/omisego/elixir-omg/pull/366

don't review/merge before #361

This is still just logic, no integrations yet

PR #372, opened 2019-01-21
Omg 329 prepare state core test

pdobacz:
1/ mechanical tidies and readability of `State.CoreTest`
2/ some regrouping of tests

**NOTE** reviewing on a per-commit basis is encouraged, as the mechanical changes (
37d5cba and 
d9e2c72) are separated from the the last more concrete one (1623885)


paulperegud:
@pnowosie  No idea how this has happened, but this PR breaks tests on master.

pdobacz:
for future reference, re 

> No idea how this has happened, but this PR breaks tests on master.

it was this PR combined with #369 . 988a017f9e26d54a8781dd626db3a40838640af9 fixes this problem

PR #377, opened 2019-01-23
refactor: tweak DeferredConfig for omg_rpc - populate explicitly

pdobacz:
without this change, when one runs only a single Watcher integration tests the `omg_rpc` doesn't get started, its populate doesn't get called.


apps/omg_watcher/test/fixtures.exs
```diff
@@ -199,11 +199,11 @@ defmodule OMG.Watcher.Fixtures do
 
     EnvAgent.save_env(server_id, env)
 
-    real_addr = Application.fetch_env!(:omg_rpc, :client) |> Map.get(:child_chain_url)
+    real_addr = Application.fetch_env!(:omg_rpc, OMG.RPC.Client) |> Keyword.fetch!(:child_chain_url)
     fake_addr = "http://#{env.ip}:#{env.port}"
 
     on_exit(fn ->
-      Application.put_env(:omg_rpc, :client, %{child_chain_url: real_addr})
+      Application.put_env(:omg_rpc, OMG.RPC.Client, child_chain_url: real_addr)
```
pgebal: I think that replaces the whole value for OMG.RPC.Client key with the keyword list provided as the third argument.
We have just one setting for that key, so that's not a problem, but we have to be careful if we add more keywords under that key.

PR #379, opened 2019-01-23
refactor: move our Mix.Tasks into an umbrella sub app for DRYing

pdobacz:


apps/xomg_tasks/mix.exs
```diff
@@ -0,0 +1,26 @@
...
+  use Mix.Project
+
+  def project do
+    [
+      app: :xomg_tasks,
+      version: OMG.Umbrella.MixProject.umbrella_version(),
+      build_path: "../../_build",
+      deps_path: "../../deps",
+      lockfile: "../../mix.lock",
+      elixir: "~> 1.6",
```
pthomalla: higher elixir version

apps/xomg_tasks/mix.exs
```diff
@@ -0,0 +1,26 @@
...
+  """
+  use Mix.Project
+
+  def project do
+    [
+      app: :xomg_tasks,
+      version: OMG.Umbrella.MixProject.umbrella_version(),
+      build_path: "../../_build",
+      deps_path: "../../deps",
+      lockfile: "../../mix.lock",
```
pthomalla: variables `build_path`, `deps_path`, `lockfile`, `elixir_paths` are unnecessary

apps/xomg_tasks/mix.exs
pdobacz: it works with 1.6 (just like the rest of subapps), so why?

apps/xomg_tasks/mix.exs
pdobacz: Why?

They are necessary, for mix to be able to understand this app when ran individually (if we want to someday). Otherwise, mix builds in the wrong dir.

Also it's a standard and we do that in other `mix.exs`.

apps/xomg_tasks/mix.exs
pthomalla: rest of project work with 1.8,
In the future could be some issue with that

apps/xomg_tasks/mix.exs
pthomalla: your run this app from header project folder, 
If someday we want running individually this be weird to sed build_path so hight

PR #380, opened 2019-01-23
Omg 371 fix discrepancies between api reference and elixir omg

pgebal:


pnowosie:
Can you please also confirm that https://github.com/omisego/elixir-omg/blob/e2f8ef77ee5dbf16ed69012fd6f1b7be6b9a5308/docs/api_specs/includes/operator_api_specs.md
`transaction.submit` response have fields named `tx_index` & `tx_hash` :question: 
I have suspicion it's wrong there.

PR #388, opened 2019-01-24
Fix docker-compose.yml watcher env (again)

kevsul:
Really fix it this time - previous effort put CHILDCHAIN_URL in the wrong place 🤦‍♂️

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -66,7 +66,6 @@ services:
       - ELIXIR_SERVICE=CHILDCHAIN
       - CONTRACT_EXCHANGER_URL=http://docker.for.mac.localhost:5000
       - ETHEREUM_RPC_URL=http://docker.for.mac.localhost:8545
-      - CHILDCHAIN_URL=http://docker.for.mac.localhost:9656
       - ETHEREUM_NETWORK=LOCALCHAIN
     restart: always
     ports:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -66,7 +66,6 @@ services:
       - ELIXIR_SERVICE=CHILDCHAIN
       - CONTRACT_EXCHANGER_URL=http://docker.for.mac.localhost:5000
       - ETHEREUM_RPC_URL=http://docker.for.mac.localhost:8545
-      - CHILDCHAIN_URL=http://docker.for.mac.localhost:9656
```
jbunce: This was changed yesterday: https://github.com/omisego/elixir-omg/blob/de71b1923c296f1a709166c3b62e070dd6b7500e/apps/omg_rpc/config/config.exs#L13

The ENV needs to be `CHILD_CHAIN_URL` and both the Childchain and the Watcher use it, so the statement needs to be present for both service definitions.

jbunce: The reason this is working at the moment (as in services get deployed) is that there are defaults if the ENV isn't set. I guess you've run into the situation where you can't query UTXOs from the Watcher? It looks like the ENV is used for both the RPC and Watcher modules.

kevsul: Sorry, I hadn't picked up that change.
Yeah, without that env var set the watcher can't talk to the childchain at all

jbunce:
[https://i.pinimg.com/originals/9a/15/66/9a1566759652442f48154fdc6ffdcf9a.jpg](url)

PR #390, opened 2019-01-24
feat: upgrade elixir 1.8 and erlang 21.2.3

pthomalla:


Dockerfile
```diff
@@ -31,8 +31,8 @@ RUN apt-get update \
 RUN wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb \
   && dpkg -i erlang-solutions_1.0_all.deb \
   && apt-get update \
-  && apt-get install -y esl-erlang=1:20.3.8.6 \
-  elixir=1.6.6-2
+  && apt-get install -y esl-erlang=1:21.2.3-1 \
```
jbunce: This will automatically make all new deployments use Elixir 1.8. Let's test this in a CI pipeline first before landing.

jbunce:
https://github.com/omisego/elixir-omg/issues/401 needs to be resolved first before this can land

Dockerfile
jbunce: Resolved in 0ce8430e231467055deb59e8c390ea9284240b55

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   test:
     docker:
-      - image: omisegoimages/blockchain-base:1.6-otp20
+      - image: omisegoimages/elixir-omg-circleci:v1.8-20190129-02
```
jbunce: This still didn't get run https://circleci.com/workflow-run/b287be1c-dc8b-44bb-b8a3-c0befa255278

apps/omg_rpc/lib/client.ex
```diff
@@ -24,11 +24,11 @@ defmodule OMG.RPC.Client do
...
   """
   @spec get_block(binary()) ::
-          {:ok, map()}
-          | {:error,
-             {:malformed_response, any()}
-             | {:client_error, any()}
-             | {:server_error, any()}}
+          {:error,
+           {:client_error, any()}
+           | {:malformed_response, false | nil | true | binary() | [any()] | number() | map()}
```
pnowosie: We need to consider this case...

pthomalla:
yes it should fall on dialyzer in elixir 1.6

PR #391, opened 2019-01-25
Omg 310 notify about ifes that involve my address

purbanow:


apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -768,29 +768,96 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
+         %{alice: alice, processor_filled: processor, transactions: [tx1, tx2]} do
+      exit_processor_request = %ExitProcessor.Request{
+        blknum_now: 5000,
+        eth_height_now: 5
+      }
+
+      %Transaction{outputs: [%{owner: tx1_owner1}, %{owner: tx1_1owner2}]} = tx1
+
+      event_1 =
+        prepare_piggyback_available_event(
```
pdobacz: after you handle the non-zero ones. I think instead of the `prepare...` function you can DRY `txbytes = Transaction.encode(tx1)` and you should be able to assert without `event_1` variable. Also I think you can afford to minimize the assertion slightly. What is most important to assert in this test, is that:
1/ there are two piggybacks available, for different txbytes
2/ owners and indices make sense **for one of the events**

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -768,29 +768,96 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
+            %{address: Crypto.zero_address(), index: 3}
+          ],
+          Transaction.encode(tx2)
+        )
+
+      assert {:ok, [^event_1, ^event_2]} = exit_processor_request |> Core.invalid_exits(processor)
+    end
+
+    @tag fixtures: [:alice, :processor_filled, :transactions]
+    test "don't detects available piggybacks because txs seen in valid block", %{
```
pdobacz: doesn't detect piggybacks.

Besides that, I'd include just one transaction in the block and make sure, that the one that gets reported is the one that **wasn't** included

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -768,29 +768,96 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
+        blknum_now: 5000,
+        eth_height_now: 5,
+        blocks_result: [Block.hashed_txs_at([recovered_tx1, recovered_tx2], 3000)]
+      }
+
+      assert {:ok, []} = exit_processor_request |> Core.invalid_exits(processor)
+    end
+  end
+
+  defp prepare_piggyback_available_event(available_inputs, available_outputs, txbytes) do
```
pdobacz: as mentioned above - by trimming down the assertions you should be able to assert without this function (not 100% sure though)

apps/omg_watcher/test/integration/in_flight_exit_test.exs
```diff
@@ -107,6 +107,14 @@ defmodule OMG.Watcher.Integration.WatcherApiTest do
 
     Eth.DevHelpers.wait_for_root_chain_block(eth_height + exiters_finality_margin)
 
+    assert %{
```
pdobacz: check if the "other" test case in this file doesn't have obsolete piggyback assertions now, which we can remove (or at least comment)

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -768,29 +768,96 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
            end)
   end
 
+  describe "available piggybacks" do
```
pdobacz: more test cases needed (remember to always think about the assertions "gist" and focus on asserting a specific behavior, so that your `asserts` are simple and small):
1/ are varying owners handled well
2/ transaction without outputs handled well?
3/ challenged IFEs emit the same piggybacks as canonical ones

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -480,12 +480,22 @@ defmodule OMG.Watcher.ExitProcessor.Core do
       get_invalid_ife_challenges(request, state)
       |> Enum.map(fn txbytes -> %Event.InvalidIFEChallenge{txbytes: txbytes} end)
 
+    available_piggybacks_events =
```
pdobacz: at this point, we need to acknowledge that the `determine_blocks_to_get` must return blocks that satisfy a set of criteria. Let's document them, in `@doc` of `determine_blocks_to_get`:

```
1/ blocks where any input to any IFE was spent
2/ blocks where any output to any IFE was spent (CC @paulperegud is this true?)
3/ blocks where the whole IFE transaction **might've** been included, to get piggyback availability and to get InvalidIFEChallenge's
```

Is this list accurate and complete though? 

The reason why I'd like this doc is that we don't want for someone to optimize this blindly, and break one of the required "block consumers"

purbanow: @paulperegud 
`2/ blocks where any output to any IFE was spent (CC @paulperegud is this true?)`

pdobacz: BTW I think my comment is obsolete, having worked on `block getting` Pepesza's doing for a while. Nevermind this and sry

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -1246,4 +1376,13 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
     # hash size * merkle tree depth
     assert byte_size(proof_bytes) == 32 * 16
   end
+
+  defp assert_events(events, expected_events) do
+    assert MapSet.new(events) == MapSet.new(expected_events)
+  end
+
+  defp invalid_exits_filtered(exit_processor_request, processor, filtered_event \\ nil) do
```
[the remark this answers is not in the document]
purbanow: Currently our fixtures +`Core.invalid_exits` setup returns always 2 `Event.PiggybackAvailable` thats why for some tests we want to make assertion agaist `[]` to filterd out those 2 events.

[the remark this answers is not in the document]
pdobacz: I get it, but still, if a test like e.g. [this](https://github.com/omisego/elixir-omg/blob/94d47f953ba563c02a06d743daa3a93eca659b1d/apps/omg_watcher/test/exit_processor/core_test.exs#L1345) returned an incorrect `NonCanonicalIFE` event, you would filter it out with this function, and the assertion would pass.

In such case you need a second version of this function that returns all events except the listed ones, e.g.

```
invalid_exits_filtered(request, processor, opts \\ [])
exclude = Keyword.get(opts, :exclude, [])
only = ... :only
# filter appropriately
```

called:
```
assert something = invalid_exits_filtered(..., exclude: [Event.PiggybackAvailable])
```

pdobacz:
added some tests and merging as Paweł is offline for a while (passes on Elixir 1.8 despite some flakiness creeping in :( )

PR #398, opened 2019-01-28
test: enrich and unbrittle Persistence tests of State.Core

pdobacz:
Instead of doing a lot of `db_updates` assertions lets run them through the persistence layer (using public APIs) and see if it works as intended.

Reduces the "reaching-into-the-guts" feeling with all the `{:put, :utxo...` etc assertions.

Now persistence per-se is tested in `State.PersistenceTest` and what is left in `State.CoreTest` is more oriented towards various details of the behavior (like flushing the mempool etc.).

Also - battle testing the approach before attacking `ExitProcessor.CoreTest` in a similar manner

### DIFF
--- a/apps/omg_api/test/state/persistence_test.exs
+++ b/apps/omg_api/test/state/persistence_test.exs
@@ -1,0 +1,217 @@
+# Copyright 2018 OmiseGO Pte Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.API.State.PersistenceTest do
+  @moduledoc """
+  Test focused on the persistence bits of `OMG.API.State.Core`
+  """
+  use ExUnitFixtures
+  use OMG.DB.Case, async: true
+
+  alias OMG.API.State.Core
+  alias OMG.API.State.Transaction
+  alias OMG.API.Utxo
+
+  import OMG.API.TestHelper
+
+  require Utxo
+
+  @eth OMG.API.Crypto.zero_address()
+  @not_eth <<1::size(160)>>
+  @zero_fees %{@eth => 0, @not_eth => 0}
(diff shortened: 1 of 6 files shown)


apps/omg_db/test/support/db_case.ex
```diff
@@ -0,0 +1,42 @@
...
+  @moduledoc """
+  Defines the useful common setup for all `...PersistenceTests`:
+   - starts temporary file handler `:briefly`
+   - creates temp dir with that
+   - initializes the low-level LevelDB storage and starts the test DB server
+  """
+
+  use ExUnit.CaseTemplate
+
+  using do
```
pnowosie: Wow, this's interestning! So by `use` this file in tests we have `setup_all` and `setup` callbacks fired by ExUnit :question: 
Nice!

apps/omg_api/test/state/persistence_test.exs
```diff
@@ -0,0 +1,217 @@
...
+    {state, db_updates}
+  end
+
+  defp persist_deposit(state, deposits, db_pid) do
+    {state, db_updates} = deposit(state, deposits)
+    assert :ok = OMG.DB.multi_update(db_updates, db_pid)
+    state
+  end
+
+  defp form(state) do
```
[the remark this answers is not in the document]
pdobacz: i wanted to be super-concise here, hence I'm not 100% in line with the original API. These are private functions which drive this test only and are taylor made for this test.

Would that make this convention acceptable? I'm thinking of a DSL-like crafting of the tests, I think also [this](https://github.com/omisego/elixir-omg/blob/master/apps/omg_api/test/ethereum_event_listener/core_test.exs) takes this approach, which I'm inclined towards.

apps/omg_api/test/state/persistence_test.exs
```diff
@@ -0,0 +1,217 @@
...
+    {:ok, {_, _, db_updates}, state} = Core.form_block(@interval, state)
+    {state, db_updates}
+  end
+
+  defp persist_form(state, db_pid) do
```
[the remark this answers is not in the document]
pdobacz: see other thread

apps/omg_db/test/support/db_case.ex
pdobacz: :thinking: it's possible that this empty using isn't necessary at all, I might have left it here by mistake

pdobacz: oh, and on `ExUnit.CaseTemplate` - yes - this is the more "standard" way of doing what we're doing with `ex_unit_fixtures`, I wanted to try it out with the intention of preferring `ExUnit.CaseTemplate` wherever it's enough (`deffixture` offers a lot more in terms of composability of the steup routines)

PR #408, opened 2019-01-30
fix: use the contracts with shorter hard-coded exit periods

pdobacz:
see also the not-going-to-be-merged long-running patch branch this refers to: https://github.com/omisego/plasma-contracts/tree/master_short_exit_period


kasima:
FYI @jbunce

PR #412, opened 2019-01-31
Exception on watcher startup

pthomalla:


apps/omg_api/lib/root_chain_coordinator/core.ex
```diff
@@ -81,13 +81,17 @@ defmodule OMG.API.RootChainCoordinator.Core do
       state = %{state | services: services}
       {:ok, state}
     else
-      :invalid_synced_height_update
+      {:invalid_synced_height_update,
```
pgebal: I think a cleaner solution would be to just log rootchain heigh and reported height on error level.

PR #420, opened 2019-02-05
Omg 369 challenge the unchallengable

pdobacz:
as title + some necessary refactors and tidies - refer to individual commits to shed more light, their rather orthogonal

### DIFF
--- a/apps/omg_watcher/lib/block_getter/core.ex
+++ b/apps/omg_watcher/lib/block_getter/core.ex
@@ -19,6 +19,7 @@ defmodule OMG.Watcher.BlockGetter.Core do
   alias OMG.API.Block
   alias OMG.API.State.Transaction
   alias OMG.Watcher.Event
+  alias OMG.Watcher.ExitProcessor
 
   use OMG.API.LoggerExt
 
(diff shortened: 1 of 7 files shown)


apps/omg_watcher/lib/block_getter/core.ex
```diff
@@ -470,29 +466,24 @@ defmodule OMG.Watcher.BlockGetter.Core do
...
+        state = state |> set_chain_status(:error) |> add_distinct_event(event)
         {{:error, :withholding}, state}
 
       true ->
         {:ok, state}
     end
   end
 
+  defp not_queued_up_yet?(number, unapplied_blocks), do: not Map.has_key?(unapplied_blocks, number)
+
```
purbanow: Maybe it will be more readable to introduce following modification:
```
defp queued_up_yet?(number, unapplied_blocks), do: Map.has_key?(unapplied_blocks, number)
```

purbanow: and then: `not queued_up_yet()`

pdobacz: ok I dug into this and realized I'd rather do more massive renames into the `block_getter` area, so I'll add a chore.

For now I'll keep as is - the idea is to have a specific and encompassing function and just a simplest `true <- specific_function() || :error` construction in the with. Removing the confusion that `not_` prefix brings in is probably a matter of the larger renaming

PR #421, opened 2019-02-05
chore: update phoenix to 1.4, plug to 1.7 and others

pdobacz:
main trigger is to fix the runtime deprecation warnings (since Elixir 1.8) which were quite obnoxious:

```
warning: deprecated time unit: :nanoseconds. A time unit should 
...
```


apps/omg_rpc/mix.exs
```diff
@@ -37,11 +37,9 @@ defmodule OMG.RPC.MixProject do
   # Type `mix help deps` for examples and options.
   defp deps do
     [
-      {:phoenix, "~> 1.3.2"},
-      {:cowboy, "~> 1.1"},
+      {:phoenix, "~> 1.3"},
```
pnowosie: Wow, `1.3` :question: - smth went wrong

apps/omg_watcher/mix.exs
```diff
@@ -41,13 +41,11 @@ defmodule OMG.Watcher.Mixfile do
 
   defp deps do
     [
-      {:phoenix, "~> 1.3.2"},
+      {:phoenix, "~> 1.3"},
+      {:plug_cowboy, "~> 1.0"},
```
pnowosie: Note to self: :heavy_check_mark: - plug_cowboy 1.0 needs plug 1.7

apps/omg_rpc/mix.exs
pdobacz: no, unless we want to ditch `1.3`. `~> 1.3` includes, among others `1.4.x`. I picked the less stringent way forward.

Do we want to rule `1.3` out the "hard way" :thinking: ?

PR #422, opened 2019-02-06
OMG-313 invalid piggybacks detect and notify

paulperegud:


apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -246,9 +254,25 @@ defmodule OMG.Watcher.ExitProcessor do
...
+      # Find tx bodies.
+      |> Core.determine_ife_blocks_to_get()
+      |> run_ife_block_getting()
+      # Compare found txes with ife.tx.
+      # If equal, persist information about position.
+      |> Core.find_ifes_in_blocks(state)
+
+    {chain_status, events} =
+      request
+      # main loop
```
pdobacz: not a `loop` really. Maybe just remove that comment? in this pipeline, the function's names are more explanatory

(we could probably do better in both pipelines, but let's not rush this refactoring, there's many other things that we could refactor here first)

(2 occurences)

pdobacz: different thing - one refactor that we probably could do here (shame on me, that I didn't already), is to put the common part of this bit (the "main loop") into `defp prepare_common_request_data(request, state)` (or some better name).

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -359,6 +376,22 @@ defmodule OMG.Watcher.ExitProcessor.Core do
     {%{state | in_flight_exits: Map.merge(ifes, updated_ifes)}, db_updates}
   end
 
+  @spec determine_ife_input_utxos_existence_to_get(ExitProcessor.Request.t(), t()) :: ExitProcessor.Request.t()
```
pdobacz: `@doc` - this is very important, as we have "this other function", which does a similar thing, so let's be precise about the behavior of both of them

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -359,6 +376,22 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+  def determine_ife_input_utxos_existence_to_get(
+        %ExitProcessor.Request{blknum_now: blknum_now} = request,
+        %__MODULE__{in_flight_exits: ifes}
+      )
+      when is_integer(blknum_now) do
+    pbs =
```
pdobacz: name

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -416,6 +453,27 @@ defmodule OMG.Watcher.ExitProcessor.Core do
     %{request | spends_to_get: spends_to_get}
   end
 
+  @spec determine_ife_spends_to_get(ExitProcessor.Request.t(), __MODULE__.t()) :: ExitProcessor.Request.t()
```
pdobacz: ditto: `@doc`

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -436,6 +494,15 @@ defmodule OMG.Watcher.ExitProcessor.Core do
     %{request | blknums_to_get: spent_blknum_result}
   end
 
+  @spec determine_ife_blocks_to_get(ExitProcessor.Request.t()) :: ExitProcessor.Request.t()
```
pdobacz: and `@doc`

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -480,6 +547,17 @@ defmodule OMG.Watcher.ExitProcessor.Core do
       get_ifes_with_competitors(request, state)
       |> Enum.map(fn txbytes -> %Event.NonCanonicalIFE{txbytes: txbytes} end)
 
+    invalid_piggybacks =
+      get_invalid_piggybacks(request, state)
+      |> Enum.map(fn {txbytes, inputs, outputs} ->
+        %Event.InvalidPiggyback{txbytes: txbytes, inputs: inputs, outputs: outputs}
+      end)
+
+    # TODO: late piggybacks are critical
```
pdobacz: change to `NOTE: late piggybacks are critical, to be implemented in OMG-408`, this deserves an item in our tracker instead of the `credo` woe

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -480,6 +547,17 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+    late_invalid_piggybacks = []
+
+    has_no_late_invalid_exits = has_no_late_invalid_exits && Enum.empty?(late_invalid_piggybacks)
```
pdobacz: maybe safer to put both `Enum.empties?` on the same line here instead of rebinding?

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -507,6 +587,190 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+         {:found_pb, true} <- {:found_pb, input_index in (bad_inputs ++ bad_outputs)} do
+      challenge_data = encode_proofs(tx, input_index, proofs)
+      {:ok, challenge_data}
+    else
+      {:found_conflicts, []} -> {:error, :no_double_spent_inputs_in_requested_ife}
+      {:found_pb, false} -> {:error, :no_double_spent_on_particular_input}
+    end
+  end
+
+  defp encode_proofs(tx, input_index, proofs) when input_index in 0..3 do
```
pdobacz: name: be specific it's about piggybacks.

Same comment for `produce_proofs_and_find_the_requested_one`

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -507,6 +587,190 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+  @spec get_input_challenge_data(ExitProcessor.Request.t(), t(), Signed.tx_bytes(), 0..7) ::
+          {:ok, input_challenge_data() | output_challenge_data()} | {:error, atom()}
+  def get_input_challenge_data(
+        request,
+        %__MODULE__{in_flight_exits: ifes} = state,
+        txbytes,
+        input_index
+      ) do
+    with {:ok, tx} <- Transaction.decode(txbytes),
+         {:known_ife, true} <- {:known_ife, Transaction.hash(tx) in Map.keys(ifes)},
```
pdobacz: follow the conventions that are already employed here (see `find_competitor` for example), so:

```
with ....
      {:ok, something} <- find_something_specific(...),
```

where `find_something_specific` returns `{:error, specific_testable_reason}` in case of a critical failure 

OR

```
with ....
      true <- specific_condition_met?()
```

where `specific_condition_met` returns either `true` or `{:error, specific_testable_reason}`

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -507,6 +587,190 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+         {:ok, proofs} <- produce_proofs_and_find_the_requested_one(request, state, tx, input_index) do
+      {:ok, proofs}
+    else
```
pdobacz: see other comment - there shouldn't be a need to revert to `else`, and even more so - a one with an all match clause, to translate error tuples

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -507,6 +587,190 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+    known_txs = get_known_txs(Enum.uniq(piggybacked_blocks ++ blocks)) ++ get_known_txs(state)
+
+    compute_proofs = fn ->
+      get_invalid_piggybacks_on_inputs(known_txs, state)
+      |> Enum.filter(fn {ife_encoded_tx, _, _, _} ->
+        ife_encoded_tx == encoded_tx
+      end)
+    end
+
+    with {:found_conflicts, [{_encoded_tx, bad_inputs, bad_outputs, proofs}]} <- {:found_conflicts, compute_proofs.()},
```
pdobacz: similar comment to this `with`

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -606,6 +870,45 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+      ) do
+    updated_ifes =
+      ifes
+      |> Enum.filter(fn {_, ife} -> ife.tx_included_at == nil end)
+      |> Enum.map(fn {hash, ife} -> {hash, ife, find_ife_in_blocks(ife, blocks)} end)
+      |> Enum.map(fn {hash, ife, position} -> {hash, %InFlightExitInfo{ife | tx_included_at: position}} end)
+      |> Map.new()
+
+    state = %{state | in_flight_exits: Map.merge(ifes, updated_ifes)}
+    {request, state}
```
pdobacz: I see we're not persisting this particular change. It may be OK, but are we conscious about this being so?

If yes, let's leave a comment somewhere (?), that this "cache-like" thing is kept in memory only.

Also, let's make a comment near where `tx_included_at` is defined, that if `nil`, it doesn't mean that it has not been included, is that (possibly) we just haven't yet put it's position to memory

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -727,6 +1033,35 @@ defmodule OMG.Watcher.ExitProcessor.Core do
   defp competitor_for(signed1, %Transaction.Signed{raw_tx: known_raw_tx}),
     do: competitor_for(signed1, known_raw_tx)
 
+  @spec get_double_spends(tx_input_info, tx_input_info) :: [{0..3, Utxo.Position.t(), 0..3}]
```
pdobacz: document in a comment `#`

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -727,6 +1033,35 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+  defp get_double_spends(inputs, known_spent_inputs) do
+    get_double_spends(index_inputs(inputs), index_inputs(known_spent_inputs))
+  end
+
+  defp index_inputs(inputs_list) when is_list(inputs_list) do
+    inputs_list
+  end
+
+  defp index_inputs(%Transaction{} = raw_tx) do
+    Transaction.get_inputs(raw_tx) |> Enum.filter(&Utxo.Position.non_zero?/1) |> Enum.with_index()
```
pdobacz: at the moment, when "gaps" in inputs are allowed, you should do this Enuming in swapped order - first apply indices then filter out.

At one point when we decide it's safe for `Transaction.get_inputs` to filter non-zeros always, we'll just ditch the filtering in here and `with_index` will be in the right spot anyway

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -760,8 +1095,11 @@ defmodule OMG.Watcher.ExitProcessor.Core do
     # the default value below is true, so that the assumption is that utxo not checked is **present**
     # TODO: rather inefficient, but no as inefficient as the nested `filter` calls in searching for competitors
     #       consider optimizing using `MapSet`
+
     utxo_positions
-    |> Enum.filter(fn pos -> !Map.get(utxo_exists?, pos, true) end)
+    |> Enum.filter(fn utxo_pos ->
```
pdobacz: are you attached into this multi-line version of this :)?

apps/omg_watcher/lib/exit_processor/in_flight_exit_info.ex
```diff
@@ -57,7 +60,10 @@ defmodule OMG.Watcher.ExitProcessor.InFlightExitInfo do
   @type t :: %__MODULE__{
           tx: Transaction.Signed.t(),
           # use utxo_position really, for convenience and tooling, even if oindex is always zero here
+          # this piece of data reflects the state of the contract
```
pdobacz: oh right, then we need better names here or we'll regret:

`contract_tx_pos` and `tx_seen_in_blocks_at` (+ comment about what `nil` means for both)

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -33,6 +33,8 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
 
   require Utxo
 
+  @moduletag :exits
```
pdobacz: leftover?

Note in general that `mix test test/exit_processor/core_test.exs` works, as does `mix test test/exit_processor/core_test.exs:123`, 123 being **any** line within test's body - for example line with the failing assertion puked out to the console by `ExUnit`

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -90,14 +92,14 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
       transactions
       |> Enum.map(&Transaction.encode/1)
 
-    [tx1_sigs, tx2_sigs] =
+    [tx1_signs, tx2_signs] =
```
pdobacz: can we undo this? we use `sigs` everywhere I think, `signs` is an English word, so it's actually more confusing

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -766,6 +806,330 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
   end
 
+  describe "evaluates new piggybacks" do
+    @tag fixtures: [:processor_filled, :in_flight_exits]
+    test "detects none if there is no piggybacks", %{processor_filled: processor} do
+      {:ok, alerts} =
+        %ExitProcessor.Request{blknum_now: 1000, eth_height_now: 5}
+        |> Core.invalid_exits(processor)
+
+      assert [] = alerts |> Enum.filter(&match?(%Event.InvalidPiggyback{}, &1))
```
pdobacz: note the `invalid_exits_filtered` (or similar name) helper that has been added, please use that. Possibly rebase is required to get that

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -766,6 +806,330 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
+    test "produces single challenge proof on double-spent piggyback input",
+         %{
+           invalid_piggyback_on_input: %{
+             state: state,
+             request: request,
+             bad_pb_output: bad_pb_output,
+             ife_txbytes: txbytes
+           }
+         } do
+      assert {:ok, [%{in_flight_input_index: ^bad_pb_output}]} =
```
pdobacz: assert all fields, similar to competitors tests etc. Are there any other behaviors related to `get_input_challenge_data` which we want to test?

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -766,6 +806,330 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
+          piggybacked_blocks_result: [Block.hashed_txs_at([recovered], tx_blknum)]
+        }
+        |> Core.find_ifes_in_blocks(state)
+
+      {:ok, alerts} = Core.invalid_exits(exit_processor_request, state)
+
+      assert [%Event.InvalidPiggyback{txbytes: ^txbytes, inputs: [], outputs: [0]}] =
+               alerts |> Enum.filter(&match?(%Event.InvalidPiggyback{}, &1))
+
+      # test and implement that we can `inflight_exit.get_output_challenge_data` similar to competitor getting
```
pdobacz: note this: if possible, I think it makes sense to assert this here (and follow the convention).

iF not possible -> remove comment.

Actually, I think it's worth to do this to make sure that getting the `challenge_data` works in all the circumstances that you detect invalid piggybacks, like, for example: https://github.com/omisego/elixir-omg/pull/422/files#diff-e406c352c14f09e333c726a8fe5c0d62R931

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -766,6 +806,330 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
+         %{
+           invalid_piggyback_on_input: %{
+             state: state,
+             request: request,
+             bad_pb_output: bad_pb_output,
+             ife_txbytes: txbytes
+           }
+         } do
+      assert bad_pb_output != 1
+      assert {:error, :no_double_spent_on_particular_input} = Core.get_input_challenge_data(request, state, txbytes, 1)
```
pdobacz: typo `double_spend`

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -766,6 +806,330 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
+          blknum_now: 4000,
+          eth_height_now: 5,
+          piggybacked_blocks_result: [Block.hashed_txs_at([recovered], tx_blknum)]
+        }
+        |> Core.find_ifes_in_blocks(state)
+
+      assert {:ok, [%Event.InvalidPiggyback{txbytes: ^txbytes, inputs: [0], outputs: []}]} =
+               invalid_exits_filtered(request, state, only: [Event.InvalidPiggyback])
+
+      proofs =
```
pdobacz: can we have this assertion more verbose? even if it makes it longer. If all you're interested in is the `decode!` being successful (I assume this is what you want in this test), then maybe just get `txbytes1, txbytes2 etc` from the pattern match and then `assert {:ok, ...} = Transaction.decode...`

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -766,6 +806,330 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
+    end
+
+    @tag fixtures: [:invalid_piggyback_on_input, :in_flight_exits, :competing_transactions]
+    test "fail when asked to produce proof for illegal oindex",
+         %{invalid_piggyback_on_input: %{state: state, request: request, ife_txbytes: txbytes}} do
+      assert {:error, :no_double_spent_on_particular_input} = Core.get_input_challenge_data(request, state, txbytes, -1)
+    end
+
+    @tag fixtures: [:alice, :bob, :processor_filled, :transactions, :in_flight_exits, :competing_transactions]
+    test "can produce the multiple challenge proofs for a input piggyback",
```
pdobacz: why do we support this? any challenge proof will do, and the API will server only one. Looks like unnecessary complexity

paulperegud: fixed

apps/omg_watcher/test/exit_processor/core_test.exs
paulperegud: fixed

apps/omg_watcher/test/exit_processor/core_test.exs
paulperegud: done

apps/omg_watcher/test/exit_processor/core_test.exs
paulperegud: done

apps/omg_watcher/test/exit_processor/core_test.exs
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: done

apps/omg_watcher/lib/exit_processor/in_flight_exit_info.ex
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: Done. Also renamed, as requested in other commit.

apps/omg_watcher/lib/exit_processor.ex
paulperegud: done - DRYed into private function

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
paulperegud: done

apps/omg_watcher/test/exit_processor/core_test.exs
paulperegud: done

apps/omg_watcher/test/exit_processor/core_test.exs
paulperegud: done

apps/omg_watcher/test/exit_processor/core_test.exs
paulperegud: done

apps/omg_watcher/lib/exit_processor/core.ex
pdobacz: see here

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -722,12 +1072,45 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+  defp index_inputs(%Transaction.Signed{raw_tx: raw_tx}), do: index_inputs(raw_tx)
+  defp index_inputs(%KnownTx{signed_tx: signed}), do: index_inputs(signed)
+
+  def get_known_txs(%__MODULE__{} = state) do
     TxAppendix.get_all(state)
     |> Enum.map(fn signed -> %KnownTx{signed_tx: signed} end)
   end
 
-  defp get_known_txs(%Block{transactions: txs, number: blknum}) do
+  def get_known_txs(%Block{transactions: txs, number: blknum}) do
```
pdobacz: why stop being `defp` - looks like it's used only in this module

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -658,6 +762,327 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
     end
   end
 
+  describe "evaluates correctness of new piggybacks" do
+    @tag fixtures: [:processor_filled]
+    test "detects none if there is no piggybacks", %{processor_filled: processor} do
```
pdobacz: do we have tests (or at least do the old tests that do) covering situations when there are valid piggybacks and this list is correctly empty?

apps/omg_watcher/test/support/integration/test_helper.ex
```diff
@@ -41,6 +41,29 @@ defmodule OMG.Watcher.Integration.TestHelper do
     |> wait_for(timeout)
   end
 
+  @spec wait_for_byzantine_events_match([match_function :: (any() -> bool())], timeout()) :: {:ok, any()} | no_return()
```
pdobacz: this helper is complex enough to `@doc` it

**EDIT** on second thought this helper looks cool but redundant. I think it's much simpler to use the original waiter (above) to wait for any `InvalidPiggyback` event to pop up and then assert that it has been fully handled (integrates fully) inside the subsequent assertion [here](https://github.com/omisego/elixir-omg/pull/422/files#diff-f77acdf5f259429a6dab36a42566b19eR117), by extending it with `outputs: [...], inputs: [...]`

paulperegud: Tried removing it, turned out it helps with flakyness caused by async processing of IFE double-spending the piggybacked input. Fixed by waiting for NonCanonicalIFE instead and checking full list next. Added comment explaining why.

PR #426, opened 2019-02-07
fix: retractable height of geth

pthomalla:


pdobacz:
let's close, as this has been solved differently at the moment

PR #427, opened 2019-02-07
Omg 386 fix ethlistener corrupt state

pdobacz:
as title + includes a more stringent handling of the height update checks in root chain coordinator, cherrypicked from 25999b1247003c01bdf58c7d90fa783f1d489ad6

### DIFF
--- a/apps/omg_api/lib/ethereum_event_listener/core.ex
+++ b/apps/omg_api/lib/ethereum_event_listener/core.ex
@@ -20,12 +20,15 @@ defmodule OMG.API.EthereumEventListener.Core do
 
   defstruct synced_height_update_key: nil,
             service_name: nil,
+            # margin between what's being exchanged with RootChainCoordinator and what's actually used to query geth
             block_finality_margin: 10,
+            # what's being exchanged with RootChainCoorinator
             synced_height: 0,
             cached: %{
               data: [],
               request_max_size: 1000,
-              events_uper_bound: 0
+              # until which height the events have been pulled and cached
+              events_upper_bound: 0
             }
 
   @type event :: %{eth_height: non_neg_integer()}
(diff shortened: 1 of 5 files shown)


apps/omg_api/lib/ethereum_event_listener/core.ex
```diff
@@ -98,21 +108,26 @@ defmodule OMG.API.EthereumEventListener.Core do
...
           cached: %{data: data} = cached_data
         } = state,
         sync_height
       ) do
-    sync = sync_height - block_finality_margin
+    sync = apply_margin(sync_height, state)
     {events, new_data} = Enum.split_while(data, fn %{eth_height: height} -> height <= sync end)
 
-    db_update = [{:put, update_key, sync}]
+    db_update = [{:put, update_key, sync_height}]
```
pthomalla: If block_finality_margin will be reduced in the settings, 
holes will be created in the processed events

apps/omg_api/lib/root_chain_coordinator/core.ex
```diff
@@ -105,7 +104,11 @@ defmodule OMG.API.RootChainCoordinator.Core do
          root_chain_height,
          new_reported_sync_height
        ) do
-    current_synced_height <= new_reported_sync_height and new_reported_sync_height <= root_chain_height
+    _ =
+      if new_reported_sync_height > root_chain_height,
+        do: Logger.warn("Eth height back to #{inspect(root_chain_height)}, had #{inspect(new_reported_sync_height)}")
+
+    current_synced_height <= new_reported_sync_height
```
pthomalla: this inequality also is unfortunately incorrect  :sob: 
second cycle of request_sync after reorg

pdobacz: can you elaborate on that? reorgs (`root_chain_height` going backwards) don't seem to affect the synced height

apps/omg_api/lib/ethereum_event_listener/core.ex
pdobacz: yeah, known issue, but this `margin` shouldn't be touched, by design.

apps/omg_api/lib/root_chain_coordinator/core.ex
pdobacz: ok I get it. Client calls `RootChainCoordinator.get_synced_info`, which just got its `root_chain_height` back-tracked and it will process on an earlier height. But the problem lies not in this equality (this equality ensures that the client doesn't backtrack, which is v. important because it would otherwise process some events twice.

Problem lies in that the RootChainCoordinator can serve a backtracked height. If it is allowed to only ever progress, it seems it'll be ok. So `RootChainCoordinator.update_root_chain_height`. I'll make this amendment

apps/omg_api/lib/ethereum_event_listener/core.ex
pthomalla: but why don't save sync instead of sync_height?

pdobacz: because it's different for services with their respective `block_finality_margin`s. If this was saved and then read from DB and sent to `RootChainCoordinator`, it would be different from where it left off.

Also there's services that behave differently, like `BlockGetter`.

Only solution would be to add `block_finality_margin` back in and hope for the best, but that seems  more awkward than the solution prior to the change you've made.

`RootChainCoordinator` was designed to provide a common view of the root chain height, which is admissible to be consumed by services, without taking their specific requirements on finality. Changing that is a deep reorganization and let's not delve into this now (we will later)

jbunce:
@purbanow @pthomalla sorry for the prod but are Piotr's changes acceptable? I'd like to get this on Ari over the weekend. Thanks.

apps/omg_api/lib/root_chain_coordinator/core.ex
pdobacz: ah, I didn't see your comment on time but I applied a similar fix, thx!

PR #448, opened 2019-02-15
Omg 329 exit proc persistence test

pdobacz:
Covers a lot of our code related to persistence of exits' state with tests.

Also improves the `State.Core` part of the persistence tests and fixes one minor bug in persisting IFEs

### DIFF
--- a/apps/omg_watcher/test/exit_processor/persistence_test.exs
+++ b/apps/omg_watcher/test/exit_processor/persistence_test.exs
@@ -62,11 +62,8 @@ defmodule OMG.Watcher.ExitProcessor.PersistenceTest do
 
     contract_statuses = [{alice.addr, @eth, 10}, {@zero_address, @eth, 10}, {alice.addr, @not_eth, 9}]
 
-    processor =
-      processor
-      |> persist_new_exits(exit_events, contract_statuses, db_pid)
-
-    assert processor == state_from(db_pid)
+    processor
+    |> persist_new_exits(exit_events, contract_statuses, db_pid)
   end
 
   @tag fixtures: [:processor_empty, :alice]
@@ -80,12 +77,9 @@ defmodule OMG.Watcher.ExitProcessor.PersistenceTest do
 
     contract_statuses = [{alice.addr, @eth, 10}, {@zero_address, @eth, 10}]
 
-    processor =
-      processor
-      |> persist_new_exits(exit_events, contract_statuses, db_pid)
-      |> persist_finalize_exits({[@utxo_pos1], [@utxo_pos2]}, db_pid)
-
-    assert processor == state_from(db_pid)
+    processor
+    |> persist_new_exits(exit_events, contract_statuses, db_pid)
+    |> persist_finalize_exits({[@utxo_pos1], [@utxo_pos2]}, db_pid)
   end
 
   @tag fixtures: [:processor_empty, :alice]
@@ -113,7 +107,7 @@ defmodule OMG.Watcher.ExitProcessor.PersistenceTest do
   end
 
   @tag fixtures: [:processor_empty, :alice, :carol]
-  test "persist started ifes and loads persisted on init",
+  test "persist multiple started ifes and loads persisted on init",
        %{processor_empty: processor, alice: alice, carol: carol, db_pid: db_pid} do
     # FIXME: dry against ExitProcessor.CoreTest? but I don't want to use fixtures here :(...
     txs = [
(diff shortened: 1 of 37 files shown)


apps/omg_watcher/test/exit_processor/persistence_test.exs
```diff
@@ -0,0 +1,240 @@
...
+  use OMG.DB.Case, async: true
+
+  alias OMG.API.DevCrypto
+  alias OMG.API.State.Transaction
+  alias OMG.API.Utxo
+  alias OMG.Watcher.ExitProcessor.Core
+
+  require Utxo
+
+  @eth OMG.API.Crypto.zero_address()
```
purbanow: Maybe it is worth to define `@eth` in one place because currently in every test file or code we write 
```
@eth OMG.API.Crypto.zero_address()
```

apps/omg_watcher/test/exit_processor/persistence_test.exs
```diff
@@ -0,0 +1,240 @@
...
+  @not_eth <<1::size(160)>>
```
purbanow: the same about this

apps/omg_watcher/test/exit_processor/persistence_test.exs
```diff
@@ -0,0 +1,240 @@
...
+    |> persist_respond_to_in_flight_exits_challenges([@utxo_pos1], db_pid)
+  end
+
+  @tag fixtures: [:processor_empty, :exits]
+  test "persist multiple challenges and challenge responses",
+       %{processor_empty: processor, db_pid: db_pid, exits: {exit_events, statuses}} do
+    processor
+    |> persist_new_exits(exit_events, statuses, db_pid)
+    |> persist_challenge_exits([@utxo_pos2, @utxo_pos1], db_pid)
+    # NOTE: see above comment
```
purbanow: Do you mean this comment?:
```
    # NOTE: this might break when respond_to_in_flight_exits_challenges is actually implemented, it works because noop

```
I think it is better to repeat comment than reference comment like this because in some point maybe someone delete above comment without deleting this one.

apps/omg_watcher/test/exit_processor/persistence_test.exs
```diff
@@ -0,0 +1,240 @@
...
+  @tag fixtures: [:processor_empty, :alice, :carol]
+  test "persist started ifes regardless of status",
+       %{processor_empty: processor, alice: alice, carol: carol, db_pid: db_pid} do
+    txs = [
+      Transaction.new([{1, 0, 0}, {1, 2, 1}], [{alice.addr, @eth, 1}]),
+      Transaction.new([{2, 1, 0}, {2, 2, 1}], [{alice.addr, @eth, 1}, {carol.addr, @eth, 2}])
+    ]
+
+    contract_statuses = [{1, <<1::192>>}, {0, <<0::192>>}]
+
```
purbanow: Maybe it is worth define somewhere `<<0::192>>}` ?

apps/omg_watcher/test/exit_processor/persistence_test.exs
```diff
@@ -0,0 +1,240 @@
...
+    hash = Transaction.hash(tx)
+    competing_tx = Transaction.new([{2, 1, 0}, {1, 0, 0}], [{alice.addr, @eth, 2}, {alice.addr, @eth, 1}])
+
+    challenge = %{
+      tx_hash: hash,
+      competitor_position: Utxo.Position.encode(@utxo_pos2),
+      call_data: %{
+        competing_tx: Transaction.encode(competing_tx),
+        competing_tx_input_index: 0,
+        competing_tx_sig: <<0::520>>
```
purbanow: Same here about `<<0::520>>`

apps/omg_watcher/test/exit_processor/persistence_test.exs
pdobacz: my proposition:
 - add `OMG.Eth.eth_pseudo_address()` as _the_ function to use to get those zero bytes, via `zero_address` of course
 - continue to use `@eth OMG.Eth.eth_pseudo_address()` for short-hand in testes

pdobacz: in the end figure out that most elegantly is to have:
  - `OMG.Eth.zero_address()` (moved from `Crypto`)
  - `OMG.Eth.RootChain.eth_pseudo_address()`

apps/omg_watcher/test/exit_processor/persistence_test.exs
pdobacz: this one is not that frequent. I think DRYing this would confuse more (new module under `support` required etc etc) than clarify

apps/omg_watcher/test/exit_processor/persistence_test.exs
pnowosie: The only cavil imho I can't see a value in making zero_address somehow related to Eth (as eth_pseudo_address). Of course it will logically be Eth by convention. But from code it's just address, constant.

pdobacz: > related to Eth

not sure what's meant. Do you mean `Eth` the module or ETH the currency in your comment?

pnowosie: Haha, this confuses me as well. I thought it as currency ;) 
Anyway, probably `zero_address` will be less confusing...

PR #456, opened 2019-02-21
OMG-239 fix: adjust multiple configuration parameters to match Ari/staging needs

pdobacz:
This mainly consists in fixing `:dev` environment, which our deployments use in a way, that allows them to be more robust, under e.g. invalid exits.

Also significant cleanups to config vars and prettify info level logs a bit

### DIFF
--- a/apps/omg_api/config/config.exs
+++ b/apps/omg_api/config/config.exs
@@ -4,7 +4,9 @@ config :omg_api,
   deposit_finality_margin: 10,
   exiters_finality_margin: 11,
   submission_finality_margin: 20,
-  ethereum_status_check_interval_ms: 6_000,
+  ethereum_events_check_interval_ms: 500,
+  block_queue_eth_height_check_interval_ms: 6_000,
+  coordinator_eth_height_check_interval_ms: 6_000,
   child_block_minimal_enqueue_gap: 1,
   fee_specs_file_path: "./fee_specs.json"
 
--- a/apps/omg_api/config/test.exs
+++ b/apps/omg_api/config/test.exs
@@ -3,5 +3,7 @@ use Mix.Config
 config :omg_api,
   deposit_finality_margin: 1,
   exiters_finality_margin: 2,
-  ethereum_status_check_interval_ms: 50,
+  ethereum_events_check_interval_ms: 50,
+  block_queue_eth_height_check_interval_ms: 100,
+  coordinator_eth_height_check_interval_ms: 100,
   fee_specs_file_path: "./../../fee_specs.json"
--- a/apps/omg_watcher/lib/db/transaction.ex
+++ b/apps/omg_watcher/lib/db/transaction.ex
@@ -142,7 +142,7 @@ defmodule OMG.Watcher.DB.Transaction do
         ]
       )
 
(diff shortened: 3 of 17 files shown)


apps/omg_api/config/test.exs
```diff
@@ -3,5 +3,7 @@ use Mix.Config
 config :omg_api,
   deposit_finality_margin: 1,
   exiters_finality_margin: 2,
-  ethereum_status_check_interval_ms: 50,
+  ethereum_events_check_interval_ms: 50,
```
pgebal: What's the reason we keep those intervals in 2:1 length ratio?

apps/omg_watcher/lib/db/transaction.ex
```diff
@@ -142,7 +142,7 @@ defmodule OMG.Watcher.DB.Transaction do
         ]
       )
 
-    _ = Logger.info("Block ##{block_number} persisted in DB done in #{insert_duration / 1000}ms")
+    _ = Logger.debug("Block \##{block_number} persisted in WatcherDB, done in #{insert_duration / 1000}ms")
```
pgebal: Do we need to exit # character?
Why don't we pass a function as argument to Logger.debug?

apps/omg_api/config/config.exs
```diff
@@ -4,7 +4,9 @@ config :omg_api,
   deposit_finality_margin: 10,
   exiters_finality_margin: 11,
   submission_finality_margin: 20,
-  ethereum_status_check_interval_ms: 6_000,
+  ethereum_events_check_interval_ms: 500,
```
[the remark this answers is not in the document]
pdobacz: I have split the responsibilities of this interval setting. It used to be related to new eth-blocks being acknowledged. Now it only determines the cadence of `EthereumEventsListener` waking up, which is good to have quite frequent in the new approach to syncing (the forced `:sync` message became a bit obsolete in the new approach).

If we devise some upgraded politics to force a `:sync` message, we might go back to the large interval. But in the "new approach" it also isn't necessary to speed up by sending these `:sync` messages all the time (because we consume ethereum events in large chunks, as opposed to a single eth_height at a time).

Hope this clarifies a bit

apps/omg_api/config/test.exs
pdobacz: hm, there isn't a hard requirement of the 2:1. It is rather a coincidence.

It sort of replicates the `block_queue_eth_height_check_interval_ms` and `coordinator_eth_height_check_interval_ms` being done "less frequently", because they need to pick up an advance of Ethereum, rather than some internal workings of our syncing like `ethereum_events_check_interval_ms` does

apps/omg_watcher/lib/db/transaction.ex
pdobacz: > Do we need to exit # character?

I think not, I'm doing it more to make it look intentional :). Then, I'm keeping it to be somewhat consistent + don't need to modify in too many places now

> Why don't we pass a function as argument to Logger.debug?
it is not necessary anymore for lazyness (since Elixir 1.7). So we went for the more succinct and less confusing syntax.

PR #461, opened 2019-02-25
alter restart strategy

InoMurko:
This PR solves the problem when the Ethereum client becomes unreachable. I didn't find any logs of restarts in Kibana though.
The restart intensity gives us roughly 1 minute for the connection to recover.

This is a temporary solution to a different problem. Processes in omg_api are all supervised on the same level and coupled to a connection that's not monitored separately.

So I would suggest a refactor and decoupling to separate hierarchies and in cases when the connection goes offline - I would suspend the API.

### DIFF
--- a/apps/omg_api/lib/application.ex
+++ b/apps/omg_api/lib/application.ex
@@ -68,7 +68,7 @@ defmodule OMG.API.Application do
       {OMG.RPC.Web.Endpoint, []}
     ]
 
-    opts = [strategy: :one_for_one]
+    opts = [strategy: :one_for_one, name: __MODULE__, max_restarts: 10000, max_seconds: 60_000]
 
     _ = Logger.info("Starting #{inspect(__MODULE__)}")
     :ok = :error_logger.add_report_handler(Sentry.Logger)


apps/omg_api/lib/application.ex
```diff
@@ -68,7 +68,7 @@ defmodule OMG.API.Application do
       {OMG.RPC.Web.Endpoint, []}
     ]
 
-    opts = [strategy: :one_for_one]
+    opts = [strategy: :one_for_one, name: __MODULE__, max_restarts: 10000, max_seconds: 60_000]
```
jbunce: What does `max_seconds` represent? The interval between restarts?

InoMurko: The combo max restarts and max seconds specify maximum restart intensity.
If more than `max_restarts` number of restarts occur in the last `max_seconds` seconds, the supervisor terminates all the child processes and then itself, ending in terminating `omg_api` application.

jbunce:
LGTM as a first pass. I'll let someone in Poland who knows more about Elixir than me hit the approve button :)

InoMurko:
There's one other thing - the inits of the child processes are not clean. Meaning that they immediately call integrated services and they're not able to start if they can't connect. A more decoupled approach would be to init them in a separate self send message or `handle_continue`.

InoMurko:
1. I can do that. But TBH, I consider branching from master a bad habit because no one deletes stale branches. So forking should be at least considered. :) I think release branches and master branches are all that we should have.
2. You're right, I have noticed today that application borders are somewhat unclean. 
3. It should be a separate PR, especially because with that approach, you're loosening the supervisor. Where before it meant that the process is automatically *ready* when it started, this is not the case if you let it start and run setup async. So yeah, you need to add *readiness*. 
4. Backoff would really be a problem to implement at the moment - because of 3. You can't backoff in the init, because that halts the supervisor.

InoMurko:
Branch from master.

PR #464, opened 2019-02-26
Omg 420 fix "tx seen in blocks at" issue

pdobacz:
review only the fix at https://github.com/omisego/elixir-omg/commit/31ac10b46d02fd05df6f5f8b519b8bced5364870

The other commit is a revert.

They both are intended to fix issues that surfaced on staging

### DIFF
--- a/apps/omg_watcher/lib/exit_processor/in_flight_exit_info.ex
+++ b/apps/omg_watcher/lib/exit_processor/in_flight_exit_info.ex
@@ -100,8 +100,29 @@ defmodule OMG.Watcher.ExitProcessor.InFlightExitInfo do
     end
   end
 
-  def make_db_update({_ife_hash, %__MODULE__{} = _ife} = update) do
-    {:put, :in_flight_exit_info, update}
+  # NOTE: we have no migrations, so we handle data compatibility here (make_db_update/1 and from_db_kv/1), OMG-421
+  def make_db_update({ife_hash, %__MODULE__{} = ife}) do
+    persistable = [
+      :tx,
+      :tx_pos,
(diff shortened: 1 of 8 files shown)


apps/omg_watcher/lib/exit_processor/in_flight_exit_info.ex
```diff
@@ -100,8 +100,85 @@ defmodule OMG.Watcher.ExitProcessor.InFlightExitInfo do
...
+      eth_height: eth_height,
+      exit_map: exit_map,
+      is_canonical: is_canonical,
+      is_active: is_active
+    }
+
+    {:put, :in_flight_exit_info, {ife_hash, value}}
+  end
+
+  defp assert_utxo_pos_type({:utxo_position, blknum, txindex, oindex})
```
pdobacz: match with `Utxo.position(...)` instead

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -45,7 +45,7 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
   @utxo_pos2 Utxo.position(@late_blknum - 1_000, 0, 1)
   @utxo_pos3 Utxo.position(1, 0, 0)
 
-  @non_zero_exit_id <<1::192>>
+  @non_zero_exit_id 0
```
pdobacz: `@non_zero_exit_id 1`?

pik694: oh, such a silly mistake...

PR #466, opened 2019-02-26
Omg 378 (elixir omg) check if new se is spent by dexisting ife challenge if yes

purbanow:


apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -298,6 +309,27 @@ defmodule OMG.Watcher.ExitProcessor do
     {:reply, canonicity_result, state}
   end
 
+  def handle_call({:create_challenge, Utxo.position(blknum, txindex, oindex) = exiting_utxo_pos}, _from, state) do
+    with spending_blknum_response = OMG.DB.spent_blknum({blknum, txindex, oindex}),
+         ife_response = Core.get_ife_based_on_utxo(exiting_utxo_pos, state),
```
pthomalla: why ```=``` and not ```<-```? 
then you can remove (if catch {:ok, val})
```
def ensure_challengeable({error,errno}_,_)
def ensure_challengeable(_,{:error,errno},_)
def ensure_challengeable(_,_,{:error, error})
```

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -471,6 +472,11 @@ defmodule OMG.Watcher.ExitProcessor.Core do
       invalid_exit_positions
       |> Enum.map(fn position -> ExitInfo.make_event_data(Event.InvalidExit, position, exits[position]) end)
 
+    # get exits which are invalid because of being spent in IFEs
+    invalid_exits_based_on_ifes_events =
+      get_invalid_exits_based_on_ifes(state)
+      |> Enum.map(fn {position, exit_info} -> ExitInfo.make_event_data(Event.InvalidExit, position, exit_info) end)
```
pthomalla: why do you need a variable invalid_exits_based_on_ifes_events? 
and why it is so far away from its use?

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -502,6 +513,40 @@ defmodule OMG.Watcher.ExitProcessor.Core do
     {chain_validity, events}
   end
 
+  @spec get_exit_info(Utxo.Position.t(), t()) :: {:ok, ExitInfo.t()} | {:ok, :not_found}
+  def get_exit_info(Utxo.position(_, _, _) = utxo_exit, %__MODULE__{exits: exits}) do
```
pthomalla: mabe ther shoud be guard is_position

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -502,6 +513,40 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+  defp get_invalid_exits_based_on_ifes(%__MODULE__{exits: exits} = state) do
+    exiting_utxo_positions =
+      get_known_txs(state)
+      |> Enum.flat_map(fn %KnownTx{signed_tx: %Transaction.Signed{raw_tx: %Transaction{} = tx}} ->
+        Transaction.get_inputs(tx)
+      end)
+
+    exits
+    |> Enum.filter(fn {utxo_pos, _exit_info} ->
+      Enum.find(exiting_utxo_positions, fn exiting_utxo_pos -> utxo_pos == exiting_utxo_pos end)
```
pthomalla: ther is Kernel.match?

apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -298,6 +309,27 @@ defmodule OMG.Watcher.ExitProcessor do
...
+         exit_response = Core.get_exit_info(exiting_utxo_pos, state),
+         {:ok, raw_spending_proof, exit_info} <-
+           Core.ensure_challengeable(spending_blknum_response, exit_response, ife_response) do
```
pthomalla: 3 times call Core, create a function

apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -298,6 +309,27 @@ defmodule OMG.Watcher.ExitProcessor do
     {:reply, canonicity_result, state}
   end
 
+  def handle_call({:create_challenge, Utxo.position(blknum, txindex, oindex) = exiting_utxo_pos}, _from, state) do
+    with spending_blknum_response = OMG.DB.spent_blknum({blknum, txindex, oindex}),
```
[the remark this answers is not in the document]
purbanow: I think the idea of `OMG.DB` is to be independent of other staff. I added here `Utxo.Position.to_db_key` which already existed

apps/omg_watcher/lib/exit_processor/core.ex
purbanow: result using `match?`;
`{true, {:utxo_position, 9000, 0, 1}, {:utxo_position, 1, 0, 0}}`
When using `match?` {:utxo_position, 9000, 0, 1} equals {:utxo_position, 1, 0, 0} I think because there is macro or somethig .
In the console(IEX) it will be true.

apps/omg_watcher/lib/exit_processor/core.ex
purbanow: I moved close to use. I make do it without extra variable but I think in this way is more readable

apps/omg_watcher/lib/exit_processor/core.ex
purbanow: I think it is enough to have:  `Utxo.position(_, _, _)`. why do you think the guard is needed?

apps/omg_watcher/lib/exit_processor.ex
purbanow: Because I need only 2 out of 3 variables to be valid. By variables I mean `
spending_blknum_response` , `ife_response`, `exit_response`

apps/omg_watcher/lib/exit_processor.ex
purbanow: What do you mean exactly?

apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -294,6 +305,27 @@ defmodule OMG.Watcher.ExitProcessor do
...
 
+  def handle_call({:create_challenge, Utxo.position(blknum, txindex, oindex) = exiting_utxo_pos}, _from, state) do
+    with spending_blknum_response <- exiting_utxo_pos |> Utxo.Position.to_db_key() |> OMG.DB.spent_blknum(),
+         %{txhash: txhash} <- OMG.Watcher.DB.Transaction.get_by_position(blknum, txindex),
+         {:ok, exit_id} <- OMG.Eth.RootChain.get_standard_exit_id(txhash, oindex),
+         {:ok, raw_spending_proof, exit_info} <-
+           Core.get_challange_data(spending_blknum_response, exiting_utxo_pos, state) do
+      spending_proof =
+        case raw_spending_proof do
+          blknum when is_number(blknum) ->
```
pthomalla: why do you check if blknum is a number?

apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -294,6 +305,27 @@ defmodule OMG.Watcher.ExitProcessor do
...
+            {:ok, hashes} = OMG.DB.block_hashes([blknum])
+            {:ok, [spending_block]} = OMG.DB.blocks(hashes)
+            spending_block
+
+          signed_tx ->
+            signed_tx
+        end
+
+      {:reply, {:ok, Core.create_challenge(exit_info, spending_proof, exiting_utxo_pos, exit_id)}, state}
```
pthomalla: why assume that Core.create_challenge never return error?

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -492,10 +493,20 @@ defmodule OMG.Watcher.ExitProcessor.Core do
       late_invalid_exits
       |> Enum.map(fn {position, late_exit} -> ExitInfo.make_event_data(Event.UnchallengedExit, position, late_exit) end)
 
+    # get exits which are invalid because of being spent in IFEs
+    invalid_exits_based_on_ifes_events =
```
pthomalla: now remove variable invalid_exits_based_on_ifes_events, because function ```get_invalid_exits_based_on_ifes``` tell about variable.
why invalid_exits_based_on_ifes_events don't return ```EventData```?

apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -294,6 +305,27 @@ defmodule OMG.Watcher.ExitProcessor do
     {:reply, canonicity_result, state}
   end
 
+  def handle_call({:create_challenge, Utxo.position(blknum, txindex, oindex) = exiting_utxo_pos}, _from, state) do
+    with spending_blknum_response <- exiting_utxo_pos |> Utxo.Position.to_db_key() |> OMG.DB.spent_blknum(),
```
pthomalla: expecting something like this
```
resonse = with {:ok, data1} <- Outsite.getData(),
{:ok, data2} <- Outsite2.getdData(),
{:ok, data3} <- Outsite3.getData(), 
  do: Core.get_challenge(data1,data2,data3)
{:reply, response}
```

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -782,4 +827,99 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+    end)
+  end
+
+  defp get_spending_transaction_with_index(%Transaction.Signed{raw_tx: tx} = tx_signed, utxo_pos) do
+    inputs = Transaction.get_inputs(tx)
+
+    if input_index = Enum.find_index(inputs, &(&1 == utxo_pos)) do
+      {tx_signed, input_index}
+    else
+      false
```
pthomalla: why not nil?
ps.
```
if expr, 
    do: value1,
  else: value2
```

apps/omg_watcher/lib/exit_processor.ex
purbanow: Because `raw_spending_proof` can be a  number or signed_tx

apps/omg_watcher/lib/exit_processor.ex
purbanow: Here we're just passing information further about errors

apps/omg_watcher/lib/exit_processor/core.ex
purbanow: I just went with convention in this function:
1. get data
2 prepare events

apps/omg_watcher/lib/exit_processor/core.ex
purbanow: I went with `false` as all functions in this file went with `false` convention instead `nil` as return value

apps/omg_watcher/lib/exit_processor.ex
purbanow: I don't see now how to get there

purbanow:
```
  with spending_blknum_response <- exiting_utxo_pos |> Utxo.Position.to_db_key() |> OMG.DB.spent_blknum(),
         %{txhash: txhash} <- OMG.Watcher.DB.Transaction.get_by_position(blknum, txindex),
         {:ok, exit_id} <- OMG.Eth.RootChain.get_standard_exit_id(txhash, oindex),
         {:ok, raw_spending_proof, exit_info} <- Core.get_challange_data(spending_blknum_response, exiting_utxo_pos, state) do
      spending_proof =
        case raw_spending_proof do
          blknum when is_number(blknum) ->
            {:ok, hashes} = OMG.DB.block_hashes([blknum])
            {:ok, [spending_block]} = OMG.DB.blocks(hashes)
            spending_block

          signed_tx ->
            signed_tx
        end

      {:reply, {:ok, Core.create_challenge(exit_info, spending_proof, exiting_utxo_pos, exit_id)}, state}
    end
```

purbanow: The run of this code(outside call to DB):
```
{:ok, hashes} = OMG.DB.block_hashes([blknum])
            {:ok, [spending_block]} = OMG.DB.blocks(hashes)
            spending_block
```
depends of `raw_spending_proof`.

apps/omg_watcher/lib/exit_processor.ex
pthomalla: My mistake, a different name would be useful.
Previous variable`Utxo.position(blknum ...`  mixes with this

apps/omg_watcher/lib/exit_processor.ex
pthomalla: And that is a problem, get_challange_data returns two different types and shell have logic.
something like this:
```
{:ok, spending_blknum_response} <- exiting_utxo_pos |> Utxo.Position.to_db_key() |> OMG.DB.spent_blknum(),
{:ok, exit_id} <- OMG.Eth.RootChain.get_standard_exit_id(txhash, oindex),
{:ok, blknums} <- Core.get_challenges_blknums(exiting_utxo_pos, state),
{:ok, hashes} <- OMG.DB.block_hashes(blknums),
{:ok, spending_blocks} <- OMG.DB.blocks(hashes), 
   do: Core.create_challenge(spending_blocks, {exit_id, exiting_utxo_pos}, state)
```

apps/omg_watcher/lib/exit_processor/core.ex
pthomalla: and false is spending transaction with index?
function name suggests that null is returned or there will throw an error, in case no spending transaction find.

pthomalla:
the integration test would come in handy

apps/omg_watcher/lib/exit_processor/core.ex
purbanow: I did:
```
  # get exits which are invalid because of being spent in IFEs
    invalid_exit_events =
      get_invalid_exits_based_on_ifes(state)
      |> Enum.map(fn {position, exit_info} -> ExitInfo.make_event_data(Event.InvalidExit, position, exit_info) end)
      |> Enum.concat(non_late_events)
      |> Enum.uniq_by(fn %Event.InvalidExit{utxo_pos: utxo_pos} -> utxo_pos end)
```

apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -294,6 +305,27 @@ defmodule OMG.Watcher.ExitProcessor do
     {:reply, canonicity_result, state}
   end
 
+  def handle_call({:create_challenge, Utxo.position(blknum, txindex, oindex) = exiting_utxo_pos}, _from, state) do
```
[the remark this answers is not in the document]
purbanow: I don't get you, can you explain a little more what you mean?

apps/omg_watcher/lib/exit_processor.ex
purbanow: Yes, it's returning two types. Your propose is not resolving this problem. Your propose cares only about one type(blknum)

pthomalla: not exactly, I propose that get_challenges_blknums can return empty hashes. create_challenge get potential spending blocks (can be empty), and do rest of the job.

apps/omg_watcher/lib/exit_processor.ex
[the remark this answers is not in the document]
purbanow: yes, you're right. I'm going to update spec

apps/omg_watcher/lib/exit_processor.ex
purbanow: Still, I think your proposal is not going to work. You're missing staff about two types blknum and transaction

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -805,6 +799,100 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+        pb_index
+      ) do
+    with {:ok, tx} <- Transaction.decode(txbytes),
+         true <- Map.has_key?(ifes, Transaction.hash(tx)) || {:error, :unknown_ife},
+         {:ok, proof} <- produce_invalid_piggyback_proof(request, state, tx, pb_index) do
+      {:ok, proof}
+    end
+  end
+
+  defp produce_invalid_piggyback_proof(
```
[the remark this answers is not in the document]
pdobacz: it is a duplicate, removing (there are several occurrences of this)

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -805,6 +799,100 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+           get_proofs_for_particular_ife(tx, pb_index, known_txs, state),
+         true <-
+           is_piggyback_in_the_list_of_known_doublespends?(pb_index, bad_inputs, bad_outputs) ||
+             {:error, :no_double_spend_on_particular_piggyback} do
+      challenge_data = prepare_piggyback_challenge_proofs(ife, tx, pb_index, proofs)
+      {:ok, hd(challenge_data)}
+    end
+  end
+
+  defp prepare_piggyback_challenge_proofs(_ife, tx, input_index, proofs)
```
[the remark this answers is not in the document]
pdobacz: we can consider fixing that, but let's defer till a cleaning-up chore perhaps, as part of larger refactors

apps/omg_watcher/lib/exit_processor.ex
pdobacz: as discussed, let's leave a todo/chore and move on

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -38,8 +41,9 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
   @early_blknum 1_000
   @late_blknum 10_000
 
-  @utxo_pos1 Utxo.position(1, 0, 0)
+  @utxo_pos1 Utxo.position(1, 3, 0)
```
pthomalla: a little bizarre concept, with position numbering.
I'm confused why pos3 is before pos1

pdobacz:
Closing in favor of the squashed #501

PR #473, opened 2019-02-28
concurrent jobs

InoMurko:
This reduces the whole dev build from 19 mins to 10 mins.
It also gives quick feedback by parallel test execution.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -13,7 +13,7 @@ defmodule OMG.Umbrella.MixProject do
         "coveralls.detail": :test,
         "coveralls.post": :test,
         "coveralls.html": :test,
-        dialyzer: :prod
+        dialyzer: :test
       ],
       dialyzer: dialyzer(),
       test_coverage: [tool: ExCoveralls],
@@ -25,7 +25,7 @@ defmodule OMG.Umbrella.MixProject do
 
   defp deps do
     [
-      {:dialyxir, "~> 1.0.0-rc.4", only: [:prod], runtime: false},
+      {:dialyxir, "~> 1.0.0-rc.4", only: [:dev, :test], runtime: false},
       {:credo, "~> 1.0.0", only: [:dev, :test], runtime: false},
       {:excoveralls, "~> 0.10.3", only: [:test], runtime: false},
       {:licensir, "~> 0.2.0", only: :dev, runtime: false},
@@ -52,17 +52,9 @@ defmodule OMG.Umbrella.MixProject do
 
   defp dialyzer do
     [
-      paths: [
-        "_build/prod/lib/omg_watcher/ebin",
-        "_build/prod/lib/omg_rpc/ebin",
-        "_build/prod/lib/omg_api/ebin",
-        "_build/prod/lib/omg_eth/ebin",
-        "_build/prod/lib/omg_db/ebin"
-      ],
       flags: [:error_handling, :race_conditions, :underspecs, :unknown, :unmatched_returns],
-      plt_add_deps: :transitive,
-      plt_add_apps: [:mix],
-      ignore_warnings: "dialyzer.ignore-warnings"
+      ignore_warnings: "dialyzer.ignore-warnings",
+      plt_add_apps: [:mix, :iex, :ex_unit, :ranch, :plug, :jason, :cowboy]
     ]
   end
 end
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2
 jobs:
-  test-1.8:
+  build:
     docker:
       - image: omisegoimages/elixir-omg-circleci:v1.8-20190129-02
         environment:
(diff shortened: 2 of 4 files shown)


pnowosie:
It seems like one step haven't completed yet - not sure what this `test-1.8` is, but I can't find it running

InoMurko:
hey @jbunce please review, as it's fiddling with deployments as well, and please take a look if you can remove the `ci/circleci: test-1.8` check in github

jbunce:
Hey Ino, thanks for looking at this. +1 for faster test times. If I've understood this correctly:

- Mix deps are cached to a CircleCI workspace 
- If the checksum of the `mix.lock` changes that indicates that the mix dependencies has been updated and therefore should not use the cache
- The workspace cache life time is 30 days if there are no changes to `mix.lock`

Some thoughts:

- I don't follow what the PLT cache does and how this would affect future builds if the `mix.lock` does not change. A quick summary is useful for reviews :)
- Do we lose visibility into useful output by changing the dialyzer step to `mix dialyzer --format short --halt-exit-status` from `mix dialyzer --halt-exit-status`?

InoMurko:
Hey @jbunce thanks!
The first part, you're 100% correct.
The second part:
PLT is a Persistent Lookup Table. There are two core PLTs, that belong to erlang and elixir specifically (internal apps etc.) and one project environment specific file, that we're building. That one is made of deps and apps, but the apps are being checked against the plt for changes. Hence we catch changes regardless if mix.lock cache.  

I don't think we loose any useful output by the way how dialyzer issues are being resolved. You either know exactly what it is, or build it yourself locally again, which takes up to 5 seconds. Dialyzer is deterministic.

mix.exs
```diff
@@ -13,7 +13,7 @@ defmodule OMG.Umbrella.MixProject do
         "coveralls.detail": :test,
         "coveralls.post": :test,
         "coveralls.html": :test,
-        dialyzer: :prod
+        dialyzer: :test
```
pdobacz: is this what we want? I recall setting this to `:prod` explicitly to omit dialyzing of test code and `test/support/` code, which is not built in `:prod` env.

(I know it caused a warning, but that was a specific warning, which perhaps could've been solved differently).

Hmmm, if I'm parsing github correctly, it seems like it passes the dialyzer step, then maybe it's an added benefit, that we cover test code with dialyzing? But anyway, I'm interested in the reasons/rationale here.

**EDIT** oh I see what you did [here](https://github.com/omisego/elixir-omg/pull/473/files#diff-9f06470aea4fd680aa740eedc568f669R5). Hm....

InoMurko: I've always tried being thorough in using dialyzer so a future step would be to clean all dialyzer warnings from tests (because it also detects _unnecessary tests_)

pdobacz: ok, let's try that, but let's remember this point to be able to back-track in case we reconsider.

I recall that dialyzing test/support code to be a huge pain once, but maybe we've given up to early.

InoMurko:
> adding my 0.03 zł - one comment about dialyzing and:
> 
> Do we want to be that aggresive about caching. I'm all for caching the PLT (I think the idea circulated around here since some time, no one ever got to doing that), but I've had bad experience with caching in CI. Is there a lot of performance gain compared to caching "just the PLT"?

Not sure I fully understand your question, but here's my take:
Old build -> around 20 minutes.
New build -> around 10 minutes.
So there's some gain in overall speed, the other part is the quick feedback we get from parallel jobs. Which I think is the most useful one. Currently if your build failed in the LINT step, you still waited 19 minutes...

Dependencies are getting cached, which is around 2 minutes (2 mins for `mix do deps.get, deps.compile, compile`, per developer, per build). And deps hardly ever change.

pdobacz:
> > adding my 0.03 zł - one comment about dialyzing and:
> > Do we want to be that aggresive about caching...
> 
> Not sure I fully understand your question, but here's my take:
> Old build -> around 20 minutes.
> New build -> around 10 minutes.
> So there's some gain in overall speed, the other part is the quick feedback we get from parallel jobs. Which I think is the most useful one. Currently if your build failed in the LINT step, you still waited 19 minutes...
> 
> Dependencies are getting cached, which is around 2 minutes (2 mins for `mix do deps.get, deps.compile, compile`, per developer, per build). And deps hardly ever change.

Got it, but I'm guessing the bulk of the 10 minute gain is the PLT gain (that + parallel execution perhaps), so I'd opt for caching only that, just to be conservative about caching and allow less surface of strange ways CI can fail (or pass incorrectly!). 

And note that we have some non-standard hocus-pocus involved around `deps` - namely `mix deps.compile plasma_contracts`. So I'd rather not have to think about CircleCI cache in case something strange happens :) and wait those 2 minutes more.

Can we perhaps try and see how slow (fast) the build is when caching only the PLT?

InoMurko:
Comparing results, @pdobacz:
No cache, Cache
Build 03:01 - 01:54
The only difference is in this step, because deps/ folder is persisted to other containers.

pdobacz:
> No cache, Cache
> Build 03:01 - 01:54

Then I'd prefer to wait that minute, to have peace of mind related to caching as little as possible.

If there's strong voices pro caching more, I can disagree and commit though.

pnowosie:
Oh and 1 more thing: we've adopted commit message naming convention. You can take a look at docs attached to [this post](https://omisego.slack.com/archives/C9KNA675J/p1547136988178700) and review commits history to get an idea.

jbunce:
Merged to `master`. Please update this for the `v0.1` branch too 👍

PR #476, opened 2019-03-04
Make appsignal report metrics

InoMurko:
Instrumenters configuration needs to be on the Phoenix level, hope that solves it.
 https://hexdocs.pm/appsignal/Appsignal.Phoenix.Instrumenter.html
The default value for active seems to be `false`.

Hope this makes metrics appear in Appsignal.

PR #477, opened 2019-03-04
fix: make appsignal report metrics

InoMurko:
Instrumenters configuration needs to be on the Phoenix level, hope that solves it.
https://hexdocs.pm/appsignal/Appsignal.Phoenix.Instrumenter.html
The default value for active seems to be false.

Hope this makes metrics appear in Appsignal.


pik694:
References #476

PR #445, opened 2019-02-15
feat: optional metadata

pthomalla:

PR #483, opened 2019-03-05
fix: allow Watcher to sync to `staging` which has exits

pdobacz:
Otherwise, one would see (on a chain that has exit finalizations), while syncing a Watcher:

```
2019-03-04 16:34:13.883 [error] module=gen_server function=error_info/7 ⋅GenServer :exit_challenger terminating
** (stop) exited in: GenServer.call(OMG.Watcher.ExitProcessor, {:challenge_exits, []}, 5000)
    ** (EXIT) exited in: GenServer.call(OMG.API.State, {:exit_utxos, [%{eth_height: 3861802, utxo_pos: 1000000000000}, %{eth_height: 3861802, utxo_pos: 10000000210000}, %{eth_height: 3861802, utxo_pos: 4000000000001}, %{eth_height: 3861802, utxo_pos: 35003000000000}, %{eth_height: 3861802, utxo_pos: 42006000000000}, %{eth_height: 3861802, utxo_pos: 1000000000}]}, 5000)
            ** (FunctionClauseError) no function clause matching in OMG.API.State.Core.utxo_exists?/2
                lib/state/core.ex:452: OMG.API.State.Core.utxo_exists?(%{eth_height: 3861802, utxo_pos: 1000000000000}, %OMG.API.State.Core{height: 48000, last_deposit_child_blknum: 124001, pending_txs: [], tx_index: 0, utxos: %{{:utxo_position, 29000, 91, 1} => %OMG.API.Utxo{amount: 1, creating_txhash: <<89, 190, 161, 124, 51, 212, 28, 130, 117, 241, 17, 19, 164, 62, 75, 96, 79, 196, 246, 10, 34, 164, 168, 51, 92,
```


pdobacz:
fixes a bug introduced in #445

apps/omg_api/lib/application.ex
```diff
@@ -51,26 +51,23 @@ defmodule OMG.API.Application do
...
         get_events_callback: &OMG.Eth.RootChain.get_in_flight_exit_starts/2,
-        process_events_callback: &ignore_validities/1
+        process_events_callback: &(&1 |> OMG.API.State.exit_utxos() |> ignore_validities())
       ),
       OMG.API.EthereumEventListener.prepare_child(
         service_name: :piggyback,
         synced_height_update_key: :last_piggyback_exit_eth_height,
         get_events_callback: &OMG.Eth.RootChain.get_piggybacks/2,
-        process_events_callback: &ignore_validities/1
+        process_events_callback: &(&1 |> OMG.API.State.exit_utxos() |> ignore_validities())
```
pgebal: I think we can leave  State.exit_utxos call inside ignore_validities. They both take one argument. If we want to make exit_utxos call more explicit we can rename ignore_validities function instead.

apps/omg_watcher/lib/supervisor.ex
```diff
@@ -100,13 +100,11 @@ defmodule OMG.Watcher.Supervisor do
         synced_height_update_key: :last_convenience_exit_processor_eth_height,
         get_events_callback: &Eth.RootChain.get_standard_exits/2,
         process_events_callback: fn exits ->
-          exits =
-            Enum.map(exits, fn %{exit_id: exit_id} ->
-              {:ok, {_, _, _, position}} = OMG.Eth.RootChain.get_standard_exit(exit_id)
-              OMG.API.Utxo.Position.decode(position)
-            end)
+          _ =
```
pgebal: Is it ok that we ignore if those inserts failed?

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -353,7 +354,7 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
              |> Core.invalid_exits(processor)
 
     # exit validly finalizes and continues to not emit any events
-    {:ok, {_, _, spends}, _} = State.Core.exit_utxos([@utxo_pos1], state)
+    {:ok, {_, _, spends}, _} = [@utxo_pos1] |> prepare_exit_finalizations() |> State.Core.exit_utxos(state)
```
pgebal: We don't need that, exit_utxos can handle an enum of Utxo.position.

apps/omg_watcher/lib/supervisor.ex
pdobacz: success of a single insert is checked within. I'll rename these to `xyz!` and get rid of this `_ = ` here, it's misleading

apps/omg_watcher/test/exit_processor/core_test.exs
pdobacz: yes, but the purpose here is to mimic the behavior of `ExitProcessor` a little bit more. It's clunky, but I'd rather do that, than end up in a similar bug in the future

apps/omg_api/lib/application.ex
pdobacz: yes, I did it this way because the function clearly was misnamed. However, I think I'm slightly inclined towards the non-DRY but explicit formulation, hmmm... but ok, let's do what you suggest

PR #484, opened 2019-03-05
fix: fix demos, make default gas/gasprice more affordable in OMG.Eth

pdobacz:
A small patch fixing some stuff that made using our Eth helpers hard on Rinkeby, where Eth isn't that abundant as in `geth --dev`.

Mainly, all our `Eth.Contract.do_sth` calls have a reasonable gas default, instead of `lots_of_gas`

### DIFF
--- a/apps/omg_api/test/integration/fixtures.exs
+++ b/apps/omg_api/test/integration/fixtures.exs
@@ -56,18 +56,21 @@ defmodule OMG.API.Integration.Fixtures do
   end
 
   deffixture alice_deposits(alice, token) do
-    deposit(alice, token)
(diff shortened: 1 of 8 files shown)


apps/omg_api/test/integration/fixtures.exs
```diff
@@ -63,11 +63,14 @@ defmodule OMG.API.Integration.Fixtures do
     deposit(stable_alice, token)
   end
 
-  defp deposit(alice, token) do
+  defp deposit(alice, token_addr) do
```
pik694: An old one. 
I think that the name suggests it only deposits the given token

PR #486, opened 2019-03-06
chore: upgrade MerkleTree

pthomalla:


apps/omg_api/lib/block.ex
```diff
@@ -87,13 +87,28 @@ defmodule OMG.API.Block do
...
+      |> Enum.join()
+
+  # |> Enum.reduce(fn x, acc -> x  <> acc end)
 
   defp merkle_hash(hashed_txs),
-    do: MerkleTree.new(hashed_txs, &Crypto.hash/1, @transaction_merkle_tree_height, @default_leaf).root.value
+    do:
+      MerkleTree.new(hashed_txs,
+        hash_function: &Crypto.hash/1,
+        hash_leaf: false,
```
pnowosie: Ups: we made a typo ;)
Let's wait for https://github.com/yosriady/merkle_tree/pull/13

pdobacz:
For posterity, as this recently raised some confusion:

Here we're abandoning the hacky branch `refactor` of our `merkle_tree` for, in favor of a tidier and less breaking `v1.5.0` release.

The behavior should be exactly the same after this change.

The only difference is that `v1.5.0` introduces a slight perf regression on validating proofs (or rather, doesn't include some optimization that `refactor` branch did), so we need to investigate that and patch on top of `v1.5.0`. Tracked in the issue referenced :point_down:

PR #488, opened 2019-03-06
Omg 381 test in flight and standard exits

pgebal:


apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -346,42 +346,87 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
   end
 
-  # NOTE: write tests - OMG-381
-  # TODO: simplify flow
-  # https://github.com/omisego/elixir-omg/pull/361#discussion_r247485778
-  @spec finalize_in_flight_exits(t(), [map()]) :: {t(), list()}
+  @spec finalize_in_flight_exits(t(), [map()]) ::
+          {:ok, t(), list()} | {:not_piggybacked, list()} | {:unknown_in_flight_exit, MapSet.t(non_neg_integer())}
   def finalize_in_flight_exits(%__MODULE__{in_flight_exits: ifes} = state, finalizations) do
+    with {:ok, ifes_by_contract_id} <- get_all_finalized_ifes_by_ife_contract_id(finalizations, ifes),
```
paulperegud: This TODO was about rewriting `finalize_in_flight_exits(ifes, finalizations)` as `Enum.reduce` with `finalize_single_exit(state, finalization)` over `finalizations`. It would look much simpler.

Consider rewriting it in that style. Tests are already in place, it should be easy.

PR #490, opened 2019-03-07
fix: omg watcher and api should have status as a umbrella dependency

InoMurko:
Without this, they're not started in deployment.


apps/omg_status/lib/status.ex
```diff
@@ -68,9 +68,9 @@ defmodule OMG.Status do
   @spec is_enabled?() :: boolean()
   defp is_enabled?() do
```
paulperegud: How about `is_enabled?()` returning `true | false | nil` and doing `is_enabled()? || false` when calling?

PR #491, opened 2019-03-07
chore: gas price change only, when try to push blocks

pthomalla:


apps/omg_api/lib/block_queue/core.ex
```diff
@@ -203,20 +202,30 @@ defmodule OMG.API.BlockQueue.Core do
...
-           gas_price_adj_params: %GasPriceParams{last_block_mined: {last_parent_height, _mined_block_num}}
+           last_parent_height: last_parent_height
          } = state
-       )
-       when parent_height == last_parent_height,
-       do: state
+       ) do
+    first_blknum = mined_child_block_num + block_interval
+
+    if parent_height == last_parent_height or
```
pdobacz: `parent_height <= last_parent_height` to consider back offs. + add a test where you check that a back of doesn't cause gas price update

apps/omg_api/lib/block_queue/core.ex
```diff
@@ -302,14 +311,11 @@ defmodule OMG.API.BlockQueue.Core do
...
     first_blknum = mined_child_block_num + block_interval
-    block_nums = make_range(first_blknum, formed, block_interval)
-
     _ = Logger.debug("preparing blocks #{inspect(first_blknum)}..#{inspect(formed)} for submission")
 
     blocks
-    |> Map.split(block_nums)
-    |> elem(0)
-    |> Map.values()
+    |> Enum.filter(fn {key, _value} -> key >= first_blknum and key <= formed end)
```
pdobacz: DRY and name this predicate. It is used twice.

apps/omg_api/test/block_queue/core_test.exs
```diff
@@ -48,14 +48,15 @@ defmodule OMG.API.BlockQueue.CoreTest do
...
 
     {:dont_form_block, state} =
       state
-      |> set_ethereum_status(1, 3, false)
+      |> set_ethereum_status(1, 3 * @child_block_interval, false)
 
     # assertions - to be explicit how state looks like
-    assert {1, 3} = state.gas_price_adj_params.last_block_mined
+    child_block_mined = 3 * @child_block_interval
+    assert {1, ^child_block_mined} = state.gas_price_adj_params.last_block_mined
```
pdobacz: oops, testing by reaching into the guts. Is there a behavior we can test here (old problem)?

**EDIT** I see more occurrences of this, so let's split out refactoring the tests to a different PR/ticket

apps/omg_api/test/block_queue/core_test.exs
```diff
@@ -382,44 +383,48 @@ defmodule OMG.API.BlockQueue.CoreTest do
...
-      # here we suddenly mine the child blocks and the GCing should happen
+      # sanity check if we haven't removed blocks to early
+      assert long_size - empty_size >= one_block_size * long_length
+
+      # here we suddenly mine the child blocks and the remove should happen
       long_mined_size =
         long
         |> set_ethereum_status(long_length, (long_length - short_length) * 1000, false)
         |> elem(1)
         |> :erlang.term_to_binary()
```
paulperegud: Use size/1 here, as in other places.

apps/omg_api/test/block_queue/core_test.exs
```diff
@@ -486,43 +492,43 @@ defmodule OMG.API.BlockQueue.CoreTest do
...
-      state1 = %{state | gas_price_adj_params: gas_params}
-      eth_gap = state1.gas_price_adj_params.eth_gap_without_child_blocks
-
-      {:do_form_block, newstate} =
-        state1
-        |> set_ethereum_status(1 + eth_gap, 0, false)
+      state = %{state | gas_price_adj_params: gas_params} |> enqueue_block(<<0>>, 6 * @child_block_interval, 1)
 
-      assert expected_max_price == newstate.gas_price_to_use
+      # many times not main block
```
paulperegud: Despite Ethereum height changing multiple times, gas price does not grow since no new blocks are mined

apps/omg_api/test/block_queue/core_test.exs
```diff
@@ -571,5 +577,21 @@ defmodule OMG.API.BlockQueue.CoreTest do
                assert {:error, :account_locked} = process_submit_result(submission, @account_locked_response, 0)
              end) =~ "[error]"
     end
+
+    @tag fixtures: [:empty_with_gas_params]
+    test "gas price change only, when try to push blocks", %{empty_with_gas_params: state} do
```
pdobacz: I'm not digging into this PR, but please double check if theres a test that `gas price changes only, when ethereum advances`. Also the test case mentioned before `gas price doesn't change when ethereum backs off, even if block in queue`.

Are there also appropriate tests for when to lower the price, according to our revised behaviors we've talked about?

Also make sure that the documentation in `moduledocs` are up to date and reflect the behaviors we're implementing here - when price is lowered, when it is raised.

apps/omg_api/test/block_queue/core_test.exs
paulperegud: Also, isn't this behavior tested below?

apps/omg_api/test/block_queue/core_test.exs
```diff
@@ -571,5 +577,21 @@ defmodule OMG.API.BlockQueue.CoreTest do
...
+      state =
+        Enum.reduce(4..10, state, fn eth_height, state ->
+          {_, state} = set_ethereum_status(state, eth_height, 0, false)
+          assert gas_price == state.gas_price_to_use
+          state
+        end)
+
+      state = state |> Core.enqueue_block(<<0>>, 6 * @child_block_interval, 1)
+      {_, state} = set_ethereum_status(state, 101, 0, false)
+      assert state.gas_price_to_use != gas_price
```
pdobacz: lets assert it changes in the correct direction, i.e. `>`

PR #498, opened 2019-03-11
fix: prefer more frequent and smaller queries for events

pdobacz:
also add some introspection around handling of queries that are dangerously lengthy


jbunce:
@pdobacz could you add an AppSignal label to this please? It would help to visualise over time the values. Should be as straight forward as adding a decorator to that function.

https://docs.appsignal.com/elixir/instrumentation/instrumentation.html#function-decorators

pdobacz:
> @pdobacz could you add an AppSignal label to this please? 

@jbunce absolutely, I suggest a separate PR to that, it might take some back and forth to get the decorator right (looks like there aren't any yet)

PR #500, opened 2019-03-11
Add support for Parity

paulperegud:
Status.
* support for parity quirks
* docs

Integration test fail if run with parity because of postgres sandbox leaking changes to database, compromising isolation between tests. Wrappers tests work fine though.

### DIFF
--- a/apps/omg_eth/lib/eth.ex
+++ b/apps/omg_eth/lib/eth.ex
@@ -30,8 +30,12 @@ defmodule OMG.Eth do
 
   import OMG.Eth.Encoding
 
+  require Logger
+
   @type address :: <<_::160>>
   @type hash :: <<_::256>>
+  @type send_transaction_opts() :: [send_transaction_option()]
+  @type send_transaction_option() :: {:passphrase, binary()}
 
   @spec node_ready() :: :ok | {:error, :geth_still_syncing | :geth_not_listening}
   def node_ready do
@@ -50,6 +54,32 @@ defmodule OMG.Eth do
   """
   @spec syncing?() :: boolean
   def syncing?, do: node_ready() != :ok
+
+  @doc """
+  Send transaction to be singed by a key managed by Ethereum node, geth or parity.
+  For geth, account must be unlocked externally.
+  If using parity, account passphrase must be provided directly or via config.
+  """
+  @spec send_transaction(map(), send_transaction_opts()) :: {:ok, hash()} | {:error, any()}
+  def send_transaction(txmap, opts \\ []) do
+    case backend() do
+      :geth ->
+        with {:ok, receipt_enc} <- Ethereumex.HttpClient.eth_send_transaction(txmap), do: {:ok, from_hex(receipt_enc)}
+
+      :parity ->
+        with {:ok, passphrase} <- get_signer_passphrase(txmap.from),
+             opts = Keyword.merge([passphrase: passphrase], opts),
+             params = [txmap, Keyword.get(opts, :passphrase, "")],
+             {:ok, receipt_enc} <- Ethereumex.HttpClient.request("personal_sendTransaction", params, []) do
+          {:ok, from_hex(receipt_enc)}
+        end
+    end
+  end
+
+  def backend do
+    Application.fetch_env!(:omg_eth, :eth_node)
+    |> String.to_existing_atom()
+  end
 
   def get_ethereum_height do
     case Ethereumex.HttpClient.eth_block_number() do
@@ -84,7 +114,7 @@ defmodule OMG.Eth do
     end
   end
 
-  @spec contract_transact(address, address, binary, [any], keyword) :: {:ok, binary} | {:error, any}
+  @spec contract_transact(address, address, binary, [any], keyword) :: {:ok, hash()} | {:error, any}
   def contract_transact(from, to, signature, args, opts \\ []) do
     data = encode_tx_data(signature, args)
 
@@ -93,8 +123,7 @@ defmodule OMG.Eth do
       |> Map.merge(Map.new(opts))
       |> encode_all_integer_opts()
 
-    with {:ok, txhash} <- Ethereumex.HttpClient.eth_send_transaction(txmap),
-         do: {:ok, from_hex(txhash)}
+    send_transaction(txmap)
   end
 
   defp encode_all_integer_opts(opts) do
@@ -128,8 +157,7 @@ defmodule OMG.Eth do
       |> Map.merge(Map.new(opts))
       |> encode_all_integer_opts()
 
-    with {:ok, txhash} <- Ethereumex.HttpClient.eth_send_transaction(txmap),
-         do: {:ok, from_hex(txhash)}
+    {:ok, _txhash} = send_transaction(txmap)
   end
 
   defp read_contracts_bin!(path_project_root, contract_name) do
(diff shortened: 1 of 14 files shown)


apps/omg_eth/lib/eth.ex
```diff
@@ -33,6 +33,29 @@ defmodule OMG.Eth do
   @type address :: <<_::160>>
   @type hash :: <<_::256>>
 
+  @doc """
+  Send transaction to be singed by a key managed by Ethereum node, geth or parity
```
boolafish: Not really related to this PR. But want to start a discussion do we really want to have Ethereum client node to manage key for us? Shouldn't it ideally all transactions should be signed from e-wallet?

paulperegud: I don't think we have a doc describing key management, and we definitely should have one.

Child chain node needs send new submitBlock transactions every 15 seconds. This key might be managed by ethereum node or by us. From security perspective it is slightly better to have Ethereum node manage it - less risk of exposing it via some crash message. Note that if our node will get owned (RCE), attacked will be able to forge and submit a block what gives him all the tokens.

Watcher does not manage its own keys - it's API is build around preparing all the data needed to do transactions on Ethereum and giving that data to the user, who is managing his own keys.

boolafish: One more thing on coupling wallet with eth node is that it will limit or add more effort if we would like to have an eth client node pools one day. 

Feel like would need to create the account on all eth nodes (not sure is good security practice or not).

apps/omg_eth/lib/eth.ex
```diff
@@ -33,6 +33,29 @@ defmodule OMG.Eth do
...
+        with {:ok, passphrase} <- get_signer_passphrase(txmap.from) do
+          opts = Keyword.merge([passphrase: passphrase], opts)
+          params = [txmap, Keyword.get(opts, :passphrase, "")]
+          Ethereumex.HttpClient.request("personal_sendTransaction", params, [])
+        end
+    end
+  end
+
+  def backend do
+    Application.get_env(:omg_eth, :eth_node, "geth")
```
boolafish: Minor: making missing config failing might be better than default to geth ?

apps/omg_eth/test/support/dev_helpers.ex
```diff
@@ -157,4 +123,59 @@ defmodule OMG.Eth.DevHelpers do
 
     fn -> WaitFor.repeat_until_ok(f) end |> Task.async() |> Task.await(timeout)
   end
+
+  # private
```
boolafish: I thought this comment means all methods below would be private. There are two `def` instead of `defp` here.

.circleci/config.yml
```diff
@@ -136,6 +136,37 @@ jobs:
...
+      - attach_workspace:
+           at: .
+
+      - run: mix local.hex --force
+      - run: mix local.rebar --force
+
+      - run:
+          name: Integration Tests with Parity
+          environment:
+            SIGNER_PASSPHRASE: ThisIsATestnetPassphrase
```
jbunce: Just as a heads up: I know this has been configured for testing but you can set this to be populated from a secret in CircleCI.

apps/omg_eth/lib/eth.ex
paulperegud: Both go-ethereum and parity provide a `signer` application - it is capable of managing keys without being Ethereum client. We can use it.

.circleci/config.yml
paulperegud: Because of an issue integration tests (database sandbox fail or it crashes and we write skipping it) I will not add parity integration test for CI for now.

apps/omg_eth/lib/eth.ex
pdobacz: I concur, the default is defined with `{:system...}` in the config file already. `Application.fetch_env!` sounds ok to me here

apps/omg_eth/lib/eth.ex
```diff
@@ -249,4 +272,16 @@ defmodule OMG.Eth do
...
   end
+
+  defp get_signer_passphrase("0x00a329c0648769a73afac7f9381e08fb43dbea72") do
+    # Parity coinbase address in dev mode, passphrase is empty
+    {:ok, ""}
+  end
+
+  defp get_signer_passphrase(_) do
+    case System.get_env("SIGNER_PASSPHRASE") do
+      nil -> {:error, :please_provide_passphrase_unlocking_ethereum_account_managed_by_parity}
```
pdobacz: how about `{:error, :passphrase_missing}` and `Logger.error("the longer message")`?

apps/omg_eth/lib/eth.ex
```diff
@@ -51,6 +51,31 @@ defmodule OMG.Eth do
...
+  """
+  @spec send_transaction(map()) :: {:ok, hash()} | {:error, any()}
+  def send_transaction(txmap, opts \\ []) do
+    case backend() do
+      :geth ->
+        with {:ok, receipt_enc} <- Ethereumex.HttpClient.eth_send_transaction(txmap), do: {:ok, from_hex(receipt_enc)}
+
+      :parity ->
+        with {:ok, passphrase} <- get_signer_passphrase(txmap.from) do
+          opts = Keyword.merge([passphrase: passphrase], opts)
```
pdobacz: no need for a nested `with`, you can:

```
with a <- f1(),
     b = f2(),
     c <- f3(), do: blah
```

apps/omg_eth/lib/eth.ex
```diff
@@ -51,6 +51,31 @@ defmodule OMG.Eth do
   @spec syncing?() :: boolean
   def syncing?, do: node_ready() != :ok
 
+  @doc """
+  Send transaction to be singed by a key managed by Ethereum node, geth or parity
+  """
+  @spec send_transaction(map()) :: {:ok, hash()} | {:error, any()}
+  def send_transaction(txmap, opts \\ []) do
```
pdobacz: please document the dual way the passphrase is fetched for parity, and the fact that it is not fetched/necessary at all for geth

apps/omg_eth/test/support/dev_mining_helper.ex
```diff
@@ -0,0 +1,78 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.Eth.DevMiningHelper do
+  @moduledoc """
+  Sends small tx every second, causing Ethereum node in `--dev` mode to create new blocks.
+  Basically helps to simulate behavior of `geth --dev --dev.period 1`. Useful with parity.
+  """
+  @devperiod 1000
```
pdobacz: `@devperiod_ms`

apps/omg_eth/test/support/dev_mining_helper.ex
```diff
@@ -0,0 +1,78 @@
...
+    {:noreply, %{state | forced_height: counter + 1}}
+  end
+
+  defp mine(addr, passphrase) do
+    %{from: addr, to: addr, value: OMG.Eth.Encoding.to_hex(1)}
+    |> OMG.Eth.send_transaction(passphrase: passphrase)
+    |> OMG.Eth.DevHelpers.transact_sync!()
+  end
+
+  defp create_spamming_account do
```
pdobacz: word `spamming` might be slightly misleading: `heartbeat`? `tic`?

apps/omg_watcher/lib/application.ex
```diff
@@ -19,6 +19,7 @@ defmodule OMG.Watcher.Application do
   use OMG.API.LoggerExt
 
   def start(_type, _args) do
+    DeferredConfig.populate(:omg_eth)
```
pdobacz: is the ubiquitous `DeferredConfig.populate(:omg_eth)` really necessary in all cases?

docs/install.md
```diff
@@ -45,6 +45,11 @@ sudo apt-get update
 sudo apt-get -y install geth
 ```
 
+### Using Parity
```
pdobacz: a different spot for this documentation perhaps? this is a doc that elaborates on the installation (bare metal) process. I think the doc that takes one through running is more appropriate (`docs/manual_..._.md` or sth).

Also it's worth to mention how one runs our integration tests with parity (`ETH_NODE=parity mix test --only integration` I presume) - I think we have the test-running invocations documented somewhere, please find and update.

For completeness one can put a quick guide to install parity, in _here_, as is done for geth just above

PR #502, opened 2019-03-12
fix: move slow tests to integration and capture exit logs

InoMurko:


apps/omg_api/test/monitor_test.exs
```diff
@@ -86,12 +89,14 @@ defmodule OMG.API.MonitorTest do
...
-    Alarm.raise({:ethereum_client_connection, :erlang.node(), __MODULE__})
-    spawn(fn -> __MODULE__.Mock.terminate(:ethereum_client_connection) end)
+    _ = Process.unlink(monitor_pid)
+    _ = Alarm.raise({:ethereum_client_connection, :erlang.node(), __MODULE__})
+    _ = spawn(fn -> __MODULE__.Mock.terminate(:ethereum_client_connection) end)
     assert pull_state_and_find_timer(monitor_pid, 1000)
   end
 
+  @tag :capture_log
+  @tag :integration
```
pnowosie: Making entire module (all test) as integration and moving file to ‚/integration’ would be cleaner for me.

apps/omg_status/test/omg_status/alarms_test.exs
```diff
@@ -37,7 +37,7 @@ defmodule OMG.Status.Alert.AlarmTest do
   end
 
   @tag timeout: 240_000
-  @tag :slow
+  @tag :integration
```
pnowosie: As above:
> Making entire module (all test) as integration and moving file to ‚/integration’ would be cleaner for me.

apps/omg_api/test/monitor_test.exs
InoMurko: Hm... This test isn't really _integration_ it's only tagged this way because it's slow and we're not able to parallel execute coveralls (_not yet at least_). So marking all of them is... I don't think it makes a difference at this point.

apps/omg_status/test/omg_status/integration/alarms_test.exs
```diff
@@ -20,6 +20,8 @@ defmodule OMG.Status.Alert.AlarmTest do
     Alarm.clear_all()
   end
 
+  @moduletag :integration
```
pnowosie: I'd like to move both `moduletags` above just below `aliases` - to keep consistent with other files

PR #503, opened 2019-03-12
fix: fix reporting unavailable piggyback as available

### LINKED ISSUE #495: Piggybacked outputs are reported as available when calling status.get endpoint in Watcher
After in-flight exit output is piggybacked, status.get endpoint still reports it as available for piggybacking.

To reproduce that follow docs/demo04 and after starting an in-flight exit:
1. piggyback onto one of the outputs
2. wait until watcher recognizes the piggyback
3. call status.get endpoint and look for event that reports piggybacked output as still available

pgebal:
Closes  #495

### DIFF
--- a/apps/omg_watcher/lib/exit_processor/core.ex
+++ b/apps/omg_watcher/lib/exit_processor/core.ex
@@ -769,7 +769,7 @@ defmodule OMG.Watcher.ExitProcessor.Core do
       inputs =
         Transaction.get_inputs(tx.raw_tx)
         |> Enum.with_index()
-        |> Enum.filter(fn {_input, index} -> InFlightExitInfo.is_piggybacked?(ife, index) end)
+        |> Enum.filter(fn {_input, index} -> InFlightExitInfo.is_input_piggybacked?(ife, index) end)
 
       {ife, inputs}
     end)
@@ -912,7 +912,7 @@ defmodule OMG.Watcher.ExitProcessor.Core do
   end
 
   @spec get_ifes_to_piggyback(ExitProcessor.Request.t(), __MODULE__.t()) ::
-          list({InFlightExitInfo.t(), Transaction.Signed.t()})
+          list(InFlightExitInfo.t())
   defp get_ifes_to_piggyback(
          %ExitProcessor.Request{blocks_result: blocks},
          %__MODULE__{in_flight_exits: ifes}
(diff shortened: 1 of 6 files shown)


apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -911,44 +911,54 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
   defp get_ifes_to_piggyback(
          %ExitProcessor.Request{blocks_result: blocks},
          %__MODULE__{in_flight_exits: ifes}
        ) do
     known_txs = get_known_txs(blocks)
 
     ifes
-    |> Map.values()
-    |> Stream.map(fn %InFlightExitInfo{tx: signed_tx} -> signed_tx end)
+    |> Stream.filter(fn {_, %InFlightExitInfo{is_active: is_active}} -> is_active end)
```
pdobacz: I think the flows can be cleaner here:
1/ keep `Map.values`, as you don't need the keys ever
2/ keep passing around `ife`, instead of the tuple `ife, signed_tx`
3/ pattern-match the `signed_tx` field out of `ife` only when needed, instead of pattern-matching it out of the tuple.

The tuple passed through the pipeline is redundant (ife contains signed tx)

apps/omg_watcher/lib/exit_processor/in_flight_exit_info.ex
```diff
@@ -215,6 +215,14 @@ defmodule OMG.Watcher.ExitProcessor.InFlightExitInfo do
     end
   end
 
+  def is_input_piggybacked?(%__MODULE__{} = ife, index) when is_integer(index) and index < @max_inputs do
```
pdobacz: if we're positive about doing this refactor then maybe `is_piggybacked` can be private, and the verbose function be used in consumers of this module

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -773,6 +773,67 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
+      finalization = %{in_flight_exit_id: @non_zero_exit_id, output_index: 0}
+
+      {:ok, processor, _} = Core.finalize_in_flight_exits(processor, [finalization])
+
+      assert {:ok, []} =
+               %ExitProcessor.Request{blknum_now: 1000, eth_height_now: 5}
+               |> invalid_exits_filtered(processor, only: [Event.PiggybackAvailable])
+    end
+
+    test "in-flight exit without piggybacked outputs does not emit piggyback available event after it's finalized" do
```
pdobacz: but is there a reason this cannot work and be tested in the logic level here? It looks like a simple condition that we can handle in our Core and test it here just fine. Looks like it'd be worth for completeness

apps/omg_watcher/test/integration/in_flight_exit_test.exs
```diff
@@ -295,8 +295,7 @@ defmodule OMG.Watcher.Integration.InFlightExitTest do
              "byzantine_events" => [
```
pdobacz: how about the comment we talked about in the upper assertion? (why only one pb available?)

docs/api_specs/status_events_specs.md
```diff
@@ -132,8 +132,7 @@ txbytes | Hex encoded string | The in-flight transaction that the event relates
 }
 ```
 
-An in-flight exit has been started and can be piggybacked. If all inputs are owned by the same address, then `available_inputs` will not be present.
-<aside class="warning"> Not Implemented Yet.</aside>
+An in-flight exit has been started and can be piggybacked. If all inputs are owned by the same address, then `available_inputs` will not be present. This event is reported only for in-flight exits from transactions that have not been included in a block.
```
pdobacz: ah when we're at it - document the behavior your fixing in this PR:
1/ piggybacked inputs/outputs don't show up as available
2/ finalized ones don't either

apps/omg_watcher/test/exit_processor/core_test.exs
pgebal: As discussed, I removed that test.

apps/omg_watcher/lib/exit_processor/in_flight_exit_info.ex
pgebal: We need is_piggybacked to be public. There is a use case in core where we don't know if we are asking for tx inputs or outputs. See `with` in `finalize_in_flight_exits`.

pdobacz: ok, nvm then, we can reassess that when we go about refactoring all this

PR #504, opened 2019-03-12
feature: observe the size of the mailbox of the processes

InoMurko:
This gives us Appsignal metrics that tell us how many messages are waiting to be processed in the queue of the process.
Example setup (it's different on the watcher node):
```
-
  title: 'VM dashboard'
  graphs:
    -
      title: Elixir.OMG.API.BlockQueue.Server.Recorder
      kind: gauge
      format: number
      draw_null_as_zero: true
      fields:
        - Elixir.OMG.API.BlockQueue.Server.Recorder
      tags:
        node: '*'
    -
      title: Elixir.OMG.API.RootChainCoordinator.Recorder
      kind: gauge
      format: number
      draw_null_as_zero: true
      fields:
        - Elixir.OMG.API.RootChainCoordinator.Recorder
      tags:
        node: '*'
    -
      title: exiter.Recorder
      kind: gauge
      format: number
      draw_null_as_zero: true
      fields:
        - exiter.Recorder
      tags:
        node: '*'
    -
      title: in_flight_exit.Recorder
      kind: gauge
      format: number
      draw_null_as_zero: true
      fields:
        - in_flight_exit.Recorder
      tags:
        node: '*'
    -
      title: piggyback.Recorder
      kind: gauge
      format: number
      draw_null_as_zero: true
      fields:
        - piggyback.Recorder
      tags:
        node: '*'
    -
      title: depositor.Recorder
      kind: gauge
      format: number
      draw_null_as_zero: true
      fields:
        - depositor.Recorder
      tags:
        node: '*'

```
The results are gathered every 5 seconds, which is rather a lot for the type of metric we're gathering but ... we'll see how it turns out.
```
DEPOSITOR_INTERVAL
IN_FLIGHT_EXIT_INTERVAL
PIGGYBACK_INTERVAL
EXITER_INTERVAL
ELIXIR.OMG.API.BLOCKQUEUE.SERVER_INTERVAL
ELIXIR.OMG.API.ROOTCHAINCOORDINATOR_INTERVAL
and others in watcher...
```

### DIFF
--- a/apps/omg_api/lib/block_queue.ex
+++ b/apps/omg_api/lib/block_queue.ex
@@ -30,6 +30,7 @@ defmodule OMG.API.BlockQueue do
   alias OMG.API.Block
   alias OMG.API.BlockQueue.Core
   alias OMG.API.BlockQueue.Core.BlockSubmission
+  alias Status.Metric.Recorder
 
   @type eth_height() :: non_neg_integer()
   @type hash() :: BlockSubmission.hash()
@@ -118,6 +119,14 @@ defmodule OMG.API.BlockQueue do
 
       interval = Application.fetch_env!(:omg_api, :block_queue_eth_height_check_interval_ms)
       {:ok, _} = :timer.send_interval(interval, self(), :check_ethereum_status)
+      parent = self()
+
+      {:ok, _} =
+        Recorder.start_link(%Recorder{
+          name: __MODULE__.Recorder,
+          fn: fn -> Process.info(parent, :message_queue_len) |> elem(1) end,
+          reporter: &Appsignal.set_gauge/3
+        })
 
       _ = Logger.info("Started BlockQueue")
       {:noreply, state}
--- a/apps/omg_api/lib/block_queue.ex
+++ b/apps/omg_api/lib/block_queue.ex
@@ -120,11 +120,7 @@ defmodule OMG.API.BlockQueue do
       interval = Application.fetch_env!(:omg_api, :block_queue_eth_height_check_interval_ms)
       {:ok, _} = :timer.send_interval(interval, self(), :check_ethereum_status)
 
-      {:ok, _} =
-        Recorder.start_link(%Recorder{
-          name: __MODULE__.Recorder,
-          parent: self()
-        })
+      {:ok, _} = Recorder.start_link(%Recorder{name: __MODULE__.Recorder, parent: self()})
 
       _ = Logger.info("Started BlockQueue")
       {:noreply, state}
--- a/apps/omg_api/lib/block_queue.ex
+++ b/apps/omg_api/lib/block_queue.ex
@@ -30,7 +30,7 @@ defmodule OMG.API.BlockQueue do
   alias OMG.API.Block
   alias OMG.API.BlockQueue.Core
   alias OMG.API.BlockQueue.Core.BlockSubmission
-  alias Status.Metric.Recorder
+  alias OMG.API.Recorder
 
   @type eth_height() :: non_neg_integer()
   @type hash() :: BlockSubmission.hash()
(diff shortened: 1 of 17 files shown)


apps/omg_api/lib/block_queue.ex
```diff
@@ -118,6 +119,14 @@ defmodule OMG.API.BlockQueue do
 
       interval = Application.fetch_env!(:omg_api, :block_queue_eth_height_check_interval_ms)
       {:ok, _} = :timer.send_interval(interval, self(), :check_ethereum_status)
+      parent = self()
+
+      {:ok, _} =
```
pdobacz: I'm hesitant, but maybe we could DRY this invocation to a one liner like `OMG.API.Recorder.start[_link](name: only_the_name_varies)`

apps/omg_api/lib/block_queue.ex
```diff
@@ -118,6 +119,14 @@ defmodule OMG.API.BlockQueue do
...
+        Recorder.start_link(%Recorder{
```
pdobacz: is doing `.start_link` safe for a seemingly non-critical service? Do we want the recorded process to go down when the `Recorder` does and vice versa?

InoMurko: The operation that Recorder does is not complex and failures are preventable - but it does expect the parent (in this example BlockQueue) to be running and provide his pid. As links are bidirectional the only other way would be to keep the Recorder pid in state and forcefully cleanup in terminate.

apps/omg_api/lib/block_queue.ex
InoMurko: Yeah. That might be better. I presume we would want to track other things on a per process basis.

apps/omg_api/lib/block_queue.ex
InoMurko: There's a third way. We keep the functionality in the process and don't have the Recorderer at all.

pdobacz: gotcha, let's leave as is then :+1:

pik694:
The PR is technically correct, but I do upvote @pdobacz 's comments and especially 
[this one](https://github.com/omisego/elixir-omg/pull/504#discussion_r265002862)

apps/omg_watcher/lib/block_getter.ex
```diff
@@ -40,18 +41,60 @@ defmodule OMG.Watcher.BlockGetter do
     GenServer.call(__MODULE__, :get_events)
   end
 
-  @spec download_block(pos_integer()) :: Core.validate_download_response_result_t()
-  defp download_block(requested_number) do
-    {:ok, {requested_hash, block_timestamp}} = Eth.RootChain.get_child_chain(requested_number)
-    response = Client.get_block(requested_hash)
+  def start_link(_args) do
```
InoMurko: I only moved functions around. API at the top, callbacks and finally private functions.

apps/omg_api/lib/ethereum_event_listener.ex
```diff
@@ -80,6 +81,15 @@ defmodule OMG.API.EthereumEventListener do
 
     {:ok, _} = schedule_get_events()
     :ok = RootChainCoordinator.check_in(height_to_check_in, service_name)
+
+    name =
+      service_name
+      |> to_string
```
pik694: very minor: `to_string()` - we rather use functions with parenthesis

pik694: in `OMG.API.Recorder` you used `Atom.to_string()` ;)

apps/omg_db/lib/leveldb_server.ex
```diff
@@ -40,13 +40,21 @@ defmodule OMG.DB.LevelDBServer do
...
   end
 
-  def init(%{db_path: db_path}) do
+  def init(%{db_path: db_path, name: name}) do
     # needed so that terminate callback is called on normal close
     Process.flag(:trap_exit, true)
 
+    name =
+      name
+      |> to_string
```
pik694: same here

PR #505, opened 2019-03-13
refactor: make in-flight exit unit test behavioral

pgebal:


pgebal:
The only visible effects of ife finalization in watcher are: something that we can recover from is persisted in db (we already have a test for that),  PiggybackAvailable events do not notify about finalized outputs (we test that that in the other PR) and ife is inactive after all piggybacked outputs are finalized (we test that in this PR).

apps/omg_watcher/test/exit_processor/core_test.exs
```diff
@@ -1789,19 +1788,15 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
-      finalization2 = %{in_flight_exit_id: ife_id, output_index: 2}
+      [_] = Core.get_active_in_flight_exits(processor)
 
-      {:ok, _, [{:put, :in_flight_exit_info, {_, exit_info}}]} =
-        Core.finalize_in_flight_exits(processor, [finalization2])
+      {:ok, processor, [{:put, :in_flight_exit_info, _}]} =
+        Core.finalize_in_flight_exits(processor, [%{in_flight_exit_id: ife_id, output_index: 2}])
 
-      assert expect_finalized_outputs(exit_info, [1, 2], [])
+      assert [] == Core.get_active_in_flight_exits(processor)
```
pdobacz: ah I only noticed this now, `get_active_in_flight_exits` and `get_in_flight_exits` - it seems like only one of these should be `def` and tested here. I think `get_in_flight_exits` should be removed and all the `CoreTest` should be adapted to test the other. `get_active_in_flight_exits` seems to be the only one used in the `ExitProcessor`

PR #506, opened 2019-03-13
refactor: decouple watcher from childchain

InoMurko:
Long term, this gives us decoupling the watcer and childchain, the ability to pack and deploy one without the other, clean application borders and dependency (startup) trees.

The new umbrella app, omg_sync, is not an OTP app (has no supervision tree or processes), but merely a stitching point so that modules can be reused.

Phase 2 is cleanup and further configuration consolidation - so that we avoid mixing application configurations.


InoMurko:
I've noticed _Watcher.Eventer_. This needs clarification.

pdobacz:
shall we close this in favor of #529 and the planned followups to that refactor?

PR #508, opened 2019-03-13
fix: allow processing exits when they were immediately challenged

pdobacz:
otherwise, when a `ExitProcessor.Core.new_exits/2` is called, when an exit already had been challenged,
there's a failed attempt to decode a `Utxo.Position.decode(0)`

The reason is that the "current" contract state of the exit already is wiped clean.

Instead, to limit the fix to the Watcher we take the necessary details of an exit from the relevant `call_data`

### DIFF
--- a/apps/omg_watcher/lib/db/eth_event.ex
+++ b/apps/omg_watcher/lib/db/eth_event.ex
@@ -74,8 +74,13 @@ defmodule OMG.Watcher.DB.EthEvent do
   """
   @spec insert_exits!([non_neg_integer()]) :: :ok
   def insert_exits!(exits) do
-    Enum.each(exits, &(&1 |> Utxo.Position.decode() |> insert_exit!()))
+    exits
+    |> Stream.map(&utxo_pos_from_exit_event/1)
+    |> Enum.each(&insert_exit!/1)
   end
+
+  @spec utxo_pos_from_exit_event(%{call_data: %{utxo_pos: pos_integer()}}) :: Utxo.Position.t()
+  defp utxo_pos_from_exit_event(%{call_data: %{utxo_pos: utxo_pos}}), do: Utxo.Position.decode(utxo_pos)
 
   @spec insert_exit!(Utxo.Position.t()) :: {:ok, %__MODULE__{}} | {:error, Ecto.Changeset.t()}
   defp insert_exit!(Utxo.position(blknum, txindex, _oindex) = position) do
--- a/apps/omg_watcher/lib/exit_processor/core.ex
+++ b/apps/omg_watcher/lib/exit_processor/core.ex
@@ -148,12 +148,17 @@ defmodule OMG.Watcher.ExitProcessor.Core do
     new_exits_kv_pairs =
       new_exits
       |> Enum.zip(exit_contract_statuses)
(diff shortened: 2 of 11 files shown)


apps/omg_watcher/lib/db/eth_event.ex
```diff
@@ -74,9 +74,14 @@ defmodule OMG.Watcher.DB.EthEvent do
   """
   @spec insert_exits!([non_neg_integer()]) :: :ok
   def insert_exits!(exits) do
-    Enum.each(exits, &(&1 |> Utxo.Position.decode() |> insert_exit!()))
+    exits
+    |> Enum.map(&parse_contract_exit_event/1)
+    |> Enum.each(&insert_exit!/1)
   end
 
+  @spec parse_contract_exit_event(%{call_data: %{utxo_pos: pos_integer()}}) :: Utxo.Position.t()
```
pnowosie: Not sure name is adequate - I thought `call_data` is not event... and since it's interested in one field maybe `get_utxo_pos` :question:

apps/omg_watcher/lib/db/eth_event.ex
```diff
@@ -74,9 +74,14 @@ defmodule OMG.Watcher.DB.EthEvent do
   """
   @spec insert_exits!([non_neg_integer()]) :: :ok
   def insert_exits!(exits) do
-    Enum.each(exits, &(&1 |> Utxo.Position.decode() |> insert_exit!()))
+    exits
+    |> Enum.map(&parse_contract_exit_event/1)
```
pnowosie:
```suggestion
    |> Stream.map(&parse_contract_exit_event/1)
```

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -148,12 +148,17 @@ defmodule OMG.Watcher.ExitProcessor.Core do
     new_exits_kv_pairs =
       new_exits
       |> Enum.zip(exit_contract_statuses)
-      |> Enum.map(fn {%{eth_height: eth_height}, {address, token, amount, utxo_pos} = contract_status} ->
-        {Utxo.Position.decode(utxo_pos),
+      |> Enum.map(fn {event, contract_status} ->
+        %{eth_height: eth_height, call_data: %{utxo_pos: utxo_pos, output_tx: tx_bytes}} = event
```
pnowosie: 1. `event` seems to be use just here - consider destructure it in params
1. it might be just Watcher's perspective, but please rename `tx_bytes -> txbytes`

apps/omg_watcher/lib/db/eth_event.ex
pdobacz: `call_data` is what we call the "extension" of the event data taken from the transaction's call data. I'll clarfify the names

apps/omg_watcher/lib/exit_processor/core.ex
pdobacz: 1. if I pattern-match the fields of `event` in the `fn` params the line becomes too long, but the formatter can't cope with that
2. :+1:

pdobacz: 2. I'll rename only here to avoid an all-file diff

Issue #479, opened 2019-03-05
Using {:system, "DATABASE_URL"} for your :url configuration is deprecated.


jbunce:
```
watcher-0 watcher warning: Using {:system, "DATABASE_URL"} for your :url configuration is deprecated.
watcher-0 watcher
watcher-0 watcher Instead define an init/2 callback in your repository that sets
watcher-0 watcher the URL accordingly from your system environment:
watcher-0 watcher
watcher-0 watcher     def init(_type, config) do
watcher-0 watcher       {:ok, Keyword.put(config, :url, System.get_env("DATABASE_URL"))}
watcher-0 watcher     end
watcher-0 watcher
watcher-0 watcher   lib/ecto/repo/supervisor.ex:55: Ecto.Repo.Supervisor.compile_config/2
watcher-0 watcher   lib/db/repo.ex:16: (module)
watcher-0 watcher   (elixir) src/elixir_compiler.erl:79: :elixir_compiler.dispatch/4
watcher-0 watcher   (elixir) src/elixir_compiler.erl:63: :elixir_compiler.compile/3
watcher-0 watcher   (elixir) src/elixir_module.erl:309: :elixir_module.eval_form/6
```

purbanow:
@jbunce , are you getting this message during a normal start of the Watcher or you did something extra?

jbunce:
@purbanow this comes from a `mix compile`

`MIX_ENV=prod`

```
childchain-0 childchain ==> omg_watcher
childchain-0 childchain Compiling 54 files (.ex)
childchain-0 childchain warning: Using {:system, "DATABASE_URL"} for your :url configuration is deprecated.
childchain-0 childchain
childchain-0 childchain Instead define an init/2 callback in your repository that sets
childchain-0 childchain the URL accordingly from your system environment:
childchain-0 childchain
childchain-0 childchain     def init(_type, config) do
childchain-0 childchain       {:ok, Keyword.put(config, :url, System.get_env("DATABASE_URL"))}
childchain-0 childchain     end
childchain-0 childchain
childchain-0 childchain   lib/ecto/repo/supervisor.ex:55: Ecto.Repo.Supervisor.compile_config/2
childchain-0 childchain   lib/db/repo.ex:16: (module)
childchain-0 childchain   (elixir) src/elixir_compiler.erl:79: :elixir_compiler.dispatch/4
childchain-0 childchain   (elixir) src/elixir_compiler.erl:63: :elixir_compiler.compile/3
[...]

PR #517, opened 2019-03-15
OMG-413 fix: enable PORT env variable for :dev environment + update docs

### LINKED ISSUE #451: Config override issue with omg_watcher
As part of the `~/config_watcher.exs` generation and setup for the `omg_watcher`, I wanted to customize the port that the watcher listens on.

When I specified the following in my `~/config_watcher.exs`:

```
use Mix.Config
config :omg_eth,
  contract_addr: "0x21c4127004b4ae732fc407dffe65d28b1d0b3594",
  txhash_contract: "0xbb2a364ded8099a9c5d11c9388d8179790fe1c497fc76796f57e159062738b95",
  authority_addr: "0xabe7a779d6b6f9cf254b62075605ef5e7129f998"

config :omg_db,
  leveldb_path: Path.join([System.get_env("HOME"), ".omg/data_watcher"])

config :omg_watcher, OMG.Watcher.Web.Endpoint,
  http: [port: 34727],
  url: [host: "localhost", port: 34727],
  debug_errors: true,
  code_reloader: true,
  check_origin: false,
  watchers: [],
  server: true
```
The following error occurred when attempting to POST to the Watcher:

```
2019-02-18 14:47:16.149 [error] ⋅GenServer OMG.Watcher.Eventer terminating
** (ArgumentError) no :pubsub server configured at, please setup :pubsub in your config.

By default this looks like:

    config :my_app, MyApp.PubSub,
      ...,
      pubsub: [name: MyApp.PubSub,
      adapter: Phoenix.PubSub.PG2]
(phoenix) lib/phoenix/endpoint.ex:510: Phoenix.Endpoint.__pubsub_server__!/1
    (omg_watcher) lib/web/endpoint.ex:16: OMG.Watcher.Web.Endpoint.broadcast!/3
    (omg_watcher) lib/eventer.ex:47: anonymous fn/1 in OMG.Watcher.Eventer.handle_cast/2
(elixir) lib/enum.ex:737: Enum."-each/2-lists^foreach/1-0-"/2
    (elixir) lib/enum.ex:737: Enum.each/2
    (omg_watcher) lib/eventer.ex:45: OMG.Watcher.Eventer.handle_cast/2
    (stdlib) gen_server.erl:616: :gen_server.try_dispatch/4
(stdlib) gen_server.erl:686: :gen_server.handle_msg/6
(stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
```

To fix this I had to add the following to my `~/config_watcher.exs` config:

`pubsub: [name: OMG.Watcher.PubSub, adapter: Phoenix.PubSub.PG2]`

This is already defined in the watchers config located here:

`apps/omg_watcher/config/config.exs`

It would appear that this config file is getting ignored/overridden by my updated `config_watcher.exs` file.

Is this the expected behaviour or should the original settings within `apps/omg_watcher/config/config.exs` be preserved if no overrides have been set in a custom config?

pdobacz:
before that customizing these ports in `dev` environment required one to override the whole Endpoint config in ~/config.exs,
which is impossible due to the way configs work in Elixir (in case of ~/config.exs it is **not** deeply merged)

fixes #451

### DIFF
--- a/apps/omg_rpc/config/dev.exs
+++ b/apps/omg_rpc/config/dev.exs
@@ -7,7 +7,7 @@ use Mix.Config
 # watchers to your application. For example, we use it
 # with webpack to recompile .js and .css sources.
 config :omg_rpc, OMG.RPC.Web.Endpoint,
-  http: [port: 9656],
+  http: [port: {:system, "PORT", 9656, {String, :to_integer}}],
   url: [host: "localhost", port: 9656],
   debug_errors: true,
   check_origin: false,
--- a/apps/omg_watcher/config/dev.exs
+++ b/apps/omg_watcher/config/dev.exs
@@ -7,7 +7,7 @@ use Mix.Config
 # watchers to your application. For example, we use it
 # with brunch.io to recompile .js and .css sources.
 config :omg_watcher, OMG.Watcher.Web.Endpoint,
-  http: [port: 7434],
+  http: [port: {:system, "PORT", 7434, {String, :to_integer}}],
   url: [host: "localhost", port: 7434],
   debug_errors: true,
   code_reloader: true,
--- a/docs/manual_service_startup.md
+++ b/docs/manual_service_startup.md
@@ -21,7 +21,9 @@ Such configuration must become part of the [Mix configuration](https://hexdocs.p
 1. (**Watcher only**) Configure PostgreSQL for `WatcherDB` database
 1. (**Watcher only**) Acquire the configuration file with root chain deployment data
 1. (**Watcher only**, optional) If running on the same machine as the child chain server, customize the location of `OMG.DB` database folder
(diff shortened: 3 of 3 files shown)


jbunce:
Adding the ability to run the services on custom ports makes sense. I'm not sure if this is the best way to implement though. Take a look at this issue which could be related: https://github.com/omisego/elixir-omg/issues/479

pdobacz:
> Adding the ability to run the services on custom ports makes sense. I'm not sure if this is the best way to implement though. Take a look at this issue which could be related: #479

I'm not sure either, but it is what we use in `:prod`, so I'm just applying the same mechanism for `:dev`. The deprecation warning you mention is Ecto-specific as is the fix that it proposes.

PR #525, opened 2019-03-18
Update and clear API specs about transactions metadata

pnowosie:
Missed API specs update


apps/omg_watcher/priv/swagger/informational_api_specs.yaml
```diff
@@ -70,7 +70,7 @@ paths:
       produces:
       - application/json
       requestBody:
-        $ref: '#/requests/GetTransactionsOpId'
+        $ref: '#/requests/GetTransactionsAllOpId'
       responses:
         200:
           $ref: '#/responses/GetTransactionsOpId'
```
pik694: Why this did not change?

pnowosie: Both routes `/account.get_transactions` and `/transaction.all` are served by the same controller function, so they should (and now) have the same request / response specs

PR #529, opened 2019-03-19
refactor: split the common part to omg from omg_api

pdobacz:
This intends to put all the functionality that is shared between the Child Chain Server and Watcher apps into another umbrella subapp (`:omg`), following the request filed in #506.

Still pending, probably better to tackle separately:
  - `OMG.State` still refers to the non-opaque API's of BlockQueue and FreshBlocks, which are child chain-only, hence in `:omg_api`. Think about how to solve their communications
  - there are some places where `OMG.API.Core.recover_tx` is used, where it shouldn't, but now it seems this can be dropped in favor of using `OMG.State.Transaction.Recovered.recover_from` instead, with some additional refactors
  - since `OMG.State` `OMG.EthereumEventListener` and `OMG.RootChainCoordinator` got sort of "promoted" to shared standalone modules, they deserve some basic happy path tests for their "imperative shell" (e.g. everything that happens outside of their `.Core` modules), so that they are covered in `omg`, not only in their client apps (`omg_api` and `omg_watcher`
  - rename `omg_api` to `omg_child_chain` or similar
  - some other renames would become due, especially `OMG.State` -> `OMG.Ledger` or similar. Also I'm thinking about ditching the `.Core` suffix for modules holding the stateless logic, but I need to ponder this some more


InoMurko:
Hey @pdobacz this is great we're tackling this!
A few points:
- You've mentioned yourself that `we're using a shared (in `omg` app) config entry` that actually is `never read in `:omg` app per se`. I think if it's never read and never used, it just causes confusion and it doesn't really enforce us catching diverging configuration. The question is really, why are they configurable and if they have to be because of deployments to different environemnts, we need to find another way of checking they match (and/or prevent them to be changed). Probably the best way at the release creation period. That should be a separate PR though. For now, I would just remove the duplicate ENV settings.
- Another step in refactoring would be to separate `omg_api` from being just the API layer and pulling out everything thats plasma related into a separate application. It's just for the sake of isolation and clarity. This way we prevent :omg_api from exiting completely, as it happens at the moment.
(same applies for :omg_watcher)
- :omg_rpc (afaik) the supervision tree is useless and not being used. The endpoints and the router... I might be missing something but if I'm not it could get removed?
- You've included `OMG.Watcher.Eventer` in the new :omg api, this means you've coupled :omg with :omg_watcher. I don't think this is the best approach. I'm not entirely sure what the purpose of Eventer is at the moment. But the only way it compiles is because there's a dependency to :omg_eth, and and :omg_api needs :omg_eth. This means that if you want to make a release for the watcher wihout :omg_api, you would still need the modules from api, otherwise it won't compile.
- The way `OMG.State` was pulled into :omg isn't entirely clean, because it references `OMG.API.BlockQueue` and `OMG.API.FreshBlocks`. Which makes me think, the `State` has functionality that isn't going to be used in the Watcher. Might be useful to tear it apart of make the State parametrized. 

Other things I've mentioned to you I think it'll give us much value is 
- Ecto needs migrations. The reasons were shown just as the new metadata columns were included in the watcher. If we had migrations, they would run before the boot of the app and we would have the problems of missing columns. Ecto also needs a update. This might be a bit more difficult task because we need to do migrations after we've been running without them for quite a while and the data is there...
- I haven't seen the data that would tell me that the way LevelDB is used is causing us problems. If it does and since LevelDB allows only one connection. Might be valuable to swap it with RocksDB + Poolboy. Maybe even before we introduce migrations for LevelDB.
- When the application looses connectivity towards `geth` the status app raises an alarm. Those alarms should be exposed to devops and developers so that they can act on it. At the moment, the only way how they can figure out things don't work is when they break completely. 
- My opinion is that much of the design is tightly coupled with Geth (by pulling). It might be valuable to create a POC and check how reliable filters are (pub/sub).

pdobacz:
@InoMurko Thank you for this. Ok, let's see:

> * You've mentioned yourself that `we're using a shared (in ` (...) For now, I would just remove the duplicate ENV settings.

which duplicate do you mean? this one particular entry is defined once, only for `omg`

and as a future PR's fix for this, maybe we leave this configuration variable in the shared `omg` and expose it via some API (e.g. `OMG.Config.deposit_finality_margin/0` or similar) for the `omg_child_chain` and `omg_watcher` to use? In any case - separate PR.

> * Another step in refactoring would be to separate `omg_api` from being just the API layer and pulling out everything thats plasma related into a separate application. It's just for the sake of isolation and clarity. This way we prevent :omg_api from exiting completely, as it happens at the moment.
>   (same applies for :omg_watcher)

In case of `omg_api`: this requires discussion, what do you mean by "plasma related" exactly, and what would be the benefit of keeping the API running, if the plasma construction is broken beyond repair.

In case of `omg_watcher` that would be definitely useful :+1:.

> * :omg_rpc (afaik) the supervision tree is useless and not being used. The endpoints and the router... I might be missing something but if I'm not it could get removed?

`omg_rpc` is the web layer of `omg_api` (in the future we have a plan to share it with `omg_watcher`) keeping all stuff related to Phoenix/cowboy specifically. The intention is to have this layer application-agnostic and have the application-aware part plugged into this, but we're sort of halfway through here.

> * You've included `OMG.Watcher.Eventer` in the new :omg api, this means you've coupled :omg with :omg_watcher.

Only as the process name, to `GenServer.cast` to it. The mechanism how `OMG.Watcher.Eventer` is plugged in can be improved and elixirized, true that.

> I don't think this is the best approach. I'm not entirely sure what the purpose of Eventer is at the moment. 

This is the adapter to the system that would push events to Watcher's client, currently Phoenix.Channels.

> But the only way it compiles is because there's a dependency to :omg_eth, and :omg_api needs :omg_eth. This means that if you want to make a release for the watcher wihout :omg_api, you would still need the modules from api, otherwise it won't compile.

I'm either missing something or this is not the case here.

> * The way `OMG.State` was pulled into :omg isn't entirely clean, because it references `OMG.API.BlockQueue` and `OMG.API.FreshBlocks`. Which makes me think, the `State` has functionality that isn't going to be used in the Watcher. Might be useful to tear it apart or make the State init parametrized.

see my opening comment to the PR, known issue, WIP. Init parametrization might be what's missing here.

> 
> Other things I've mentioned to you I think it'll give us much value is
> 
> * Ecto needs migrations. The reasons were shown just as the new metadata columns were included in the watcher. If we had migrations, they would run before the boot of the app and we would have the problems of missing columns. Ecto also needs a update.

Hm? Thought we do have migrations, they just weren't applied in the recent `metadata` case. In any case, this feels like a separate issue, or does it change the way this refactor should be handled?

> This might be a bit more difficult task because we need to do migrations after we've been running without them for quite a while and the data is there...

not yet, we're still in the phase, where the data is expendable.

> * I haven't seen the data that would tell me that the way LevelDB is used is causing us problems. If it does and since LevelDB allows only one connection. Might be valuable to swap it with RocksDB + Poolboy. Maybe even before we introduce migrations for LevelDB.

I'm not sure I understand this, but this swapping out is on the roadmap, tracked separately.

> * When the application looses connectivity towards `geth` the status app raises an alarm. Those alarms should be exposed to devops and developers so that they can act on it. At the moment, the only way how they can figure out things don't work is when they break completely.

separate issue or how does this weigh in on the refactor at hand?

> * My opinion is that much of the design is tightly coupled with Geth (by pulling). It might be valuable to create a POC and check how reliable filters are (pub/sub).

again, separate issue?

pdobacz:
OTOH, my question to you @InoMurko is how well does this handle what #506 was about, especially on the end of `Monitor`, `Alarm`. Does this PR satisfy the "first step in HA improvements" that you envision, which I took #506 to be?

InoMurko:
1. Shared config issue
If we look at `:omg_db` mix.exs and the config.exs, values are defined in both. Same goes for the :omg app. And thats what I meant that it makes things inconsistent. 

2. Separating :omg_api from `plasma related stuff` was meant that we would pull them out as a separate application supervision tree (`EthereumEventListener`, `RootChainCoordinator` and `BlockQueue.Server`). The reason is, that you want to keep the API running, even though you are going to return `not running http status`. But your auto heal mechanisms on the plasma level are gonna bring the service back online.

3. `OMG.RPC.Web.Endpoint` now I understand. Hm... thats not the way I would do this - if you wanted it to be included in a separate parts of the application (watcher, childchain) than those applications should take it in as a dependency. I did not expect the app would take it's supervision structure from one app into the other. 
Point 2 and 3 are actually quite related.

To regards to migrations, you're right. There are there. I didn't find them because I was looking at how they're run and didn't look at the root mix. 🤦 

I've written everything else only for the sake of us being aware of them when you're doing refactoring. So you can ignore them if you don't see value in them.

pdobacz:
ok notes after the call:
  - all clarified
  - green this PR, another round of discussion, merge (for now punt the EventerAPI and BlockQueueAPI and keep the current faux-pub-sub)
  - list, prioritize and tackle next steps of the refactors elsewhere :boat:. More or less this list:

> 
>   - place `monitor`/`alarm` and the likes in proper spots
>   - ditch configs in `mix.exs` - everything should work without these ;)
>   - make the APIs responsive even if the app isn't ready
>   - put the supervision tree of omg_rpc into omg_rpc (after the above is done)
>   - use omg_rpc for omg_watcher
>   - tidy EventerAPI stuff being called twice for ExitProcessor.finalize, this isn't necessary
>   - have proper subscribe mechanism for the `Eventer` and `BlockQueue`, instead of the `cast
>   - alarms should be exposed to devops and developers
>   - POC to use filters instead of polls on Geth
>

apps/omg/lib/state/transaction/recovered.ex
```diff
@@ -0,0 +1,145 @@
...
+  @spec recover_from!(binary) :: Transaction.Recovered.t()
+  def recover_from!(encoded_signed_tx) do
+    {:ok, recovered} = Transaction.Recovered.recover_from(encoded_signed_tx)
+    recovered
+  end
+
+  @doc """
+  Checks if input spenders and recovered transaction's spenders are the same and have the same order
+  """
+  @spec all_spenders_authorized(t(), list()) :: :ok
```
InoMurko: missing tuple error return statement in specs

apps/omg/lib/state/transaction/recovered.ex
```diff
@@ -0,0 +1,145 @@
...
+    |> Enum.reduce({:ok, []}, fn sig, acc -> get_spender(hash_no_spenders, sig, acc) end)
+  end
+
+  defp get_spender(_hash_no_spenders, _sig, {:error, _} = err), do: err
+
+  defp get_spender(hash_no_spenders, sig, {:ok, spenders}) do
+    recovered_address = Crypto.recover_address(hash_no_spenders, sig)
+
+    case recovered_address do
+      {:ok, spender} -> {:ok, spenders ++ [spender]}
```
InoMurko: `[spender|spenders]` should be faster

apps/omg_watcher/lib/utxo_selection.ex
```diff
@@ -17,7 +17,7 @@ defmodule OMG.Watcher.UtxoSelection do
   Provides Utxos selection and merging algorithms.
   """
 
-  alias OMG.API.{Crypto, State.Transaction}
+  alias OMG.{Crypto, State.Transaction}
```
InoMurko: hm... we normally don't write aliases this way

InoMurko:
Left three comments, otherwise seems like an excellent first tackle!

apps/omg_watcher/lib/utxo_selection.ex
pdobacz: yeah, I'm not a fan either, but I was just moving what I found, I don't want to add that to the bulk of changes here.

To enforce a certain style, we'd probably need a `credo` check

PR #530, opened 2019-03-19
Omg 438 cant challenge an invalid exit from deposit utxo

pthomalla:


apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -385,6 +382,34 @@ defmodule OMG.Watcher.ExitProcessor do
...
 
+  defmacro is_deposit(blknum) do
+    {:ok, interval} = OMG.Eth.RootChain.get_child_block_interval()
+
+    quote do
+      rem(unquote(blknum), unquote(interval)) != 0
+    end
+  end
+
+  defp get_standard_exit_id(
```
pdobacz: put all the machinery (this and `is_deposit`) to calculate the tx hash into `Core` and under unit tests pls. For non-deposit txs it must circle back into here, so it might require a bit more figuring out, but this functionality should be covered by tests.

apps/omg_watcher/test/integration/invalid_exit_test.exs
```diff
@@ -85,6 +86,49 @@ defmodule OMG.Watcher.Integration.InvalidExitTest do
     IntegrationTest.wait_for_byzantine_events([], @timeout)
   end
 
+  @tag fixtures: [:watcher_sandbox, :stable_alice, :child_chain, :token, :stable_alice_deposits]
```
pdobacz: why not merge this with the non-deposit case as we discussed? if it is possible and I'm not missing anything here, let's not bloat the integration tests.

Having and challenging an invalid deposit-exit in the previous test scenario seems to be easy to get at 1st glance (just open and challenge another standard exit)

apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -361,12 +359,13 @@ defmodule OMG.Watcher.ExitProcessor do
...
   end
 
-  def handle_call({:create_challenge, Utxo.position(blknum, txindex, oindex) = exiting_utxo_pos}, _from, state) do
+  def handle_call({:create_challenge, Utxo.position(blknum, _txindex, oindex) = exiting_utxo_pos}, _from, state) do
     with spending_blknum_response <- exiting_utxo_pos |> Utxo.Position.to_db_key() |> OMG.DB.spent_blknum(),
-         %{txhash: txhash} <- OMG.Watcher.DB.Transaction.get_by_position(blknum, txindex),
-         {:ok, exit_id} <- OMG.Eth.RootChain.get_standard_exit_id(txhash, oindex),
-         {:ok, raw_spending_proof, exit_info} <-
-           Core.get_challenge_data(spending_blknum_response, exiting_utxo_pos, state) do
+         {:ok, {block_hash, _block_timestamp}} <- Eth.RootChain.get_child_chain(blknum),
```
pdobacz: you can fetch the block hash consistently with the other occurrence here: `{:ok, hashes} = OMG.DB.block_hashes([raw_blknum])` - see below

apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -361,12 +359,13 @@ defmodule OMG.Watcher.ExitProcessor do
...
+         {:ok, [block]} <- OMG.DB.blocks([block_hash]),
+         {:ok, raw_spending_proof, %ExitInfo{exit_txhash: exit_txhash} = exit_info} <-
```
pdobacz: return `exit_txhash` separately, not inside the `ExitInfo` struct: `{:ok, raw_spending_proof, exit_info, exit_txhash} <- ...`. No reason to leverage `ExitInfo` here, which is used totally differnet by the `ExitProcessor`

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -1274,13 +1274,27 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+  def get_challenge_data(spending_blknum_response, exiting_utxo_pos, block, %__MODULE__{exits: exits} = state) do
     with %ExitInfo{} = exit_info <- Map.get(exits, exiting_utxo_pos, {:error, :exit_not_found}),
+         exit_txhash <- get_standard_exit_txhash(exit_info, exiting_utxo_pos, block),
          ife_response = get_ife_based_on_utxo(exiting_utxo_pos, state),
          {:ok, raw_spending_proof} <- ensure_challengeable(spending_blknum_response, ife_response),
-         do: {:ok, raw_spending_proof, exit_info}
+         do: {:ok, raw_spending_proof, %{exit_info | exit_txhash: exit_txhash}}
+  end
+
+  defp get_standard_exit_txhash(
```
pdobacz: comment both clauses of `get_standard_exit_txhash`

1/ `# get the hash of a deposit transaction`
2/ `# get the hash of a transaction from a block`

Issue #526, opened 2019-03-18
`/transaction.submit` change proposition: remove the need to RLP encode signatures


pnowosie:
We wanted to change [Child chain](https://omisego.github.io/elixir-omg/docs-ui/?url=operator_api_swagger.json#/ChildChain/submit) and corresponding [Watcher's](https://omisego.github.io/elixir-omg/docs-ui/?url=watcher_api_swagger.json#/Transaction/submit) `/transaction.submit` endpoint to receive:
* transaction - RLP & HEX encoded bytes of (raw) transaction
* signatures - array of HEX encoded bytes of signatures

:exclamation: Endpoint version which accepts just `transaction` will be removed :exclamation: 

Currently these endpoints receive transaction with signatures together, RLP-encoded.

__Reason for change:__ 
We've currently introduced [Informational Watcher helper](https://omisego.github.io/elixir-omg/docs-ui/?url=informational_api_swagger.json#/Transaction/createTransaction) `/transaction.create` which can help clients to compose transaction for a provided payment order. Helper will respond with RLP-encoded transaction bytes which is ready to sign by the UTXO owner(s).  
If the client could just produce signatures (2 standard crypto operations) and not be forced to bind the transaction & signatures together, it would simplify the tech stack on client side.

More versatile clients who needs more control over UTxOs being used in transaction have to still produce correct transaction (which includes RLP encoding). However they could save to not RLP encode transaction & signatures together.

__Rollout plan__
We would introduce parameter `signatures` to :point_up: endpoints, so there will be: 
1. possibility to send full RLP-encoded transaction with signatures using just `transaction` request field,
2. send raw transaction and signatures separately using `transaction` and `signatures` fields in the request body.

As we believe the 2nd version is better and is still the time to move, we will opt for **obsolete and remove 1st version** for the `v1.0`.

I looking forward for any feedback. Thanks :exclamation:

Pongch:
Hey @pnowosie, here's my 2 cents:

I think having two classes of clients is a great idea. Ideally client responsibility should be reduced to simply signing and submitting transaction without the overheads of building + RLP encoding (which thus far has taken quite an effort to implement). This will also mean more client libraries implementations in the future 💯 

[...]

PR #531, opened 2019-03-19
Omg 399 illegality of tx with holes - #1

pnowosie:
This is __part one__ of the task - provides fix for child chain, there will be also fix for plasma contract repo.

The idea of the fix is to stop improperly constructed transactions leak into the system and make them invalid while transaction is decoded from RLP.

### DIFF
--- a/apps/omg_api/lib/state/transaction.ex
+++ b/apps/omg_api/lib/state/transaction.ex
@@ -133,26 +133,22 @@ defmodule OMG.API.State.Transaction do
   def reconstruct(_), do: {:error, :malformed_transaction}
 
   defp reconstruct_inputs(inputs_rlp) do
-    {:ok,
-     Enum.map(inputs_rlp, fn [blknum, txindex, oindex] ->
-       %{blknum: parse_int(blknum), txindex: parse_int(txindex), oindex: parse_int(oindex)}
-     end)}
+    Enum.map(inputs_rlp, fn [blknum, txindex, oindex] ->
+      %{blknum: parse_int(blknum), txindex: parse_int(txindex), oindex: parse_int(oindex)}
+    end)
+    |> inputs_without_gaps()
   rescue
     _ -> {:error, :malformed_inputs}
   end
 
   defp reconstruct_outputs(outputs_rlp) do
-    outputs =
-      Enum.map(outputs_rlp, fn [owner, currency, amount] ->
-        with {:ok, cur12} <- parse_address(currency),
-             {:ok, owner} <- parse_address(owner) do
-          %{owner: owner, currency: cur12, amount: parse_int(amount)}
-        end
-      end)
-
-    if error = Enum.find(outputs, &match?({:error, _}, &1)),
-      do: error,
-      else: {:ok, outputs}
+    Enum.map(outputs_rlp, fn [owner, currency, amount] ->
+      with {:ok, cur12} <- parse_address(currency),
+           {:ok, owner} <- parse_address(owner) do
+        %{owner: owner, currency: cur12, amount: parse_int(amount)}
+      end
+    end)
+    |> outputs_without_gaps()
   rescue
     _ -> {:error, :malformed_outputs}
   end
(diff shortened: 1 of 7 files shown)


apps/omg_api/lib/state/transaction.ex
```diff
@@ -223,4 +219,31 @@ defmodule OMG.API.State.Transaction do
   """
   @spec get_outputs(t()) :: list(output())
   def get_outputs(%__MODULE__{outputs: outputs}), do: outputs
+
+  defp inputs_without_gaps([input | _] = inputs) do
```
pdobacz: It looks like `init` can be chipped off in `check_for_gaps`, I think it would be a bit more readable thanks to that.

Next and more general: is there a good reason why `inputs_without_gaps` is so different from `outputs_...`? I'd be great if they only differed in their `check_for_gasp` invocation.

Even nexter: hey, have you considered doing `Enum.chunk` to slice the `inputs`/`outputs` into all `{n, n+1}` pairs and see if there is no `{empty, nonempty}` pair? I think this would spare us the `reduce` and be very readable

apps/omg_api/test/state/transaction_test.exs
```diff
@@ -204,4 +204,59 @@ defmodule OMG.API.State.TransactionTest do
...
+             |> Transaction.decode()
+  end
+
+  test "Decoding transaction without outputs is successful" do
+    assert {:ok, _} =
+             create_decoded([{1, 0, 0}], [])
+             |> Transaction.decode()
+  end
+
+  defp create_decoded(inputs, outputs),
```
pdobacz: `create_encoded`?

apps/omg_api/test/state/transaction_test.exs
```diff
@@ -204,4 +204,59 @@ defmodule OMG.API.State.TransactionTest do
 
     assert {:error, :malformed_metadata} = Transaction.Signed.decode(ExRLP.encode([sigs, inputs, outputs, ""]))
   end
+
+  @tag fixtures: [:utxos, :alice]
+  test "Decoding transaction with gaps in inputs returns error",
+       %{utxos: [input1, input2 | _], alice: %{addr: alice_addr}} do
+    assert {:error, :inputs_contain_gaps} ==
```
pdobacz: I'm not 100% certain, but I'd suggest testing on a more "public API level", i.e. `TestHelper.create_signed |> Transaction.Recovered.recover_from` pair (actually `TestHelper.create_encoded |> decode |> ...`, but the former is better for now, as I'll likely refactor `Transaction.Recovered.recover_from` to take bytes as input while #529)

The behavior that we really care about is what happens when processing

pdobacz:
Let's note https://github.com/omisego/elixir-omg/issues/526#issuecomment-474739946, and maybe stop for a second and look more closely at the particular use case of using gaps in outputs that @chriskohlhoff has. If there's a strong case to support these, maybe it's still time to reconsider.

apps/omg_api/test/state/transaction_test.exs
pnowosie: To make this testable as you suggested @pdobacz I've introduced quite [unpleasant hack](https://github.com/omisego/elixir-omg/blob/b261e5ba4c5e462a9e15bfc1d7f0bef1c8004541/apps/omg_api/lib/state/transaction/recovered.ex#L42). 
I want to make sure that `no gaps check` is really low that no incorrect transaction can sneak in.

pnowosie:
> Let's note [#526 (comment)](https://github.com/omisego/elixir-omg/issues/526#issuecomment-474739946), and maybe stop for a second and look more closely at the particular use case of using gaps in outputs that @chriskohlhoff has. If there's a strong case to support these, maybe it's still time to reconsider.

I'll review the contract code that this check is really needed for outputs. However having in mind how tough contract implementation is - probably having this restriction guaranteed would help us keep it at the sane level...

apps/omg_api/test/state/transaction_test.exs
pdobacz: ah I hear you now. Yeah let's do this till #529 lands (and remember to fix differently, if it doesn't :) )

PR #532, opened 2019-03-20
docs: set targetgaslimit when starting dev geth

### LINKED ISSUE #493: Deploying Rootchain contract to local geth in dev mode fails with “exceeds block gas limit” error
Setting up a local env, running `geth  --dev --dev.period 1`.
`OMG.Eth.DevHelpers.prepare_env` fails with:
```
** (FunctionClauseError) no function clause matching in OMG.Eth.DevHelpers.deploy_sync!/1

    The following arguments were given to OMG.Eth.DevHelpers.deploy_sync!/1:

        # 1
        {:error, %{"code" => -32000, "message" => "exceeds block gas limit"}}

    Attempted function clauses (showing 1 out of 1):

        def deploy_sync!({:ok, txhash} = transaction_submission_result)

    test/support/dev_helpers.ex:116: OMG.Eth.DevHelpers.deploy_sync!/1
    test/support/dev_helpers.ex:51: OMG.Eth.DevHelpers.prepare_env!/2
    (stdlib) erl_eval.erl:680: :erl_eval.do_apply/6
...
```

The cause is that `geth --dev` lowers the block gas limit with each block and can quickly get down to under 6180000, which is what DevHelpers uses when it deploys the RootChain contract.

Two possible workarounds:
1. Restart geth with ` --dev.period 5`. This means that the block gas limit will decrease more slowly, giving you time to deploy the contract.
2. Lower the value for `@ gas_contract_rootchain` in `apps/omg_eth/test/support/deployer.ex` to around 5500000. This _should_ still be enough gas to deploy the contract (although it may not...).

N.B. Starting geth with `--targetgaslimit 8000000` looks like it should solve the problem, but currently `--targetgaslimit ` doesn't appear to have any effect

pgebal:
Closes #493

### DIFF
--- a/docker-compose-non-mac.yml
+++ b/docker-compose-non-mac.yml
@@ -2,7 +2,7 @@ version: "2.1"
 services:
   geth:
     image: ethereum/client-go:v1.8.21
-    entrypoint: /bin/sh -c "apk add curl && geth --dev --dev.period 1 --rpc --rpcapi personal,web3,eth,net --rpcaddr 0.0.0.0 --rpcvhosts=* --rpcport=8545"
+    entrypoint: /bin/sh -c "apk add curl && geth --targetgaslimit "6200000" --dev --dev.period 1 --rpc --rpcapi personal,web3,eth,net --rpcaddr 0.0.0.0 --rpcvhosts=* --rpcport=8545"
     ports:
      - "8545:8545"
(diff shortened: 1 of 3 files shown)


kevsul:
How about adding ` --targetgaslimit "6200000"` to the docker-compose* files as well?

PR #536, opened 2019-03-22
Improving docker startup

pnowosie:
* removed unnecessary child chain postgres cointainer
* add running ecto migrations before Watcher start in `launcher.py`


docker-compose.yml
```diff
@@ -11,21 +11,6 @@ services:
...
 
-  childchain_db:
-    image: postgres
-    restart: always
-    environment:
-      POSTGRES_USER: omisego_dev
-      POSTGRES_PASSWORD: omisego_dev
-      POSTGRES_DB: omisego_dev
-    ports:
-      - "5432:5432"
```
pnowosie: I was tempting to change `watcher_db` host port to std - but I thing it's actually better in the case when local postgres is running (and took std port)

launcher.py
```diff
@@ -399,6 +405,26 @@ def initialise_watcher_postgres_database(self) -> bool:
         )
         return False
 
+    def update_watcher_postgres_database(self) -> bool:
```
pnowosie: Running it locally I can see `ecto.reset` command logs, but can't find `ecto.migrate`'s one. I'd expect to see smth like:
`2019-03-22 12:24:42.824 [info] module=Ecto.Migrator function=log/2 ⋅Already up⋅`

We'll see...

PR #539, opened 2019-03-25
fix: protecting OMG.DB from refactors in OMG.Block

pdobacz:
a continuation of #464 movement, which turned out to be necessary for Blocks and Utxo values too

### DIFF
--- a/apps/omg/lib/utxo.ex
+++ b/apps/omg/lib/utxo.ex
@@ -50,8 +50,6 @@ defmodule OMG.Utxo do
     %{owner: owner, currency: currency, amount: amount, creating_txhash: creating_txhash}
   end
 
-  def from_db_value(%__MODULE__{} = utxo), do: from_db_value(Map.from_struct(utxo))
-
   def from_db_value(%{owner: owner, currency: currency, amount: amount, creating_txhash: creating_txhash})
       when is_binary(owner) and is_binary(currency) and is_integer(amount) and is_nil_or_binary(creating_txhash) do
     value = %{owner: owner, currency: currency, amount: amount, creating_txhash: creating_txhash}
--- a/apps/omg/lib/utxo.ex
+++ b/apps/omg/lib/utxo.ex
@@ -38,11 +38,23 @@ defmodule OMG.Utxo do
     end
   end
 
+  defmacrop is_nil_or_binary(binary) do
+    quote do
+      is_binary(unquote(binary)) or is_nil(unquote(binary))
+    end
+  end
+
   # NOTE: we have no migrations, so we handle data compatibility here (make_db_update/1 and from_db_kv/1), OMG-421
-  def to_db_value(%__MODULE__{} = utxo) do
-    Map.take(utxo, [:owner, :currency, :amount, :creating_txhash])
+  def to_db_value(%__MODULE__{owner: owner, currency: currency, amount: amount, creating_txhash: creating_txhash})
+      when is_binary(owner) and is_binary(currency) and is_integer(amount) and is_nil_or_binary(creating_txhash) do
+    %{owner: owner, currency: currency, amount: amount, creating_txhash: creating_txhash}
   end
 
   def from_db_value(%__MODULE__{} = utxo), do: from_db_value(Map.from_struct(utxo))
-  def from_db_value(utxo_map), do: struct!(__MODULE__, utxo_map)
+
+  def from_db_value(%{owner: owner, currency: currency, amount: amount, creating_txhash: creating_txhash})
+      when is_binary(owner) and is_binary(currency) and is_integer(amount) and is_nil_or_binary(creating_txhash) do
+    value = %{owner: owner, currency: currency, amount: amount, creating_txhash: creating_txhash}
+    struct!(__MODULE__, value)
+  end
 end
--- a/apps/omg_watcher/lib/exit_processor.ex
+++ b/apps/omg_watcher/lib/exit_processor.ex
@@ -28,6 +28,7 @@ defmodule OMG.Watcher.ExitProcessor do
   alias OMG.State.Transaction
   alias OMG.Utxo
 
+  alias OMG.Block
   alias OMG.DB
   alias OMG.Eth
   alias OMG.Watcher.ExitProcessor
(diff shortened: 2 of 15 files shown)


apps/omg/lib/utxo.ex
```diff
@@ -38,11 +38,23 @@ defmodule OMG.Utxo do
     end
   end
 
+  defmacrop is_db_tx_hash(binary) do
```
pnowosie: Rename to ‚is_nil_or_binary’ ?

apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -372,7 +373,8 @@ defmodule OMG.Watcher.ExitProcessor do
           raw_blknum when is_number(raw_blknum) ->
             {:ok, hashes} = OMG.DB.block_hashes([raw_blknum])
             {:ok, [spending_block]} = OMG.DB.blocks(hashes)
-            spending_block
+            %Block{} = spending_block_struct = Block.from_db_value(spending_block)
```
pik694: do we really need to check the type here?

apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -465,21 +467,19 @@ defmodule OMG.Watcher.ExitProcessor do
     spend_blknum
   end
 
-  defp run_block_getting(%ExitProcessor.Request{blknums_to_get: blknums} = request) do
+  defp run_block_getting(%ExitProcessor.Request{blknums_to_get: blknums} = request),
+    do: %{request | blocks_result: do_block_getting(blknums)}
+
+  defp run_ife_block_getting(%ExitProcessor.Request{piggybacked_blknums_to_get: blknums} = request),
```
pik694: Are the blknums piggybacked?
I understand the idea, nevertheless the name is quite not accurate.

pik694: what about `fetching` instead of `getting`

apps/omg_watcher/lib/exit_processor.ex
```diff
@@ -465,21 +467,19 @@ defmodule OMG.Watcher.ExitProcessor do
...
-  end
 
-  defp run_ife_block_getting(%ExitProcessor.Request{piggybacked_blknums_to_get: blknums} = request) do
-    _ = Logger.debug("piggybacked_blknums_to_get: #{inspect(blknums)}")
-    {:ok, hashes} = OMG.DB.block_hashes(blknums)
-    _ = Logger.debug("piggybacked_hashes: #{inspect(hashes)}")
-    {:ok, blocks} = OMG.DB.blocks(hashes)
-    _ = Logger.debug("piggybacked_blocks_result: #{inspect(blocks)}")
-    %{request | piggybacked_blocks_result: blocks}
+    blocks |> Enum.map(&(%Block{} = Block.from_db_value(&1)))
```
pik694: Another not necessary type check (IMO).

apps/omg_watcher/lib/exit_processor/core.ex
```diff
@@ -1280,7 +1280,7 @@ defmodule OMG.Watcher.ExitProcessor.Core do
    - a block number which can be used to retrieve needed information to challenge or an ife which spends inputs
    - the relevant exit information
   """
-  @spec get_challenge_data(tuple(), Utxo.Position.t(), Block.t() | :not_found, t()) ::
+  @spec get_challenge_data(tuple(), Utxo.Position.t(), map() | :not_found, t()) ::
```
pik694: Consider also renaming the input parameter into `db_block`

apps/omg_watcher/lib/exit_processor.ex
pdobacz: I don't like the name either, but let's defer this to a larger refactor of EP (which hopefully will come at one point :clock1:)

apps/omg_watcher/lib/exit_processor.ex
pdobacz: you're right, I'm not sure why I did it this way, I recall there being one spot where it made a difference, and then missed some change which made it irrelevant. Thx!

apps/omg_watcher/lib/exit_processor/competitor_info.ex
```diff
@@ -49,16 +50,21 @@ defmodule OMG.Watcher.ExitProcessor.CompetitorInfo do
...
+    }
+
     {:put, :competitor_info, {tx_hash, value}}
   end
 
-  def from_db_kv({tx_hash, %__MODULE__{} = competitor}), do: from_db_kv({tx_hash, Map.from_struct(competitor)})
-
   def from_db_kv(
-        {tx_hash, %{tx: %Transaction.Signed{} = tx, competing_input_index: index, competing_input_signature: signature}}
+        {tx_hash, %{tx: %{} = signed_tx_map, competing_input_index: index, competing_input_signature: signature}}
```
pik694: why not add a guard `when is_map(signed_tx_map)`? I think this is more consistent

apps/omg_watcher/lib/exit_processor/in_flight_exit_info.ex
```diff
@@ -180,6 +177,28 @@ defmodule OMG.Watcher.ExitProcessor.InFlightExitInfo do
     {ife_hash, struct!(__MODULE__, ife_map)}
   end
 
+  # NOTE: the databases currently don't hold the `signed_tx_bytes` field, hence dropping this here and in the other fun.
+  # NOTE: non-private because `CompetitorInfo` holds `Transaction.Signed` objects too
+  def from_db_signed_tx(%{raw_tx: raw_tx_map, sigs: sigs}) when is_list(sigs) do
```
pik694: there is only one name for `to_db_value` methods and pattern matching chooses the correct one, but methods `from_db_*_tx` have different names. Could you choose one approach?

pdobacz: in fact, I can't, because `from_*` would need to know what type it reads, while it is always a map. So the caller needs to decide, what it expects. OTOH calling it as `from_db_value(Transaction.Signed, map)` would go against the other convention, where these are always single argument functions. So this way or another, we're breaking some conventions

apps/omg_watcher/lib/exit_processor/competitor_info.ex
pdobacz: I didn't, thinking it wouldn't match against a struct then, but it seems `is_map(struct) == true` so ok

PR #542, opened 2019-03-25
feat: parity rpc specific errors

paulperegud:
When complaining about duplicated transactions Parity uses different error code and different message. This PR handles that.


apps/omg_api/lib/block_queue/core.ex
```diff
@@ -450,6 +450,15 @@ defmodule OMG.API.BlockQueue.Core do
 
       {:error, %{"code" => -32_000, "message" => "nonce too low"}} ->
         process_nonce_too_low(submission, newest_mined_blknum)
+
+      # parity error code for duplicated tx
+      {:error, %{"code" => -32_010, "message" => "Transaction with the same hash was already imported."}} ->
```
InoMurko: couldn't this handling get abstracted away in a relevant module?

paulperegud: Leaving a TODO for that.

PR #543, opened 2019-03-25
OMG-218 feature: getting metrics on leveldb usage (read,multiread,write)

InoMurko:


apps/omg_db/test/omg_db/recorder_test.exs
```diff
@@ -0,0 +1,80 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.RecorderTest do
+  @moduledoc """
+  A smoke test of the LevelDB support. The intention here is to **only** test minimally, that the pipes work.
```
pdobacz: looks like copy pasted material

apps/omg_db/test/omg_db/recorder_test.exs
```diff
@@ -0,0 +1,80 @@
...
+
+  For more detailed persistence test look for `...PersistenceTest` tests throughout the apps.
+
+  Note the excluded moduletag, this test requires an explicit `--include wrappers`
+  """
+  use ExUnitFixtures
+  use OMG.DB.Case, async: true
+  alias OMG.DB
+  @moduletag :wrappers
```
pdobacz: if this test isn't just redundant testing the `wrappers`, which it looks like the case here, this tag isn't needed - let's run these tests always as part of the `mix test` invocation

apps/omg_db/test/omg_db/recorder_test.exs
```diff
@@ -0,0 +1,80 @@
...
+
+  test "if multi update counter gets incremented", %{db_pid: db_pid} do
+    :ok =
+      DB.multi_update(
+        [{:put, :block, %{hash: "xyz"}}, {:put, :block, %{hash: "vxyz"}}, {:put, :block, %{hash: "wvxyz"}}],
```
pdobacz: if this test doesn't require 3 puts, let's simplify (and also collapse into a single line)

apps/omg_db/test/omg_db/recorder_test.exs
```diff
@@ -0,0 +1,80 @@
+# Copyright 2018 OmiseGO Pte Ltd
```
pdobacz: re this file's location and the others moved: 

per convention it should be
```
apps/omg_db/test/application_test.exs
apps/omg_db/test/db_test.exs
apps/omg_db/test/recorder_test.exs
```

I think I can recall your proposition for this move (`mix test test/omg_db`, wasn't it?), but maybe it's better to discuss the new convention and potentially adopt in a separate PR? 

Also, isn't there another way to do this kind of test selection :thinking: ?

apps/omg_db/lib/leveldb_server.ex
```diff
@@ -53,92 +52,92 @@ defmodule OMG.DB.LevelDBServer do
...
+      |> Enum.map(fn key -> get(key, db_ref, name) end)
       |> LevelDBCore.decode_values(:block_hash)
 
     {:reply, result, state}
   end
 
-  def handle_call(:in_flight_exits_info, _from, %__MODULE__{db_ref: db_ref} = state) do
-    result = get_all_by_type(:in_flight_exit_info, db_ref)
+  def handle_call(:in_flight_exits_info, _from, %__MODULE__{db_ref: db_ref, name: name} = state) do
+    result = get_all_by_type(:in_flight_exit_info, db_ref, name)
```
pdobacz: since we have two fields of `state` used now, maybe we can consider passing `state` around and pattern matching the details DRYly in functions like `get_all_by_type`?

apps/omg_db/test/omg_db/recorder_test.exs
InoMurko: I am not aware how one can run only tests specific to an application. But yeah. I’ve moved the tests because I wanted to run db tests only locally 😁

InoMurko: Looking at the phoenix framework tets location, https://github.com/phoenixframework/phoenix/tree/master/test, they’re located under another subfolder-app name.

PR #545, opened 2019-03-26
feat: handle parity's "won't replace, gas price is too low" error

paulperegud:


apps/omg_api/lib/block_queue/core.ex
```diff
@@ -441,22 +441,32 @@ defmodule OMG.API.BlockQueue.Core do
...
       {:error, %{"code" => -32_000, "message" => "replacement transaction underpriced"}} ->
         _ = Logger.debug("Submission #{inspect(submission)} is known, but with higher price - ignored")
         :ok
 
+      # parity version
+      {:error,
+       %{
+         "code" => -32_010,
+         "message" =>
+           "Transaction gas price is too low. There is another transaction with same nonce in the queue." <> _
```
pdobacz: 2 nitpicks here

1/ if we're `" <> _`-ing the portion of the matched message, let's do more of that to fit on one line :)
2/ at this point maybe let's dry the two `_ = Logger.debug("Submission #{inspect(submission)}...` invocations into functions `log_known_tx(submission)` and `log_known_tx_higher_price(submission)`

Issue #535, opened 2019-03-22
MacOS Mojave setup instruction (w/o docker)


pnowosie:
# Environment

MacOS Mojave 10.14.3

Elixir
```
> elixir --version
Erlang/OTP 21 [erts-10.2.3] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.8.1 (compiled with Erlang/OTP 21)
```

Solidity
```
> solc --version
solc, the solidity compiler commandline interface
Version: 0.4.25+commit.59dbf8f1.Darwin.appleclang
```

## Elixir-omg compilation
Clone latest master into directory of you choice.
Fetch dependencies with `mix deps.get`

Compilation of dependencies should go well, probably you would need to install some requirements (e.g. openssl, automake...) your system is missing, but installing latest version should do the trick.

Only exception to above is `eleveldb` which is not maintained for some time now.
For this you need to add commandline tools (c++ includes) ❓ 
Replace also all occurence of `10.8` to `10.7` in `<repo_root>\deps/eleveldb/c_src/leveldb/build_config.mk`
Then try to compile eleveldb with
```
> CXXFLAGS=-stdlib=libc++ CPATH=/Library/Developer/CommandLineTools/CommandLineTools/usr/include/c++/v1 mix deps.compile eleveldb
```

pik694:
As said in #555, for now macOS users can simply apply the following patch:
https://drive.google.com/file/d/1JRuR2_nuJnxhFoZNCIBXafZGYRbncuOZ/view?usp=sharing

PR #555, opened 2019-03-27
macOS Mojave compilation

pik694:
`eleveldb` from `hexpm` does not compile on macOS, I propose an update of this dependency so it compiles flawlessly on macOS Mojave.


mix.exs
```diff
@@ -25,6 +25,7 @@ defmodule OMG.Umbrella.MixProject do
 
   defp deps do
     [
+      {:eleveldb, git: "https://github.com/basho/eleveldb.git", branch: "develop", manager: :make, override: true},
```
pdobacz: two comments here:
1/ give a comment why this is overriden :)
2/ can we override to sth less volatile (tag perhaps? a stable branch? - `develop` seems like we could easily go into a broken set of dependencies, if one accidentally upgrades `mix.lock`)

pik694: I'm afraid that I could only refer to the exact commit, as the fix on `eleveldb` side was applied no sooner than 20 days ago and there is none tag including it.

InoMurko:
The way I got it to work on OSX is by sticking with older command line tools `Command Line Tools (macOS 10.13) for Xcode 9.4.1`

mix.exs
pdobacz: hm, would there be a way to solve this mojave-specific problem with less possibility to backfire to non-mojave users? :thinking: (I just saw Ino's comment - maybe this?)

InoMurko: I second  @pdobacz comment. This might be a bit wild :)

pnowosie:
> older command line tools `Command Line Tools (macOS 10.13) for Xcode 9.4.1`
Yeah @InoMurko we also did some ugly hacks #535

pnowosie:
Closes #535

pik694:
After discussion with @pdobacz, we decided that we will postpone merging this PR until `eleveldb` with fixes is released (tagged on GitHub). Until now, we will prepare a temporary patch, so any macOS user can apply it.

achiurizo:
Closing this as stale( also since the discussion decided to wait on `eleveldb` updates with new releases. )

PR #562, opened 2019-03-28
Omg 430 fix the annoying non failing error messages in integration tests

pthomalla:
Use technic suggests in Phoenix.Ecto.SQL.Sandbox, (they alsow add Sandbox in sourse code)
```
  ## Example
  This plug should only be used during tests. First, set a flag to
  enable it in `config/test.exs`:
      config :your_app, sql_sandbox: true
  And use the flag to conditionally add the plug to `lib/your_app/endpoint.ex`:
      if Application.get_env(:your_app, :sql_sandbox) do
        plug Phoenix.Ecto.SQL.Sandbox
      end
  It's important that this is at the top of `endpoint.ex`, before any other plugs.
```
unfortunately, this technic pollute a little bit our source code :/
it's becase this two line:
https://github.com/omisego/elixir-omg/blob/d9608b968de79161a1060f5e4ffd56271f6a3db2/apps/omg_watcher/lib/omg_watcher/supervisor.ex#L39-L40
can't set `mod` without `self()` from option (posibility not discribe in documentation, but available).
and this muste be set just after create child `Watcher.DB.Repo` to avoid race.
https://github.com/omisego/elixir-omg/blob/d9608b968de79161a1060f5e4ffd56271f6a3db2/apps/omg_watcher/lib/omg_watcher/supervisor.ex#L51-L60
race:
  `OMG.Watcher.BlockGetter` connect with DB and get init value from it
   meanwhile, Sandbox tries to change mode.


pdobacz:
@pthomalla can you provide some explanation of what this fixes, why was it broken, and how does the fix work?

Its going to be very useful to parse the changes to review and for future reference.

Also 2 changes look a tiny bit controversial: 1/ putting the sandbox related stuff into `watcher`'s supervisor production code 2/ baking the Ecto adapter into the module, as opposed to the config

apps/omg_watcher/lib/omg_watcher/supervisor.ex
```diff
@@ -22,33 +22,58 @@ defmodule OMG.Watcher.Supervisor do
 
   alias OMG.Watcher
 
+  if Application.get_env(:omg_watcher, :sql_sandbox) do
+    defmodule Sandbox do
```
pnowosie: I can understand that we need to inject this mode change just after repo is alive. However I thought error appears only when mode is changed during [this](https://github.com/omisego/elixir-omg/blob/ff0c73cc4b08e7971005b2192c4cf44ac1ea1bcd/apps/omg_watcher/lib/omg_watcher/block_getter.ex#L63) block_getter's call. 
I think can think a little more to avoid this huge changes which are only here to kill some annoying log...

apps/omg_watcher/lib/omg_watcher/web/endpoint.ex
```diff
@@ -18,6 +18,11 @@ defmodule OMG.Watcher.Web.Endpoint do
 
   socket("/socket", OMG.Watcher.Web.Socket, websocket: [])
 
+  # It's important that this is at the top of `endpoint.ex`, before any other plugs.
+  if Application.get_env(:omg_watcher, :sql_sandbox) do
+    plug(Phoenix.Ecto.SQL.Sandbox)
```
pnowosie: From my best understanding we don't use ecto through Phoenix, so it's enough to plug the sandbox to the repo. Phoenix controllers are unaware of ecto existence. Please revert.

apps/omg_watcher/mix.exs
```diff
@@ -43,8 +43,9 @@ defmodule OMG.Watcher.Mixfile do
     [
       {:phoenix, "~> 1.3"},
       {:plug_cowboy, "~> 1.0"},
-      {:phoenix_ecto, "~> 3.3"},
-      {:postgrex, ">= 0.13.5"},
+      {:phoenix_ecto, "~> 4.0"},
```
pnowosie: I like this version bump 🙌

apps/omg_watcher/test/fixtures.exs
```diff
@@ -144,7 +138,7 @@ defmodule OMG.Watcher.Fixtures do
         name: Watcher.Supervisor
       )
 
-    :ok = SQL.Sandbox.checkout(DB.Repo)
+    :ok = SQL.Sandbox.checkout(DB.Repo, ownership_timeout: 180_000)
```
pnowosie: Why we need this - this fixture is used for quick controller's tests. There were no issues with them.

apps/omg_watcher/lib/omg_watcher/supervisor.ex
```diff
@@ -22,22 +22,48 @@ defmodule OMG.Watcher.Supervisor do
...
+  end
+
+  @children_just_behind_repo if(Mix.env() == :test, do: [{__MODULE__.Sandbox, []}], else: [])
+
   def start_link do
     Supervisor.start_link(__MODULE__, :ok, name: __MODULE__)
   end
 
   def init(:ok) do
+    childeren_need_be_on_front =
```
pnowosie:
```suggestion
    top_children =
```

apps/omg_watcher/lib/omg_watcher/supervisor.ex
```diff
@@ -22,22 +22,48 @@ defmodule OMG.Watcher.Supervisor do
...
+      [
+        {OMG.InternalEventBus, []},
+        %{
+          id: Watcher.DB.Repo,
+          start: {Watcher.DB.Repo, :start_link, []},
+          type: :supervisor
+        }
+      ] ++ @children_just_behind_repo
```
pnowosie:
```suggestion
      ] ++ @children_run_after_repo
```

apps/omg_watcher/lib/omg_watcher/supervisor.ex
pnowosie: Add comment what we want to achieve, e.g. link to this :pr:

pnowosie:
Please add additional assertion in the :watcher fixture `DB.Blocks.all() == []` to avoid accidental DB polution

PR #567, opened 2019-03-31
feature: introspection of development env

InoMurko:
This is a wip in order to start the discussion:
- how to handle cookies
- how to distinguish deployment environments

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -45,7 +45,7 @@ services:
 
   childchain:
     image: elixir-omg:dockercompose
-    entrypoint: /bin/bash -c "./launcher.py && mix xomg.child_chain.start --config ~/config.exs"
+    entrypoint: /bin/bash -c "./launcher.py && elixir --erl '-sname childchain' -S mix xomg.child_chain.start --config ~/config.exs"
     environment:
       - MIX_ENV=dev
       - EXIT_PERIOD_SECONDS=86400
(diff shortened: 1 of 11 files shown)


InoMurko:
You would now be able to do something like
`iex --sname introspection --cookie secret_cookie --remsh "watcher@9dc33376b9d0"`
@jbunce the cookie env keys are ERL_CC_COOKIE (ch-ch)
ERL_W_COOKIE (watcher)

docker-compose.yml
```diff
@@ -45,7 +45,7 @@ services:
 
   childchain:
     image: elixir-omg:dockercompose
-    entrypoint: /bin/bash -c "./launcher.py && mix xomg.child_chain.start --config ~/config.exs"
+    entrypoint: /bin/bash -c "./launcher.py && elixir --erl '-sname childchain' -S mix xomg.child_chain.start --config ~/config.exs"
```
pdobacz: maybe give a hint about how to do this for the non-docker setup (in `docs/manual_....md`)? Also possibly a hint on how to use this in general in `README.md`

InoMurko: It's a bit difficult at the moment, because they all presume you boot into the app in an interactive way anyhow (`iex`).

InoMurko: It might be better for us to wait and rewrite these sections once we get distillery out? And then write these for developers (`how to develop`, and `how to deploy`)?

pdobacz: okay, sounds good! I forgot that we're suggesting `iex` there

PR #569, opened 2019-04-01
refactor: move emitting exit_finalized events to ExitProcessor

pdobacz:
The functionality to emit the stub of `ExitFinalized` events got placed in `State` following the example of transaction events. But exit finalized don't seem to have a good reason to not be emitted from the `ExitProcessor` for much cleaner code.

Also, fixed a broken test in `ExitProcessor`: [here](https://github.com/omisego/elixir-omg/compare/emit_exit_finalized_from_exit_processor?expand=1#diff-91398068aa8ecf8b7913c807e662c304R344) one should be testing processing of valid exits, as the test title suggests.

### DIFF
--- a/apps/omg_watcher/lib/omg_watcher/exit_processor.ex
+++ b/apps/omg_watcher/lib/omg_watcher/exit_processor.ex
@@ -232,15 +232,11 @@ defmodule OMG.Watcher.ExitProcessor do
         Utxo.Position.decode(utxo_pos)
       end)
 
-    {:ok, exit_event_triggers, db_updates_from_state, validities} = State.exit_utxos(exits)
+    {:ok, db_updates_from_state, validities} = State.exit_utxos(exits)
+    {new_state, exit_event_triggers, db_updates} = Core.finalize_exits(state, validities)
 
-    _ = if not Enum.empty?(exit_event_triggers), do: Logger.info("Finalized exits: #{inspect(validities)}")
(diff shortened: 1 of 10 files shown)


apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -232,15 +232,11 @@ defmodule OMG.Watcher.ExitProcessor do
         Utxo.Position.decode(utxo_pos)
       end)
 
-    {:ok, exit_event_triggers, db_updates_from_state, validities} = State.exit_utxos(exits)
+    {:ok, db_updates_from_state, validities} = State.exit_utxos(exits)
+    {new_state, exit_event_triggers, db_updates} = Core.finalize_exits(state, validities)
 
-    _ = if not Enum.empty?(exit_event_triggers), do: Logger.info("Finalized exits: #{inspect(validities)}")
+    EventerAPI.emit_events(exit_event_triggers)
```
pgebal: Let's leave the logging. We can decrease logging level to debug.

pdobacz: https://github.com/omisego/elixir-omg/pull/569/files#diff-f1ce1871bf4b96d60bda0fbd02e3891cR226 there's logging, there were two invocations of logging, I only removed one :)

PR #573, opened 2019-04-02
Omg 443 unique deposit

pik694:
Update `plasma-contracts` dependency up to the version with fixed unique deposit id bug.


boolafish:
we might want to update the doc as mentioned in the issue: https://github.com/omisego/plasma-contracts/issues/80

Probably this doc (or more?): https://github.com/omisego/elixir-omg/blob/master/docs/mvp_morevp_interaction.md

pik694:
Oh, right. I completely forgot about this.

docs/mvp_morevp_interaction.md
```diff
@@ -60,11 +60,22 @@ In each we explain possible ways of preventing the double spend.
 10. Mallory submits IFE -> Mallory piggybacks output -> Mallory waits for finalization -> Mallory does SE on output  
     ! same as (9.)
 
-### Exploring `standardExitId(hash(tx), oindex)`
+### Exploring `standardExitId(txbytes, utxo_pos)`
 Many of solutions described above are possible only if there is a way to have one-to-one mapping between in-flight exits and standard exits.
-To achieve that, we change the way `inFlightExitId` and `standardExitId` are computed, so both are a function of hash(tx).
+To achieve that, we change the way `inFlightExitId` and `standardExitId` are computed, so both are a function of hash(tx) ([see](#Distinction-between-deposit-and-regular-transactions)).
```
boolafish: `so both are a function of hash(tx)` might need to be changed as well?

docs/mvp_morevp_interaction.md
```diff
@@ -60,11 +60,22 @@ In each we explain possible ways of preventing the double spend.
...
+### Exploring `standardExitId(txbytes, utxo_pos)`
 Many of solutions described above are possible only if there is a way to have one-to-one mapping between in-flight exits and standard exits.
-To achieve that, we change the way `inFlightExitId` and `standardExitId` are computed, so both are a function of hash(tx).
+To achieve that, we change the way `inFlightExitId` and `standardExitId` are computed, so both are a function of hash(tx) ([see](#Distinction-between-deposit-and-regular-transactions)). 
 This way whenever one type of exit comes in we can check if other type of exit was performed.
 
+
+#### Distinction between deposit and regular transactions
+In case the tx is a deposit, the `standardExitId` is a function of hash(tx concat tx_pos). 
+This distinction was made as when one deposits the same amount twice (or more) the tx_bytes are the same, so exit ids is not unique for the similar deposits,
```
boolafish: I would like to emphasize "without tx_pos"
eg. `....the tx_bytes are the same, so without concatenating tx_pos, the exit ids would be the same and not unique for the similar deposits...`

Also, I would link the issue to this paragraph as well as there has some valuable discussion on why we chosen this design: https://github.com/omisego/plasma-contracts/issues/80

docs/mvp_morevp_interaction.md
```diff
@@ -60,11 +60,22 @@ In each we explain possible ways of preventing the double spend.
...
-To achieve that, we change the way `inFlightExitId` and `standardExitId` are computed, so both are a function of hash(tx).
+To achieve that, we change the way `inFlightExitId` and `standardExitId` are computed, so both are a function of hash(tx) ([see](#Distinction-between-deposit-and-regular-transactions)). 
 This way whenever one type of exit comes in we can check if other type of exit was performed.
 
+
+#### Distinction between deposit and regular transactions
+In case the tx is a deposit, the `standardExitId` is a function of hash(tx concat tx_pos). 
+This distinction was made as when one deposits the same amount twice (or more) the tx_bytes are the same, so exit ids is not unique for the similar deposits, 
+therefore we concatenate the tx_bytes with the tx_pos of the deposit.
+Unfortunately, we cannot `standardExitId` consistently for both regular txs and deposits, as we would loose one-to-one mapping between SE and IFE exit id,
```
boolafish: ...we cannot make `standardExitId` consistent....

docs/mvp_morevp_interaction.md
```diff
@@ -86,3 +97,5 @@ A: I don't see a way to do it yet.
 Q: When IFE/piggyback on output is added, what is the cost of checking if standard exits were performed from output(s)?  
 A: Number of outputs is known on both IFE and piggyback; single SLOAD per output?  
 
+Q: What about mapping between IFE and SE from a deposit?
```
boolafish: I think it is not clear what the link between the question and answer here. I prefer the way you put in issue:
eg.
Q: Is there any issue/concern from having different exit id for deposit tx and regular tx?
A: The only problem is the fact that if we started an IFE from a deposit, we could not check whether someone tried to also SE the deposit. But this is not a problem since we cannot start an IFE from the deposit: `sum of inputs < sum of outputs`.

PR #574, opened 2019-04-03
Phoenix.PubSub for :omg sending stuff to :omg_api and :omg_watcher

### LINKED ISSUE #556: Better separation of concerns between child chain and watcher
The purpose of this piece of work is to be able to cleanly deploy two applications - childchain and watcher.
We can achieve that by having clean application borders.

I propose that we utilize Publish–subscribe pattern through out the repository. 
1. OMG application needs to be free from either childchain (omg_api) or watcher (omg_watcher)
2. `OMG.State` is coupled with `OMG.EventerAPI` witch calls `OMG.Watcher.Eventer`. This makes OMG modules coupled with the Watcher application.
3. `OMG.State` is coupled with `OMG.BlockQueueAPI` witch calls `OMG.API.BlockQueue.Server`. This makes OMG modules coupled with the Childchain application.

One additional thing to be sure to cover - subscriptions need to be persisted between process restarts by either application configuration or Process.monitor (and re-subscription in case of failure).

pdobacz:
Fixes #556.

Blocks to enqueue to `OMG.API.BlockQueue` and events to push to `OMG.Watcher.Eventer` are now delivered through `OMG.InternalEventBus` via `Phoenix.PubSub`.

re

> One additional thing to be sure to cover - subscriptions need to be persisted between process restart

I'm not entirely sure, but I think the `link: true` option on subscribing handles that, i.e. when the pubsub adapter goes down, so would the subscriber, allowing it to restart and resubscribe.

The contender tool to have pubsub was https://github.com/otobus/event_bus, but `Phoenix.PubSub` was chosen because:
  - supports cross-node messages
  - isn't a new dependency to have, as we're already using with `Phoenix.Channels` used in the Watcher


apps/omg/lib/omg/state.ex
```diff
@@ -172,53 +170,47 @@ defmodule OMG.State do
...
-  Returns `db_updates` due and relies on the caller to do persistence
   """
   def handle_call({:close_block, eth_height}, _from, state) do
-    {:ok, {_block, event_triggers, db_updates}, new_state} = do_form_block(state, eth_height)
-
-    EventerAPI.emit_events(event_triggers)
+    {:ok, {block, event_triggers, db_updates}, new_state} = do_form_block(state, eth_height)
 
+    publish_block_to_event_bus(block, event_triggers)
     {:reply, {:ok, db_updates}, new_state}
```
InoMurko: one interesting thing that genserver allows you to do is not block the reply by returning
`{:reply, reply, new_state, timeout() | :hibernate | {:continue, term()}}`
so you would do something like
`{:reply, reply, new_state, {:continue, {:publish_block_to_event_bus, block, event_triggers}}}`

do you find this useful?

apps/omg_api/lib/omg_api/block_queue.ex
```diff
@@ -147,17 +152,20 @@ defmodule OMG.API.BlockQueue do
...
-    def handle_cast({:enqueue_block, %Block{number: block_number, hash: block_hash} = block}, %Core{} = state) do
+    def handle_info(
+          {:internal_event_bus, :enqueue_block, %Block{number: block_number, hash: block_hash} = block},
+          %Core{} = state
+        ) do
       {:ok, parent_height} = Eth.get_ethereum_height()
       state1 = Core.enqueue_block(state, block_hash, block_number, parent_height)
       _ = Logger.info("Enqueuing block num '#{inspect(block_number)}', hash '#{inspect(Base.encode16(block_hash))}'")
 
       FreshBlocks.push(block)
```
InoMurko: FreshBlocks push is handled in a separate PR right?

InoMurko:
Looks great! We need to document the topics and data thats being pushed once we're done with this. Might come useful when adding new features!

apps/omg/lib/omg/internal_event_bus.ex
```diff
@@ -0,0 +1,50 @@
...
+    args
+    |> Keyword.put_new(:name, OMG.InternalEventBus)
+    |> PubSub.PG2.child_spec()
+  end
+
+  @doc """
+  Subscribes the current process to the internal bus topic
+  """
+  def subscribe(topic, opts \\ []) do
+    PubSub.subscribe(OMG.InternalEventBus, topic, opts)
```
InoMurko: __MODULE__ ?

apps/omg/lib/omg/state.ex
```diff
@@ -172,53 +170,47 @@ defmodule OMG.State do
...
     {:noreply, new_state}
   end
 
   defp do_form_block(state, eth_height \\ nil) do
     {:ok, child_block_interval} = Eth.RootChain.get_child_block_interval()
     Core.form_block(child_block_interval, eth_height, state)
   end
+
+  defp publish_block_to_event_bus(block, event_triggers) do
+    :ok = OMG.InternalEventBus.broadcast("events", {:emit_events, event_triggers})
```
InoMurko: Do we know enough about scaling to N nodes that we can decide whether we want to broadcast to all nodes in the cluster or use `direct_broadcast`?
Perhaps we don't...

apps/omg/lib/omg/state.ex
pdobacz: did my homework on `handle_continue`, this is good to have in the toolbox!

Here however I'm not sure on needed. A small drawback is the reduced clarity of what's going on. Also both genserver messages may be bulky (it does matter, even for a sent-to-self message, doesn't it?).

InoMurko: Yep, (to-self) go to the same mailbox.

apps/omg_api/lib/omg_api/block_queue.ex
pdobacz: no need to handle - both are on `OMG.API` so no separation of concerns concerns here. It makes some sense for `FreshBlocks` to get the block from the `BlockQueue`, so I decided to keep it simple. But you find it very unelegant or surprising, we can subscribe `FreshBlocks` too I think

apps/omg/lib/omg/state.ex
pdobacz: I think we don't, but I think `direct_broadcast` is safer for `enqueue_block`, whereas `broadcast` for `events` - intuitively speaking at least. So maybe as future proofing let's do like that

PR #587, opened 2019-04-06
fix: remove utils dependency, ensure it works without it

InoMurko:
Keeping the utils deps lightweight. 
If we keep `ecto_sql` as a dependency in omg_utils, that gets pulled into childchain as well. But ch-ch doesn't interface towards PG at all...


apps/omg_utils/lib/omg_utils/http_rpc/response.ex
```diff
@@ -55,9 +55,9 @@ defmodule OMG.Utils.HttpRPC.Response do
   def sanitize(value), do: value
 
   defp do_filter(map_or_struct) do
-    if Code.ensure_loaded?(Ecto) do
+    if :code.is_loaded(Ecto) do
       Enum.filter(map_or_struct, fn
-        {_, %Ecto.Association.NotLoaded{}} -> false
+        {_, %{__struct__: Ecto.Association.NotLoaded}} -> false
```
pthomalla: I have some mixed feelings about it.
What if we use sanitize a little bit differently, and add some option to this module?
`use OMG.Utils.HttpRPC.Encode, filter: [%Ecto.Association.NotLoaded{}]`

InoMurko: Does it work without a dependency to ecto?

pthomalla: make small mistake, I thing about this `use OMG.Utils.HttpRPC.Response, filter: [%Ecto.Association.NotLoaded{}]`
should work, I did not try such tricks a lot.
In Watcher where we use it wee had ecto dependence already.
thanks to this we can get rid of the reverse dependence

Issue #591, opened 2019-04-08
Endpoints taking in encoded txbytes might crash the ExitProcessor


pdobacz:
Spotted in `v0.1.3`, reproduced in c561caff0548c5c797f1645d64cf904b94abda88 in `v0.1.4`

The process OTP-restarts and continues, but the invalid input should be handled and an error message provided, with no crash.

For example:

```
iex(1)> 2019-04-08 10:34:30.443 [error] module=gen_server function=error_info/7 ⋅GenServer OMG.Watcher.ExitProcessor terminating
** (MatchError) no match of right hand side value: {:error, :malformed_transaction}
    (omg_watcher) lib/exit_processor/core.ex:773: OMG.Watcher.ExitProcessor.Core.get_ife/2
    (omg_watcher) lib/exit_processor/core.ex:617: OMG.Watcher.ExitProcessor.Core.get_competitor_for_ife/3
    (omg_watcher) lib/exit_processor.ex:279: OMG.Watcher.ExitProcessor.handle_call/3
    (stdlib) gen_server.erl:661: :gen_server.try_handle_call/4
    (stdlib) gen_server.erl:690: :gen_server.handle_msg/6
    (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Last message (from #PID<0.4945.0>): {:get_competitor_for_ife, <<0>>}⋅
2019-04-08 10:34:30.448 [info] module=OMG.DB function=exit_infos/1 ⋅Reading exits' info, this might take a while. Allowing 60000 ms⋅
2019-04-08 10:34:30.463 [error] ⋅#PID<0.4945.0> running OMG.Watcher.Web.Endpoint (cowboy_protocol) terminated
Server: localhost:7434 (http)
Request: POST /in_flight_exit.get_competitor
** (exit) exited in: GenServer.call(OMG.Watcher.ExitProcessor, {:get_competitor_for_ife, <<0>>}, 5000)
    ** (EXIT) an exception was raised:
        ** (MatchError) no match of right hand side value: {:error, :malformed_transaction} 
            (omg_watcher) lib/exit_processor/core.ex:773: OMG.Watcher.ExitProcessor.Core.get_ife/2
            (omg_watcher) lib/exit_processor/core.ex:617: OMG.Watcher.ExitProcessor.Core.get_competitor_for_ife/3
            (omg_watcher) lib/exit_processor.ex:279: OMG.Watcher.ExitProcessor.handle_call/3
[...]

PR #593, opened 2019-04-08
fix: handle malformed txbytes as inputs to IFE-related endpoints

pdobacz:
port of #592 from `v0.1` to `master`.


pdobacz:
for some reason merge of #592 didn't close #591. So let's try again: fixes #591

PR #596, opened 2019-04-08
Fix /transaction.all slow responses

### LINKED ISSUE #589: Fix/Research slow /transaction.all endpoint
The endpoint seems to try to load the whole table when querying a list of 200 transactions (see [explorer](http://quest.ari.omg.network/address/0x6f4a4981bc761973eb81e67731e566242c53275a) and end up crashing. I'm thinking about trying to add some indexes.

T-Dnzt:
closes #589

# Overview

This PR switches from using a double left join to using a single inner with a bunch of added indices for increased performance. 

# Note

Also cherry picked [this commit](https://github.com/omisego/elixir-omg/commit/946197ccb372b9be6971abe6587d133e777a63a4) to get it into 0.1 (it was preventing Docker from starting).

### DIFF
--- a/apps/omg_watcher/lib/db/transaction.ex
+++ b/apps/omg_watcher/lib/db/transaction.ex
@@ -21,11 +21,12 @@ defmodule OMG.Watcher.DB.Transaction do
 
   alias OMG.API.State.Transaction
   alias OMG.API.Utxo
-  alias OMG.Watcher.DB
+  alias OMG.Watcher.{DB, DB.TxOutput}
 
   require Utxo
 
-  import Ecto.Query, only: [from: 2]
+  # , only: [from: 2, where: 3, join: 5]
+  import Ecto.Query
 
   @type mined_block() :: %{
           transactions: [OMG.API.State.Transaction.Recovered.t()],
@@ -67,7 +68,7 @@ defmodule OMG.Watcher.DB.Transaction do
     DB.Repo.one(query)
   end
 
-  def get_by_filters(address, blknum, limit) do
+  def get_by_filters(address, blknum \\ nil, limit \\ nil) do
     query_get_last(limit)
     |> query_get_by_address(address)
     |> query_get_by_blknum(blknum)
@@ -83,16 +84,14 @@ defmodule OMG.Watcher.DB.Transaction do
     )
   end
 
-  defp query_get_by_address(base, nil), do: base
+  defp query_get_by_address(query, nil), do: query
 
-  defp query_get_by_address(base, address) do
-    from(
-      tx in base,
-      distinct: true,
-      left_join: output in assoc(tx, :outputs),
-      left_join: input in assoc(tx, :inputs),
-      where: output.owner == ^address or input.owner == ^address
-    )
+  defp query_get_by_address(query, address) do
+    query
+    |> join(:inner, [t], o in TxOutput, t.txhash == o.creating_txhash or t.txhash == o.spending_txhash)
+    |> where([t, o], o.owner == ^address)
+    |> select([t, o], t)
+    |> distinct(true)
   end
 
   defp query_get_by_blknum(base, nil), do: base
--- a/apps/omg_watcher/lib/db/transaction.ex
+++ b/apps/omg_watcher/lib/db/transaction.ex
@@ -21,7 +21,7 @@ defmodule OMG.Watcher.DB.Transaction do
 
   alias OMG.API.State.Transaction
   alias OMG.API.Utxo
-  alias OMG.Watcher.{DB, DB.TxOutput}
+  alias OMG.Watcher.DB
 
   require Utxo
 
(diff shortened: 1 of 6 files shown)


apps/omg_watcher/priv/repo/migrations/20190408131000_add_missing_indices_to_txoutputs.exs
```diff
@@ -0,0 +1,11 @@
+defmodule OMG.Watcher.Repo.Migrations.AddMissingIndicesToTxOuputs do
+  use Ecto.Migration
+
+  def change do
+    create index(:txoutputs, [:creating_txhash, :spending_txhash])
+    create index(:txoutputs, [:creating_deposit])
+    create index(:txoutputs, [:spending_txhash])
```
T-Dnzt: This is @kasima – add an NB about why a `spending_txhash` index alone.

apps/omg_watcher/lib/db/transaction.ex
```diff
@@ -21,11 +21,11 @@ defmodule OMG.Watcher.DB.Transaction do
 
   alias OMG.API.State.Transaction
   alias OMG.API.Utxo
-  alias OMG.Watcher.DB
+  alias OMG.Watcher.{DB, DB.TxOutput}
```
pnowosie: I've seen a lot changes actually expanding this kind into separate lines - it's more alchemist ;)

apps/omg_watcher/lib/db/transaction.ex
```diff
@@ -67,7 +67,7 @@ defmodule OMG.Watcher.DB.Transaction do
     DB.Repo.one(query)
   end
 
-  def get_by_filters(address, blknum, limit) do
+  def get_by_filters(address, blknum \\ nil, limit \\ nil) do
```
pnowosie: Please revert - seem to be just for test, and I'm afraid of consequences

apps/omg_watcher/lib/db/transaction.ex
```diff
@@ -83,16 +83,14 @@ defmodule OMG.Watcher.DB.Transaction do
     )
   end
 
-  defp query_get_by_address(base, nil), do: base
+  defp query_get_by_address(query, nil), do: query
```
pnowosie: Nice! Please rename 👇 following code to keep it consistent

apps/omg_watcher/priv/repo/migrations/20190408131000_add_missing_indices_to_txoutputs.exs
```diff
@@ -0,0 +1,11 @@
...
+    create index(:txoutputs, [:spending_exit])
```
pnowosie: this ':spending_exit' would be super sparse - consider partial index like [here](https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher/priv/repo/migrations/20190315095855_alter_transactions_table_add_partitial_index.exs)

apps/omg_watcher/test/db/transaction_test.exs
```diff
@@ -24,6 +24,22 @@ defmodule OMG.Watcher.DB.TransactionTest do
 
   require Utxo
 
+  describe "get_by_filters/3" do
```
pnowosie: Please revert, already tested [here](https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher/test/omg_watcher/web/controllers/transaction_test.exs)

apps/omg_watcher/priv/repo/migrations/20190408131000_add_missing_indices_to_txoutputs.exs
T-Dnzt: Composite indexes will be used by the optimizer when either the first column is used, or both columns are used in the query. However, if a query only uses the second column (`spending_txhash`), the composite index will not be used, hence the addition of another index for it.

apps/omg_watcher/test/db/transaction_test.exs
T-Dnzt: But those are unit tests for the `Transaction` schema 🤔Controller tests are not enough to ensure that each function at the schema level works as expected and is thoroughly tested. It also makes sense to build a pyramid, a lot of small unit tests testing a lot of edge cases at the bottom (which I didn't really do because I'm quite unfamiliar with the fixture approach you're using and didn't want to mess up everything), and a few high level integration/controller tests testing the happy path.

apps/omg_watcher/lib/db/transaction.ex
T-Dnzt: Removing it, but I do feel having them optional makes the code clearer in terms of expectations (since it isn't handled until 2 functions down in pattern matching).

T-Dnzt:
@pnowosie Merging without removing the tests (because we need the feature kinda right now), but happy to discuss further and potentially submit another PR :)

kasima:
This has  generally improved response times across all of the watcher endpoints. 🙌 

Performance graphs - OmiseGO MoreVP Watcher v20190131 - AppSignal 2019-04-09 14-24-27


## TODO
- Cherry pick this into `master`

apps/omg_watcher/test/db/transaction_test.exs
pdobacz: yeah, we relied on controller tests, since the whole thing was super-simplistic, now as it evolves we can reconsider, in case the controller tests get too unwieldy.

PR #607, opened 2019-04-09
Remove content-type requirement for /status.get

### LINKED ISSUE #601: /status.get requires a `Content-Type` header while not requiring a body
In order to fix the Swagger documentation and avoid adding some hacks to it, the `/status.get` endpoint should not require a `Content-Type` to be set since no `body` needs to be sent. 

To expand on that, since most endpoints don't actually require a body (a lot of params are optional), the enforcing of the `Content-Type` header presence should not really be there if there's no body, but this issue only focuses on the status endpoint for the watcher (because it __never__ needs a body).

T-Dnzt:
closes #601 ~~(@pnowosie: I'd keep the issue to clean implementation, but not hold this PR open)~~ 

# Overview

This PR removes the `Content-Type` requirement for the `/status.get` endpoint. The solution implemented in this PR is __not optimal__ and just a quick fix. The content-type / body parsing should be left to Phoenix instead of being done inside a custom plug function.

### DIFF
--- a/apps/omg_watcher/test/web/controllers/status_test.exs
+++ b/apps/omg_watcher/test/web/controllers/status_test.exs
@@ -49,4 +49,24 @@ defmodule OMG.Watcher.Web.Controller.StatusTest do
 
     started_apps |> Enum.each(fn app -> :ok = Application.stop(app) end)
   end
+
+  @tag fixtures: [:watcher_sandbox, :root_chain_contract_config]
+  test "status endpoint returns expected response format without worrying about the Content-Type header" do
+    response_body = TestHelper.rpc_call("status.get", nil, 200, "text/plain")
+    %{"version" => "1.0", "success" => true, "data" => data} = response_body
+
+    assert %{
+             "last_validated_child_block_number" => last_validated_child_block_number,
+             "last_mined_child_block_number" => last_mined_child_block_number,
+             "last_mined_child_block_timestamp" => last_mined_child_block_timestamp,
+             "eth_syncing" => eth_syncing,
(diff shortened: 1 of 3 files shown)


apps/omg_watcher/test/web/controllers/status_test.exs
```diff
@@ -49,4 +49,24 @@ defmodule OMG.Watcher.Web.Controller.StatusTest do
 
     started_apps |> Enum.each(fn app -> :ok = Application.stop(app) end)
   end
+
+  @tag fixtures: [:watcher_sandbox, :root_chain_contract_config]
+  test "status endpoint returns expected response format without worrying about the Content-Type header" do
+    response_body = TestHelper.rpc_call("status.get", nil, 200, "text/plain")
+    %{"version" => "1.0", "success" => true, "data" => data} = response_body
+
+    assert %{
```
pdobacz: oh, actually I have a suggestion - let's have this test test only what it intends to test. Either just end the test on line 56 (with `assert`) or compare data from regular `TestHelper.success?("status.get")` call, instead of duplicating the detailed assertions

pnowosie:
:heavy_check_mark: AFAIC
This probably was merged into v0.1 - for the master I'd like to clean it a little, so I'm opting to keep associated issue open. Can we close this PR now?  (CC: @T-Dnzt @pdobacz )

pdobacz:
@pnowosie

> This probably was merged into v0.1

 note that this PR wants to merge into `v0.1`.

pnowosie:
Merged on @T-Dnzt request

Issue #608, opened 2019-04-09
Refactor ExitProcessor and its tests


pdobacz:
As it was noted, `ExitProcessor` either is rotting fast or has already rotted.

This issue is to track the tidying process, comprising of (among others, first draft list):
 - [x] tidies and dries to the imperative shell `exit_processor.ex` #609 
 - [x] streamlining the processing of standard exit challenges, making logic more testable and tested #618 
 - [x] reducing the test brittleness by swapping out heavy use of fixtures for helper functions #631 
 - [x] drying of `exit_processor/core_test.exs` using the above #631 
 - [x] separate testing of compatibility with `OMG.State.Core` from the rest of tests #631 
 - [x] refactors and drying in the logic module `exit_processor/core.ex` #684
 - [x] (optionally, if still required) splitting code/tests across more files #837 
 - [x] tidy handling of input-output indices (0-3 vs 0-7) #684
 - [x] increase readability and dry the invalid piggyback detection and proof building #684
 - ~reduce interactions with `Eth`, to consider:~
    - ~drop checking `status` of exits~ - can't do that - if an old exit is invalid during a small period of time (while syncing) it would cause `unchallenged_exit` and halt
    - ~store `exit_id` (or `txbytes`) from `call_data` and store in `OMG.DB`~ - meh, better do it after ALD lands
 - [x] investigate and improve performance issues related (possibly different issue) #754

pdobacz:
as the last of planned tidies finally got merged, closing this :tada:

PR #609, opened 2019-04-09
tidy exit processor pt1 - `ExitProcessor`

pdobacz:
Partially covers #608.

This PR focuses on drying `exit_processor.exs` code, especially the interactions with the data storage.

Also some readability improvements.

Probably reviewing commit-by-commit (and final state) is recommended

### DIFF
--- a/apps/omg_watcher/lib/omg_watcher/exit_processor.ex
+++ b/apps/omg_watcher/lib/omg_watcher/exit_processor.ex
@@ -273,19 +273,8 @@ defmodule OMG.Watcher.ExitProcessor do
 
   def handle_call({:finalize_in_flight_exits, finalizations}, _from, state) do
     _ = if not Enum.empty?(finalizations), do: Logger.info("Recognized ife finalizations: #{inspect(finalizations)}")
-
-    case Core.finalize_in_flight_exits(state, finalizations) do
-      {:ok, state, db_updates} ->
-        {:reply, {:ok, db_updates}, state}
-
-      {:unknown_piggybacks, unknown_piggybacks} ->
-        _ = Logger.error("Outputs not piggybacked: #{inspect(unknown_piggybacks)}")
-        {:stop, :unknown_piggybacks, state}
-
-      {:unknown_in_flight_exit, unknown_ifes} ->
-        _ = Logger.error("Unknown in-flight exits: #{inspect(unknown_ifes)}")
-        {:stop, :unknown_in_flight_exit, state}
-    end
+    {:ok, state, db_updates} = Core.finalize_in_flight_exits(state, finalizations)
+    {:reply, {:ok, db_updates}, state}
   end
 
   def handle_call(:check_validity, _from, state) do
--- a/apps/omg_watcher/lib/omg_watcher/exit_processor.ex
+++ b/apps/omg_watcher/lib/omg_watcher/exit_processor.ex
@@ -116,7 +116,11 @@ defmodule OMG.Watcher.ExitProcessor do
   end
 
   @doc """
-  Checks validity and causes event emission to `OMG.Watcher.Eventer`. Works with `OMG.State` to discern validity
+  Checks validity of all exit-related events and returns the list of actionable items.
+  Works with `OMG.State` to discern validity.
+
+  This function may also update some internal caches to make subsequent calls not redo the work,
+  but under unchanged conditions, it should have unchanged behavior from POV of an outside caller.
   """
   def check_validity do
     GenServer.call(__MODULE__, :check_validity)
@@ -280,13 +284,10 @@ defmodule OMG.Watcher.ExitProcessor do
 
       {:unknown_in_flight_exit, unknown_ifes} ->
         _ = Logger.error("Unknown in-flight exits: #{inspect(unknown_ifes)}")
-        {:stop, :unknown_in_flight_exit, Elixir.Agent.Server, state}
+        {:stop, :unknown_in_flight_exit, state}
     end
   end
 
-  @doc """
-  Combine data from `ExitProcessor` and `OMG.State` to figure out what to do about exits
-  """
   def handle_call(:check_validity, _from, state) do
     new_state = update_with_ife_txs_from_blocks(state)
 
@@ -375,9 +376,9 @@ defmodule OMG.Watcher.ExitProcessor do
     request
     |> run_status_gets()
     |> Core.determine_utxo_existence_to_get(state)
-    |> run_utxo_exists()
+    |> get_utxo_existence()
     |> Core.determine_spends_to_get(state)
-    |> run_spending_block_getting()
+    |> get_spending_blocks()
   end
 
   # based on in-flight exiting transactions, updates the state with witnesses of those transactions' inclusions in block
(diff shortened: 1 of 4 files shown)


apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -268,64 +269,46 @@ defmodule OMG.Watcher.ExitProcessor do
...
-
-      {:unknown_piggybacks, unknown_piggybacks} ->
-        _ = Logger.error("Outputs not piggybacked: #{inspect(unknown_piggybacks)}")
-        {:stop, :unknown_piggybacks, state}
-
-      {:unknown_in_flight_exit, unknown_ifes} ->
-        _ = Logger.error("Unknown in-flight exits: #{inspect(unknown_ifes)}")
-        {:stop, :unknown_in_flight_exit, Elixir.Agent.Server, state}
-    end
+    {:ok, state, db_updates} = Core.finalize_in_flight_exits(state, finalizations)
```
pik694: I assume that we no longer have that `case`, as if this returned something other than `{:ok, ...}`, then something really bad must have happened

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -268,64 +269,46 @@ defmodule OMG.Watcher.ExitProcessor do
...
+    {:reply, {:ok, db_updates}, state}
   end
 
   @doc """
   Combine data from `ExitProcessor` and `OMG.State` to figure out what to do about exits
   """
   def handle_call(:check_validity, _from, state) do
```
pik694: I am not sure whether the atom `:check_validity` is relevant to what the handler does

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -268,64 +269,46 @@ defmodule OMG.Watcher.ExitProcessor do
...
-    {state1, request} = prepare_validity_check(state)
-    {chain_status, events} = Core.invalid_exits(request, state1)
-    {:reply, {chain_status, events}, state}
+    state1 =
+      %ExitProcessor.Request{}
```
pik694: is it a part of some greater refactor? If no then why not pass default argument?

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -268,64 +269,46 @@ defmodule OMG.Watcher.ExitProcessor do
...
+      |> update_with_ife_txs_from_blocks(state)
+
+    response =
+      %ExitProcessor.Request{}
+      |> fill_request_with_data(state1)
```
pik694: same here

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pik694: is it a part of some greater refactor? If no then why not pass default argument?

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pik694: I assume that we no longer have that `case`, as if this returned something other than `{:ok, ...}`, then something really bad must have happened

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -268,64 +269,46 @@ defmodule OMG.Watcher.ExitProcessor do
...
   end
 
   @doc """
   Combine data from `ExitProcessor` and `OMG.State` to figure out what to do about exits
   """
   def handle_call(:check_validity, _from, state) do
-    {state1, request} = prepare_validity_check(state)
-    {chain_status, events} = Core.invalid_exits(request, state1)
-    {:reply, {chain_status, events}, state}
+    state1 =
```
pik694: maybe `updated_state` or `new_state`. I really do not like naming with enumeration

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -268,64 +269,46 @@ defmodule OMG.Watcher.ExitProcessor do
...
       %ExitProcessor.Request{}
-      # TODO: same comment as above in get_competitor_for_ife
-      |> run_status_gets()
-      |> Core.determine_utxo_existence_to_get(state)
-      |> run_utxo_exists()
-      |> Core.determine_spends_to_get(state)
-      |> run_spend_getting()
-      |> Core.determine_blocks_to_get()
-      |> run_block_getting()
+      |> fill_request_with_data(state)
```
pik694: what about `create_request_environment` or sth similar?

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -334,22 +317,23 @@ defmodule OMG.Watcher.ExitProcessor do
...
       |> Core.get_input_challenge_data(state, txbytes, input_index)
 
     {:reply, response, state}
   end
 
   def handle_call({:get_output_challenge_data, txbytes, output_index}, _from, state) do
-    {state1, request} = prepare_validity_check(state)
-    response = Core.get_output_challenge_data(request, state1, txbytes, output_index)
-    {:reply, response, state}
+    state1 =
```
pik694: as above

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pdobacz: yes, finalizing non-existent ifes or non-piggybacked outputs is an impossible state, so we prefer a simple crash then.

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pdobacz: it is `GenServer.called` by `check_validity/0` above and it calls `Core.check_validity`.

When it comes to the name itself - the call in general checks the validity of the exits - it checks standard and in-flight exits for all possible kinds of "invalidities" (aka `byzantine_events`).

`check_validity` looks like a reasonable name to me, but do we have any alternative proposition?

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pdobacz: ah right, this is some brain fart here. 

But I kept the explicit `%ExitProcessor.Request` to be piped to `fill_request_with_data`, because otherwise it is much less readable - I want to emphasize that we're starting off with a clean request struct and then piping it through the process.

But for `update_with_ife_txs_from_blocks` the argument makes no sense

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pdobacz: ah replied above

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pdobacz: see the explicitness/readability argument above

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pik694: You are right, but the name suggests that the function might be pure (just checks the validity), but it is not that important so we wanted to waste time on it :wink:

pdobacz: oh, but this is a v. good point! To an external user, `check_validity` is what is being done. It's still "pure" in the sense that the stuff that is non-pure is just caching being done.

I'll keep the name, but put docs in stating, that this f might also do some amount of caching for subsequent calls

PR #614, opened 2019-04-11
Discuss and document the `elixir-omg`-`plasma-contracts` versioning/branching

pdobacz:


docs/branching.md
```diff
@@ -0,0 +1,32 @@
...
+
+Rationale:
+- the history and the relations between the versions are readable and accountable,
+- we can predictably sync our respective watchers/run child chain servers against deployed contracts
+- and at the same time be able to move with `master`s quickly
+
+("->" means that the `mix.exs` holds the following branch)
+
+- `elixir-omg/v0.1` -> `plasma-contracts/v0.1`, that's understood, the `mix.lock` points to a specific frozen version, no changes expected there
+- `elixir-omg/master` -> `plasma-contracts/master`, mix lock points to the contracts ~deployed on `devel`~ we want to CI-check with `elixir-omg`. So it might be newer than what contract is deployed on `devel`
```
kasima: Confusing language here.

> mix lock points to the contracts ~deployed on `devel`~

and

> it might be newer than what contract is deployed on `devel`

Are you trying to say that we will try to point to the contract deployed on the `development` environment, except contract deployment is a manual process right now.

Ideally, for `development` environment, the contract gets re-deployed when there is a dependency update. This requires some tooling work and probably want to decouple contract deployments out of the elixir-omg before doing this.

docs/branching.md
```diff
@@ -0,0 +1,32 @@
...
+- `elixir-omg/v0.2` -> `plasma-contracts/v0.2`, the commit that fixes *the branch `plasma-contracts/v0.2` in `elixir-omg`* will need to be constantly rebased onto `master`, as long we keep looping features into `elixir-omg/v0.2`. *The mix lock will always point to the contract commit currently deployed on `devel`*.
```
kasima: Rebasing changes on release branch onto master may cause a lot of grief, since we'll be replaying the changes on a moving base. It will be easier to simply merge.

What do you mean by "looping features into elixir-omg/v0.2"? From master?

docs/branching.md
```diff
@@ -0,0 +1,32 @@
...
+
+When want to integrate to a newer contract, we'll:
+- make a PR to `elixir-omg/master` bumping the contract version in `mix.lock` and (possibly) providing compatibility with any possible breaking changes.
+This will run CI checks on the new integration
+- PR is merged
+- this is the moment when `elixir-omg/master` might not sync to `devel` anymore, if changes to contract were breaking!
+If they weren't breaking, we can end here, all should sync fine
+- if changes were breaking:
+- fast forward `plasma-contracts/v0.2` to the desired commit on `master`
```
kasima: Can we use the same branch pairs here? If this for a PR in `elixir-omg/master`, then wouldn't it be pointing to a fixed SHA on `plasma-contracts`? Not necessarily a release branch (`v0.2`).

docs/branching.md
```diff
@@ -0,0 +1,32 @@
...
+- have a PR/commit/rebase to bump `mix.lock` version of `plasma-contracts` in `elixir-omg/v0.2`.
+
+Also, if we want to upgrade contracts to be potentially promoted to `staging-v0.2`/`Ari-v0.2`, we should test on `devel` first, so we'd like to reset `devel` under such circumstances too (even if changes had been non-breaking above).
```
kasima: In this situation, the contracts have been tested on the `development` environment for a while already. Promotion would only mean a new release of `staging`. `staging` is for testing a release.

docs/branching.md
kasima: Sorry, missed the markdown strike-through. Ignore the first part.

docs/branching.md
pdobacz: I'm not sure I know what you mean here.

Let me rephrase perhaps - for the sake of explaining myself and also to get a better grip of the process (it's quite dizzying...)

I think the difference between `v0.2` and `master` in this example, is that `master` moves the quickest in both repos.

In contracts - `master` is always a set of building/test-passing code with the latest features/updates/fixes.

In elixir-omg - `master` - same applies to ch-ch/watcher. Now the trick is, that `elixir-omg/master` doesn't always use the topmost `plasma-contracts/master` SHA - because of the `mix.lock` and to have reproducible CI etc etc. So to move that forward with this, there must be this explicit PR to `elixir-omg/master`.

Now when this PR passes CI, we know that _should the `development` contract be redeployed from scratch_ on merge, everything should work fine. But we are not sure whether this change is backwards compatible, i.e. the new `elixir-omg/master` will work with the currently deployed `development`.

If it is expected to work, then nothing needs to happen - we can just continue to use the "old" `development` with the new `master`.

But if either it isn't or we want to test the new `v0.2` on a deployed env, we must redeploy the `development`. This is where `plasma-contracts/v0.2` would "come into play" - it sits at the spot, which is the SHA that is currently deployed on `development`. And now _should we decide to feature freeze the contract here_ this is the spot where `v0.2` would remain forever, ultimately being used to deploy `staging`/`Ari` (maybe we could allow to have well-reviewed hot-fixes sneak in there, but not sure).

I've rephrased the last step of the process a bit to maybe be clearer.

docs/branching.md
pdobacz: Hm, maybe I've been unclear here.

There are two circumstances when we want to redeploy the `development`:
  - we must, because there's some important feature landing, which is non-backwards-compatible - like the `exit_id` fix Piotr did recently
  - we want, but don't have to. For example, 3 bugs were fixed in the contract, but the contract remained compatible with `elixir-omg` throughout these changes, so we could've continued to use the "old `development`" for some time. But then we want these bugfixes in `staging`/`Ari` so, despite still being technically compatible, we want to upgrade `development` and test the contract there.

So this sentence says "Okay, you know now what we do about non-backwards-compatible changes to the contract. Let me tell you about other situations where we want to redeploy".

docs/branching.md
kasima: Yes, that's what I mean, to have the contract re-deploy automatically. There's some yak shaving we need to do to decouple the deployment tooling.

docs/branching.md
kasima: That language works better, thanks.

boolafish:
(Please don't consider this comment as any kind of blocker, just never fully understand the reason behind yet, and is curious about this.)

Can I ask a (might be dumb) question that why are we preferring the branching-releasing approach instead of continuous delivery?

pdobacz:
@boolafish 

> why are we preferring the branching-releasing approach instead of continuous delivery

The apps (child chain server, watcher etc.) can and actually do CD (I think).

The contract code is the problematic part, since delivering it means wiping the state of the child chain clean (since we don't have the upgrade mechanism yet - and even then an upgrade to contract code might remain a manual process, at least for the mainnety deployments).

Not sure if this answers your question though

boolafish:
@pdobacz thanks! Now it's clear to me : )

docs/branching.md
kasima: > If it is expected to work, then nothing needs to happen - we can just continue to use the "old" development with the new master.

This is not a state we want to be in. It's going to cause confusion.

If `mix.lock` is updated with a new SHA on `plasma-contracts` repo, want to enforce that the _development_ environment *always* matches the `elixir-omg/master`, including its dependencies. This may require manual deployment steps now, but the goal is to automate this.

> But if either it isn't or we want to test the new v0.2 on a deployed env, we must redeploy the development.

There is an assumption here that is making things confusing. If we need to test `elixir-omg/v0.2` on a deployed env, we won't use the _development_ environment. We would provision a new environment to test, as necessary.

Fewer variables for these environments means less confusion. We want to declare that _development_ deploys `elixir-omg/master` and its dependencies, always.

kasima:
Simplified proposal in #692.

PR #616, opened 2019-04-12
Performance metrics

pthomalla:


apps/omg_db/lib/db.ex
```diff
@@ -29,64 +30,75 @@ defmodule OMG.DB do
...
-    GenServer.call(server_name, {:blocks, blocks_to_fetch})
+  @decorate transaction_event(:LevelDB)
+  def blocks(blocks_to_fetch, server_name \\ @server_name) do
+    case blocks_to_fetch do
+      [] -> {:ok, []}
+      blocks_to_fetch -> GenServer.call(server_name, {:blocks, blocks_to_fetch})
+    end
   end
 
+  @decorate transaction_event(:LevelDB)
```
pdobacz: I'm still thinking about the naming convention for the event groups and namespaces. I think there's a problem with camel-case entries like `LevelDB` as they suggest a module, while there is no such module strictly speaking.

Perhaps we should consider using:
1/ `__MODULE__` for event groups always, consistently
2/ something more top-level for transactions' namespaces? `web`/`background` is the appsignals default, which I agree is pretty coarse, but maybe it will we what is most useful actually :thinking: ?

We also need to take AppSignal's UX into account - switching namespaces filters is very clunky, so the less the better IMHO

apps/omg_db/lib/omg_db/leveldb_server.ex
```diff
@@ -64,7 +64,26 @@ defmodule OMG.DB.LevelDBServer do
     end
   end
 
-  def handle_call({:multi_update, db_updates}, _from, state) do
+  def handle_call({:multi_update, db_updates}, _from, state), do: multi_update(db_updates, state)
```
pdobacz: I think we have two options here:

1/ add a `OMG.DBServer` behavior which implements this, which has callbacks like `multi_update`, which calls `GenServer.call(__MODULE__...)` and is in turn called by `OMG.DB`. Then you'll decorate the callback implementations in `LevelDBServer` - preferred option but more work
2/ rename to `do_multi_update` and alike and keep the `do_` functions decorated private - lazy option, just to be consistent

We can also consider
3/ anticipate doing (1/) in a separate PR and keep the current name, but leave a `TODO behind`, copying (1/) into the body of the TODO

apps/omg_watcher/lib/omg_watcher/api/account.ex
```diff
@@ -18,11 +18,13 @@ defmodule OMG.Watcher.API.Account do
   """
 
   alias OMG.Watcher.DB
+  use Appsignal.Instrumentation.Decorators
 
   @doc """
   Returns a list of amounts of currencies that a given address owns
   """
   @spec get_balance(OMG.Crypto.address_t()) :: list(DB.TxOutput.balance())
+  @decorate transaction(:Watcher_API_Account)
```
pdobacz: no snakel case + see other comments on naming conventions. In this particular case I think the default `web` namespace is totally ok.

Oh, actually, you don't need a transaction at all - Phoenix does this already!

Find & fix all occurrences

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -155,8 +157,12 @@ defmodule OMG.Watcher.BlockGetter do
           Core.t()
         ) :: {:noreply, Core.t()} | {:stop, :normal, Core.t()}
   def handle_info(msg, state)
+  def handle_info(:producer, state), do: producer(state)
+  def handle_info({_ref, {:downloaded_block, response}}, state), do: downloaded_block(response, state)
+  def handle_info({:DOWN, _ref, :process, _pid, :normal} = _process, state), do: {:noreply, state}
+  def handle_info(:sync, state), do: sync(state)
 
-  def handle_info(:producer, state) do
+  defp producer(state) do
```
pdobacz: weren't you intending to decorate these new functions?

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -98,15 +99,16 @@ defmodule OMG.Watcher.BlockGetter do
     {:reply, Core.chain_ok(state), state}
   end
 
-  def handle_cast(
-        {:apply_block,
+  def handle_cast({:apply_block, %BlockApplication{} = to_apply}, state), do: apply_block(to_apply, state)
```
pdobacz: similar comment as with `LevelDBServer.multi_update`, just here it's simpler (no need for a behavior etc), so maybe let's just invert the approach: have `apply_block` call `GenServer.cast` and decorate that.

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -98,15 +99,16 @@ defmodule OMG.Watcher.BlockGetter do
...
+
+  defp apply_block(
```
pdobacz: solving `apply_block`'s metrics is the issue I mentioned that needs to be figured out.

Somewhere probably here, you need to define a `background`-namespace transaction that will encompass processing of a single block - so sounds like

```
@decorate transaction 
def apply_block
```

or
```
@decorate transaction 
defp do_apply_block
```

InoMurko:
This approach couples us with Appsignal quite a lot - what if we go Telemetry way?
```
In a nutshell, you register a custom module and function to be invoked for certain events, which are executed whenever there is such event. Event name is a list of atoms. Each event is composed of a numeric value and can have metadata attached to it. Let's see an example.
```

apps/omg_watcher/lib/omg_watcher/api/account.ex
InoMurko: Why is this a Transaction and others are transaction events? If I understood the docs correctly you need to start a transaction first, which is followed by transaction events...

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -44,6 +45,7 @@ defmodule OMG.Watcher.ExitProcessor do
   Accepts events and processes them in the state - new exits are tracked.
   Returns `db_updates`
   """
+  @decorate transaction_event(:ExitProcessor)
```
InoMurko: Under which transaction are these transaction events collected?

apps/omg_watcher/lib/omg_watcher/api/account.ex
pdobacz: what others do you mean? 

I guess we should use `event` whenever we somehow know, that a transaction is "ongoing", and this PR attempts to do that, with some polishes required (more on such suggestions in my comments).

Here I think (and in all other APIs called directly by Phoenix controllers), we can safely assume that we're running under phoenix (or some other "entrypoint") and the txs are already ongoing. 

OTOH, I'm not sure whether we'd like to have an `event` here, instead of just nothing. I suggested "nothing", but come to think of it - an `event` would allow to see the portion of the request handling spent in here, as contrasted with Phoenix overhead. Not sure which is better - we can for now go with a "rich" version (with `event`) and later ditch the events, if they're noisy or overheady.

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pdobacz: should be driven by `EthereumEventListeners` who are effectively managing the background jobs here, and calling into here.

Come to think of it, the above is quite secret knowledge. It would be cool to have sth like
```
@decorate transaction_or_event(...)
```

But otherwise, I think the way to go now would just be to `event` **everywhere** except where we're sure we're either handling an incoming request (`web` namespace) or acting on our own by starting a background job (`background` namespace)

InoMurko: I'm gonna list why I don't like these annotations:
- they seem to couple us with appsignal 
- we don't know what the performance impact of decorations is and it seems like there's no way to turn them off (which is scary!)
- there ought to be a cleaner way to gather and measure events (Telemetry maybe?) and have a pluggable backend for metrics

apps/omg_watcher/lib/omg_watcher/api/account.ex
InoMurko: It seems like the flow of what a transaction is and what an event is somewhat hidden in the macros. implicit implementation means we will have difficulties measuring the performance impact and because of that, I really think we ought to consider other approaches

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pdobacz: @InoMurko

>I'm gonna list why I don't like these annotations:

Generally, I agree with your 3 concerns here, but wouldn't we rather just get AppSignal metrics up and running quickly, and possibly move to something more elegant at a later stage?

I've just given it a quick scan, and it seems going the Telemetry way (+ AppSignal's custom metrics, to have the similar final effect, correct?), would require us to collect all the low level measurement values manually.

BTW, you suggested instrumenting using decorators in the first place, which part of the execution specifically made you change your opinion?

Regardless, I'll let you guys (@pthomalla) figure out the best first approach here.

pthomalla: we could make our `decorate`, where we can turn off some measurement

apps/omg_db/lib/db.ex
```diff
@@ -20,6 +20,7 @@ defmodule OMG.DB do
 
   ### Client (port)
 
+  use OMG.Utils.Metrics
```
InoMurko: please specify what you're importing

apps/omg_utils/mix.exs
```diff
@@ -25,5 +25,5 @@ defmodule Utils.MixProject do
   defp elixirc_paths(:prod), do: ["lib"]
   defp elixirc_paths(_), do: ["lib", "test/support"]
 
-  defp deps, do: [{:ecto_sql, "~> 3.1"}]
+  defp deps, do: [{:ecto_sql, "~> 3.1"}, {:appsignal, "~> 1.9"}]
```
InoMurko: why is ecto_sql included again?

apps/omg_watcher/lib/omg_watcher/db/txoutput.ex
```diff
@@ -211,6 +222,7 @@ defmodule OMG.Watcher.DB.TxOutput do
   end
 
   @spec get_sorted_grouped_utxos(OMG.Crypto.address_t()) :: %{OMG.Crypto.address_t() => list(%__MODULE__{})}
+  @decorate measure_event()
```
InoMurko: wouldn't it make more sense, to measure this event in `OMG.Watcher.Web.Controller.Transaction.create/2`?

apps/omg_watcher/lib/omg_watcher/api/in_flight_exit.ex
```diff
@@ -69,6 +72,7 @@ defmodule OMG.Watcher.API.InFlightExit do
 
   This delegates directly to `OMG.Watcher.ExitProcessor` see there for details
   """
+  @decorate measure_event()
```
InoMurko: Wouldn't it make more sense to measure the event from `OMG.Watcher.Web.Controller.InFlightExit.prove_canonical/2`?

apps/omg_watcher/lib/omg_watcher/db/txoutput.ex
pthomalla: it catches by phoenix (and make transaction with namespace web)

InoMurko: but how does the value differ if this is actually the first thing it calls? why do we need to measure it separately?

apps/omg_watcher/lib/omg_watcher/api/in_flight_exit.ex
pthomalla: yes, it's more sense.
and this is already happening, then it is opened appsignal transaction with namespace web

InoMurko: so we're measuring the same thing twice, just the namespace is different?

apps/omg_utils/mix.exs
pthomalla: I worked on the older version of the master branch, 
I am already doing rebase

apps/omg_watcher/lib/omg_watcher/db/txoutput.ex
pthomalla: thanks to the appsignal event, we can divide the time to execute appsignal transaction into smaller parts and check what really takes us the most time

apps/omg_watcher/lib/omg_watcher/api/in_flight_exit.ex
pthomalla: o start the measurements you have to call appsignal transaction, and the event can check how the transaction time will spread

apps/omg/lib/omg/ethereum_event_listener.ex
```diff
@@ -91,7 +92,9 @@ defmodule OMG.EthereumEventListener do
     {:noreply, {initial_state, callbacks_map}}
   end
 
-  def handle_info(:sync, {%Core{} = core, _callbacks} = state) do
+  def handle_info(:sync, state), do: do_sync(state)
+  @decorate measure_start()
+  def do_sync({%Core{} = core, _callbacks} = state) do
```
InoMurko: could this be private?

InoMurko: could this be private?

apps/omg/lib/omg/state.ex
```diff
@@ -36,6 +38,7 @@ defmodule OMG.State do
     GenServer.start_link(__MODULE__, :ok, name: __MODULE__)
   end
 
+  @decorate measure_event()
```
InoMurko: sometimes the decorator are above specs, sometimes below...

apps/omg/lib/omg/state.ex
```diff
@@ -189,19 +195,26 @@ defmodule OMG.State do
 
   Does its on persistence!
   """
+  @decorate measure_event()
```
InoMurko: is measuring this event twice, necessary?

apps/omg_watcher/lib/omg_watcher/api/status.ex
```diff
@@ -36,6 +38,7 @@ defmodule OMG.Watcher.API.Status do
   last mined child block number and it's timestamp, and a flag indicating if watcher is syncing with Ethereum.
   """
   @spec get_status() :: {:ok, t()} | {:error, atom}
+  @decorate measure_event()
```
InoMurko: as this function body is quite big ... would the result of the measurement tell you anything meaningful? would the result tell you where in the body it spent a lot of time?

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -46,6 +47,8 @@ defmodule OMG.Watcher.BlockGetter do
     GenServer.start_link(__MODULE__, :ok, name: __MODULE__)
   end
 
+  def apply_block(%BlockApplication{} = to_apply), do: GenServer.cast(__MODULE__, {:apply_block, to_apply})
```
InoMurko: why is this necessary? wouldn't it be better for the same execution to cast the message? dynamic function invocation is 6x slower then normal...

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -206,7 +213,7 @@ defmodule OMG.Watcher.BlockGetter do
 
       _ = Logger.debug("Synced height is #{inspect(synced_height)}, got #{length(blocks_to_apply)} blocks to apply")
 
-      Enum.each(blocks_to_apply, &GenServer.cast(__MODULE__, {:apply_block, &1}))
+      Enum.each(blocks_to_apply, &apply_block(&1))
```
InoMurko: reason for change?

apps/omg_watcher/lib/omg_watcher/db/txoutput.ex
```diff
@@ -112,13 +116,16 @@ defmodule OMG.Watcher.DB.TxOutput do
     }
   end
 
+  @decorate measure_event()
```
InoMurko: isn't this already handled? https://hexdocs.pm/appsignal/Appsignal.Ecto.html

apps/omg/lib/omg/state.ex
pthomalla: and what do you prefer, before spec or after?

apps/omg/lib/omg/state.ex
pthomalla: why twice? I do not understand

apps/omg_watcher/lib/omg_watcher/api/status.ex
pthomalla: yes, it tells me how much time spend in RootChain, and ExitProcessor and BlockGetter
(in this modules is also measure_event)

apps/omg_watcher/lib/omg_watcher/block_getter.ex
pthomalla: at first, I thought that I cant decorate GenServer.cast/GenServer.call. 
And at this point, we could remove this.

Dynamic function? could You send me some article about why elixir execute this function so slow? 
I understand that this way will have one pater matching more. 
but that shouldn't make such a big difference

apps/omg_watcher/lib/omg_watcher/block_getter.ex
pthomalla: GenServer.cast/GenServer.call with decorate are very tricky :D 
I will restore the use of GenServer.cast

apps/omg_watcher/lib/omg_watcher/db/txoutput.ex
pthomalla: a bit yes,
but it seems to me that in ecto there is no time for processing into structs (creating query, and transform to result), only the time of working on the database

apps/omg_watcher/lib/omg_watcher/block_getter.ex
InoMurko: http://erlang.org/doc/efficiency_guide/functions.html#function-calls

apps/omg/lib/omg/state.ex
InoMurko: before spec

apps/omg/lib/omg/state.ex
InoMurko: Actually, it's three times, with little difference 
1. https://github.com/omisego/elixir-omg/pull/616/files/c7d58abfb73bd2f4e76f5c9340320461f24cd446#diff-cfee1fb662913b3948db6872bbce2278R198
2. https://github.com/omisego/elixir-omg/pull/616/files/c7d58abfb73bd2f4e76f5c9340320461f24cd446#diff-cfee1fb662913b3948db6872bbce2278R212
3. https://github.com/omisego/elixir-omg/pull/616/files/c7d58abfb73bd2f4e76f5c9340320461f24cd446#diff-731c57d623f5cd277dc2fcaf9abcae43R33

apps/omg_watcher/lib/omg_watcher/block_getter.ex
InoMurko: why not call a local function?

apps/omg/lib/omg/state.ex
pthomalla: ok, 
1. whole time to form_block
2. multi_update (only time DB update use by multiple functions)
3. more tricky multi_update with waiting for a response (waiting to get into the gen server with long message queues, can be problematic. can tell us about the change of approach sending messages)

apps/omg_watcher/lib/omg_watcher/block_getter.ex
pthomalla: I had the same dilemma :D
1. GenServer timer, it can fire if whole apply_block will make here
2. GenServer approach is to make small task
3. supposedly, the message queue is quite fast (erlangu / elixir puts quite a lot of effort in it )

InoMurko: not sure I understand

apps/omg/lib/omg/state.ex
InoMurko: 2 and 3 are the same

apps/omg_watcher/lib/omg_watcher/block_getter.ex
pdobacz: let me jump in, as I just answered the same question :).

The reason for this is that we want to handle block application on a per-block basis. Mainly it is for more granularity - application of every block takes the getter from one valid state to another, so we wanted to replicate that in the GenServer messages.

Also a long range of large blocks to apply would result in the call to `sync` block the Process for a long time, chunking it into blocks keeps it more responsive - eg. after we received and sent some blocks to `apply_block`, we might want to immediately start pulling new ones (I think this is how this works, but not 100% sure)

There's also some historical reasons - we used to `stop` this `GenServer` under some logic-driven circumstances, this idea was rightfully abandoned, but the splitting of block application remains.

InoMurko: Not arguing with the logical implementation: just how the measurement was done :)

apps/omg_utils/lib/omg_utils/metrics.ex
```diff
@@ -0,0 +1,47 @@
...
+  use Decorator.Define,
+    measure_start: 0,
+    measure_event: 0
+
+  @discard_namespace_metrics Application.get_env(:metrics, :discard, [])
+
+  def measure_start(body, context) do
+    namespace = context.module |> Module.split() |> List.last() |> String.to_existing_atom()
+
+    if Enum.find(@discard_namespace_metrics, &match?(^&1, namespace)),
```
pdobacz: let's keep at the back of our heads that this might be slow, if there are a lot of `discards`. But for now I think we can keep it simple.

apps/omg_utils/lib/omg_utils/metrics.ex
```diff
@@ -0,0 +1,47 @@
...
+      do: body,
+      else: Appsignal.Instrumentation.Decorators.transaction(namespace, body, context)
+  end
+
+  def measure_event(body, context) do
+    namespace = context.module |> Module.split() |> List.last() |> String.to_existing_atom()
```
pdobacz: this is not a `namespace` for AppSignal, it is an event group, IIRC

apps/omg_utils/lib/omg_utils/metrics.ex
```diff
@@ -0,0 +1,47 @@
...
+    measure_event: 0
+
+  @discard_namespace_metrics Application.get_env(:metrics, :discard, [])
+
+  def measure_start(body, context) do
+    namespace = context.module |> Module.split() |> List.last() |> String.to_existing_atom()
+
+    if Enum.find(@discard_namespace_metrics, &match?(^&1, namespace)),
+      do: body,
+      else: Appsignal.Instrumentation.Decorators.transaction(namespace, body, context)
```
pdobacz: what about my comment on the namespaces/event groups? 

> I'm still thinking about the naming convention for the event groups and namespaces. I think there's a problem with camel-case entries like LevelDB as they suggest a module, while there is no such module strictly speaking.
> 
> Perhaps we should consider using:
> 1/ __MODULE__ for event groups always, consistently
> 2/ something more top-level for transactions' namespaces? web/background is the appsignals default, which I agree is pretty coarse, but maybe it will we what is most useful actually thinking ?
> 
> We also need to take AppSignal's UX into account - switching namespaces filters is very clunky, so the less the better IMHO

are we dead set to have such wealth of namespaces? what are the benefits

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -156,7 +158,12 @@ defmodule OMG.Watcher.BlockGetter do
         ) :: {:noreply, Core.t()} | {:stop, :normal, Core.t()}
   def handle_info(msg, state)
 
-  def handle_info(:producer, state) do
+  def handle_info(:producer, state), do: do_producer(state)
+  def handle_info({_ref, {:downloaded_block, response}}, state), do: do_downloaded_block(response, state)
+  def handle_info({:DOWN, _ref, :process, _pid, :normal} = _process, state), do: {:noreply, state}
+  def handle_info(:sync, state), do: do_sync(state)
+
+  defp do_producer(state) do
```
pdobacz: missed comment again:

> weren't you intending to decorate these new functions?

apps/omg_utils/lib/omg_utils/metrics.ex
pthomalla: this is done during compilation

apps/omg_utils/lib/omg_utils/metrics.ex
pthomalla: We only use the last parts from the module name. (appsignal does not let you use a dot)
appsignal default namespace for transaction is http_request.

We are limited to names that can be found in atoms, 
without string pattern (:"any name" ps. module name is :"Elixir.module_name")

We can still find replacement signs for dots, e.g. ```@```

apps/omg_utils/lib/omg_utils/metrics.ex
pdobacz: ah right :+1:

apps/omg_utils/lib/omg_utils/metrics.ex
pdobacz: again: why not use just `background` vs `http_request`, the two default namespaces AppSignal provides? The question was "what are the benefit of using your multiple custom namespaces?", nothing about dots.

and where is the example result to look at?

apps/omg_watcher/lib/omg_watcher/block_getter.ex
pdobacz: is there a reason only `do_producer` is decorated out of the new functions broken apart?

apps/omg_utils/lib/omg_utils/metrics.ex
pthomalla: image
when we use custom names, 
we have more information (lock at ExitProcessor, from BlockGetter we use 39.69% and from EthereumEventListener only 1.09%). 
more precise cause of usage

pdobacz: OK, I'm not sure which is better, let's just proceed like this and try in practice, unless anyone has further objections.

I suggest to leave a comment inside the moduledoc:

```
NOTE: the namespace and event group naming convention here is tentative.
It is possible we'll revert to standard coarser division into `web` and `background` namespaces Appsignal suggests
```

PR #604, opened 2019-04-09
Refactor swagger docs

mederic-p:
# Overview

This PR introduces a refactoring of the 2 watcher api specs.
There was some inconsistencies in the documentation and some endpoints/params were outdated/not available.

The refactoring consists in a split into different files for better readability and maintainability.
When developing and updating the spec, these files should be updated and then the full spec should be re-generated based on these files.

For more details, you can check the [added documentation](https://github.com/omisego/elixir-omg/blob/refactor-swagger-doc/apps/omg_watcher/priv/swagger/swagger.md).

PR #624, opened 2019-04-17
Update api docs with latest specs

mederic-p:
This PR updates the 3 specs files for `0.1` with their latest design from [#604](https://github.com/omisego/elixir-omg/pull/604).

> Note: We should can until [#604](https://github.com/omisego/elixir-omg/pull/604) is merged before merging this one in case there are more changes


pnowosie:
Since #604 is accepted I'm merging it. Thank you @mederic-p again ❗️

PR #625, opened 2019-04-17
fix: swap dependencies with existing ones

InoMurko:


InoMurko:
The MerkleTree dependency was restored and fixed and the update of the dependency postponed.

Issue #626, opened 2019-04-17
Fix performance issues with producing merkle proofs in `merkle_tree` v1.5.0


pdobacz:
See https://github.com/omisego/elixir-omg/pull/486#issuecomment-483994525

It seems that producing merkle proofs is underperforming, while it seems to have been quick in the hacky https://github.com/omisego/merkle_tree/tree/refactor fork branch.

It should have no negative effects on the app (these proofs are pulled relatively rarely, and the delay is on the order of magnitude of ms). It does however make our unit test slooowish, so let's fix this for productivity's sake.

pdobacz:
fixed in #631 / aadf98bae455f1852e8254b03f9dfdc643fbe75b

PR #631, opened 2019-04-18
tidy exit processor pt3 - `ExitProcessor.CoreTest`

pdobacz:
Continuing work towards #608 

This PR aims at DRYing, splitting up (a bit) and simplifying the tests of `ExitProcessor.Core`.

Use of `fixtures` has been greatly reduced. Firstly, there's a lot less reliance on shared fixtures between tests, so that modifying the fixture to add a more specific test isn't necessary - one can now much more easily setup the new tests using helper functions etc. The "rich" fixtures like `processor_filled` are intended to be used whenever we can take them "as they are" and rely only on the generic setup they contain.

This effort can be continued. Things left to do:
  - divide up the large remaining test into `describe` blocks or separate test files, to make the setup more focused and readable. The way this test is arranged should be thought through carefully
  - put all the tests in their respectful categories; right now their very garbled. There is some repetition, and probably some specific behaviors might still be untested.

No tests where removed or changed in terms of logic. Some tests were split out to `state_interaction_test.exs`. The order of tests hasn't been altered to make it more reviewable

The new approach to test setup required us to optimize the `merkle_tree` library, which this PR also includes

### DIFF
--- a/apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
+++ b/apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
@@ -460,19 +460,14 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
              Core.determine_utxo_existence_to_get(%ExitProcessor.Request{blknum_now: @late_blknum}, empty)
   end
 
-  @tag fixtures: [:alice, :processor_empty, :in_flight_exit_events, :contract_ife_statuses]
+  @tag fixtures: [:alice, :processor_empty, :transactions]
   test "ifes and standard exits don't interfere", %{
     alice: alice,
     processor_empty: processor,
-    in_flight_exit_events: [one_ife | _],
-    contract_ife_statuses: [one_ife_status | _]
+    transactions: [one_ife | _]
   } do
     standard_exit_tx = TestHelper.create_recovered([{1, 0, 0, alice}], @eth, [{alice, 10}])
-
-    {processor, _} =
-      processor
-      |> start_se_from(standard_exit_tx, @utxo_pos1)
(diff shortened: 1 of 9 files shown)


pdobacz:
OH, after review/discussion, I'll squash some of the commits, as indicated in the commit msgs

apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
```diff
@@ -1550,22 +1521,19 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
     @tag fixtures: [:alice, :processor_filled, :transactions, :competing_transactions]
     test "handle two competitors, when the younger one already challenged",
          %{alice: alice, processor_filled: processor, transactions: [tx1 | _], competing_transactions: [comp | _]} do
-      txbytes = Transaction.raw_txbytes(tx1)
-      other_txbytes = Transaction.raw_txbytes(comp)
-
-      %{signed_tx: %{sigs: [other_signature, _]}} =
-        other_recovered = OMG.TestHelper.sign_recover!(comp, [alice.priv, alice.priv])
+      txbytes = txbytes(tx1)
+      {comp_txbytes, comp_signature} = {txbytes(comp), sig(comp)}
```
paulperegud: welcome to the dark side ;-)

pdobacz: do you mean the abbreviations or the tuple magic? :)

If the former - I was also a bit reluctant at first, but the brevity benefits + using a rather repeating (and scope-limited) abbreviations `tx`, `comp`, `sig`... so here I am :joy:

pdobacz:
forgot to mention it fixes #626

PR #635, opened 2019-04-19
feature: raise an alarm when contracts are not deployed properly or c…

InoMurko:
…lient is still syncing


InoMurko:
Will review this in the afternoon. It’s getting old!

PR #640, opened 2019-04-25
feat: watcher application metrics

pthomalla:
regarding the balance for the fee and the total number of tokens in our contract.
Possible solutions:

1. adding them to the structure in state/core.ex.
Additional fields:
  -- balances: %{Crypto.address_t() => non_neg_integer()}
  -- fee_balances: %{Crypto.address_t() => non_neg_integer()}
balances and fee_balances are closely related to state/core .ex and state.ex
if we put it elsewhere, there is a greater risk disperse control over the balance.
Minus the state size increases. The balance / fee_balances alone can be calculated from historical data.

2. a separate GenServer that focuses only on fee and balance, isolation from state balance and fee_balance.
watcher could only count balance
Minus connecting to everything under what state is connected

after a short consultation,
We decided at first, to add new module State.Core.Metrics, 
and in it calculate metric balance without fee_balances in intervals of time.
(Calculation from whole utxo map)

~1m utxos average time of completion
real_utxos: 999_998
unique user:  1024, time: 0.150868 s
balance time: 0.36 s
median: 122,  time: 0.83776 s

### DIFF
--- a/apps/omg/config/config.exs
+++ b/apps/omg/config/config.exs
@@ -6,7 +6,7 @@ config :omg,
   deposit_finality_margin: 10,
   ethereum_events_check_interval_ms: 500,
   coordinator_eth_height_check_interval_ms: 6_000,
-  metrics_interval: 60_000,
+  metrics_collection_interval: 60_000,
   client_monitor_interval_ms: 500
 
 config :omg, :eip_712_domain,
--- a/apps/omg/config/config.exs
+++ b/apps/omg/config/config.exs
@@ -6,6 +6,7 @@ config :omg,
   deposit_finality_margin: 10,
   ethereum_events_check_interval_ms: 500,
   coordinator_eth_height_check_interval_ms: 6_000,
+  metrics_interval: 60_000,
   client_monitor_interval_ms: 500
 
 config :omg, :eip_712_domain,
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -90,7 +90,7 @@ defmodule OMG.State do
     # Get utxos() is essential for the State and Blockgetter. And it takes a while. TODO - measure it!
     # Our approach is simply blocking the supervision boot tree
     # until we've processed history.
-    {:ok, _} = :timer.send_interval(Application.fetch_env!(:omg, :metrics_interval), self(), :send_metrics)
+    {:ok, _} = :timer.send_interval(Application.fetch_env!(:omg, :metrics_collection_interval), self(), :send_metrics)
     {:ok, DB.utxos(), {:continue, :setup}}
   end
 
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -90,6 +90,7 @@ defmodule OMG.State do
     # Get utxos() is essential for the State and Blockgetter. And it takes a while. TODO - measure it!
     # Our approach is simply blocking the supervision boot tree
     # until we've processed history.
+    {:ok, _} = :timer.send_interval(Application.fetch_env!(:omg, :metrics_interval), self(), :send_metrics)
     {:ok, DB.utxos(), {:continue, :setup}}
   end
 
@@ -119,6 +120,13 @@ defmodule OMG.State do
       end
 
     {:ok, _} = Recorder.start_link(%Recorder{name: __MODULE__.Recorder, parent: self()})
+
+    {:noreply, state}
+  end
+
+  def handle_info(:send_metrics, state) do
+    _ = Core.Metrics.calculate(state)
+    |> Enum.map(fn {key, value} -> Appsignal.set_gauge(key, value) end)
 
     {:noreply, state}
   end
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -125,8 +125,9 @@ defmodule OMG.State do
   end
 
   def handle_info(:send_metrics, state) do
-    _ = Core.Metrics.calculate(state)
-    |> Enum.map(fn {key, value} -> Appsignal.set_gauge(key, value) end)
+    _ =
+      Core.Metrics.calculate(state)
+      |> Enum.map(fn {key, value} -> Appsignal.set_gauge(key, value) end)
 
     {:noreply, state}
   end
--- a/apps/omg/lib/omg/state/core/metrics.ex
+++ b/apps/omg/lib/omg/state/core/metrics.ex
@@ -1,0 +1,43 @@
+# Copyright 2019 OmiseGO Pte Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.State.Core.Metrics do
+  @moduledoc """
+  Counting business metrics sent to appsignal
+  """
+
+  alias OMG.Eth.Encoding
+  alias OMG.State.Core
+
+  def calculate(%Core{utxos: utxos}) do
+    [
+      {"unique_users", unique_users(utxos)}
+      | balance(utxos)
+        |> Enum.map(fn {currency, amount} -> {"balance_" <> Encoding.to_hex(currency), amount} end)
+    ]
+  end
+
+  defp unique_users(utxos) do
+    utxos
+    |> Enum.map(fn {_, %OMG.Utxo{owner: owner}} -> owner end)
+    |> Enum.uniq()
+    |> Enum.count()
+  end
+
+  defp balance(utxos) do
+    Enum.reduce(utxos, %{}, fn {_, %{currency: currency, amount: amount}}, acc ->
+      Map.update(acc, currency, amount, &(&1 + amount))
+    end)
+  end
+end
--- a/apps/omg/lib/omg/state/core/metrics.ex
+++ b/apps/omg/lib/omg/state/core/metrics.ex
@@ -23,8 +23,7 @@ defmodule OMG.State.Core.Metrics do
   def calculate(%Core{utxos: utxos}) do
     [
       {"unique_users", unique_users(utxos)}
-      | balance(utxos)
-        |> Enum.map(fn {currency, amount} -> {"balance_" <> Encoding.to_hex(currency), amount} end)
+      | Enum.map(balance(utxos), fn {currency, amount} -> {"balance_" <> Encoding.to_hex(currency), amount} end)
     ]
   end
 
--- a/apps/omg/test/omg/state/core/metrics_test.exs
+++ b/apps/omg/test/omg/state/core/metrics_test.exs
@@ -1,0 +1,47 @@
+# Copyright 2019 OmiseGO Pte Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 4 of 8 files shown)


apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -46,6 +46,7 @@ defmodule OMG.Watcher.ExitProcessor do
   Returns `db_updates`
   """
   def new_exits(exits) do
+    Appsignal.increment_counter("exit.started", length(exits))
```
pdobacz: I think all of these (and more) can be handled by a single line in `EthereumEventListener` - you count by the listener name (`service name`) and increment by `length(events)`, when they are consumed. Seems like the effect would be even richer then (you're missing out deposits)

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
```diff
@@ -48,8 +48,14 @@ defmodule OMG.Watcher.Web.Controller.Transaction do
   """
   def submit(conn, params) do
     with {:ok, tx} <- expect(params, "transaction", :hex) do
+      Appsignal.increment_counter("transaction.succeed", 1)
```
pdobacz: this doesn't bring us much info (this is the watcher and the endpoint is just a bypass to the child chain), but we can keep it.

Note however, that for the child chain server's counterpart, we'd probably want something more elaborate (`transaction.succeed` and `transaction.failed.<the_exact_reason_of_failure>` maybe? or more generally `transaction.failed_statelessly` vs `transaction.failed_statefully` would be an interesting diagnostic, that is, on which stage did the transaction get rejected).

Having that in mind, it could be interesting to see some tps measure of transactions successfully processed in the Watcher (e.g. when it's syncing etc). So I guess this measurement in `controllers/transaction.ex` doesn't really measure the transaction count in a very interesting way

pnowosie:
@pthomalla - what's the time unit in your sample output ☝️ in :pr: description ❓ 
(please update there)
```unique user: 1024, time: 0.150868
balance time: 0.36
median: 122, time: 0.83776```

**Comment by pnowosie** [comment]
Regarding commit messages I'd propose to stick more to the convention, so after the category (fix|style|feat...) start sentence from a verb in present time imperative form, e.g. `add something`, `fix a bug`...

File: apps/omg/test/omg/state/core/metrics_test.exs [hunk 3]
```diff
@@ -0,0 +1,47 @@
...
+
+  require Utxo
+
+  @eth OMG.Eth.RootChain.eth_pseudo_address()
+  @not_eth <<1::size(160)>>
+  @tag fixtures: [:alice, :bob, :carol]
+
+  test "calculate metrics from state", %{alice: alice, bob: bob, carol: carol} do
+    utxos = %{
+      Utxo.position(2_835, 4076, 3) => %OMG.Utxo{amount: 700_000_000, currency: @eth, owner: alice},
```
pnowosie: Can you pls make these utxo positions looks more real - someone could get wrong assumptions about block numbers and deposits. So non-deposit position blknum is divisible by 1_000 and deposit ones has 0, 0 in tx & o-index

apps/omg/lib/omg/state.ex
```diff
@@ -117,6 +118,14 @@ defmodule OMG.State do
     {:noreply, state}
   end
 
+  def handle_info(:send_metrics, state) do
```
pnowosie: Do we have a way not to block `State` to collect this❓ 
Is it a good practice to get a current state and do computation asynchronously 
(CC: @paulperegud , @InoMurko )

apps/omg/config/config.exs
```diff
@@ -6,6 +6,7 @@ config :omg,
   deposit_finality_margin: 10,
   ethereum_events_check_interval_ms: 500,
   coordinator_eth_height_check_interval_ms: 6_000,
+  metrics_interval: 60_000,
```
pnowosie:
```suggestion
  metrics_collection_interval: 60_000,
```

apps/omg/lib/omg/state/core/metrics.ex
```diff
@@ -0,0 +1,43 @@
...
+  """
+
+  alias OMG.Eth.Encoding
+  alias OMG.State.Core
+
+  def calculate(%Core{utxos: utxos}) do
+    [
+      {"unique_users", unique_users(utxos)}
+      | balance(utxos)
+        |> Enum.map(fn {currency, amount} -> {"balance_" <> Encoding.to_hex(currency), amount} end)
```
InoMurko: why force piping? in this case, without it is much more readable. 
`Enum.map(balance(utxos), fn {currency, amount} -> {"balance_" <> Encoding.to_hex(currency), amount} end)`

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
```diff
@@ -47,10 +47,27 @@ defmodule OMG.Watcher.Web.Controller.Transaction do
   Submits transaction to child chain
   """
   def submit(conn, params) do
-    with {:ok, tx} <- expect(params, "transaction", :hex) do
-      API.Transaction.submit(tx)
-      |> api_response(conn, :submission)
+    elem =
+      with {:ok, tx} <- expect(params, "transaction", :hex) do
```
InoMurko: why use a with in this case? you're calling one function and not a series of steps
```
result = case expect(params, "transaction", :hex) do
{:ok, tx}->api_response(API.Transaction.submit(tx)conn, :submission)
error -> error
end
case result do
whatever....
end
```

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
```diff
@@ -47,10 +47,27 @@ defmodule OMG.Watcher.Web.Controller.Transaction do
...
+      {:error, {:validation_error, _, _}} ->
+        Appsignal.increment_counter("transaction.failed.validation", 1)
+
+      {:error, {error, _}} when is_atom(error) ->
+        Appsignal.increment_counter("transaction.fail." <> Atom.to_string(error), 1)
+
+      {:error, _} ->
+        Appsignal.increment_counter("transaction.fail.unidentified", 1)
+
+      _ ->
```
InoMurko: what is this supposed to handle?

apps/omg/lib/omg/state.ex
InoMurko: Depends how busy the process is.

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
InoMurko: question still stands, what is `_`? are you sure you're not catching `everything else`?

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
```diff
@@ -63,4 +64,22 @@ defmodule OMG.Watcher.Web.Controller.Transaction do
...
     end
   end
+
+  defp increment_metrics_counter(response) do
+    case response do
+      {:error, {:validation_error, _, _}} ->
+        Appsignal.increment_counter("transaction.failed.validation", 1)
+
+      {:error, {error, _}} when is_atom(error) ->
+        Appsignal.increment_counter("transaction.fail." <> Atom.to_string(error), 1)
```
InoMurko: are you sure this works?
also from Appsignal docs:
```
We recommend naming your metrics something easily recognizable and without too many dynamic elements. While you can wildcard parts of the metric name for dashboard creation, we recommend you only use this for small grouping and not use IDs for metric names.
```

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
```diff
@@ -51,6 +51,7 @@ defmodule OMG.Watcher.Web.Controller.Transaction do
       API.Transaction.submit(tx)
       |> api_response(conn, :submission)
     end
+    |> increment_metrics_counter()
```
InoMurko: why abuse with statements and pipes so much? there are better ways to write this

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
pthomalla: I select here `with` because it is shorter

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
pthomalla: everything else is success

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
InoMurko: just copying the elixir handbook... `with clause is used to combine matching clauses`

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
InoMurko: I hope you're correct ;)

InoMurko: my advice would be that you always prefer explicit over implicit. makes it more readable and reliable. especially since there's no dialyzer specs.

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
pthomalla: there are three error `:client_error`, `:server_error`,`:malformed_response` <-- answer for Appsignal it's small group for error.
Question is if we don't want this in a different way (is there `one_of_atoms` like funkcion?)

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
pthomalla: for me, shorter is better, and `elixir likes pipes`

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
```diff
@@ -63,4 +64,22 @@ defmodule OMG.Watcher.Web.Controller.Transaction do
       |> api_response(conn, :create)
     end
   end
+
+  defp increment_metrics_counter(response) do
+    case response do
+      {:error, {:validation_error, _, _}} ->
+        Appsignal.increment_counter("transaction.failed.validation", 1)
+
+      {:error, {error, _}} when is_atom(error) ->
```
pthomalla: @InoMurko What do you think about:
```suggestion
      {:error, {error, _}} when error == :client_error or error == :server_error or error == :malformed_response ->
```
or 
```suggestion
      {:error, {error, _}} when error == :client_error 
                           when error == :server_error 
                           when  error == :malformed_response ->
```
or maybe something else

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
pthomalla: in order to calm you down, I suggest this solution:
```elixir
  defp increment_metrics_counter(response) do
    case response do
      {:error, {:validation_error, _, _}} ->
        Appsignal.increment_counter("transaction.failed.validation", 1)

      {:error, {error, _}} when is_atom(error) ->
        Appsignal.increment_counter("transaction.fail." <> Atom.to_string(error), 1)

      %Plug.Conn{} ->
        Appsignal.increment_counter("transaction.succeed", 1)

      _ ->
        Appsignal.increment_counter("transaction.fail.unidentified", 1)
    end

    response
  end
```

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
pdobacz: I'll repeat a comment I've made in a discussion we've had with Paweł - do we even want this counter? Mind this is counting whatever a bypass endpoint to a child chain is doing, so close to nothing.

pthomalla: I will add these errors to 'transaction.fail.unidentified'

apps/omg_watcher/lib/omg_watcher/web/controllers/transaction.ex
InoMurko: that's alright, I like pipes as well! but only when they're needed. 
https://github.com/lexmag/elixir-style-guide#needless-pipeline

https://github.com/christopheradams/elixir_style_guide#avoid-single-pipelines

InoMurko:
In terms of latest development where we're planning to move to Datadog... do we still need this PR and what do we do with it?

pdobacz:
I'd suggest getting fully ramped up on Datadog and then either reusing or scrapping this. I think there's some generic code in here with potential to reuse.

pdobacz:
I'd suggest getting fully ramped up on Datadog and then either reusing or scrapping this. I think there's some generic code in here with potential to reuse.

InoMurko:
Closed because of Datadog work. Keeping the branch though.

PR #641, opened 2019-04-25
fix: make ExitProcessor tolerate spending blocks not being found

pdobacz:
Fixes the situation which looks like the following:

```
2019-04-25 11:45:53.709 [error] module=gen_server function=error_info/7 ⋅GenServer OMG.Watcher.ExitProcessor terminating
** (FunctionClauseError) no function clause matching in OMG.Watcher.ExitProcessor.Core.get_known_txs/1
    (omg_watcher) lib/exit_processor/core.ex:738: OMG.Watcher.ExitProcessor.Core.get_known_txs([:not_found])
    (omg_watcher) lib/exit_processor/core.ex:523: OMG.Watcher.ExitProcessor.Core.get_ifes_with_competitors/2
    (omg_watcher) lib/exit_processor/core.ex:487: OMG.Watcher.ExitProcessor.Core.invalid_exits/2
    (omg_watcher) lib/exit_processor.ex:257: OMG.Watcher.ExitProcessor.handle_call/3
    (stdlib) gen_server.erl:661: :gen_server.try_handle_call/4
    (stdlib) gen_server.erl:690: :gen_server.handle_msg/6
    (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Last message (from OMG.Watcher.BlockGetter): :check_validity⋅
```

or like this:

```
April 24th 2019, 15:09:05.2482019-04-24 13:09:05.248 [error] module=gen_server function=error_info/7 ⋅GenServer OMG.Watcher.ExitProcessor terminating
April 24th 2019, 15:09:05.248    :not_found.number()
April 24th 2019, 15:09:05.248    (elixir) lib/enum.ex:1327: Enum."-map/2-lists^map/1-0-"/2
April 24th 2019, 15:09:05.248    (elixir) lib/enum.ex:2337: Enum.sort_by/3
April 24th 2019, 15:09:05.248    (elixir) lib/enum.ex:2337: anonymous fn/2 in Enum.sort_by/3
April 24th 2019, 15:09:05.248    (omg_watcher) lib/exit_processor/core.ex:745: OMG.Watcher.ExitProcessor.Core.get_known_txs/1
April 24th 2019, 15:09:05.248    (omg_watcher) lib/exit_processor/core.ex:511: OMG.Watcher.ExitProcessor.Core.get_ifes_with_competitors/2
[...]

PR #642, opened 2019-04-25
fix: make ExitProcessor tolerate spending blocks not being found

pdobacz:
ported from #641 see there for description.

Here the implementation is slightly different, we're doing the filtering uniformly on all flavors of the requests


apps/omg_watcher/lib/omg_watcher/exit_processor/core.ex
```diff
@@ -560,6 +564,24 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
+  @spec handle_spent_blknum_result(list(spent_blknum_result_t()), list(Utxo.Position.t())) :: list(pos_integer())
+  def handle_spent_blknum_result(spent_blknum_result, spent_positions_to_get) do
+    {not_founds, founds} =
+      Stream.zip(spent_positions_to_get, spent_blknum_result)
+      |> Enum.split_with(fn {_utxo_pos, result} -> result == :not_found end)
+
+    {_, blknums_to_get} = Enum.unzip(founds)
+
+    warn? = !Enum.empty?(not_founds)
+    _ = if warn?, do: Logger.warn("UTXO doesn't exists but no spend registered (spent in exit?) #{inspect(not_founds)}")
```
pnowosie: I propose to inline this redundant variable and use just ‚Enum.any(not_founds)’

pdobacz: ah this line stirs up emotions.

I did the `warn?` to avoid a 5-line-spanning, unreadable split `mix format` proposes if I inline.

OTOH `Enum.any?([false]) != !Enum.empty?([false])` so I'm not sure if this is a viable shortcut here

PR #643, opened 2019-04-26
Fix wrong filename for 0.1 operator api spec

mederic-p:
The current link for the operator api spec is wrong: https://developer.omisego.co/elixir-omg/docs-ui/?url=0.1/operator_api_swagger.yaml

This PR updates it to point to: https://developer.omisego.co/elixir-omg/docs-ui/?url=0.1/operator_api_specs.yaml


pdobacz:
fixes #676

PR #644, opened 2019-04-26
DevHelpers should use passprase from config

pnowosie:


InoMurko:
Where does this requirement come from?

pdobacz:
Hmmm, this doesn't look complete, since we only advise (and require) `SIGNER_PASSPHRASE` to be present when using parity, if I'm not mistaken.

I'm not sure how the testnet passphrases were intended to be handled, so I'd wait around for Pepesza to tell how this is supposed to work. In general, this should work fine for `geth` and `parity` and then for signing done by the "faucet", authority_addr and the test-accounts used by integration tests (which are also imported into the nodes). So this might overall be not that simple.

InoMurko:
Also, to comment on @pdobacz, if this is meant for self provisioned networks AND for testnet and mainnet this ought to be included in release generation (look for release tasks). 
Byt would prefer to have a ticket and acceptance criteria defined

pnowosie:
Due to some concerns we can go without this :pr:

PR #645, opened 2019-04-26
Deploying with faucet addres instead of first

pnowosie:
Issue raised by @jbunce - cannot deploy contract because of insufficient funds.
After this changes it tries first to get faucet address as deployer and fallbacks to first address if not provided.

### DIFF
--- a/apps/omg_eth/test/support/dev_helpers.ex
+++ b/apps/omg_eth/test/support/dev_helpers.ex
@@ -47,8 +47,8 @@ defmodule OMG.Eth.DevHelpers do
 
     with {:ok, _} <- Application.ensure_all_started(:ethereumex),
          {:ok, authority} <- create_and_fund_authority_addr(opts),
-         {:ok, depoyer_addr} <- get_deployer_address(opts),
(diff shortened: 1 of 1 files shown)


apps/omg_eth/test/support/dev_helpers.ex
```diff
@@ -47,8 +47,8 @@ defmodule OMG.Eth.DevHelpers do
 
     with {:ok, _} <- Application.ensure_all_started(:ethereumex),
          {:ok, authority} <- create_and_fund_authority_addr(opts),
-         {:ok, [addr | _]} <- Ethereumex.HttpClient.eth_accounts(),
-         {:ok, txhash, contract_addr} <- Eth.Deployer.create_new(OMG.Eth.RootChain, root_path, from_hex(addr)),
+         {:ok, depoyer_addr} <- get_deployer_address(opts),
```
pdobacz: typo

apps/omg_eth/test/support/dev_helpers.ex
```diff
@@ -136,6 +136,11 @@ defmodule OMG.Eth.DevHelpers do
 
   # private
 
+  defp get_deployer_address(opts) do
```
InoMurko: maybe a function comment why this is necessary for future reference ;)

PR #646, opened 2019-04-26
fix: finalize in-flight exits in Watcher

### LINKED ISSUE #630: Watcher allows for spending in-flight exited utxos
In Watcher utxos exited via in-flight exit are available for spend after the exit is finalized. It means that operator can produce transactions that stem from exited utxos and Watcher accepts them as valid.

pgebal:
Partly solves #630. Exited utxo can no longer be spent, but IFEs are not reflected in convenience API. That will be solved in a separate PR.

Closes #630.

### DIFF
--- a/apps/omg/lib/omg/state/core.ex
+++ b/apps/omg/lib/omg/state/core.ex
@@ -56,9 +56,9 @@ defmodule OMG.State.Core do
           | [in_flight_exit()]
 
   @type in_flight_exit() :: %{in_flight_tx: binary()}
-  @type piggyback() :: %{txhash: Transaction.tx_hash(), output_index: non_neg_integer}
+  @type piggyback() :: %{tx_hash: Transaction.tx_hash(), output_index: non_neg_integer}
 
-  @type validities_t() :: {list(Utxo.Position.t()), list(Utxo.Position.t())}
+  @type validities_t() :: {list(Utxo.Position.t()), list(Utxo.Position.t() | piggyback())}
 
   @type utxos() :: %{Utxo.Position.t() => Utxo.t()}
 
@@ -403,18 +403,19 @@ defmodule OMG.State.Core do
   end
 
   def exit_utxos([%{tx_hash: _} | _] = piggybacks, %Core{utxos: utxos} = state) do
-    piggybacks
-    |> Enum.map(fn %{tx_hash: tx_hash, output_index: oindex} ->
-      # oindex in contract is 0-7 where 4-7 are outputs
-      oindex = oindex - 4
+    {piggybacks_of_unknown_utxos, piggybacks_of_known_utxos} =
+      piggybacks
+      |> Enum.map(&find_utxo_matching_piggyback(&1, utxos))
+      |> Enum.split_with(fn {_, position} -> position == nil end)
 
-      utxos
-      |> Map.to_list()
-      |> Enum.find(&match?({Utxo.position(_, _, ^oindex), %Utxo{creating_txhash: ^tx_hash}}, &1))
-    end)
-    |> Enum.filter(&(&1 != nil))
-    |> Enum.map(fn {position, _} -> position end)
-    |> exit_utxos(state)
+    {:ok, {db_updates, {valid, invalid}}, state} =
+      piggybacks_of_known_utxos
+      |> Enum.map(fn {_, {position, _}} -> position end)
+      |> exit_utxos(state)
+
+    {unknown_piggybacks, _} = Enum.unzip(piggybacks_of_unknown_utxos)
+
+    {:ok, {db_updates, {valid, invalid ++ unknown_piggybacks}}, state}
   end
 
   def exit_utxos(exiting_utxos, %Core{utxos: utxos} = state) do
@@ -427,6 +428,15 @@ defmodule OMG.State.Core do
     new_state = %{state | utxos: Map.drop(utxos, valid)}
 
     {:ok, {db_updates, validities}, new_state}
+  end
+
+  defp find_utxo_matching_piggyback(%{tx_hash: tx_hash, output_index: oindex} = piggyback, utxos) do
+    # oindex in contract is 0-7 where 4-7 are outputs
+    oindex = oindex - 4
+
+    position = Enum.find(utxos, &match?({Utxo.position(_, _, ^oindex), %Utxo{creating_txhash: ^tx_hash}}, &1))
+
+    {piggyback, position}
   end
 
   @doc """
--- a/apps/omg/test/omg/state/core_test.exs
+++ b/apps/omg/test/omg/state/core_test.exs
@@ -518,9 +518,11 @@ defmodule OMG.State.CoreTest do
   test "exits utxos given in various forms", %{alice: alice, state_alice_deposit: state} do
     # this test checks whether all ways of calling `exit_utxos/1` work on par
     # this is _very important_ to support all clients of that functions, whose inputs come in different flavors
+    %Transaction.Recovered{tx_hash: tx_hash} = tx = create_recovered([{1, 0, 0, alice}], @eth, [{alice, 7}, {alice, 3}])
+
     state =
       state
-      |> Core.exec(create_recovered([{1, 0, 0, alice}], @eth, [{alice, 7}, {alice, 3}]), :ignore)
+      |> Core.exec(tx, :ignore)
       |> success?
 
     utxo_pos_exits = [Utxo.position(@blknum1, 0, 0), Utxo.position(@blknum1, 0, 1)]
@@ -541,6 +543,9 @@ defmodule OMG.State.CoreTest do
              utxo_pos_exits
              |> Enum.map(&Utxo.Position.encode/1)
              |> Core.exit_utxos(state)
+
+    piggybacks = [%{tx_hash: tx_hash, output_index: 4}, %{tx_hash: tx_hash, output_index: 5}]
+    assert exit_utxos_response_reference == Core.exit_utxos(piggybacks, state)
   end
 
   @tag fixtures: [:alice, :state_alice_deposit]
@@ -622,6 +627,13 @@ defmodule OMG.State.CoreTest do
     utxo_pos_exit_1 = Utxo.position(@blknum1, 0, 0)
 
     assert {:ok, {[], {[], [^utxo_pos_exit_1]}}, ^state} = Core.exit_utxos([utxo_pos_exit_1], state)
+  end
+
+  @tag fixtures: [:state_empty]
+  test "notifies about invalid in-flight exit", %{state_empty: state} do
+    piggyback = %{tx_hash: 1, output_index: 5}
+
+    assert {:ok, {[], {[], [^piggyback]}}, ^state} = Core.exit_utxos([piggyback], state)
   end
 
   @tag fixtures: [:alice, :state_empty]
--- a/apps/omg_watcher/lib/omg_watcher/exit_processor.ex
+++ b/apps/omg_watcher/lib/omg_watcher/exit_processor.ex
@@ -290,8 +290,16 @@ defmodule OMG.Watcher.ExitProcessor do
 
   def handle_call({:finalize_in_flight_exits, finalizations}, _from, state) do
     _ = if not Enum.empty?(finalizations), do: Logger.info("Recognized ife finalizations: #{inspect(finalizations)}")
-    {:ok, state, db_updates} = Core.finalize_in_flight_exits(state, finalizations)
-    {:reply, {:ok, db_updates}, state}
+
+    {:ok, exits} = Core.prepare_utxo_exits_for_in_flight_exit_finalizations(state, finalizations)
+
+    # NOTE: it's not straightforward to track from utxo position returned when exiting utxo in State to ife id
+    # See issue #671 https://github.com/omisego/elixir-omg/issues/671
+    {invalidities, state_db_updates} = Enum.reduce(exits, {%{}, []}, &collect_invalidities_and_state_db_updates/2)
+
+    {:ok, state, db_updates} = Core.finalize_in_flight_exits(state, finalizations, invalidities)
+
+    {:reply, {:ok, state_db_updates ++ db_updates}, state}
   end
 
   def handle_call(:check_validity, _from, state) do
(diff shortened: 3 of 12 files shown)


apps/omg_eth/lib/eth.ex
```diff
@@ -175,7 +175,7 @@ defmodule OMG.Eth do
     end
   end
 
-  defp event_topic_for_signature(signature) do
+  def event_topic_for_signature(signature) do
```
InoMurko: why is this public?

pgebal: It shouldn't be, this code is not for review yet.

pgebal: @InoMurko Fixes are uploaded, feel free to have a look.

apps/omg/lib/omg/state/core.ex
```diff
@@ -405,7 +405,6 @@ defmodule OMG.State.Core do
   def exit_utxos([%{tx_hash: _} | _] = piggybacks, %Core{utxos: utxos} = state) do
     piggybacks
     |> Enum.map(fn %{tx_hash: tx_hash, output_index: oindex} ->
-      # oindex in contract is 0-7 where 4-7 are outputs
```
pdobacz: I think this comment might have been helpful. This subtraction is otherwise rather cryptic

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -273,8 +273,12 @@ defmodule OMG.Watcher.ExitProcessor do
 
   def handle_call({:finalize_in_flight_exits, finalizations}, _from, state) do
     _ = if not Enum.empty?(finalizations), do: Logger.info("Recognized ife finalizations: #{inspect(finalizations)}")
-    {:ok, state, db_updates} = Core.finalize_in_flight_exits(state, finalizations)
-    {:reply, {:ok, db_updates}, state}
+
+    {:ok, state, db_updates, input_exits, output_exits} = Core.finalize_in_flight_exits(state, finalizations)
+    {:ok, input_exits_state_updates, _} = State.exit_utxos(input_exits)
+    {:ok, output_exits_state_updates, _} = State.exit_utxos(output_exits)
```
pdobacz: there still is a large step missing here, which is taking validities into account, to cover the case where an IFE finalizes invalidly. But that's another large leap to take, so perhaps another issue/PR. And it would require to introduce the concept of a late invalid IFE, so that SEs and IFEs finally behave on par.

Maybe as a bare minimum, grab the invalid utxos (second list in the `_`-out tuple in the return), concatenate both flavors (inputs outputs) and if that list is non-empty (some IFE in-out-puts finalized invalidly) print a `Logger.warn` with some message and positions (make sure to inspect).

Leaving a TODO or a NOTE and reference to an issue is also desired

apps/omg_watcher/lib/omg_watcher/exit_processor/core.ex
```diff
@@ -366,7 +366,9 @@ defmodule OMG.Watcher.ExitProcessor.Core do
   end
 
   @spec finalize_in_flight_exits(t(), [map()]) ::
-          {:ok, t(), list()} | {:unknown_piggybacks, list()} | {:unknown_in_flight_exit, MapSet.t(non_neg_integer())}
+          {:ok, t(), list(), list(), list()}
```
pdobacz: let's put a `doc` here explaining what the args and returns are. This just got way more complicated.

**Among other** explanations, make sure to include what the kinds of lists are the two new lists - that they "fit" to whatever `State.Core` likes to receive in `exit_utxos` while handling IFEs.

pdobacz: Next, add tests under `state_interaction_test.exs`, that check that whatever you're feeding `State.Core` makes sense.

You can even add a test, where some IFE finalizes invalidly and check if you're getting the Logger warning for completeness (`ExUnit.CaptureLog`). Later we'd swap this test for test of more proper handling.

This combined should give you decent test coverage of the IFE finalization

apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
```diff
@@ -1326,21 +1326,47 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
   end
 
   describe "in-flight exit finalization" do
-    test "deactivate in-flight exit after all piggybacked outputs are finalized",
-         %{processor_empty: processor, transactions: [tx | _]} do
+    test "deactivate in-flight exit after all piggybacked transaction inputs are finalized",
```
pdobacz: leave this test as it was and add a new test (or tests) that test your new behaviors

pdobacz: another comment related to this. Judging by the test description, you're triggering exiting in `State` only when all piggybacks have finalized? I think this is too late - you should exit in `State` when a particular piggybacked in/output finalizes immediately.

If this is just about the test description, and you're exiting immediately, then the implementation is fine; splitting the test and/or amending the description will do!

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pgebal: done

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -438,4 +436,26 @@ defmodule OMG.Watcher.ExitProcessor do
...
+    {:ok, output_exits_state_updates, {_, output_invalidities}} = State.exit_utxos(output_exits)
+
+    exit_invalidities = input_invalidities ++ output_invalidities
+
+    {state, db_updates} =
+      if Enum.empty?(exit_invalidities) do
+        {state, db_updates}
+      else
+        _ = Logger.warn("Invalid in-flight exit finalization: #{inspect(exit_invalidities)}")
+        Core.activate_in_flight_exit(state, ife_id)
```
pgebal: I'd like to talk about that. Ping me when you read this.

apps/omg/lib/omg/state/core.ex
```diff
@@ -403,18 +403,29 @@ defmodule OMG.State.Core do
...
-    |> exit_utxos(state)
+    {piggybacks_of_unknown_utxos, piggybacks_of_known_utxos} =
+      piggybacks
+      |> Enum.map(fn %{tx_hash: tx_hash, output_index: oindex} = piggyback ->
+        # oindex in contract is 0-7 where 4-7 are outputs
+        oindex = oindex - 4
+
+        position =
+          utxos
+          |> Map.to_list()
```
InoMurko: code like this, with massive pipes... is this really necessary?

apps/omg/lib/omg/state/core.ex
```diff
@@ -403,18 +403,29 @@ defmodule OMG.State.Core do
...
-      utxos
-      |> Map.to_list()
-      |> Enum.find(&match?({Utxo.position(_, _, ^oindex), %Utxo{creating_txhash: ^tx_hash}}, &1))
-    end)
-    |> Enum.filter(&(&1 != nil))
-    |> Enum.map(fn {position, _} -> position end)
-    |> exit_utxos(state)
+    {piggybacks_of_unknown_utxos, piggybacks_of_known_utxos} =
+      piggybacks
+      |> Enum.map(fn %{tx_hash: tx_hash, output_index: oindex} = piggyback ->
```
InoMurko: maybe these big pipes are not the optimal approach to this? what if you pull the implementation out into a private function and document the function separately + specs?

pdobacz: ah, I concur about pulling the implementation out to a `defp`, but I'd leave all 3 pipes in the diff in :)

pgebal: done

apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
```diff
@@ -1325,22 +1325,126 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
+      ife_id2 = 2
+      tx_hash1 = Transaction.raw_txhash(tx1)
+      tx_hash2 = Transaction.raw_txhash(tx2)
+
+      processor =
+        processor
+        |> start_ife_from(tx1, status: {1, ife_id1})
+        |> start_ife_from(tx2, status: {1, ife_id2})
+
+      {processor, _} = Core.new_piggybacks(processor, [%{tx_hash: tx_hash1, output_index: 0}])
```
pdobacz: with the increased amounts of calls to `Core.new_piggybacks`, I think it's time to do a refactoring due:

Follow the same approach as `start_ife_from(tx, status: {1, ife_id})` -> `piggyback_ife_from(tx, output_index)` so that we can append the `|> piggyback_ife_from(...)` to the pipes like above.

Direct calls to `Core.new_piggybacks` would stay on the few occasions where it is more convenient/ilustrative or where we want to do sth with the db updates

apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
```diff
@@ -1325,22 +1325,126 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
+
+      {:unknown_piggybacks, ^expected_unknown_piggybacks} =
+        Core.prepare_utxo_exits_for_in_flight_exit_finalizations(processor, [finalization1, finalization2])
+    end
+  end
+
   describe "in-flight exit finalization" do
-    test "deactivate in-flight exit after all piggybacked outputs are finalized",
+    test "exits piggybacked transaction inputs",
+         %{
```
pdobacz: single line the `%{...} do`

apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
```diff
@@ -1356,8 +1460,8 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
         |> Core.new_piggybacks([%{tx_hash: tx_hash, output_index: 1}])
 
       finalization = %{in_flight_exit_id: ife_id, output_index: 1}
-      {:ok, processor, _} = Core.finalize_in_flight_exits(processor, [finalization])
-      {:ok, ^processor, []} = Core.finalize_in_flight_exits(processor, [finalization])
+      {:ok, processor, _} = Core.finalize_in_flight_exits(processor, [finalization], %{})
+      {:ok, ^processor, _} = Core.finalize_in_flight_exits(processor, [finalization], %{})
```
pdobacz: here I think the assertion that no more stuff is written do db was relevant. Pls revert the `[]`

apps/omg_watcher/test/omg_watcher/exit_processor/persistence_test.exs
```diff
@@ -253,7 +253,7 @@ defmodule OMG.Watcher.ExitProcessor.PersistenceTest do
   end
 
   defp persist_finalize_ifes(processor, finalizations, db_pid) do
-    {:ok, processor, db_updates} = Core.finalize_in_flight_exits(processor, finalizations)
+    {:ok, processor, db_updates} = Core.finalize_in_flight_exits(processor, finalizations, [])
```
pdobacz: usually the last arg is a map

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -434,4 +442,22 @@ defmodule OMG.Watcher.ExitProcessor do
     {:ok, spend_blknum} = position |> Utxo.Position.to_db_key() |> OMG.DB.spent_blknum()
     spend_blknum
   end
+
+  defp collect_invalidities_and_state_db_updates(
+         {ife_id, {input_exits, output_exits}},
+         {invalidities_by_ife_id, state_db_updates}
+       ) do
+    {:ok, input_exits_state_updates, {_, input_invalidities}} = State.exit_utxos(input_exits)
```
pdobacz: leave also a comment here
```
# we can't call `State.exit_utxos(input_exits ++ output_exits)` because the types of these enumerable items are distinct
```

apps/omg_watcher/test/omg_watcher/exit_processor/core/state_interaction_test.exs
```diff
@@ -182,6 +182,74 @@ defmodule OMG.Watcher.ExitProcessor.Core.StateInteractionTest do
     assert Utxo.position(1, 0, 0) in spends_to_get
   end
 
+  test "can work with State to exit utxos from in-flight transactions", %{
+    processor_empty: processor,
```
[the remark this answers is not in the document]
pgebal: Done

apps/omg/test/omg/state/core_test.exs
```diff
@@ -624,6 +624,13 @@ defmodule OMG.State.CoreTest do
     assert {:ok, {[], {[], [^utxo_pos_exit_1]}}, ^state} = Core.exit_utxos([utxo_pos_exit_1], state)
   end
 
+  @tag fixtures: [:state_empty]
```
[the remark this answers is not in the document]
pgebal: Ad 1. That's tested in the test above this one.
Ad 2. Done.

apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
pgebal: Done

apps/omg_watcher/test/omg_watcher/exit_processor/persistence_test.exs
pgebal: Done

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pgebal: Done

apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
pgebal: This was done on purpose. Those db_updates returned were in fact current state of ifes after finalizations are applied to the core. Even if nothing changes, db_updates will be produced. There should be a check that those updates are the same as in the line above.

After giving it a second thought, it feels more nautral to expect (as you expected) that those updates are "real" updates and are only produced when ifes are in fact changing. I refactored that.

apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
pgebal: Done. I left calls that use multiple piggybacks in argument list and those that test Core.new_piggybacks/2

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -273,8 +273,16 @@ defmodule OMG.Watcher.ExitProcessor do
...
+
+    {:ok, exits} = Core.prepare_utxo_exits_for_in_flight_exit_finalizations(state, finalizations)
+
+    # NOTE: it's not straightforward to track from utxo position returned when exiting utxo in State to ife id
+    # See issue #671 https://github.com/omisego/elixir-omg/issues/671
+    {invalidities, state_db_updates} = Enum.reduce(exits, {%{}, []}, &collect_invalidities_and_state_db_updates/2)
+
+    {:ok, state, db_updates} = Core.finalize_in_flight_exits(state, finalizations, invalidities)
+
+    {:reply, {:ok, state_db_updates ++ db_updates}, state}
```
InoMurko: maybe you can prepend these lists?

PR #647, opened 2019-04-27
feature: parallel coveralls

InoMurko:
I've seen that integration tests can take around 20 minutes now so I thought it might be a good time to try and reduce the time.
I've split the integration + wrapper tags into child chain, watcher and common (everything else) and we run them in parallel, gather the results and send off to coveralls. It reduces the runs significantly. Also, we now compile the source only once, cache the _build folder just before the test run.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,7 +123,7 @@ jobs:
             if [[ -z "$COVERALLS_REPO_TOKEN" ]]; then
               mix coveralls.html --parallel --umbrella --include child_chain --exclude watcher --exclude common --exclude test --trace
             else
-              mix coveralls.circle --parallel --umbrella --include child_chain --exclude watcher --exclude common --trace ||
+              mix coveralls.circle --parallel --umbrella --include child_chain --exclude watcher --exclude common --exclude test --trace ||
                 # if mix failed, then coveralls_merge won't run, so signal done here and return original exit status
                 (retval=$? && curl -k https://coveralls.io/webhook?repo_token=$COVERALLS_REPO_TOKEN -d "payload[build_num]=$CIRCLE_WORKFLOW_WORKSPACE_ID&payload[status]=done" && exit $retval)
             fi
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -139,9 +139,6 @@ jobs:
           keys:
             - v1-mix-cache-test-compile-{{ checksum "mix.lock" }}
       - run:
-          name: Compile
-          command: mix compile
-      - run:
           name: Integration Tests & Coveralls Part Child Chain
           command: |
             # Don't submit coverage report for forks, but let the build succeed
@@ -178,9 +175,6 @@ jobs:
       - restore_cache:
           keys:
             - v1-mix-cache-test-compile-{{ checksum "mix.lock" }}
-      - run:
-          name: Compile
-          command: mix compile
       - run:
           name: Integration Tests & Coveralls Part Watcher
           command: |
@@ -219,9 +213,6 @@ jobs:
           keys:
             - v1-mix-cache-test-compile-{{ checksum "mix.lock" }}
       - run:
-          name: Compile
-          command: mix compile
-      - run:
           name: Integration Tests & Coveralls Part Common
           command: |
             # Don't submit coverage report for forks, but let the build succeed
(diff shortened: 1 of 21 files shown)


jbunce:
LGTM. @pdobacz any comments?

apps/omg_child_chain/test/omg_child_chain/integration/happy_path_test.exs
```diff
@@ -31,13 +31,14 @@ defmodule OMG.ChildChain.Integration.HappyPathTest do
...
   @moduletag :integration
+  @moduletag :child_chain
   # bumping the timeout to two minutes for the tests here, as they do a lot of transactions to Ethereum to test
   @moduletag timeout: 120_000
 
   @eth OMG.Eth.RootChain.eth_pseudo_address()
   @interval OMG.Eth.RootChain.get_child_block_interval() |> elem(1)
 
-  @tag fixtures: [:alice, :bob, :omg_child_chain, :token, :alice_deposits]
+  @tag fixtures: [:alice, :bob, :omg_child_chain, :token, :alice_deposits, :eth_node]
```
pdobacz: are you sure this is required and isn't a leftover of some kind? `:eth_node` should be already stood up when `:omg_child_chain` is. Let's maybe try removing the explicit fixture once more.

InoMurko:
A further addition would be to split watcher_* tests in two (they run the longest currently).

.circleci/config.yml
```diff
@@ -118,36 +189,37 @@ jobs:
...
-            mix coveralls.post \
-              --umbrella \
-              --trace \
-              --include integration \
-              --include wrappers \
-              --sha $CIRCLE_SHA \
-              --branch $CIRCLE_BRANCH \
-              --message "$GIT_COMMIT_DESCRIPTION"
+            if [[ -z "$COVERALLS_REPO_TOKEN" ]]; then
+              mix coveralls.html --parallel --umbrella --include common --exclude watcher --exclude child_chain --exclude test --trace
```
pdobacz: if someone adds a tests and tags it `:integration`, but forgets to tag it `:watcher` or `:common` etc, it would get excluded, correct? Same with `:wrappers`. 

how about we have a fourth run of `mix coveralls`, which does `--exclude common/watcher/child_chain` (all three) but does `--include integration --include wrappers`, to scoop all those tests left without tags? Or maybe we remove `:common` altogether, and have that encompass all of the tests not marked explicitly as `watcher` or `child_chain` with `--exclude watcher --exclude child_chain --include integration --include wrappers`

apps/omg_child_chain/test/omg_child_chain/integration/happy_path_test.exs
InoMurko: I've looked at the :omg_child_chain fixture and I can't find anything that would start :eth_node... Where are you seeing it?

I've added :eth_node as one of the last commits, there were runs that failed before. For example: https://circleci.com/gh/omisego/elixir-omg/5416

.circleci/config.yml
InoMurko: Your proposal would run all the tests again, no?
If you're worried about forgetting (I wouldn't be because we're doing PR's for that) maybe the best approach is to remove integration and wrapper tags altogether. That would force the developer to choose something.

apps/omg_child_chain/test/omg_child_chain/integration/happy_path_test.exs
pdobacz: it is started by it's dependency fixtures `:omg_child_chain` -> `:root_chain_config` -> `:contract` IIRC.

The error you link to... :thinking: hmm strange, why would this happen? Anyway, this looks as unavailability of the child chain server rather than of the eth node. And apparently it does fail for the second tx sent.

InoMurko: In my opinion it’s because of pub/sub alarms that put the rpc in 503 mode. 
EthereumClientMonitor eventually raises and alarm that the client isn’t reachable and that puts the API offline.

InoMurko: If we look at the `warn` logs in the failing test above we can see that:
```
  * test deposit, spend, restart, exit etc works fine2019-04-28 21:46:15.063 [warn] module=OMG.EthereumClientMonitor function=handle_event/2 ⋅Health check raised :ethereum_client_connection alarm.⋅
2019-04-28 21:46:15.063 [warn] module=OMG.RPC.Plugs.Health function=handle_event/2 ⋅Alarm ethereum_client_connection was raised.⋅
```

.circleci/config.yml
pdobacz: ah, so `include` takes precedence, right.

Then out of the two options I think current approach is better and let's not forget to tag ;).

Maybe we'll think of some fool-proof way of tagging later.

apps/omg_child_chain/test/omg_child_chain/integration/happy_path_test.exs
pdobacz: but why would such explicit inclusion of the fixture affect this situation? or why would this become an issue in this particular PR? This doesn't fit together - the eth node should be up and running at this point of test, until the teardown code is executed. 

The contracts got deployed fine and deposits were made, so why would the ethereum node disappear suddenly, causing the alarm to fire?

InoMurko: Looking into this further - my current take is that what happens is that there is a race condition between stopping the applications and EthereumClientMonitor.

InoMurko:
@jbunce do I need to port this to v0.1?

jbunce:
@InoMurko Yes please, and https://github.com/omisego/elixir-omg/pull/652 when landed

kasima:
Nice, @InoMurko! Changed the required checks.

PR #650, opened 2019-04-29
fix: Support Gorli contracts & add support for the Watcher

jbunce:
- Fixes support for the Childchain
- Adds support for the Watcher
- PEP8 compliant

### DIFF
--- a/launcher.py
+++ b/launcher.py
@@ -514,7 +514,7 @@ def get_environment_variables() -> dict:
             'RINKEBY_AUTHORITY_ADDRESS',
             '0xe5153ad259be60003909492b154bf4b7f1787f70'
         )
-    elif os.getenv('ETHEREUM_NETWORK') == 'GORLI':
+    elif os.getenv('ETHEREUM_NETWORK') == 'GÖRLI':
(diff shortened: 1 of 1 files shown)


launcher.py
```diff
@@ -496,18 +501,32 @@ def get_environment_variables() -> dict:
...
+            '0x84a86f06b97e4c2d694ba507e7fcd8cf78adc4fbd596b1d3626ec7ba8242450d' # noqa E501
+        )
+        RINKEBY_CONTRACT['authority_addr'] = os.environ.get(
+            'RINKEBY_AUTHORITY_ADDRESS',
+            '0xe5153ad259be60003909492b154bf4b7f1787f70'
+        )
+    elif os.getenv('ETHEREUM_NETWORK') == 'GÖRLI':
+        GÖRLI_CONTRACT['contract_addr'] = os.environ.get(
+            'GÖRLI_CONTRACT_ADDRESS',
+            '0x607ba3407d9aab7dec4dfe67993060b9949ad6e1'
```
kasima: What are these default addresses?

jbunce: They exist for if the ENV isn't passed.

PR #651, opened 2019-04-29
fix: GÖRLI -> GORLI

jbunce:
Kubernetes only accepts pod names in characters from the English alphabet:

```
Invalid value: "GÖRLI_CONTRACT_ADDRESS": a valid environment variable name must consist of alphabetic characters, digits, '_', '-', or '.', and must not start with a digit (e.g. 'my.env-name',  or 'MY_ENV.NAME',  or 'MyEnvName1', regex used for validation is '[-._a-zA-Z][-._a-zA-Z0-9]*')
```

### DIFF
--- a/launcher.py
+++ b/launcher.py
@@ -11,7 +11,7 @@ import requests
 
 
 RINKEBY_CONTRACT = {}
-GÖRLI_CONTRACT = {}
+GORLI_CONTRACT = {}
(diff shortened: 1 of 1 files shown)


InoMurko:
@kasima waiting for your UTF8 comment. 😀🤓

PR #652, opened 2019-04-29
feat: Add CD for services connected to Parity

jbunce:
As title


.circleci/config.yml
```diff
@@ -269,6 +269,9 @@ jobs:
             kubectl set image statefulset childchain childchain=$DOCKER_IMAGE
             while true; do if [ "$(kubectl get pods childchain-0 -o jsonpath=\"{.status.phase}\" | grep Running)" ]; then break; fi; done
             kubectl set image statefulset watcher watcher=$DOCKER_IMAGE
+            kubectl set image statefulset childchain-parity childchain=$DOCKER_IMAGE
```
InoMurko: I presume geth is the default state? Should we make that explicit?

jbunce: Not until we've decided that Parity is going to be the de facto Ethereum client :). At the moment Parity is experimental.

Issue #628, opened 2019-04-17
Add support for LevelDB replacement


InoMurko:
Abstracting the DB layer so that we're able to plugin alternatives (ETS, RocksDB)

InoMurko:
Done.

PR #653, opened 2019-04-29
628 rocksdb support

InoMurko:
Issue: https://github.com/omisego/elixir-omg/issues/628
This (draft) PR is trying to get rocksdb and keeping leveldb support so that we're able to test both and migrate when we feel comfortable.
RocksDB supports multithreading so having a pool of connections/workers would be beneficial.
Needs more testing, especially for prefix search (also test for Leveldb and maybe do a perf comparison!)

Currently, rocksdb is pointing int to my gitlab fork because of some parallelism compilation flags that are causing issues. When this gets merged or solved we can point to  a release: https://gitlab.com/barrel-db/erlang-rocksdb/merge_requests/112


apps/omg_db/mix.exs
```diff
@@ -32,6 +32,8 @@ defmodule OMG.DB.MixProject do
     [
       {:appsignal, "~> 1.0"},
       # version caused by dependency in merkle_patricia_tree from blockchain
+      {:rocksdb,
```
InoMurko: will get swapped out once https://gitlab.com/barrel-db/erlang-rocksdb/merge_requests/112#note_167748150 gets merged!

apps/omg_db/config/config.exs
```diff
@@ -4,10 +4,13 @@ use Mix.Config
...
-  type: :leveldb,
+  type: :rocksdb,
   # leveldb
   leveldb_path: Path.join([System.get_env("HOME"), ".omg/data"]),
   # leveldb
   server_module: OMG.DB.LevelDB.Server,
   # leveldb
-  server_name: OMG.DB.LevelDB.Server
+  server_name: OMG.DB.LevelDB.Server,
+  r_server_module: OMG.DB.RocksDB.Server,
```
InoMurko: I did not want to consolidate the settings now because these will change with the use of poolboy.

apps/omg_db/lib/omg_db/rocksdb/core.ex
```diff
@@ -0,0 +1,132 @@
...
+defmodule OMG.DB.RocksDB.Core do
+  @moduledoc """
+  Responsible for converting type-aware, logic-specific queries and updates into leveldb specific queries and updates
+  """
+
+  # adapter - testable, if we really really want to
+
+  @single_value_parameter_names OMG.DB.single_value_parameter_names()
+
+  @keys_prefixes %{
```
pdobacz: I'm a bit nervous about these prefixes. Let's sanitize them a bit, when we have the clean chance. I was thinking about this, but never wanted to plunge into. I suggest the following:

```
# two-letter prefixes always for various buckets of data
@keys_prefixes %{
    block: "bl",
    block_hash: "bh",
    utxo: "ut",
    exit_info: "ei",
    in_flight_exit_info: "if",
    competitor_info: "ci",
    spend: "sp"
  }
```

what do you think?

apps/omg_db/lib/omg_db/rocksdb/server.ex
```diff
@@ -0,0 +1,206 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.DB.RocksDB.Server do
+  @moduledoc """
+  Handles connection to leveldb
+  """
+
+  # All complex operations on data written/read should go into OMG.DB.LevelDB.Core
```
pdobacz: fix leveldb here in the comment

apps/omg_db/lib/omg_db/rocksdb/server.ex
```diff
@@ -0,0 +1,206 @@
...
+  # WARNING, terminate below will be called only if :trap_exit is set to true
+  def terminate(_reason, %__MODULE__{db_ref: db_ref}) do
+    :ok = :rocksdb.close(db_ref)
+  end
+
+  # Argument order flipping tools :(
+  # write options
+  # write_options() = [{sync, boolean()} | {disable_wal, boolean()} | {ignore_missing_column_families, boolean()} |
+  # {no_slowdown, boolean()} | {low_pri, boolean()}]
+  @spec write(Exleveldb.write_actions(), t) :: :ok | {:error, any}
```
pdobacz: Exleveldb here

apps/omg_db/mix.exs
pdobacz: notice the comment above is about `:exleveldb` (I'm not sure if it is relevant anymore)

apps/omg_db/test/omg_db/db_test.exs
```diff
@@ -27,6 +27,7 @@ defmodule OMG.DBTest do
```
pdobacz: commenting on the `@moduledoc` :point_up:  - will this become a test agnostic of the driver/backend (the moduledoc says `LevelDB` support). Maybe the wording can be updated there?

apps/omg_db/test/omg_db/db_test.exs
```diff
@@ -63,6 +64,97 @@ defmodule OMG.DBTest do
...
+  test "if multi reading competitor infos returns writen results", %{db_dir: _dir, db_pid: pid} do
+    db_writes = create_write(:competitor_info, pid)
+    {:ok, competitors_info} = DB.competitors_info(pid)
+    [] = competitors_info -- db_writes
+  end
+
+  test "if multi reading and writting does not pollute returned values", %{db_dir: _dir, db_pid: pid} do
+    db_writes = create_write(:exit_info, pid)
+    {:ok, exits} = DB.exit_infos(pid)
+    [] = exits -- db_writes
```
pdobacz: add `assert` to emphasize what we're actually testing here (and improved error printouts I suppose)

apps/omg_db/test/omg_db/rocks_db_test.exs
```diff
@@ -0,0 +1,164 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.RocksDBTest do
+  @moduledoc """
+  A smoke test of the LevelDB support. The intention here is to **only** test minimally, that the pipes work.
```
pdobacz: RocksDB (+ see other comment to possibly update these `moduledocs` accordingly)

PR #656, opened 2019-04-30
Dependency compliance eip-712 signature test

pnowosie:
Tests that elixir-generated signatures can be successfully verified by contract code.


apps/omg/test/omg/typed_data_hash_test.exs
```diff
@@ -18,8 +18,8 @@ defmodule OMG.TypedDataHashTest do
   * tier 1: acknowledged third party (Metamask) signatures we can verify (recover address from)
   * tier 2: final structural hash on prepared transaction that gives the same signatures as above
   * tier 3: intermediate results of hashing (domain separator, structural hashes of inputs & outputs)
-  * tier 4: end-to-end test of generating signatures in elixir code and verifying them in solidity library (comming
-            soon, TODO: ref :pr: here)
+  * tier 4: end-to-end test of generating signatures in elixir code and verifying them in solidity library (
+    done in https://github.com/omisego/elixir-omg/pull/656)
```
pdobacz: if we want to leave a link here, link to code (test module name?) seems more helpful

apps/omg_eth/test/omg_eth/signature_test.exs
```diff
@@ -0,0 +1,105 @@
...
+  @moduledoc """
+  Tests that elixir-generated signatures can be successfully verified by contract code.
+  NOTE: These are integration tests as they require Ethereum node running and contract deployment,
+  however it deployed contract once for all tests and then calls a function on it which should be
+  quick enough so we can afford it runs with unit tests.
+  """
+
+  alias OMG.DevCrypto
+  alias OMG.Eth
+  alias OMG.State.Transaction
```
pdobacz: the dependencies are broken (cycle).

Such tests don't belong in the `OMG.Eth`, from the perspective of `elixir-omg` they're testing whether `elixir-omg` conforms to rules set by its dependency (`plasma-contracts`). These tests belong to wherever the signature creating/validation takes place (`:omg` probably)

apps/omg_eth/test/omg_eth/signature_test.exs
```diff
@@ -0,0 +1,105 @@
...
+  alias OMG.TestHelper
+
+  use ExUnitFixtures
+  use ExUnit.Case, async: false
+
+  @alice TestHelper.generate_entity()
+  @bob TestHelper.generate_entity()
+  @eth OMG.Eth.RootChain.eth_pseudo_address()
+  @token TestHelper.generate_entity().addr
+
```
pdobacz: this definitely requires an `:integration` and `:common` module tags. Reflecting on the note from moduledoc, unit tests should be available also on machines without `eth_node`.

apps/omg_eth/test/omg_eth/signature_test.exs
```diff
@@ -0,0 +1,105 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.Eth.SignatureTest do
+  @moduledoc """
+  Tests that elixir-generated signatures can be successfully verified by contract code.
```
pdobacz: to be precise this should test something stronger - the elixir-generated signatures are just test fixture here - remember that no signatures would be generated in elixir in production.

This test should test that EIP-712-compliant signatures generated _somehow_ (via Elixir code as it happens) are treated the same by both Elixir signature code and contract signature code.

So the `defp verify` here should rather get a signature and check whether contract verification is successful iff Elixir verification is successful.

EDIT after seeing the PR in `plasma-contracts` - you can even consider returning the signer address from `verify` in contract and compare this to the address Elixir gives you, instead of comparing in the contract.

Possibly then the `verify` name will not be appropriate anymore.

apps/omg_eth/test/support/deployer.ex
```diff
@@ -45,4 +45,14 @@ defmodule OMG.Eth.Deployer do
     Eth.deploy_contract(from, bytecode, [], [], opts)
     |> Eth.DevHelpers.deploy_sync!()
   end
+
+  def create_new(OMG.Eth.Eip712, path_project_root, from, opts) do
+    defaults = @tx_defaults |> Keyword.put(:gas, @gas_contract_token)
```
pdobacz: use a dedicated gas attribute instead of `@gas_contract_token` (value can be the same ofc)

apps/omg_eth/test/omg_eth/signature_test.exs
pdobacz: We can consider finding a place for other Elixir-vs-contract kind of tests that might follow - your test being the first of such family of tests. `OMG.PrimitivesConformance.SignatureTest`? super clunky :/, maybe there's a better name to call such tests.

apps/omg_eth/test/omg_eth/signature_test.exs
boolafish: > remember that no signatures would be generated in elixir in production.

Looking at this comment makes me really curious the purpose of the "end-to-end" test here? (Also, how does it "end-to-end", where to where?)

If Elixir is not what we are interested in testing here, then it means the goal here is to test the contract. Then the test should be in the contract repo and probably not here (also, this was mentioned in previous contract CR).

pnowosie: Changing :pr: title to hopefully less confusing (thanks @boolafish pointing it out)
I see the purpose of test is both `child chain` & `contract` verifies signatures - or depends on EIP-712 hashing.
Do they both works the same :question: When we upgrade dependency in elixir-omg, will this ability to verify signatures in both places remains unchanged :question:

apps/omg_eth/test/support/deployer.ex
```diff
@@ -45,4 +46,14 @@ defmodule OMG.Eth.Deployer do
     Eth.deploy_contract(from, bytecode, [], [], opts)
     |> Eth.DevHelpers.deploy_sync!()
   end
+
+  def create_new(OMG.Eth.Eip712, path_project_root, from, opts) do
```
pik694: please refactor `create_new` by extracting the common code of all the variants

apps/omg/test/omg/dependency_conformance/signature_test.exs
```diff
@@ -0,0 +1,100 @@
...
+    on_exit(exit_fn)
+    [contract: signtest_addr]
+  end
+
+  test "signature test empty transaction", context do
+    contract = context[:contract]
+    tx = Transaction.new([], []) |> DevCrypto.sign([@alice.priv])
+    sig = tx.sigs |> Enum.at(0)
+
+    assert true == verify(contract, tx, sig)
```
pik694: you can leave `assert verify(...)`

apps/omg_eth/test/omg_eth/signature_test.exs
pdobacz: For future reference: relevant PR for contracts: https://github.com/omisego/plasma-contracts/pull/99

PR #657, opened 2019-04-30
fix: crash reports

InoMurko:


InoMurko:
https://github.com/omisego/elixir-omg/pull/658

PR #658, opened 2019-04-30
fix: make odd length string errors go away

pdobacz:
It seems the errors happened when omg_performance "simple perftest" tests ran.

After changes from cd094b589a6881cc38696dc7006c792a6b1415c5 we started ramping up the whole :omg_child_chain, including the Ethereum-machinery.

This machinery complained about the ethereum configs being malformed.

This proposes to revert to the selective start approach prior to the mentioned change, including some readability and doc improvements.

NOTE: to reproduce the odd length string error, tag `happy_path_test` and `performance_test` with `tag: :this` and do:

```
mix test --only this --seed 294090
```

### DIFF
--- a/apps/omg_performance/lib/performance.ex
+++ b/apps/omg_performance/lib/performance.ex
@@ -50,7 +50,6 @@ defmodule OMG.Performance do
 
   alias OMG.Crypto
   alias OMG.Integration.DepositHelper
-  alias OMG.Status.Alert.Alarm
   alias OMG.TestHelper
   alias OMG.Utxo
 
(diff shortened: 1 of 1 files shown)


InoMurko:
interestingly, if you do this: `ExUnit.start(capture_log: true)` it captures the error logs as well
https://github.com/omisego/elixir-omg/pull/657/files

runs:
https://circleci.com/gh/omisego/elixir-omg/5733
https://circleci.com/gh/omisego/elixir-omg/5743

now which approach is better. hard to say. if I play the devils advocate: 
- in this specific case, the idea was to run the simple perf tests without all `EthereumListeners`. But they get shutdown because there's no blockchain client running and the alarm captures them and then prevents them from restarting. So that's covered.
- it might make the test maintenance difficult

pdobacz:
> it captures the error logs as well

aye, but it is a global setting, which I think we should use only when testing things we specifically want to print some logs out. Those error/warn logs which pop up in test runs sometimes lead to interesting findings, I'd very much like to keep them pestering us (and us fixing the issues :) )

> in this specific case, the idea was to run the simple perf tests without all EthereumListeners

We're after not only terminating those services, but never starting them. Note also, that this affected how the perf test actually run, not only perf test's smoke test. We want to run a minimal artificial subset of the production services and reason about that specifically.

> it might make the test maintenance difficult

What I see as a more proper alternative here is that we parametrize the production supervisor children with an option like `:start_eth_services`, and set it to `false` when running things like that. But I'm not sure whether we want to invest in that ATM.

I am guessing we're pretty safe - if we miss some service to start during perftesting, we'll know that rather immediately and explicitly.

PR #665, opened 2019-05-03
refactor: testing out unreachable code

InoMurko:


InoMurko:
tested, it works on v0.1

PR #666, opened 2019-05-04
setup containers with releases

InoMurko:
This is part 1 of work that needs to be done to move our deployment towards OTP releases.
All images are based on Alpine 3.8 - https://github.com/omisego-images/docker-elixir-omg
Integration tests run towards custom alpine with geth and solc. It would be good to move away from this model though.

Processes and flows have been adjusted in a way similar to eWallet... thanks @sirn !!! But with separate repositories so that we're not coupled with their work.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -332,31 +332,24 @@ jobs:
     steps:
       - checkout
       - run: make docker-child_chain CHILD_CHAIN_IMAGE_NAME=$CHILD_CHAIN_IMAGE_NAME
-      #- notify_slack_failure
-      #- notify_slack
 
   publish_watcher:
     executor: metal_watcher
     steps:
       - checkout
       - run: make docker-watcher WATCHER_PROD_IMAGE_NAME=$WATCHER_PROD_IMAGE_NAME
-      #- notify_slack_failure
-      #- notify_slack
 
   deploy_watcher:
     executor: deploy
     steps:
       - checkout
       #- run: sh .circleci/ci_deploy.sh
-      #- notify_slack_failure
-      #- notify_slack_deploy
+
   deploy_child_chain:
     executor: deploy
     steps:
       - checkout
       #- run: sh .circleci/ci_deploy.sh
-      #- notify_slack_failure
-      #- notify_slack_deploy
 
   coveralls_merge:
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,12 +34,12 @@ executors:
   metal_child_chain:
     machine: true
     environment:
-      CHILD_CHAIN_PROD_IMAGE_NAME: "omisego/child_chain"
+      IMAGE_NAME: "omisego/child_chain"
 
   metal_watcher:
     machine: true
     environment:
-      WATCHER_PROD_IMAGE_NAME: "omisego/watcher"
+      IMAGE_NAME: "omisego/watcher"
 
 commands:
   setup_elixir-omg_workspace:
@@ -247,12 +247,14 @@ jobs:
     steps:
       - checkout
       - run: make docker-child_chain CHILD_CHAIN_IMAGE_NAME=$CHILD_CHAIN_IMAGE_NAME
+      - run: sh .circleci/ci_publish.sh
 
   publish_watcher:
     executor: metal_watcher
     steps:
       - checkout
       - run: make docker-watcher WATCHER_PROD_IMAGE_NAME=$WATCHER_PROD_IMAGE_NAME
+      - run: sh .circleci/ci_publish.sh
 
   deploy_watcher:
     executor: deploy
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -502,14 +502,14 @@ workflows:
                 - v0.1
       # Publish in case of master branch.
       - publish_child_chain:
-          requires: [child_chain_coveralls_and_integration_tests, watcher_coveralls_and_integration_tests, common_coveralls_and_integration_tests, test, dialyzer, lint]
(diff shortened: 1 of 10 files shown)


apps/omg_child_chain/config/dev.exs
```diff
@@ -3,5 +3,4 @@ use Mix.Config
 
 config :omg_child_chain,
   ethereum_events_check_interval_ms: 500,
-  block_queue_eth_height_check_interval_ms: 1_000,
-  coordinator_eth_height_check_interval_ms: 1_000
```
InoMurko: it was not used at all

Dockerfile.child_chain
```diff
@@ -0,0 +1,67 @@
...
+        bash \
+        libressl \
+        libressl-dev \
+        lksctp-tools
+
+COPY rootfs /
+
+# USER directive is not being used here since privileges are dropped via
+# s6-setuigid in /entrypoint. s6-overlay is required to be run as root.
+ARG user=child_chain
```
jbunce: Are we rolling with `child_chain`? I know it's not related to this PR but there's a lot of inconsistency in this project between Childchain/child_chain/Child Chain etc

Dockerfile.child_chain
```diff
@@ -0,0 +1,67 @@
...
+
+ARG release_version
+
+ADD _build/prod/rel/child_chain/releases/${release_version}/child_chain.tar.gz /app
+RUN chown -R "${uid}:${gid}" /app
+WORKDIR /app
+
+# Child Chain app is using PORT environment variable to determine which port to run
+# the application server.
+ENV PORT 9656
```
jbunce: This isn't going to be something that changes often, and the reference to the variable is used directly below. I'd remove this.

Makefile
```diff
@@ -0,0 +1,125 @@
...
+#
+
+docker-child_chain-prod:
+	docker run --rm -it \
+		-v $(PWD):/app \
+		-v $(IMAGE_BUILD_DIR)/deps:/app/deps \
+		-u root \
+		--entrypoint /bin/sh \
+		$(IMAGE_BUILDER) \
+		-c "cd /app && make build-child_chain-prod"
```
jbunce: Not sure on this mixture of hyphens and underscores. Makes me think childchain is better.

Dockerfile.child_chain
InoMurko: I don't know what to do! :D

Makefile
InoMurko: I don't know what to do at this stage. A was told by @pdobacz the correct naming is two words.

jbunce:
Will this still allow for ENVs to be collected at runtime? Presumably this clears the way for Distillery releases too?

Dockerfile.child_chain
InoMurko: https://github.com/omisego/elixir-omg/blob/master/apps/omg_rpc/config/prod.exs#L7

InoMurko:
> Will this still allow for ENVs to be collected at runtime? Presumably this clears the way for Distillery releases too?

Yes @jbunce the work that applies ENVs has been done via Config Providers (more about them here https://hexdocs.pm/distillery/extensibility/config_providers.html). In code, they can be observed as `OMG.Eth.ReleaseTasks.SetContract`. These modules (`https://github.com/omisego/elixir-omg/blob/master/rel/config.exs#L42`) kick in just before the application release starts, does what you tell it to do and applies the configuration.

Dockerfile.child_chain
pnowosie: There was a discussion and resulted to `child chain` is a correct form. So in this context with underscore.

Makefile
pdobacz: `child-chain` is two words, so...

Dockerfile.child_chain
pdobacz: > lot of inconsistency in this project between Childchain/child_chain/Child Chain

I'd say we're pretty consistent with ChildChain/child_chain/Child chain. Two words all of these. I just searched: `Childchain` is very rare, `Rootchain` doesn't show up, there's several `childchains` and `rootchains`. We can groom this at some point.

InoMurko: I agree. I don't think I've made a naming mistake in this PR. Everything is `child_chain`!

.circleci/config.yml
```diff
@@ -281,49 +228,45 @@ jobs:
...
-      - image: omisegoimages/elixir-omg-circleci:v1.8-20190129-02
-        environment:
-          MIX_ENV: dev
-    working_directory: ~/repo
+  publish_child_chain:
+    executor: metal_child_chain
+    steps:
+      - checkout
+      - run: make docker-child_chain CHILD_CHAIN_IMAGE_NAME=$CHILD_CHAIN_IMAGE_NAME
+      #- run: sh .circleci/ci_publish.sh
```
pdobacz: is there a plan for the commented lines here or just some leftovers to be deleted? If the former, maybe worth to comment

apps/omg_child_chain/config/dev.exs
pdobacz: :+1: whoops! missed that or maybe a git-merge fail? thx for catching!

mix.exs
```diff
@@ -10,9 +10,7 @@ defmodule OMG.Umbrella.MixProject do
       deps: deps(),
       preferred_cli_env: [
         coveralls: :test,
-        "coveralls.detail": :test,
-        "coveralls.post": :test,
-        "coveralls.html": :test,
+        "coveralls.circle": :test,
```
pdobacz: I use `coveralls.details` and `.html` sometimes locally, please leave those in. 

`.post` is pre-circle era so can go

.circleci/config.yml
InoMurko: yep, that comes in as part 2 when I setup deployment!

PR #670, opened 2019-05-07
refactor: move configuration to env

InoMurko:
This would simplify the release process because we would create only one MIX_ENV=prod release and control that release with environment specific variables.

The fallback values are from *dev*.

### DIFF
--- a/apps/omg/config/config.exs
+++ b/apps/omg/config/config.exs
@@ -4,8 +4,9 @@ use Mix.Config
 
 config :omg,
   deposit_finality_margin: 10,
-  ethereum_events_check_interval_ms: 500,
-  coordinator_eth_height_check_interval_ms: 6_000,
+  ethereum_events_check_interval_ms: {:system, "ETHEREUM_EVENTS_CHECK_INTERVAL_MS", 500, {String, :to_integer}},
+  coordinator_eth_height_check_interval_ms:
+    {:system, "COORDINATOR_ETH_HEIGHT_CHECK_INTERVAL_MS", 1_000, {String, :to_integer}},
   client_monitor_interval_ms: 500
 
 config :omg, :eip_712_domain,
--- a/apps/omg/config/dev.exs
+++ b/apps/omg/config/dev.exs
@@ -1,7 +1,2 @@
-# dev config necessary to load project in iex
+# Intentionally blank!
 use Mix.Config
-
-config :omg,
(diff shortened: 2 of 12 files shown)


pdobacz:
Hm, this is interesting, I want to explore the idea some more:

1. How did you chose the set of config parameters to use this? There are many entries left with current handling - maybe we should let this approach be used for all configurable parameters
2. Taking this further - wouldn't we rather migrate to some non-`config.exs` configuration medium (it was planned anyway, maybe it's time to take this leap?). Maybe leverage sth like [vapor](https://github.com/keathley/vapor)? `config.exs` would remain only for strictly compile-time parameters
3. Regardless of 1 and 2 - I think we should provide solid default arguments for all setups our users might need - at a minimum I'd provide the old values used in `:prod` as default to be used in `:prod` with the possibility to override via ENV.

InoMurko:
1. These parameters define the behavior of both applications. I don't see anything else. Maybe those that define environment and logging but I wanted to add them separately (they just increase the amount of output logged).
2. I would discourage using anything that's not standardized, especially with the changes that elixir 1.9 is bringing (built in release creation)
3. All default arguments for 1. and 2. would be provided via docker compose environment specific file.

pdobacz:
> 1. These parameters define the behavior of both applications. I don't see anything else.

hm, what about `child_block_minimal_enqueue_gap` or `deposit_finality_margin` for example - what if one wants to configure that? I'm trying to figure out which do we chose to override via ENV and which not and why

> 2. All default arguments for 1. and 2. would be provided via docker compose environment specific file.

yeah, but what about those who chose not to use docker compose?

InoMurko:
Configuration like `child_block_minimal_enqueue_gap` or `deposit_finality_margin` we're the same in `prod` and in `dev` hence I didn't wan't to touch them. So that ought to be a separate discussion. Currently, they would be hardcoded in the releseas `sys.config` and manually editable there. We can have that discussion - what ought to be configurable and what not. I think you're the only one that really knows that :) and the behavioral implications it has.

2. If you don't use docker-compose, the release would work similarly to how it does now if you don't provide contract environment variables `OMG.Eth.ReleaseTasks.SetContract`. It would exit. 
But generally, nothing beats a good deployment readme that explains how the environment needs to be setup and what the implications are.

InoMurko:
This is closed because of a different approach: prod configs are default!

PR #675, opened 2019-05-09
Replay prevention for EIP-712  structural signatures

### LINKED ISSUE #621: Verification of network parameters during startup
[EIP-712](https://github.com/omisego/elixir-omg/issues/551) introduces 2 parameters that should be verified against network application is working on.
These parameters are
* __contract address__ - passed in configuration item: `:omg, :eip712Domain, :verifying_contract`
* __chain_id__ - passed  in configuration item: `:omg, :eip712Domain, :chain_id`

These params are used to correctly compute structured hash & verify signature. To assert that config values are up-to-date, application (child chain & Watcher) should crash on startup.

Think also about automated deployment - how to ensure proper values are passed to configuration file.

### LINKED ISSUE #677: Change signing typedData Ouput to use 'currency' instead of 'token'
Currently the `Output` spec in the signing typed data has a member named `token`.
However, everywhere else (e.g. getUtxos, getBalance) we use 'currency'.
This means the user has to change from 'currency' to 'token' when signing, which is confusing.

It would be better to be consistent and use 'currency' everywhere.

pnowosie:
* Removes support for `chain_id` in signature as not applicable in plasma transactions
* Change eip-712-domain by removing `chain_id` and use of real `child chain address` for replay prevention 
* Renaming `token` to `currency` in eip-712 type definition (thanks to @kevsul )

Closes #677
Closes #621


pnowosie:
Note to self:
- [ ] after omisego/plasma-contracts#100 :pr: is merged
- [ ] rebase into master & revert [this](https://github.com/omisego/elixir-omg/pull/675/commits/6e39c7be0805d07af3657d0cfad094fffc783d1c#diff-07a5e1eff81cafd8b9dbe4fbe68c223aR44)

PR #678, opened 2019-05-10
fix: not raise exceptions during get_challenge_data

pthomalla:


apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -351,13 +351,18 @@ defmodule OMG.Watcher.ExitProcessor do
...
+    request = %ExitProcessor.Request{se_exiting_pos: exiting_utxo_pos}
+
     response =
-      %ExitProcessor.Request{se_exiting_pos: exiting_utxo_pos}
-      |> Core.determine_standard_challenge_queries(state)
-      |> fill_request_with_standard_challenge_data()
-      |> Core.determine_exit_txbytes(state)
-      |> fill_request_with_standard_exit_id()
-      |> Core.create_challenge(state)
+      with %ExitProcessor.Request{} = standard_challenge_queries <-
```
pdobacz: let's have `Core.determine_standard_challenge_queries` return `{:ok, ...}` and match against the `:ok` like so:
```
with {:ok, request_with_queries} <- Core.determine_standard_challenge_queries(request, state),
```

This properly indicates that the call might return an `{:error, ...}` as it does. OTOH, matching the specific structure in the `with` is superfluous. Also note I've changed the variable name

PR #684, opened 2019-05-13
Refactor ExitProcessor.Core (tidy exit processor pt4)

pdobacz:
references #608

Solves multiple issues with `ExitProcessor.Core`:
  - dries consumption of ethereum events
  - dries and clarifies the flows related to piggyback invalidity checks
  - adds some tests (changes to `core_test.exs` are only additive or cleanups)
  - separates some code into submodules
  - dries the double spend functionality which was implemented three-fold (for competitors, for piggybacks and for ife-se cross checks)
  - makes handling of piggyback indices (with either the magical "4" added or not) cleaner - now "4" is added only on the "outside". Inside of `ExitProcessor.Core` "natural" indices are used + an explicit indication whether `input` or `output` is meant, as necessary

### DIFF
--- a/apps/omg_watcher/lib/omg_watcher/exit_processor/core.ex
+++ b/apps/omg_watcher/lib/omg_watcher/exit_processor/core.ex
@@ -352,8 +352,8 @@ defmodule OMG.Watcher.ExitProcessor.Core do
     updated_ifes =
       challenges
       |> Enum.reduce(ifes_to_update, fn %{tx_hash: tx_hash, output_index: output_index}, acc ->
-        with {:ok, {ife, _}} <- Map.fetch(acc, tx_hash),
-             {:ok, updated_ife} <- InFlightExitInfo.challenge_piggyback(ife, output_index) do
+        with {ife, _} = Map.fetch!(acc, tx_hash),
+             updated_ife = InFlightExitInfo.challenge_piggyback(ife, output_index) do
           # mark as updated
           %{acc | tx_hash => {updated_ife, true}}
         else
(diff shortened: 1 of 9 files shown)


pdobacz:
This beefy set of changes is comprised of some pure refactors, some polishes to tests + adding tests as well as some simplifying fixes not being just refactors.

The break down is as follows (to ease reviewing):
  - 2 fixes a18fc4c and 28a30d6
  - then some tests added in 8b78ddc to fill gaps in coverage spotted on the way
  - then a large refactor of incoming events handling https://github.com/omisego/elixir-omg/compare/8b78ddc9ba16c535999bb891cb5c99c81bd025be...187e606878f4c52e23641820e40e2666909b6a58
  - then another refactor of double-spend finding/proving, readability, tidying the input/output indices in https://github.com/omisego/elixir-omg/compare/187e606878f4c52e23641820e40e2666909b6a58...eae5fcf78d747ac8be1d19209890e92bc1b599c4
  - next moving some code to submodule in 2f1277310c82acd702bc3b3bfa8993f4d3b5735a
  - another low hanging fruit tidy e9e888f976859b9d5c81af35969ae0f74f01aa5f
  - finally some refactoring of test (Core api changed too but there's no logic change to tests) to make more succinct tests: 75db6e113d7787f65ea3e6ee837d5ce20bc472e4

In terms of `exit_processor/core.ex` taking a look at the final outcome might be also a worthwhile review.

apps/omg_watcher/test/omg_watcher/exit_processor/core/standard_exit_challenge_test.exs
```diff
@@ -194,6 +193,23 @@ defmodule OMG.Watcher.ExitProcessor.Core.StandardExitChallengeTest do
                |> Core.create_challenge(processor)
     end
 
+    test "creates challenge: tx utxo double spent not in IFE, but there is an unrelated IFE open",
```
paulperegud: Is this a negative test? Unrelated IFE being open does not affect normal SE double-spend challenge?

paulperegud: > tx utxo double spent not in IFE

is confusing a bit

apps/omg_watcher/lib/omg_watcher/exit_processor/core.ex
```diff
@@ -98,27 +98,15 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
-    Private
-    """
-    defstruct [:signed_tx, :utxo_pos]
-
-    @type t() :: %__MODULE__{
-            signed_tx: Transaction.Signed.t(),
-            utxo_pos: Utxo.Position.t()
-          }
-  end
+  @type pb_type_t() :: :input | :output
```
paulperegud: As a improvement, value tagging can be used here.

```
  @type pb_type_t() :: :input | :output
  @type pb_t() :: {pb_type_t(), non_neg_integer()}
```

apps/omg_watcher/lib/omg_watcher/exit_processor/core.ex
```diff
@@ -1017,7 +891,7 @@ defmodule OMG.Watcher.ExitProcessor.Core do
     |> Stream.filter(fn %InFlightExitInfo{is_active: is_active} -> is_active end)
     # TODO: expensive!
     |> Stream.filter(fn %InFlightExitInfo{tx: %Transaction.Signed{raw_tx: raw_tx}} ->
-      !is_among_known_txs?(raw_tx, known_txs)
+      !is_among_known_txs?(known_txs, raw_tx)
```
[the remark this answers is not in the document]
paulperegud: `!` operator on truthy value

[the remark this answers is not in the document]
pdobacz: https://hexdocs.pm/elixir/Kernel.html#!/1

maybe calling the function `find_amoung_known_txs` will make it slightly less surprising? We know `Enum.find` would return `nil` and that `!` behaves with `nil` so then it's more to the point

PR #686, opened 2019-05-14
refactor: prefix and configuration cleanup

InoMurko:
This addresses some of the configuration issues of using two databases with same configuration and sorts out the key prefixes used for iterator scans.
Also adds same measurements as we have for LDB.

### DIFF
--- a/apps/omg_db/lib/omg_db/rocksdb/server.ex
+++ b/apps/omg_db/lib/omg_db/rocksdb/server.ex
@@ -20,6 +20,8 @@ defmodule OMG.DB.RocksDB.Server do
   # All complex operations on data written/read should go into OMG.DB.RocksDB.Core
 
   use GenServer
+  use OMG.Utils.Metrics
+
   alias OMG.DB.RocksDB.Core
   alias OMG.DB.RocksDB.Recorder
   require Logger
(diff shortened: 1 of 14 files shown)


apps/omg_db/lib/omg_db/rocksdb/server.ex
```diff
@@ -61,18 +63,40 @@ defmodule OMG.DB.RocksDB.Server do
       |> String.to_atom()
 
     {:ok, _recorder_pid} = Recorder.start_link(%Recorder{name: recorder_name, parent: self(), table: table})
-    setup = [{:create_if_missing, true}, {:prefix_extractor, {:fixed_prefix_transform, 1}}]
+    setup = [{:create_if_missing, false}, {:prefix_extractor, {:fixed_prefix_transform, 5}}]
```
pdobacz: hm, `{:create_if_missing, false}` is a tiny bit surprising, why?

apps/omg_db/lib/omg_db/rocksdb/server.ex
```diff
@@ -91,17 +118,19 @@ defmodule OMG.DB.RocksDB.Server do
     {:reply, result, state}
   end
 
-  def handle_call(:utxos, _from, state) do
+  defp do_utxos(state) do
```
pdobacz: hm, I just noticed that `measure_event()` is missing here. I don't see a reason why (it's missing also in `LevelDB` implementation).

@pthomalla was there a good reason to leave it out or just missed?

If missed, then let's add (in both impls) for consistency, when we have this cleanup PR at hand.

apps/omg_db/lib/omg_db/rocksdb/server.ex
InoMurko: hmm.. this is the `init`, so you had to call `init_storage` first already, no? thats how ldb works.

apps/omg_db/lib/omg_db/rocksdb/server.ex
InoMurko: yeah, indeed it's missing from ldb as well. @pthomalla please advise. I can add this in this PR no probs.

apps/omg_db/lib/omg_db/rocksdb/server.ex
pdobacz: ah, I forgot about that split. I see `leveldb/server` has it like this, thx & nvm

apps/omg_db/lib/omg_db/rocksdb/server.ex
pthomalla: Because there is no additional logic, to get_all_by_type 
but for consistency, we should add it here too.

PR #692, opened 2019-05-16
Simplified branching and deployment process

kasima:
Simplification and clarification of the branching and deployment strategy from #614.

/cc @pdobacz @achiurizo @jbunce


docs/branching.md
```diff
@@ -0,0 +1,98 @@
...
+- `elixir-omg/master` will point to the `plasma-contracts/master`
+- A release branch in `elixir-omg` will point to the corresponding release branch in `plasma-contracts`. For example:
+  - `elixir-omg/v0.1` -> `plasma-contracts/v0.1`
+
+### For mix.lock
+- Always points to a specific SHA that in the history of the `plasma-contracts` branch referenced in `mix.exs`
+
+
+## Deployment Scenarios
+
```
achiurizo: It might be worthwhile to add a summary difference for each of our environments or linking into another document for further reading.
From [some internal documentation](https://github.com/omisego/devops/wiki/DevOps-Wiki), we could do a brief description like:

```suggestion

We have 3 types of environments that we deploy the project into:

### Development environment

Development environment is for builds that land in the master branch. This environment should be considered unstable as all commits to master will be deployed here. Continuous deployment is configured for this cluster. There are two deployment instances:

* One set of services connected to Rinkeby with Geth
* one connected to Görli with Parity.

### Staging environment

This environment should be treated as an always up environment as we will have integration from the other services that require blockchain service interaction. Continuous deployment is configured for this cluster.
```

The above is a partial suggestion and could be detailed further.

docs/branching.md
```diff
@@ -0,0 +1,98 @@
...
+### 1 - Single production deployment, ongoing development
+
+This is the active scenario most of the time.
+
+Branches and environments:
+- `master` is automatically deployed to **development** environment
+- `v0.1` is automatically deployed to **staging-v0-1** environment
```
achiurizo: nitpick: Is this a set environment naming convention? how do we feel about `staging/v0.1` as the environment name? Is that valid?

pdobacz:
Nice! By taking away the:

> If it is expected to work, then nothing needs to happen - we can just continue to use the "old" development with the new master.

it simplifies a lot (compared to #614). The benefit lost (more stable `development` environments) is probably not worth the confusion, so I'm all for this model.

Also hashing it out combined with the network upgrade flow like you did clarifies a lot. Thanks for putting this effort out of a rabbit hole :).

Let me sit on this till tmr and most likely we'll go with this.

docs/branching.md
```diff
@@ -0,0 +1,98 @@
...
+- Manually deploy to **staging-v0-2** and **production-v0-2** environments, _if absolutely necessary_
+
+### 4 - Two production deployments, ongoing development
+
+We will have two production environments during the network upgrade, so that users have the opportunity to exit the old environment and deposit into the new environment.
+
+Everything the same as Scenario 3 except - Deploying new contracts to `v0.2`
+- :stop_sign: _NOPE_
+
+Once this phase ends, we take down the older `production-v0-1` and `staging-v0-1` and return to Scenario 1. We may want to consider continuing to run a watcher for an old version a longer period of time.
```
boolafish: curious, what's the reason to keep watcher when operator is stopping already? (To see the exit games for 2 weeks?_

pdobacz: > curious, what's the reason to keep watcher when operator is stopping already? (To see the exit games for 2 weeks?_

 This too I think, but not only.

The chain is stopped but there still might be invalid exits we'd like to have an eye on. Also this is what people will do - "their" watchers will continue to do their job connected to the defunct child chain, until all tokens are exited. So we're mimicking other people's situation here.

docs/branching.md
jbunce: That sounds reasonable to me

docs/branching.md
```diff
@@ -0,0 +1,98 @@
...
+- TODO – Automate contract deployments in **development**
+
+Deploying new contracts in the release branch:
+- :stop_sign: - _NOPE_
+- We cannot deploy any `elixir-omg` code that is incompatible with the currently deployed contracts in **staging** and **production**
+
+### 2 - Production deployment, validating a new version for network upgrade
+
+Branches and environments:
+- `master` is automatically deployed to **development** environment
```
jbunce: If we're deploying the same code to two different environments couldn't we consolidate `development` and `staging` into once cluster? That would save on infrastructure costs. We can still have the concept of the two but they'd live on one cluster.

pdobacz: I don't think this is the same code. `staging` get's deployed from `v0.2` pushes, so significantly less frequently and, starting from cutting point, from different code entirely.

jbunce: As we're on `v0.2` now, it might be good to include that in this document too.

pdobacz: in this context `v0.1` and `v0.2` are used as "two example stable branches". The original doc from #692 had:

> This takes `v0.1` and `v0.2` as an example of two versions, where one is deployed in `staging`/`Ari` and the other is the upcoming upgrade and is deployed in `development`.

maybe we should add such (possibly groomed) explanation at the top?

And to answer your comment specifically - in this "phase" - "Production deployment, validating a new version for network upgrade", there is not yet a `v0.2` branch cut, the development towards `v0.2` is done in `master` branch

pdobacz:
I've left one more suggestion, as a follow-up to Jake's comment. After we possibly sort that out - shall we merge and proceed with bringing the repo up to speed?

docs/branching.md
kasima: Not tied to this environment name format but was intending to be able to use them as subdomains

docs/branching.md
kasima: Our deployment details are kept in our devops documentation. Will detail there.

docs/branching.md
kasima: The thought here is to deploy the same code (`master`) to two environments. The assumption was that **staging** is as close to **production** as possible (resources, tooling, monitoring, etc.), so we get a more realistic sense of how `master` is going to perform. This is more of an infrastructure validation step. @jbunce – certainly open to an argument about this not being necessary.

The other goal here is to delay cutting the new release branch (in this example, `v0.2`) as long as possible, so that we minimize the period we may need to port changes between `master` and `v0.2`. As stated in "Scenario 3", we cut the release branch when we're ready to provision the new **production** environment, which looks like how things went with Samrong 🙌.

_NB – Acknowledged that as the team grows and we work on features more concurrently, we'll probably have to cut the release branch sooner, which puts us in perhaps a "Scenario 2.5" – But I I'm trying to keep it simple here._

jbunce: Why deploy the same code to two different places? For me this:

- Adds more complexity in remembering which environment contract you deposited into
- Creates redundant deployments that aren't used, therefore burning unnecessary cycles (and therefore money)

I can definitely see a use case once the `v0.2` branch is cut to have:

`master` -> `development`
`v0.2` -> staging

Perhaps until then we can roll with a single deployment environment in `development`?

kasima: > Why deploy the same code to two different places?

Assuming that *staging* mirrors *production* more closely than *development*, the purpose is to vet the code against a *production*-like infrastructure. It's certainly an intermediary step.

But perhaps I'm being too cautious and overly concerned with managing two branches. If it makes things simpler, I'm fine to deploy *staging* once we cut a new release branch. I'll update to reflect that.

PR #694, opened 2019-05-16
feat: Add support to launcher.py for Infura

jbunce:
Deployments with Infura as the Ethereum connectivity backend fail as the client version RPC call is not available. This PR skips that check.


launcher.py
```diff
@@ -41,8 +41,13 @@ def start(self):
         )
         self.update_appsignal_deployment()
         self.check_chain_data_path()
-        self.ethereum_client = check_ethereum_client(self.ethereum_rpc_url)
-        logging.info('Ethereum client is {}'.format(self.ethereum_client))
+        if 'infura' not in self.ethereum_rpc_url:
```
pdobacz: 1. this could be a separate function
2. `self.ethereum_client` is unset in the `infura` case. Maybe it's safer to either not set it at all or set it to something meaningful in the `infura` case

Issue #679, opened 2019-05-10
security critical Watcher (don't use PostgreSQL)


pthomalla:
moving everything that uses PostgreSQL to levelDB
 - [x] get_utxos #696 
 - [x] compose standard exit #792
 - [x] get in flight exit

PR #696, opened 2019-05-17
feat: add get_extable_utxos to watcher security-critical API

pthomalla:


apps/omg_watcher/test/omg_watcher/web/controllers/utxo_test.exs
```diff
@@ -27,6 +26,162 @@ defmodule OMG.Watcher.Web.Controller.UtxoTest do
...
+                 "blknum" => 11_000,
+                 "txindex" => 0,
+                 "oindex" => 0,
+                 "owner" => ^carol_enc
+               }
+             ] = TestHelper.get_exitable_utxos(carol.addr)
+    end
+
+    @tag fixtures: [:blocks_inserter, :db_initialized, :alice]
+    test "outputs with value zero are not inserted into DB, the other has correct oindex", %{
```
pnowosie: remove

apps/omg/lib/omg/state/core.ex
```diff
@@ -445,4 +428,33 @@ defmodule OMG.State.Core do
...
+      end)
+
+    Enum.concat(db_updates_new_utxos, db_updates_spent_utxos)
+  end
+
+  @spec standard_exitable(list(term), Crypto.address_t()) :: list(term)
+  def standard_exitable(utxos_query_result, address) do
+    Stream.filter(utxos_query_result, fn {_, %{owner: owner}} -> owner == address end)
+    |> Stream.map(fn {{blknum, txindex, oindex}, utxo} ->
+      utxo |> Map.put(:blknum, blknum) |> Map.put(:txindex, txindex) |> Map.put(:oindex, oindex)
```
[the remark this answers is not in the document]
pthomalla: this isn't update of map :/

apps/omg/test/omg/state/core_test.exs
```diff
@@ -736,6 +736,40 @@ defmodule OMG.State.CoreTest do
     |> fail?(:utxo_not_found)
   end
 
+  @tag fixtures: [:alice, :bob, :carol]
+  test "getting user utxos from utxos_query_result", %{
```
pdobacz: 1. add more unit-testey cases here (like no utxos that belong to an address within the query result and empty query result).
2. add (or expand existing) test to the persistence tests of state, to make sure that the `utxos_query_result` is in line with the `Core` function

apps/omg_watcher/lib/omg_watcher/api/account.ex
```diff
@@ -36,4 +36,12 @@ defmodule OMG.Watcher.API.Account do
   def get_utxos(address) do
     DB.TxOutput.get_utxos(address)
   end
+
+  @spec get_exitable_utxos(OMG.Crypto.address_t()) :: list(%DB.TxOutput{})
+  def get_exitable_utxos(address) do
```
pdobacz: document using `@doc` mentioning the intent and properities and differences to `get_utxos`.

Also document the new endpoint in the appropriate API documents

apps/omg_watcher/test/fixtures.exs
```diff
@@ -217,6 +228,13 @@ defmodule OMG.Watcher.Fixtures do
     }
   end
 
+  defp conditional_update_key_value_storage({blknum, recovered_txs}) do
+    if OMG.DB.ping() do
+      db_updates = OMG.State.Core.db_update_utxos(blknum, recovered_txs)
```
pdobacz: don't do this without a good reason (see other comment)

apps/omg_watcher/test/fixtures.exs
```diff
@@ -179,13 +179,24 @@ defmodule OMG.Watcher.Fixtures do
       %{owner: bob.addr, currency: @eth, amount: 100, blknum: 2}
     ])
 
+    if OMG.DB.ping() do
```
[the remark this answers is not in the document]
pdobacz: it looks like all these modifications ans hacks are unnecessary.

You've got your logic tested in `State.CoreTest` (+ `State.PeristenceTest`). Your endpoint only uses the DB query and logic so adding an additional check in one of our integration tests will in my opinion do.

apps/omg/test/omg/state/persistence_test.exs
```diff
@@ -174,6 +183,34 @@ defmodule OMG.State.PersistenceTest do
     state
   end
 
+  defp persist_standard_exitable_utxos(%Core{utxos: utxos} = state, addresses, db_pid) do
```
pdobacz: oh wait, this is overdone. We're after "making sure that the logic in `Core` works with the results of the query".

just add a new test (or pick one exiting with some interesting utxos already in, but only if there's a sensible one to pick from) and in that test check whether

```
{:ok, utxos_query_result} = OMG.DB.utxos(db_pid)
Core.standard_exitable_utxos(utxos_query_result, address.addr)
```
works and makes sense (e.g. contains two utxos of alice or something), done.

The necessity to reach into the `%Core{utxos: utxos}` should be a test smell to you here, not to mention the necessity to maintain this test later on.

apps/omg_db/lib/db.ex
```diff
@@ -29,7 +29,7 @@ defmodule OMG.DB do
 
   @callback multi_update(term()) :: :ok | {:error, any}
   @callback blocks(block_to_fetch :: list()) :: {:ok, list(term)}
-  @callback utxos() :: {:ok, list(term)}
+  @callback utxos() :: {:ok, list({utxo_pos_db_t, OMG.Utxo.t()})}
```
pdobacz: a-a, `:omg_db`  can't depend on `OMG....`

apps/omg_watcher/lib/omg_watcher/api/account.ex
```diff
@@ -29,11 +29,22 @@ defmodule OMG.Watcher.API.Account do
...
-
-  TODO: For now uses Postgres data, but should be adapted to OMG.DB (in security-critical only mode)
   """
   @spec get_utxos(OMG.Crypto.address_t()) :: list(%DB.TxOutput{})
   def get_utxos(address) do
     DB.TxOutput.get_utxos(address)
   end
+
+  @doc """
+  Returns all utxos owner by `address`
```
pdobacz: fix the language here

apps/omg_watcher/test/fixtures.exs
```diff
@@ -227,6 +235,11 @@ defmodule OMG.Watcher.Fixtures do
         eth_height: 1
       })
 
+    {:ok, {_, _, db_updates}, _} =
+      OMG.State.Core.form_block(0, %OMG.State.Core{pending_txs: Enum.reverse(recovered_txs), height: blknum})
```
pdobacz: don't use such deep internals in testing. What if one decides to keep `pending_txs` already reversed?!

If I was forced to have those `db_updates` I'd go through a couple of **public** calls from `OMG.State.Core`, or even just write them out manually.

But see my other comment - I honestly think this is not necessary here.

apps/omg_watcher/test/omg_watcher/web/controllers/utxo_test.exs
```diff
@@ -27,6 +26,162 @@ defmodule OMG.Watcher.Web.Controller.UtxoTest do
   @eth OMG.Eth.RootChain.eth_pseudo_address()
   @eth_hex Encoding.to_hex(@eth)
 
+  describe "standard_exitable" do
```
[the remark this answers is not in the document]
pdobacz: agree on this, this is overtested. If you want useful controllers tests here, do just these:

- "no utxos are returned for non-existing addresses" (unhappy path handling, just in case, but really I think this isn't necessary)
- "badly encoded address works well when getting exitable utxos"

none of these requires any `OMG.DB` setup.

And if you absolutely want to test that those two endpoints behave the same, do a single test that compares the result of `get_utxos` with `get_exitable_utxos` and do the `OMG.DB` setup there only (and don't use internals, see other comment).

**But** that would be much nicer to solve with a single line in an existing integration test (which I suggested adding and is missing still...)

apps/omg_watcher/lib/omg_watcher/api/account.ex
pdobacz: > Also document the new endpoint in the appropriate API documents

:point_up: what about this? I only see a rename in the API documentation

apps/omg/lib/omg/state/core.ex
```diff
@@ -455,4 +438,38 @@ defmodule OMG.State.Core do
...
+
+    Enum.concat(db_updates_new_utxos, db_updates_spent_utxos)
+  end
+
+  @doc """
+    Filter user utxos from db response.
+    It may take a while for a large response from db
+  """
+  @spec standard_exitable_utxos(list({OMG.DB.utxo_pos_db_t(), OMG.Utxo.t()}), Crypto.address_t()) ::
+          list({Utxo.Position.db_t(), Utxo.t()})
```
pdobacz: I don't think this is what get's returned here. Funny that dialyzer wouldn't complain

apps/omg_watcher/lib/omg_watcher/api/account.ex
pthomalla: https://github.com/omisego/elixir-omg/blob/f45dd822c6949ace5d137ad735400afc78ddd36f/apps/omg_watcher/priv/swagger/watcher_api_specs.yaml#L49
it is already documented

pdobacz: ah, I see now. But anyway I think its `account.get_exitable_utxos`

pdobacz:
part of #679

apps/omg_watcher/test/omg_watcher/web/controllers/account_test.exs
```diff
@@ -82,4 +87,203 @@ defmodule OMG.Watcher.Web.Controller.AccountTest do
...
+             "object" => "error",
+             "code" => "operation:bad_request",
+             "description" => "Parameters required by this operation are missing or incorrect.",
+             "messages" => %{
+               "validation_error" => %{
+                 "parameter" => "address",
+                 "validator" => ":hex"
+               }
+             }
+           } == TestHelper.no_success?("account.get_utxos", %{"address" => 1_234_567_890})
```
pdobacz: check if same applies to `account.get_exitable_utxos`

apps/omg_watcher/test/support/integration/test_helper.ex
```diff
@@ -66,10 +66,17 @@ defmodule OMG.Watcher.Integration.TestHelper do
   @doc """
   We need to wait on both a margin of eth blocks and exit processing
   """
-  def wait_for_exit_processing(exit_eth_height, timeout \\ 5_000) do
+  def wait_for_exit_processing(exit_eth_height, process_exits_args, timeout \\ 5_000) do
```
pdobacz: these functions were intended to just wait, now you're waiting for the exit processing (as in `ExitProcessor`) and then calling `processExit` (different meaning of"process" here) in the contract. 

Now, the behavior is surprising (you modify the contract and SUT!).

Please undo and solve separately and test more explicitly in the integration test you've modified:

```
...
startExit
wait_for_exit_processing
check get_utxos
processExit
wait some more using a different specific function (wait_for_exit_finalization_processing?)
check get_utxos and get_exitable_utxos
```

there's just a single occurrence of this so no need to over-generalize. The other test modified can be reverted to old state, thanks to the old meaning of `wait_for_exit_processing`

apps/omg_db/lib/db.ex
```diff
@@ -43,7 +43,7 @@ defmodule OMG.DB do
   @callback initiation_multiupdate(GenServer.server()) :: :ok | {:error, any}
   @callback multi_update(term(), GenServer.server()) :: :ok | {:error, any}
   @callback blocks(block_to_fetch :: list(), GenServer.server()) :: {:ok, list()} | {:error, any}
-  @callback utxos(GenServer.server()) :: {:ok, list(term)} | {:error, any}
+  @callback utxos(GenServer.server()) :: {:ok, list({utxo_pos_db_t, OMG.Utxo.t()})} | {:error, any}
```
pdobacz: https://github.com/omisego/elixir-omg/pull/696#discussion_r285969820

> a-a, :omg_db can't depend on OMG....

PR #707, opened 2019-05-22
feat: Implement Spandex for tracing

jbunce:
Not sure if this is the correct way to implement Spandex. Let me know :)

### DIFF
--- a/apps/omg_child_chain/config/dev.exs
+++ b/apps/omg_child_chain/config/dev.exs
@@ -5,3 +5,9 @@ config :omg_child_chain,
   ethereum_events_check_interval_ms: 500,
   block_queue_eth_height_check_interval_ms: 1_000,
   coordinator_eth_height_check_interval_ms: 1_000
+
+config :omg_child_chain, OMG.ChildChain.Tracer,
+  service: :omg_child_chain,
+  adapter: SpandexDatadog.Adapter,
+  disabled?: false,
+  env: "DEV"
--- a/apps/omg_child_chain/lib/api.ex
+++ b/apps/omg_child_chain/lib/api.ex
@@ -19,6 +19,7 @@ defmodule OMG.ChildChain do
   Should handle all the initial processing of requests like state-less validity, decoding/encoding
   (but not transport-specific encoding like hex).
   """
+  require Spandex
 
   alias OMG.Block
   alias OMG.ChildChain.FeeServer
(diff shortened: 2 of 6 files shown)


apps/omg_child_chain/config/dev.exs
```diff
@@ -5,3 +5,9 @@ config :omg_child_chain,
   ethereum_events_check_interval_ms: 500,
   block_queue_eth_height_check_interval_ms: 1_000,
   coordinator_eth_height_check_interval_ms: 1_000
+
+config :omg_child_chain, OMG.ChildChain.Tracer,
```
InoMurko: hm... why dev? I think this should be a general setup and should be in `config.exs`

jbunce: Because production only gets deployed as `PROD`. If there's a `DEV` deployment as well developers can write to DD and see the impact of local changes to performance. I'm fine with moving it into `config.exs` if it makes more sense though.

InoMurko: Where is this being sent to? How does it integrate?

jbunce: The sole adapter is Datadog (https://github.com/spandex-project/spandex)

apps/omg_child_chain/lib/api.ex
```diff
@@ -36,13 +37,16 @@ defmodule OMG.ChildChain do
           {:ok, %{txhash: Transaction.tx_hash(), blknum: pos_integer, txindex: non_neg_integer}}
           | {:error, submit_error()}
   def submit(transaction) do
+    Tracer.start_span("transaction")
```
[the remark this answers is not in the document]
pdobacz: :+1: on the telemetry, but same comments as before - maybe let's just test out the results in a live env and then polish out the implementation.

Having said that, it seems it would be possible to just swap out the implementation of our `@decorate measure_event()` and `decorate measure_that_other_thing` which were implemented in the previous metrics effort by @pthomalla.

I think transfering all the metrics captured currently to spandex could be a two liner modification using https://github.com/spandex-project/spandex#decorators

I think `decorate trace` becomes an AppSignal "transaction" and `decorate span` becomes an AppSignal "event", am I parsing the spandex' README correctly?

chrishunt:
📋 cc https://github.com/omisego/elixir-omg/issues/765

jbunce:
Abandoning in favour of #765

PR #708, opened 2019-05-22
feat: return rich syncing information from /status.get

pdobacz:
On a couple of occasions we wished `status.get` returned richer information about the status the watcher's syncing is in.

Also, such information (like how old is the head of ethereum **as seen by the watcher**) is security-critical information - we're only secure and able to spend and use the child chain if we're fully up-to-date with ethereum.

So this is a proposition to put more stuff into the `status.get` response:

```
iex(30)> ~c(echo '{}' | http POST #{watcher_url}/status.get) |>  :os.cmd() |>  Poison.decode!()
%{
  "data" => %{
    "byzantine_events" => [],
!! NEW !! ===> "contract_addr" => "0x3c651506beb2ce64806f8fcf03522a211de19133",
    "eth_syncing" => false,
    "in_flight_exits" => [],
    "last_mined_child_block_number" => 1000,
    "last_mined_child_block_timestamp" => 1558535130,
!! NEW !! ===> "last_seen_eth_block_number" => 4427041,
!! NEW !! ===> "last_seen_eth_block_timestamp" => 1558535190,
    "last_validated_child_block_number" => 1000, !! NEW !! ===> "last_validated_child_block_timestamp" => 1558535130,
!! NEW !! ===> "services_synced_heights" => [
!! NEW !! ===> %{"height" => 4427041, "service" => "block_getter"},
!! NEW !! ===> %{"height" => 4427029, "service" => "challenges_responds_processor"},
!! NEW !! ===> %{"height" => 4427029, "service" => "competitor_processor"},
!! NEW !! ===> %{"height" => 4427031, "service" => "convenience_deposit_processor"},
!! NEW !! ===> %{"height" => 4427029, "service" => "convenience_exit_processor"},
!! NEW !! ===> %{"height" => 4427031, "service" => "depositor"},
!! NEW !! ===> %{"height" => 4427029, "service" => "exit_challenger"},
!! NEW !! ===> %{"height" => 4427029, "service" => "exit_finalizer"},
!! NEW !! ===> %{"height" => 4427029, "service" => "exit_processor"},
!! NEW !! ===> %{"height" => 4427029, "service" => "ife_exit_finalizer"},
!! NEW !! ===> %{"height" => 4427029, "service" => "in_flight_exit_processor"},
!! NEW !! ===> %{"height" => 4427029, "service" => "piggyback_challenges_processor"},
!! NEW !! ===> %{"height" => 4427029, "service" => "piggyback_processor"},
!! NEW !! ===> %{"height" => 4427041, "service" => "root_chain_height"}
    ]
  },
  "success" => true,
  "version" => "1.0"
}
```

As a related bonus, some more logs about syncing are now being displayed on application startup

### DIFF
--- a/apps/omg_watcher/lib/omg_watcher/api/status.ex
+++ b/apps/omg_watcher/lib/omg_watcher/api/status.ex
@@ -18,6 +18,7 @@ defmodule OMG.Watcher.API.Status do
   """
 
   alias OMG.Eth
+  alias OMG.RootChainCoordinator
   alias OMG.State
   alias OMG.Watcher.BlockGetter
   alias OMG.Watcher.Event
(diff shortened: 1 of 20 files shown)


apps/omg_watcher/lib/omg_watcher/api/status.ex
```diff
@@ -40,23 +41,33 @@ defmodule OMG.Watcher.API.Status do
...
+        last_seen_eth_block_timestamp: eth_block_timestamp,
         eth_syncing: Eth.syncing?(),
         byzantine_events: events_processor ++ events_block_getter,
-        in_flight_exits: in_flight_exits
+        in_flight_exits: in_flight_exits,
+        contract_addr: contract_addr,
+        services_synced_heights: services_synced_heights
       }
 
       {:ok, status}
```
InoMurko: if a `with` statement fails, you'll not be able to debug this.

dialyzer.ignore-warnings
```diff
@@ -1,3 +1,7 @@
 apps/omg_performance/
 apps/omg/test
 apps/omg_eth/test
```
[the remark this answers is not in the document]
pdobacz: FWIW I think this is an issue purely with a dependency, which is linked here, so we'll know to un-ignore when that's resolved. I'll track the issue and undo the ignore when the resolution is released there (or if it turns out to be ours then I'll fix on our side).

No chance of piling up here.

apps/omg_watcher/lib/omg_watcher/api/status.ex
pdobacz: true that, probably the original `with` was misused here.

Shall we pattern match the `:ok`s and fail the request if the services relied upon fail?

pdobacz: On second thought, I think it would be appropriate and consistent to return `{:error, ...}`, when the external services relied upon don't behave well.

But I agree, that the number of these calls would make the usefulness of the `with`-driven return low. How about we chunk it into some logical groups and decorate the final `{:error, ...}` returned with some useful information about which part specifically failed. Lemme propose some solution here.

InoMurko: A simple but effective approach is that you wrap the right hand side in a tuple, for example:
`with {:get_user, {:ok, user}} <- {:get_user, User.get_user(id)}`
because when your right hand side fails, you know which clause failed and you can at least log that.

pdobacz: eh, I can't seem to make up my mind today. 

For now I figured that all the things that can fail are services (nothing the user could fix), so I went with crashing the call. I pushed this version out now, but I can quite easily do the "structure `with`" back again.

Sorry to have complicated things.

Take a look how it is done now (care to take on the main review for this one, since we're already at it?)

PR #745, opened 2019-05-29
feat: more responsiveness for BlockGetter by adding BlockGetter.Status

### LINKED ISSUE #541: status.get endpoint might timeout and crash if Watcher is very busy
For example, when a large block is being processed, this can happen:

```March 25th 2019, 12:35:28.390    Request: POST /status.get

    March 25th 2019, 12:35:28.390        ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started⋅

    March 25th 2019, 12:35:28.390    2019-03-25 11:35:28.389 [error] ⋅#PID<0.8864.707> running OMG.Watcher.Web.Endpoint (cowboy_protocol) terminated

    March 25th 2019, 12:35:28.390    Server: watcher.ari.omg.network:80 (http)

    March 25th 2019, 12:35:28.390    ** (exit) exited in: GenServer.call(Sentry.TaskSupervisor, {:start_task, [{:nonode@nohost, #PID<0.8864.707>, #PID<0.8864.707>}, [#PID<0.8864.707>], :monitor, {:erlang, :apply, [#Function<3.26443835/0 in Sentry.Client.do_send_event/3>, []]}], :temporary, nil}, :infinity)

    March 25th 2019, 12:35:28.389    2019-03-25 11:35:28.388 [error] module=OMG.Watcher.Web.View.ErrorView function=template_....................ErrorView, view_template: "500.json"}⋅
    March 25th 2019, 12:35:28.387    2019-03-25 11:35:28.386 [info] module=Phoenix.Logger function=phoenix_error_render/3 request_id=2m7j1013m44465ffnkehhqr1 ⋅Converted exit {:timeout, {GenServer, :call, [OMG.Watcher.BlockGetter, :get_events, 5000]}} to 500 response⋅
```

See in particular
```{:timeout, {GenServer, :call, [OMG.Watcher.BlockGetter, :get_events, 5000]}}```.

Probably we should push data that `status.get` shows into some other process on a cadence decoupled from the API and fetch the `status.get` response from that new process?

This is the going to be same story with `ExitProcessor` - the detection of byzantine stuff shouldn't be done on every `status.get` as it can get quite involved.

### PR DESCRIPTION
**Pull request by pdobacz**
Fixes #541 

## Overview

Allows for `status.get` queries to succeed even when BlockGetter is doing some blocking heavy lifting

## Changes

- introduce `OMG.Watcher.BlockGetter.Status` which gets the relevant status of `OMG.Watcher.BlockGetter` pushed to on every change of the latter. Then this status is returned from `OMG.Watcher.BlockGetter.Status`, as opposed to `OMG.Watcher.BlockGetter` which might be busy processing blocks.

Some attempts to make `OMG.Watcher.BlockGetter` more responsive in general have been made, but no easy solution was found yet, so going forward with a simple patch fixing the problem at hand. We can go back to doing the responsiveness later.

## Testing

 - the integration test of the `status.get` endpoint touches this new service in our CI
 - to check that the busyness of the watcher syncing doesn't interfere anymore, sync to v0.2 development (it has a couple of ~1000tx blocks). During the application (!) of those blocks, `status.get` is responsive, as opposed to before the change here.

### CODE REVIEW

File: apps/omg_watcher/lib/omg_watcher/block_getter/supervisor.ex [hunk 1]
```diff
@@ -30,6 +30,7 @@ defmodule OMG.Watcher.BlockGetter.Supervisor do
     # such a situation will cause error when trying to execute already executed transaction
     children = [
       {OMG.State, []},
+      {OMG.Watcher.BlockGetter.Status, []},
```
pdobacz: I was hesitant whether this is better, or just a direct `start_link` within `OMG.Watcher.BlockGetter`'s init, to make them even more tightly bound

InoMurko: yep, that would make more sense. For what it's doing, an `Agent` would suffice as well.

apps/omg_watcher/lib/omg_watcher/block_getter/status.ex
```diff
@@ -0,0 +1,43 @@
...
+
+  def start_link(_args) do
+    GenServer.start_link(__MODULE__, :ok, name: __MODULE__)
+  end
+
+  @doc """
+  Overwrites the currently stored status with the provided one
+  """
+  @spec update(Core.chain_ok_response_t()) :: :ok
+  def update(status), do: GenServer.call(__MODULE__, {:update, status})
```
InoMurko: is strong consistency relevant? a cast would make it more decoupled.

pdobacz: I think consistency is good in this case, so I'd rather start with `call` and only `cast` if needed. I'd like the `.Status` to present status of the parent as accurately as possible, and the `update` is trivial.

By `call` we ensure that things like `status.get` will get a consistent result.

apps/omg_watcher/lib/omg_watcher/block_getter/supervisor.ex
pdobacz: ok, will change that.

I was considering `Agent` but went with `GenServer` only for the sake of limiting the toolset currently employed. But if you're making this point I'll go with `Agent`

apps/omg_watcher/lib/omg_watcher/block_getter/status.ex
InoMurko: OK. In this case, BlockGetter is now doing *more* work than it did before (it's waiting for the state update of the child process on several occasions) and blocking his own internal work.

pdobacz: but this is not about reducing BlockGetter's work, quite the contrary - this is about having BlockGetter be the unresponsive heavy lifter and put its queryable state somewhere else for responsiveness.

Also as I said, BlockGetter would wait on the super-responsive `.Status` slave, to do trivial work so that's not much blocking.

Issue #579, opened 2019-04-04
Standardize application bootup


InoMurko:
Distillery should be used to create builds for watcher and ch-ch.
Some progress:
https://github.com/omisego/elixir-omg/pull/540
OMG-439-otp_release

InoMurko:
Part 1 merged https://github.com/omisego/elixir-omg/pull/585

InoMurko:
While https://github.com/omisego/elixir-omg/pull/617 got merged, this is on hold until devops can test.

InoMurko:
I'm the devops now :) working on it again...

InoMurko:
In the process of doing this ticket, contracts were bumped from `2ff791e420b0702afe1e1514a6cd2af82cd6df4d` to `f1493aec965bf3072920bc8cf1b281782c148560`. Contracts were not redeployed, because there was no contract changes (only added truffle deployment support).

InoMurko:
DONE

PR #746, opened 2019-05-29
feature: make local watcher run a proper OTP release

InoMurko:
#579 

## Overview

Ha! The local watcher now pulls in a the latest image created from master, tagged as *latest*!

## Changes

- Getting the proper ethereum client address from env, and setting it in the sys.config via config providers.
- adding a command `full_local` that wipes the DB's every run. Ecto reset and KV wipe.

## Testing

make docker-watcher && docker-compose -f docker-compose-watcher-mac.yml up

### DIFF
--- a/apps/omg_watcher/lib/omg_watcher/block_getter/supervisor.ex
+++ b/apps/omg_watcher/lib/omg_watcher/block_getter/supervisor.ex
@@ -37,7 +37,7 @@ defmodule OMG.Watcher.BlockGetter.Supervisor do
       }
     ]
 
-    opts = [strategy: :one_for_all, max_restarts: 500, max_seconds: 5]
+    opts = [strategy: :one_for_all]
 
     _ = Logger.info("Starting #{inspect(__MODULE__)}")
     Supervisor.init(children, opts)
--- a/apps/omg_watcher/lib/omg_watcher/block_getter/supervisor.ex
+++ b/apps/omg_watcher/lib/omg_watcher/block_getter/supervisor.ex
@@ -37,7 +37,7 @@ defmodule OMG.Watcher.BlockGetter.Supervisor do
       }
     ]
 
-    opts = [strategy: :one_for_all, max_restarts: 1000, max_seconds: 60]
+    opts = [strategy: :one_for_all, max_restarts: 500, max_seconds: 5]
 
     _ = Logger.info("Starting #{inspect(__MODULE__)}")
     Supervisor.init(children, opts)
--- a/docker-compose-watcher-mac.yml
+++ b/docker-compose-watcher-mac.yml
@@ -15,16 +15,13 @@ services:
       timeout: 3s
       retries: 5
 
-  elixir-omg:
-    build: .
-    image: elixir-omg:dockercompose
-
   watcher:
-    image: elixir-omg:dockercompose
-    entrypoint: /bin/bash -c "./launcher.py && elixir --erl '-sname watcher' -S mix xomg.watcher.start --convenience --config ~/config_watcher.exs "
+    image: omisego/watcher:latest
+    command: "full_local"
+    build:
+      context: ./
+      dockerfile: Dockerfile.watcher
     environment:
-      - MIX_ENV=dev
-      - ELIXIR_SERVICE=WATCHER
       - ETHEREUM_RPC_URL=https://rinkeby.infura.io/v3/<your_api_key>
       - CHILD_CHAIN_URL=http://samrong.omg.network
       - ETHEREUM_NETWORK=RINKEBY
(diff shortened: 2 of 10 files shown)


InoMurko:
Need to port this to linux as well... battling with PG connectivity though.

InoMurko:
Need to port this to linux as well... battling with PG connectivity though.

docker-compose-watcher-mac.yml
```diff
@@ -15,24 +15,21 @@ services:
...
-      - MIX_ENV=dev
-      - ELIXIR_SERVICE=WATCHER
       - ETHEREUM_RPC_URL=https://rinkeby.infura.io/v3/<your_api_key>
       - CHILD_CHAIN_URL=http://samrong.omg.network
       - ETHEREUM_NETWORK=RINKEBY
       - RINKEBY_CONTRACT_ADDRESS=0x740ecec4c0ee99c285945de8b44e9f5bfb71eea7
       - RINKEBY_TXHASH_CONTRACT=0x29f8cd44b4b94a148f779105f0e09e06f762b411ebef6c499281b74d45818c1c
       - RINKEBY_AUTHORITY_ADDRESS=0x41863dafbdf8cfc2a33fc38c0b525b6343d857b3
       - DATABASE_URL=postgres://omisego_dev:omisego_dev@docker.for.mac.localhost:5433/omisego_dev
-    restart: always
```
InoMurko: Why was this here? @jbunce

apps/omg_watcher/lib/omg_watcher/block_getter/supervisor.ex
```diff
@@ -37,7 +37,7 @@ defmodule OMG.Watcher.BlockGetter.Supervisor do
       }
     ]
 
-    opts = [strategy: :one_for_all, max_restarts: 1000, max_seconds: 60]
+    opts = [strategy: :one_for_all, max_restarts: 500, max_seconds: 5]
```
InoMurko: I have lowered this because it was set to unrealistic numbers and would have never been propagated upwards. For example, if you don't set the `Infura` API key and let it crash with requests, it would never have propagated the error to the supervisor where an alarm is already raised for Ethereum connectivity.

pdobacz: good call, this was annoying to say the least. Maybe we can drive it even lower, since proper mitigations of eth-node connectivity issues have been put in place?

InoMurko: 3 and 5? (the default)

pdobacz: ok but just to confirm the "proper mitigations" part:

if, say, 6 of the `EthereumEventListeners` bork on `timeout` from geth or anything **at the same time** - the restarts will be held off till after geth is known to be responsive again and it won't kill the supervisor?

If we're not certain of this, then maybe a figure lower but still safely above the #`EthereumEventListeners` is a better idea.

OTOH if we such geth-related problems are going to be caught, then the default is probably the best, for simplicity sake.

InoMurko: `if, say, 6 of the EthereumEventListeners bork on timeout from geth at the same time - the restarts will be held off till after geth is known to be responsive again and it won't kill the supervisor?` - Yes.

pdobacz: perfect! then let's just do defaults here. (same for child chain server's supervisor or is it up-to-date already?)

docker-compose-watcher-mac.yml
jbunce: This needs to be here. It means the Docker container will be restarted if the BEAM dies.

InoMurko: Which is odd, because the beam never dies. :-)
It either never starts, which means it just burns useless cycles that are related to wrong configuration. Or it starts and it works.

docker-compose-watcher-mac.yml
```diff
@@ -15,24 +15,21 @@ services:
...
+      - NODE_HOST=127.0.0.1
```
jbunce: What's this ENV used for?

InoMurko:
> LGTM! any modifications to docs needed? (ones which explained how to use the dockerized watcher?)

Docs stay the same!

rootfs/watcher_entrypoint
```diff
@@ -30,6 +30,15 @@ importas -iu default_cookie default_cookie
...
 s6-env ERLANG_COOKIE=$ERLANG_COOKIE
 
+## NODE_DNS
+##
+## Used in clustering; by default wallet is configured to discover nodes via
+## DNS. Registering nodes to the cluster is a responsibility of cluster tools
+## e.g. Kubernetes.
+##
+importas -D localhost NODE_DNS NODE_DNS
+s6-env NODE_DNS=$NODE_DNS
```
jbunce: What's S6 being used for here? This is a different way from running services and I'd like to make sure we definitely need it.

docker-compose-watcher-mac.yml
jbunce: I remember a time where it did ;)

docker-compose-watcher-mac.yml
InoMurko: It's for identifying the node name

rootfs/watcher_entrypoint
InoMurko: Basically it cleans up the process for you, so you don't have 15 bash process hang around because you logged into the container trying to debug something… if you `docker exec -it container erl -remsh "..."`, cleanup is required.
It also allows us to not the run the app as root (suid tools like s6-setuidgid).

PR #747, opened 2019-05-29
feature: swap DB type via env

InoMurko:
Slack.

## Overview

RocksDB is implemented, but not tested "in the wild".

## Changes

- Pull the DB type from ENV.

## Testing
Hey, If it works, it works.

### DIFF
--- a/apps/omg_db/config/config.exs
+++ b/apps/omg_db/config/config.exs
@@ -3,7 +3,7 @@ use Mix.Config
 # see [here](README.md) for documentation
 
 config :omg_db,
-  type: :leveldb,
+  type: {:system, "DB_TYPE", :leveldb, {String, :to_atom}},
   path: Path.join([System.get_env("HOME"), ".omg/data"]),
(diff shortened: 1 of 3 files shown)


apps/omg_db/config/config.exs
```diff
@@ -3,7 +3,7 @@ use Mix.Config
 # see [here](README.md) for documentation
 
 config :omg_db,
-  type: :leveldb,
+  type: {:system, "DB_TYPE", :leveldb, {String, :to_atom}},
```
pdobacz: I just spotted that this config var is undocumented, see https://github.com/omisego/elixir-omg#omgdb-configuration---omg_db-app

InoMurko: The question is whether we want to put it there? It's not something we will support indefinitely.

InoMurko: Type is only there to ease the transition, no?

pdobacz: ah right, nvm then, haven't thought of that

Issue #724, opened 2019-05-23
Starting Watcher after fully syncing with 1.5M txs


pnowosie:
When Watcher starts for the 2nd time after it's fully sync it suffers from several timeouts and UTXO set is loaded several times.

__How to reproduce__
* Feed ch-ch with 1.5M transactions (e.g. for test #723 )
* Sync Watcher, stop it
* Start Watcher again 

[Full log in gist](https://gist.github.com/pnowosie/5df534c92ea7596d75436cf1e3e651c2)

```
2019-05-23 09:23:23.504 [info] module=OMG.DB.LevelDB function=utxos/1 ⋅Reading UTXO set, this might take a while. Allowing 600000 ms⋅
2019-05-23 09:23:52.774 [info] module=OMG.RootChainCoordinator function=handle_continue/2 ⋅Starting Elixir.OMG.RootChainCoordinator service.⋅
2019-05-23 09:23:52.774 [info] module=OMG.EthereumEventListener function=handle_continue/2 ⋅Starting EthereumEventListener for depositor.⋅
2019-05-23 09:23:52.774 [info] module=OMG.EthereumEventListener function=handle_continue/2 ⋅Starting EthereumEventListener for convenience_deposit_processor.⋅
2019-05-23 09:23:52.775 [info] module=OMG.DB.LevelDB function=exit_infos/1 ⋅Reading exits' info, this might take a while. Allowing 60000 ms⋅
2019-05-23 09:23:57.785 [error] module=gen_server function=error_info/7 ⋅GenServer :convenience_deposit_processor terminating
** (stop) exited in: GenServer.call(OMG.DB.LevelDB.Server, {:get_single_value, :last_convenience_deposit_processor_eth_height}, 5000)
    ** (EXIT) time out
    (elixir) lib/gen_server.ex:989: GenServer.call/3
    (omg) lib/omg/ethereum_event_listener.ex:71: OMG.EthereumEventListener.handle_continue/2
    (stdlib) gen_server.erl:637: :gen_server.try_dispatch/4
    (stdlib) gen_server.erl:388: :gen_server.loop/7
    (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Last message: {:continue, :setup}⋅
2019-05-23 09:23:57.785 [error] module=gen_server function=error_info/7 ⋅GenServer :depositor terminating
[...]

PR #750, opened 2019-05-31
fix: fix starting watcher

pthomalla:
:clipboard: Fixes https://github.com/omisego/elixir-omg/issues/724

## Overview
When Watcher starts for the 2nd time after it's fully sync 
it suffers from BlockGetter.Supervisor children several restart.

## Changes
BlockGetter waits for the ExitProcessor to start

## Testing
Feed ch-ch with 1.5M transactions (e.g. for test #723 )
Sync Watcher, stop it
Start Watcher again

### DIFF
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -87,15 +87,16 @@ defmodule OMG.State do
   Start processing state using the database entries
   """
   def init(:ok) do
-    # Get utxos() is essential for the State and Blockgetter. And it takes a while. TODO - measure it!
+    # Get data essential for the State and Blockgetter. And it takes a while. TODO - measure it!
     # Our approach is simply blocking the supervision boot tree
     # until we've processed history.
-    {:ok, DB.utxos(), {:continue, :setup}}
-  end
-
-  def handle_continue(:setup, {:ok, utxos_query_result}) do
+    {:ok, utxos_query_result} = DB.utxos()
     {:ok, height_query_result} = DB.get_single_value(:child_top_block_number)
     {:ok, last_deposit_query_result} = DB.get_single_value(:last_deposit_child_blknum)
+    {:ok, [utxos_query_result, height_query_result, last_deposit_query_result], {:continue, :setup}}
+  end
+
+  def handle_continue(:setup, [utxos_query_result, height_query_result, last_deposit_query_result]) do
     {:ok, child_block_interval} = Eth.RootChain.get_child_block_interval()
 
     {:ok, state} =
--- a/apps/omg_watcher/lib/omg_watcher/block_getter.ex
+++ b/apps/omg_watcher/lib/omg_watcher/block_getter.ex
@@ -72,6 +72,10 @@ defmodule OMG.Watcher.BlockGetter do
     maximum_block_withholding_time_ms = Application.fetch_env!(:omg_watcher, :maximum_block_withholding_time_ms)
     maximum_number_of_unapplied_blocks = Application.fetch_env!(:omg_watcher, :maximum_number_of_unapplied_blocks)
 
+    # TODO rethink posible solutions see issue #724
+    # if we do not wait here, `ExitProcessor.check_validity()` may timeouts,
+    # which causes State and BlockGetter to reboot, fetches entire UTXO set again, and then timeout...
+    _ = :sys.get_status(ExitProcessor, 10 * 60_000)
     exit_processor_initial_results = ExitProcessor.check_validity()
 
     {:ok, state} =
@@ -238,6 +242,7 @@ defmodule OMG.Watcher.BlockGetter do
       :nosync ->
         :ok = check_in_to_coordinator(state.synced_height)
         :ok = update_status(state)
+        {:ok, _} = schedule_sync_height()
         {:noreply, state}
 
       {:error, _} = error ->


pnowosie:
Why we aren't using PR description from template here? E.g. #744 :question:

apps/omg/lib/omg/state.ex
```diff
@@ -81,19 +81,15 @@ defmodule OMG.State do
...
   def init(:ok) do
     # Get utxos() is essential for the State and Blockgetter. And it takes a while. TODO - measure it!
     # Our approach is simply blocking the supervision boot tree
     # until we've processed history.
-    {:ok, DB.utxos(), {:continue, :setup}}
-  end
-
-  def handle_continue(:setup, {:ok, utxos_query_result}) do
+    {:ok, utxos_query_result} = DB.utxos()
+    # don't use handle continue, BlockGetter using get_status() in setup
```
pnowosie: This comment doesn't apply to following lines but also to above `getting utxos`.
Can you merge it into previous comment? You can also verbosely describe the issue root cause in the :PR: description and refer the :PR: here.

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -113,6 +111,9 @@ defmodule OMG.Watcher.BlockGetter do
          } = to_apply},
         state
       ) do
+    exit_processor_results = ExitProcessor.check_validity()
```
pnowosie: Looks dangerous as it changes the flow of processing exits. Have you considered this carefully :question:

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -72,6 +72,9 @@ defmodule OMG.Watcher.BlockGetter do
     maximum_block_withholding_time_ms = Application.fetch_env!(:omg_watcher, :maximum_block_withholding_time_ms)
     maximum_number_of_unapplied_blocks = Application.fetch_env!(:omg_watcher, :maximum_number_of_unapplied_blocks)
 
+    # TODO rethink posible solutions see issue #724
+    # wait for ExitProcessor stand up
```
pdobacz: let's have a more detailed synopsis of the breaking scenario here, what happens if we don't wait

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -72,6 +72,9 @@ defmodule OMG.Watcher.BlockGetter do
...
+    _ = :sys.get_status(ExitProcessor, :infinity)
```
pdobacz: this may wait a lot, but not `infinity` - come up with some sensible waiting time - I'd pick the pessimistic duration/timeouts of the `ExitProcessor` standing up + add small margin to that.

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -72,6 +72,10 @@ defmodule OMG.Watcher.BlockGetter do
     maximum_block_withholding_time_ms = Application.fetch_env!(:omg_watcher, :maximum_block_withholding_time_ms)
     maximum_number_of_unapplied_blocks = Application.fetch_env!(:omg_watcher, :maximum_number_of_unapplied_blocks)
 
+    # TODO rethink posible solutions see issue #724
+    # if we do not wait, in check_validity may timeout occurs,
```
pnowosie:
```suggestion
    # if we do not wait here, `ExitProcessor.check_validity()` may timeouts,
```

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -72,6 +72,10 @@ defmodule OMG.Watcher.BlockGetter do
...
+    # which causes reboot State and BlockGetter
```
pnowosie:
```suggestion
    # which causes State and BlockGetter to reboot, fetches entire UTXO set again, and then timeout...
```

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -72,6 +72,10 @@ defmodule OMG.Watcher.BlockGetter do
     maximum_block_withholding_time_ms = Application.fetch_env!(:omg_watcher, :maximum_block_withholding_time_ms)
     maximum_number_of_unapplied_blocks = Application.fetch_env!(:omg_watcher, :maximum_number_of_unapplied_blocks)
 
+    # TODO rethink posible solutions see issue #724
+    # if we do not wait here, `ExitProcessor.check_validity()` may timeouts,
+    # which causes State and BlockGetter to reboot, fetches entire UTXO set again, and then timeout...
+    _ = :sys.get_status(ExitProcessor, 10 * 60_000)
     exit_processor_initial_results = ExitProcessor.check_validity()
```
InoMurko: ugh... this is a rather brute force approach. with a bit of reactivness - pub sub this would be far more decoupled and scalable

InoMurko: Is there a time factor involved that doesn't allow you to do this properly?

pthomalla: We should't have dependencies here from ExitProcessor.
but that would make it difficult for the code with several messages sent between the genserver.

We need to think more about what dependencies our genserves should have

pnowosie: There isn't time factor but I don't think pub-sub is better here.
Bringing more complexity won't help us improve this module in the future. As now I consider it good enough.
Thanks for the comment any way, I appreciate your involvement ❗️

PR #751, opened 2019-05-31
fix: Revert removal of --no-start

jbunce:
## Overview

As title


jbunce:
Git is drunk. Commits are `Verified` now. I guess this is because the PR contains commits that weren't. Admin pushing this into the repo.

PR #754, opened 2019-05-31
fix: remove the quadratic cost of finding double spends for IFE txs

pdobacz:
Slightly related to #541, but more so to the choppy syncs of a late `development` chain, due to expensive checking of in-flight exits

## Overview

The initial implementation of in-flight exit double spend seeking involved operations costing roughly `O(#ife^2)` and `O(#ife*#txs_in_some_blocks)`, which could've exploded badly. We change the implementation (tests untouched), to use several mappings allowing us to have no quadratic operations.

## Changes

- `KnownTx` module now encompasses operations and preparations steps concerning the transactions we are scanning for double spends. The mappings are prepared there
- to find double spends we leverage mainly the `input_spent => txs_spending_it` mapping allowing us to quickly fetch double spends for an ife/se

## Testing

unit tests should cover these behaviors decently.


apps/omg_watcher/lib/omg_watcher/exit_processor/known_tx.ex
```diff
@@ -0,0 +1,110 @@
...
+
+  Expects an `Enumberable` of `KnownTx`s
+  Duplicates are possible.
+  """
+  @spec group_txs_by_input(Enumerable.t()) :: known_txs_by_input_t
+  def group_txs_by_input(all_known_txs) do
+    all_known_txs
+    |> Stream.map(&{&1, Transaction.get_inputs(&1.signed_tx)})
+    |> Stream.flat_map(fn {known_tx, inputs} -> for input <- inputs, do: {input, known_tx} end)
+    |> Enum.group_by(&elem(&1, 0), &elem(&1, 1))
```
paulperegud: I don't really understand the usage of `Stream` module here. In every case execution is made eager by some call at the end, and also all of initial input is always consumed. Is this about memory allocation for intermediate lists?

pdobacz: a late reply but yes, it is about mem for intermediate steps, as we anticipate the input enumerable can be large.

Issue #636, opened 2019-04-23
Enable specdiffs in Dialyzer and fix the issues


InoMurko:
Dialyzer needs to have stricter settings:
```
-      flags: [:error_handling, :race_conditions, :underspecs, :unknown, :unmatched_returns],
+      flags: [:specdiffs, :error_handling, :race_conditions, :underspecs, :unknown, :unmatched_returns],
```

PR #756, opened 2019-06-01
Enable `specdiffs` for Dialyzer and fix some dialyzer issues

achiurizo:
:clipboard: #636
## Overview

This enables `specdiffs` for our dialyzer settings and also makes fixes to some of our dialyzer issues and whitelists the remaining for later.

## Changes

* Enables the [specdiffs setting](ce3744e0)
* Updates [dialyzer to rc6](ce3744e0)
* Fixes [Utxo.Position dialyzer issues and adds a is_position guard](2cde6a27)
* Fixes some of the [OMG.Crypto dialyzer issues](bf8fd924) and adds it back to the [ignore list](45050f61)

## Testing

You can test changes via:

```bash
mix dialyzer
```


apps/omg/lib/omg/utxo/position.ex
```diff
@@ -25,63 +25,67 @@ defmodule OMG.Utxo.Position do
   alias OMG.Utxo
   require Utxo
 
+  import Utxo, only: [is_position: 3]
+
   @type t() :: {
           :utxo_position,
           # blknum
-          pos_integer,
+          non_neg_integer,
```
achiurizo: I relaxed this type due to dialyzer suggesting success typing for [get_position/1](https://github.com/omisego/elixir-omg/pull/756/files#diff-e24a6d0e13aa92e737647e776f74cadbR78) was able to return a `non_neg_integer` and wasn't entirely sure if `pos_integer` was necessary.

apps/omg/lib/omg/crypto.ex
```diff
@@ -37,7 +37,7 @@ defmodule OMG.Crypto do
   @doc """
   Recovers address of signer from binary-encoded signature.
   """
-  @spec recover_address(hash_t(), sig_t()) :: {:ok, address_t()} | {:error, :signature_corrupt}
+  @spec recover_address(hash_t(), sig_t()) :: {:ok, address_t()} | {:error, :signature_corrupt | binary}
```
InoMurko: FYI, dialyzier is unable to be specific when it comes to binary types. In this case, when we go down the rabbit hole of `:libsecp256k1.ecdsa_recover_compact/4` the error return tuples resolve to: `"Compression flag invalid"`, `"Recovery id invalid 0-3"`, `"ecdsa_signature_parse_compact returned 0"`, `"ecdsa recovery problem"` or `"ecdsa pubkey serialize error"`. Thats the reason why we generally prefer `atom` only tuples.

dialyzer.ignore-warnings
```diff
@@ -1,3 +1,25 @@
 apps/omg_performance/
 apps/omg/test
 apps/omg_eth/test
+
+# TODO
+apps/omg_eth/
```
InoMurko: I would suggest we're very specific which errors we're ignoring. This would capture every issue, not only those coming because of added `specdiffs`.

apps/omg/lib/omg/utxo/position.ex
```diff
@@ -25,63 +25,67 @@ defmodule OMG.Utxo.Position do
...
   def oindex(Utxo.position(_, _, oindex)), do: oindex
 
-  @spec get_position(pos_integer()) :: {pos_integer, non_neg_integer, non_neg_integer}
-  defp get_position(encoded) do
+  @spec get_position(pos_integer()) :: {non_neg_integer, non_neg_integer, non_neg_integer}
+  defp get_position(encoded) when is_integer(encoded) and encoded > 0 do
     blknum = div(encoded, @block_offset)
     txindex = encoded |> rem(@block_offset) |> div(@transaction_offset)
-    oindex = rem(encoded, @transaction_offset)
+    oindex = encoded |> rem(@transaction_offset) |> trunc() |> abs()
```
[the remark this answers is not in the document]
achiurizo: Dialyzer found a success typing for `char` here on this value. Not sure what the best way to resolve that.

apps/omg/lib/omg/crypto.ex
```diff
@@ -37,7 +37,7 @@ defmodule OMG.Crypto do
   @doc """
   Recovers address of signer from binary-encoded signature.
   """
-  @spec recover_address(hash_t(), sig_t()) :: {:ok, address_t()} | {:error, :signature_corrupt}
+  @spec recover_address(hash_t(), sig_t()) :: {:ok, address_t()} | {:error, :signature_corrupt | atom}
   def recover_address(<<digest::binary-size(32)>>, <<packed_signature::binary-size(65)>>) do
```
InoMurko: hey @achiurizo my comment https://github.com/omisego/elixir-omg/pull/756#discussion_r289621724 here was not to say that binary() is wrong, but that dialyzer is unable to be specific what the contents of a binary are.
there's not much we can do because down the rabbit hole, thats how `:libsecp256k1.ecdsa_recover_compact/4` works

PR #759, opened 2019-06-04
590 Remove 'Content-Type' header enforcing

### LINKED ISSUE #590: Allow to omit content-type header if no param sent in watcher API

pnowosie:
:clipboard: Fixes #590 
## Overview

Following HTTP-RPC we required every request specifies header `Content-Type: application/json`. This causes issues with REST clients on paths that passes no data in the body.

## Changes
Removes header enforcing.

- [apps/omg_watcher/lib/omg_watcher/web/router.ex](https://github.com/omisego/elixir-omg/blob/d4a529836c56d8c1ac70d9d850d8d9737a493a0b/apps/omg_watcher/lib/omg_watcher/web/router.ex)

## Testing

No need to test - it's HTTP protocol behavior.

### DIFF
--- a/apps/omg_watcher_rpc/lib/web/router.ex
+++ b/apps/omg_watcher_rpc/lib/web/router.ex
@@ -17,7 +17,6 @@ defmodule OMG.WatcherRPC.Web.Router do
 
   pipeline :api do
     plug(:accepts, ["json"])
-    plug(:enforce_json_content)
   end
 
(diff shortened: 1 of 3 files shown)


pnowosie:
> Would it be more beneficial to negate the test?

I have mixed feelings. Let say I followed DoD and leave a test reversed as a demonstration.
It's useless therefore I added a note in test's moduledoc

Issue #721, opened 2019-05-23
OMG-363 - Watcher get.block by number


achiurizo:
:clipboard: [PLAN-51](https://omisego.atlassian.net/browse/PLAN-51)

As a block explorer, 
I would like to receive the content of all transactions inside a child chain block based on block number,
so that I have a better way to visualize transactions based on historical data.

pdobacz:
hm, wait, we have https://developer.omisego.co/elixir-omg/docs-ui/?url=master/informational_api_swagger.json#/Transaction/getTransactionsAll where we can filter by blknum.

 - It doesn't server all the details (we agreed at one point that for details one should rather iterate on `transaction.get(hash)`
 - it is subject to the limits (200 max I think), so it would require pagination to be fully usable I suppose
 - it doesn't mimic `block.get` from the child chain, which was the plan at some point, but I think this is not what we want here because we want to "receive the content of all transactions inside a child chain block"

So either I'm missing something or perhaps the existing endpoint will do (with the extension of pagination)

pdobacz:
as discussed, only pagination is missing #725

PR #763, opened 2019-06-06
725 transaction list pagination

### LINKED ISSUE #725: Add transaction.all pagination
As a consumer of the transction.all API
I can get paginated results
So I can potentially get all the results instead of just the first 200 matches

pnowosie:
:clipboard: Fixes #725, #721

## Overview

Add pagination support for `/transaction.all` and  `/account.get_transactions` endpoints.

## Changes

1. Additional module `Paginator` that wraps the data returned by the DB query with information about the `page` and `limit`
1. Change in controller and DB layers to accept additional parameters
1. Pagination is suported by Ecto & Postgres - no special features, dependencies needed
1. Change to the response to support additional `data_paging` field sibling to `data` in standard API response

**NOTE:** As a side-effect data sanitization was moved from controllers layer to views layer. This gives us more flexibility in the views as they work on elixir datatypes not limited to maps and strings. This way is cleaner. [Commit](https://github.com/omisego/elixir-omg/commit/864e74fa5d0cb3872db0a1c7f1edb366f03548dd)
  

## Testing

New tests are added. One can test with [swagger](https://developer.omisego.co/elixir-omg/docs-ui/?url=master/informational_api_specs.yaml#/Transaction/transactions_all)


apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -438,18 +468,46 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
...
         {2000,
          [
-           Test.create_recovered([{1_000, 1, 0, alice}], @eth, [{bob, 1}])
+           Test.create_recovered([{5, 0, 0, alice}], @eth, [{bob, 10}]),
+           Test.create_recovered([{1_002, 0, 0, bob}], @eth, [{alice, 5}])
          ]}
       ])
 
-      assert [_, _, _] = TestHelper.success?("transaction.all")
+      assert {[tx1, tx2], %{"limit" => 2, "page" => 2}} = transaction_all_with_paging(%{limit: 2, page: 2})
```
pgebal: I don't understand that test. We basically put 4 transactions into the db, and get back four in total in calls for all transactions. What do we really want to test here?

pnowosie: Nice catch :+1:

apps/omg_watcher/lib/omg_watcher/db/transaction.ex
```diff
@@ -202,16 +202,16 @@ defmodule OMG.Watcher.DB.Transaction do
...
-    case Keyword.drop(constrains, allowed_constrains) do
+  defp filter_constraints(constraints, allowed_constraints) do
+    case Keyword.drop(constraints, allowed_constraints) do
       [{out_of_schema, _} | _] ->
         _ =
-          Logger.warn("Constrain on #{inspect(out_of_schema)} does not exist in schema and was dropped from the query")
+          Logger.warn("Constraint on #{inspect(out_of_schema)} does not exist in schema and was dropped from the query")
 
-        constrains |> Keyword.take(allowed_constrains)
+        constraints |> Keyword.take(allowed_constraints)
```
InoMurko: `Keyword.take(constraints, allowed_constraints)`

apps/omg_watcher_rpc/lib/web/validators/constraints.ex
```diff
@@ -12,30 +12,30 @@
...
       limit: [:pos_integer, :optional],
       page: [:pos_integer, :optional]
     ]
 
-    constrains
-    |> Enum.reduce_while({:ok, []}, fn {constrain, validators}, {:ok, list} ->
-      with {:ok, value} when not is_nil(value) <- expect(params, Atom.to_string(constrain), validators) do
-        {:cont, {:ok, [{constrain, value} | list]}}
+    constraints
+    |> Enum.reduce_while({:ok, []}, fn {constraint, validators}, {:ok, list} ->
```
InoMurko: `Enum.reduce_while(constraints, {:ok, []}, fn {constraint, validators}, {:ok, list} ->`

apps/omg_utils/lib/omg_utils/paginator.ex
```diff
@@ -0,0 +1,51 @@
...
+      constraints
+      |> Keyword.take([:page, :limit])
+      |> Keyword.put_new(:page, @first_page)
+      |> Keyword.update(:limit, max_limit, &min(&1, max_limit))
+      |> Map.new()
+
+    %__MODULE__{data: [], data_paging: data_paging}
+  end
+
+  @dialyzer {:nowarn_function, set_data: 2}
```
InoMurko: move it to the dialyzer ignore file

apps/omg_watcher_rpc/lib/web/validators/constraints.ex
InoMurko: also, this can be written much better with a case clause
```
Enum.reduce_while(constraints, {:ok, []}, fn {constraint, validators}, {:ok, list} ->
  case expect(params, Atom.to_string(constraint), validators) do
    {:ok, nil} -> {:cont, {:ok, list}}
    {:ok, value} -> {:cont, {:ok, [{constraint, value} | list]}}
    error -> {:halt, error}
  end
end
```

apps/omg_utils/lib/omg_utils/paginator.ex
pnowosie: I tried and failed - new dialyzer output format don't much the old - so you need a translator... or smth

apps/omg_watcher/lib/omg_watcher/db/transaction.ex
pnowosie: the matter of taste I prefer functional approach

InoMurko: What makes you think this is functional and the other isn’t?

apps/omg_watcher_rpc/lib/web/validators/constraints.ex
pnowosie: Indeed case is more readable - :+1:

apps/omg_watcher/lib/omg_watcher/db/transaction.ex
InoMurko: https://elixir-lang.org/getting-started/enumerables-and-streams.html#the-pipe-operator -> `Its purpose is to highlight the data being transformed by a series of functions.`

You're not chaining function calls, you're abusing the operator. Functional or non functional, it has nothing to do with it.

apps/omg_utils/lib/omg_utils/paginator.ex
InoMurko: hm... maybe @achiurizo can shed some light how he created the ignore file?

PR #768, opened 2019-06-08
fix: get cors settings from proper app

InoMurko:
CORS headers missing from response

## Overview

/

## Changes

- Fetch settings from proper app

## Testing

/


kasima:
Nice! :raised_hands:

TODO:
- throw a functional test at it?

PR #770, opened 2019-06-08
Api response proper version

pnowosie:
:clipboard: Reported by @jbunce 

## Overview

Changes API version in standard response to corresponds to OMG network version.

## Changes

Quick fix changes the string in response template.
Proper fix is to use configuration value of https://github.com/omisego/elixir-omg/blob/591cff16177d3e588b3c8be43983d67ba3546728/apps/omg/config/config.exs#L13 which requires contract redeployment


pnowosie:
It will backport to v0.2.1 through `master` with #810

Issue #766, opened 2019-06-07
Integrate Sentry for Exception Handling


achiurizo:
:clipboard: https://github.com/omisego/elixir-omg/pull/520,  https://github.com/omisego/devops/issues/79

Make sure we have moved all our exception handling and reporting over to Sentry.

PR #771, opened 2019-06-09
Add Sentry for exception reporting

achiurizo:
:clipboard: #766 

## Overview

Adds Sentry to `elixir-omg`.

## Changes

* Add [Phoenix reporting](https://github.com/getsentry/sentry-elixir#setup-with-plug-or-phoenix)
* Add [crashed process exception reporting](https://github.com/getsentry/sentry-elixir#capture-crashed-process-exceptions)
* Add [source code reporting](https://github.com/getsentry/sentry-elixir#reporting-exceptions-with-source-code)

## Testing

Need help to test it further. So far used Sentry's [included mix task](https://github.com/getsentry/sentry-elixir#testing-your-configuration) to test configuration:

```bash
SENTRY_DSN=whatever mix sentry.send_test_event
```


apps/omg/lib/omg/application.ex
```diff
@@ -20,9 +20,11 @@ defmodule OMG.Application do
 
   use Application
   alias OMG.Alert.AlarmHandler
+  require Logger
 
   def start(_type, _args) do
     :ok = AlarmHandler.install()
+    {_, _} = Logger.add_backend(Sentry.LoggerBackend)
```
achiurizo: Not sure how to do this the right way but the [documentation](https://github.com/getsentry/sentry-elixir#capture-crashed-process-exceptions) and [an issue](https://github.com/getsentry/sentry-elixir/issues/345) shows that it's only needed to be loaded once.

achiurizo: Not sure what the best way to do this is but their documentation suggests that this only needs to be loaded once otherwise it will return an error.

Dockerfile
```diff
@@ -70,6 +70,7 @@ RUN mix deps.clean --all
 RUN mix deps.get
 
 RUN mix compile
+RUN mix sentry.recompile
```
InoMurko: why is this needed?

apps/omg/lib/omg/application.ex
InoMurko: Once per node?

Dockerfile
achiurizo: @InoMurko https://github.com/getsentry/sentry-elixir#reporting-exceptions-with-source-code

apps/omg/lib/omg/application.ex
achiurizo: @InoMurko once per application instance yes

InoMurko: Hmmm. That doesn’t make too much sense. Are you sure? Where did you notice that?
Logger is elixir built in logging library that supports multiple backends. But they normally don’t need to be installed moe then once.

achiurizo: @InoMurko I don't think I explained myself well here. This line needs to only be included once, but I am unsure where is a good place within our (umbrella?) project cc https://github.com/getsentry/sentry-elixir/issues/345 for where I was looking for guidance.

apps/omg_watcher_rpc/lib/application.ex
```diff
@@ -50,6 +50,8 @@ defmodule OMG.WatcherRPC.Application do
       name: OMG.WatcherRPC.RootSupervisor
     ]
 
+    {_, _} = Logger.add_backend(Sentry.LoggerBackend)
```
chrishunt: Does it make sense to match `{:ok, _}` on these? If not, should we handle errors?

InoMurko: good point, definitely matching on `:ok` makes sense!  handling errors not so, if it doesn't work there's not much we can do and we want to fail completely.

achiurizo: Yep agreed. The reason why I went about this route is I wasn't entirely sure where the best place to put this code so and did a general match instead so I could place it everywhere and not have it complain when I loaded it more than once. 🤕  If you all have suggestions where I should load this that would be great, otherwise I'll try to create another `application.ex` for the umbrella app. re: https://github.com/getsentry/sentry-elixir/issues/345.

apps/omg/lib/omg/application.ex
InoMurko: Aha! Alright, makes sense now. If you want to understand how umbrella applications are being divided between release modes (_wathcer_, _child_chain_) you may want to look at rel/config.exs. That would explain which umbrella applications are included in one or the other modus operandi. I suggest you either include in only in _omg_, or in both `omg_watcher_rpc` and `omg_child_chain_rpc`.

dialyzer.ignore-warnings
```diff
@@ -5,6 +5,8 @@ apps/omg_eth/test
 
 # TODO
 
+Unknown function 'Elixir.Sentry':capture_exception/2
```
InoMurko: If you add `:sentry` in the PLT app list in `mix.exs` you don't need to add these ignore here.
But there's an open issues that will produce an umatched_return: https://github.com/getsentry/sentry-elixir/pull/344/files


Also, I think you forgot 
```
use Plug.ErrorHandler
use Sentry.Plug
```
to add in the integration.

apps/omg/lib/omg/application.ex
```diff
@@ -20,9 +20,11 @@ defmodule OMG.Application do
 
   use Application
   alias OMG.Alert.AlarmHandler
+  require Logger
 
   def start(_type, _args) do
     :ok = AlarmHandler.install()
+    {:ok, _} = Logger.add_backend(Sentry.LoggerBackend)
```
pthomalla: move this option to `config/config.exs`
```elixir
 config :logger,
    backends: [Sentry.LoggerBackend, :console]
```
tests will pass, and backend will be available even application is restarted.

apps/omg_child_chain_rpc/lib/omg_child_chain_rpc/application.ex
```diff
@@ -32,6 +32,7 @@ defmodule OMG.ChildChainRPC.Application do
       {OMG.ChildChainRPC.Web.Endpoint, []}
     ]
 
+
```
pthomalla: unnecessary enter

apps/omg_status/lib/omg_status/application.ex
```diff
@@ -31,6 +31,7 @@ defmodule OMG.Status.Application do
         []
       end
 
+
```
pthomalla: unnecessary enter

apps/omg_watcher_rpc/lib/application.ex
```diff
@@ -50,6 +50,7 @@ defmodule OMG.WatcherRPC.Application do
       name: OMG.WatcherRPC.RootSupervisor
     ]
 
+
```
pthomalla: unnecessary enter

config/config.exs
```diff
@@ -18,6 +18,21 @@ config :logger, :console,
   discard_threshold: 2000,
   metadata: [:module, :function, :request_id]
 
+config :logger,
+  backends: [Sentry.LoggerBackend, :console]
+
+config :sentry,
```
InoMurko: Just remembered... have you found any docs how Sentry works with a elixir release?

apps/omg_child_chain_rpc/lib/omg_child_chain_rpc/web/endpoint.ex
```diff
@@ -15,6 +15,7 @@
 defmodule OMG.ChildChainRPC.Web.Endpoint do
   use Phoenix.Endpoint, otp_app: :omg_child_chain_rpc
   use Appsignal.Phoenix
+  use Sentry.Phoenix.Endpoint
```
pthomalla: add the same line to the watcherapps/omg_watcher_rpc/lib/web/endpoint.ex

config/config.exs
achiurizo: @InoMurko nope I haven't looked for any so far. Can follow up with another PR to improve this integration.

Issue #712, opened 2019-05-23
Create scripts to setup the elixir-omg project


achiurizo:
:clipboard: [PLAN-244](https://omisego.atlassian.net/browse/PLAN-244)

As a developer of elixir-omg, I have easy set up scripts for my development environment, So that I can get started quickly

jbunce:
From the root of the `elixir-omg` directory: `docker-compose -f docker-compose-full-services-mac.yml up` will start everything for you. If there's something that needs to be improved please let me know :)

PR #773, opened 2019-06-10
Chore: Update installation instructions for Linux

chrishunt:
:clipboard: https://github.com/omisego/elixir-omg/issues/712

## Overview

This updates the installation instructions for Linux. Tested on (my) fresh linode.

## Changes

- Remove Vagrant because I don't think we recommend this anymore
- Add a few missing dependencies
- Specify solidity version `0.4.x` is required (contracts cannot be compiled with `0.5.x` right now)
  - Add instructions for installing latest `0.4.x` release manually
- Update PostgreSQL instructions and move to `install.md` from `manual_service_startup.md`
- Add Parity install instructions
- Add `mix deps.compile` (needed for `libsecp256k1-dev`)

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -7,7 +7,7 @@ Only **Linux** platforms are supported now. These instructions have been tested 
 ## Prerequisites
 * **Erlang OTP** `>=20` (check with `elixir --version`)
 * **Elixir** `>=1.6` (check with `elixir --version`)
-* **solc** `~>0.4.24` (check with `solc --version`)
+* **solc** `~>0.4` (check with `solc --version`)
 
 ### Optional prerequisites
 * **`httpie`** - to run HTTP requests from `docs/demoxx.md` demos
--- a/docs/install.md
+++ b/docs/install.md
@@ -1,7 +1,4 @@
-# 1. Installation via Vagrant
-Refer to https://github.com/omisego/xomg-vagrant.
-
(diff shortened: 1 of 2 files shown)


docs/install.md
```diff
@@ -10,15 +7,34 @@ Only **Linux** platforms are supported now. These instructions have been tested
 ## Prerequisites
 * **Erlang OTP** `>=20` (check with `elixir --version`)
 * **Elixir** `>=1.6` (check with `elixir --version`)
-* **solc** `>=0.4.24` (check with `solc --version`)
+* **solc** `~>0.4.24` (check with `solc --version`)
```
achiurizo: Should we have this match with the latter instructions grabbing [0.4.26](https://github.com/omisego/elixir-omg/pull/773/files#diff-64407d3e6d5ef382426bdfbcae040c6aR76)? 

Also FWIW, the contracts seem to mix minor versions from [0.4.0](https://github.com/omisego/plasma-contracts/blob/master/contracts/ByteUtils.sol), [0.4.18](https://github.com/omisego/plasma-contracts/blob/master/contracts/MintableToken.sol). so maybe relax this to `~>0.4`?

chrishunt: Good call, updated in https://github.com/omisego/elixir-omg/pull/773/commits/4592b31f27dfe200d1aaec2705841d8d3426baee

PR #780, opened 2019-06-12
fix: Sleep time increase to allow Watcher to deploy

jbunce:
## Overview

Test runner does not succeed because the Watcher is not yet ready. This bumps the sleep timer. 

## Changes

sleep 200

## Testing

Yolo?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -318,7 +318,7 @@ jobs:
             kubectl set image statefulset childchain-samrong childchain=$DOCKER_IMAGE
             while true; do if [ "$(kubectl get pods childchain-samrong-0 -o jsonpath=\"{.status.phase}\" | grep Running)" ]; then break; fi; done
             kubectl set image statefulset watcher-samrong watcher=$DOCKER_IMAGE
-            sleep 120 # TODO: Get the readiness status from each statefulset
+            sleep 200 # TODO: Get the readiness status from each statefulset
       - run:
           name: Functional Tests
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -318,7 +318,7 @@ jobs:
             kubectl set image statefulset childchain-samrong childchain=$DOCKER_IMAGE
             while true; do if [ "$(kubectl get pods childchain-samrong-0 -o jsonpath=\"{.status.phase}\" | grep Running)" ]; then break; fi; done
             kubectl set image statefulset watcher-samrong watcher=$DOCKER_IMAGE
-            sleep 120 # TODO: Get the readiness status from each statefulset
+            sleep 200 # TODO: Get the readiness status from each statefulset
```
InoMurko: hmmm sleeps are `ewww`. There's no TIME in distributed systems! :D

Are we able to pull every couple of seconds?

jbunce: I agree but I can't find the bloody JSON object for Readiness probes

InoMurko: isn't `status.get` enough?

jbunce: Yes, but this needs to come from Kubernetes and the not application. This is especially pertinent when we move to automated contract with service deploys where we won't know the ephemeral IPv4 that the new service is using. Needs to come from the orchestrator itself.

InoMurko: so what else can we do to avoid sleeps? this spends valuable execution time every single run...

PR #783, opened 2019-06-13
782 ch cors

InoMurko:
https://github.com/omisego/elixir-omg/issues/782
https://github.com/omisego/devops/issues/94
## Overview

Enabled CORS on child chain and enables tests for CORS headers.

## Changes

Added dependency to child_chain_rpc app and enabled tests on all controllers.

## Testing
/

### DIFF
--- a/apps/omg_child_chain_rpc/test/support/web/test_helper.ex
+++ b/apps/omg_child_chain_rpc/test/support/web/test_helper.ex
@@ -25,15 +25,25 @@ defmodule OMG.ChildChainRPC.Web.TestHelper do
       conn(method, path, params_or_body)
       |> put_req_header("content-type", "application/json")
 
-    response = request |> send_request
+    response = send_request(request)
 
     assert response.status == 200
+    # CORS check
+    assert ["*"] == get_resp_header(response, "access-control-allow-origin")
 
+    required_headers = [
+      "access-control-allow-origin",
+      "access-control-expose-headers",
+      "access-control-allow-credentials"
+    ]
+
+    for header <- required_headers do
+      assert header in Keyword.keys(response.resp_headers)
+    end
+
+    # CORS check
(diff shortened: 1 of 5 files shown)


apps/omg_child_chain_rpc/test/support/web/test_helper.ex
```diff
@@ -25,15 +25,25 @@ defmodule OMG.ChildChainRPC.Web.TestHelper do
...
 
-    response = request |> send_request
+    response = send_request(request)
 
     assert response.status == 200
+    # CORS check
+    assert ["*"] == get_resp_header(response, "access-control-allow-origin")
 
+    required_headers = [
+      "access-control-allow-origin",
```
jbunce: Are these the only headers we use? Is this a test for CORS is enabled, or additionally all the CORS headers we use?

InoMurko: We use a default configuration coming from CORS plug dependency:
https://github.com/mschae/cors_plug#configuration

InoMurko: We don't support HTTP OPTIONS. But for POST, if CORS is enabled, the endpoints should return:
```
On GET, POST, ... requests:

    Access-Control-Allow-Origin
    Access-Control-Expose-Headers
    Access-Control-Allow-Credentials

```

InoMurko: We're not setting any `origin` -> is that a requirement?

kevsul: I don't think it's necessary right now.
We discussed adding it to config, letting whoever's deploying the server specify the origin.
But that kind of fine grained control is probably overkill at this stage - if someone want this, they'll know how to do it

PR #792, opened 2019-06-17
feat: make compose_utxo_exit security-critical

pthomalla:
## Overview

endpoint /utxo.get_exit_data uses now compose exit uses key-value store instead of postgres
:clipboard: https://github.com/omisego/elixir-omg/issues/679

## Overview

Create an endpoint for composing utxo exit without using Postgres.

## Changes
- drawing an abstract to calculate compose output exit
- use abstract calculate in omg_watcher/db/txoutput
- use API.Utxo in API.InFlightExit


pnowosie:
Please update #679 checkboxes to reflect what is done ATM, if something isn't needed / not relevant also add a note.

apps/omg_watcher/lib/omg_watcher/api/in_flight_exit.ex
```diff
@@ -106,7 +105,7 @@ defmodule OMG.Watcher.API.InFlightExit do
           {<<>>, <<>>}
 
         utxo_pos ->
-          with {:ok, %{proof: proof, txbytes: txbytes}} <- DB.TxOutput.compose_utxo_exit(utxo_pos),
+          with {:ok, %{proof: proof, txbytes: txbytes}} <- OMG.Watcher.API.Utxo.compose_utxo_exit(utxo_pos),
```
pnowosie: Please alias OMG.Watcher.API for better visibility (probably 1st place API reuse other API module)
```suggestion
          with {:ok, %{proof: proof, txbytes: txbytes}} <- API.Utxo.compose_utxo_exit(utxo_pos),
```

apps/omg_watcher/test/omg_watcher/integration/in_flight_exit_test.exs
```diff
@@ -51,6 +51,11 @@ defmodule OMG.Watcher.Integration.InFlightExitTest do
         [{alice, 5}, {bob, 15}]
       )
 
+    in_flight_tx =
```
pnowosie: Please add comment to the `compose_deposit_exit` about the consequences, so you can't compose exit from spent deposit. Also add comment here that we need to prepare these txs before utxos get spend (and refer to the function comment)

apps/omg_watcher_rpc/test/fixtures.exs
```diff
@@ -70,15 +70,34 @@ defmodule OMG.WatcherRPC.Fixtures do
     :ok
   end
 
-  @doc "run only database in sandbox and endpoint to make request"
-  deffixture phoenix_ecto_sandbox do
+  deffixture web_endpoint do
```
pnowosie: How about?
```suggestion
  deffixture phoenix_without_ecto do
```

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/utxo_test.exs
```diff
@@ -23,19 +23,53 @@ defmodule OMG.WatcherRPC.Web.Controller.UtxoTest do
 
   @eth OMG.Eth.RootChain.eth_pseudo_address()
 
-  @tag fixtures: [:initial_blocks]
-  test "getting exit data returns properly formatted response" do
+  @tag fixtures: [:phoenix_ecto_sandbox, :db_initialized]
+  test "get_exit_data should return error when there is no txs in specfic block" do
+    assert %{
+             "code" => "get_utxo_exit:no_deposit_for_given_blknum",
```
pnowosie: I think we can remap this error to the below, so the return value should be 
```
  "code" => "exit:invalid",
  "description" => "Utxo was spent or does not exist.",
```
This is basically the same, the fact it's deposit utxo is not relevant

See: https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher_rpc/lib/web/controllers/fallback.ex

apps/omg_watcher/lib/omg_watcher/api/utxo.ex
```diff
@@ -42,4 +40,60 @@ defmodule OMG.Watcher.API.Utxo do
   def create_challenge(utxo) do
     ExitProcessor.create_challenge(utxo)
   end
+
+  @decorate measure_event()
+  @spec compose_utxo_exit(Utxo.Position.t()) :: {:ok, exit_t()} | {:error, :utxo_not_found}
```
pdobacz: @pthomalla @pnowosie these functions here (and probably the on for IFEs) seems rather straightforward to (unit)test without engaging global state (`OMG.DB`). Why don't we follow the core-shell distinction here? We can then also increase the number of paths to exit composing tested a lot. We can even allow some lee-way and have the `if Utxo.Position.is_deposit?` in here, but the rest of the flow is easy to split.

pnowosie:
I can see this wasn't done https://github.com/omisego/elixir-omg/pull/792#issuecomment-503468512
Also please update the PR description - about what it deliver, what changes...

apps/omg_watcher/lib/omg_watcher/api/utxo.ex
```diff
@@ -42,4 +40,19 @@ defmodule OMG.Watcher.API.Utxo do
...
+  def compose_utxo_exit(decoded_utxo_pos) when is_deposit(decoded_utxo_pos) do
```
pnowosie: As we know `API` works on domain structures - I'd drop `decoded` as obvious.
```suggestion
  def compose_utxo_exit(utxo_pos) when is_deposit(utxo_pos) do
```

apps/omg_watcher/lib/omg_watcher/api/utxo.ex
```diff
@@ -42,4 +40,19 @@ defmodule OMG.Watcher.API.Utxo do
...
+    OMG.DB.utxos() |> Core.get_deposit_utxo(decoded_utxo_pos) |> Core.compose_deposit_exit(decoded_utxo_pos)
+  end
+
+  def compose_utxo_exit(Utxo.position(blknum, _, _) = decoded_utxo_pos) do
```
pnowosie: Also drop `decoded_utxo_pos` -> `utxo_pos`

apps/omg_watcher/lib/omg_watcher/api/core.ex
```diff
@@ -0,0 +1,86 @@
...
+  """
+
+  alias OMG.Block
+  alias OMG.State.Transaction
+  alias OMG.Utxo
+  use OMG.Utils.Metrics
+  require Utxo
+
+  @decorate measure_event()
+  def compose_output_exit([%{txindex: _} | _] = txs, Utxo.position(_, txindex, _) = decoded_utxo_pos) do
```
pnowosie: Please add @spec
Also I can't see a value in `[%{txindex: _} | _] = txs` this only assumes that 1st elem has `txindex`. If we add spec it'd became better documented

apps/omg_watcher/lib/omg_watcher/api/core.ex
```diff
@@ -0,0 +1,86 @@
...
+    if txs |> Enum.any?(&match?(%{txindex: ^txindex}, &1)) do
+      sorted_tx_bytes =
+        txs
+        |> Enum.sort_by(& &1.txindex)
```
pnowosie: I have a filling that all callers (to this function) already provide `txs` already sorted...
However I'm not sure we can assume this fact as granted... OK, never mind - just thinking aloud ;)

apps/omg_watcher/test/omg_watcher/api/core_test.exs
```diff
@@ -0,0 +1,104 @@
...
+                 owner: alice.addr
+               },
+               position
+             )
+  end
+
+  @tag fixtures: [:alice]
+  test "compose output exit", %{alice: alice} do
+    tx_encode = fn number ->
+      TestHelper.create_signed([{1_000, number, 0, alice}], @eth, [{alice, 10}])
```
pnowosie: I believe we have test helper [`create_encoded`](https://github.com/omisego/elixir-omg/blob/64d02e994ea06a9c8cda3650f5894050c7eacffa/apps/omg/test/support/test_helper.ex#L94), but this function is still helpful

apps/omg_watcher/test/omg_watcher/api/core_test.exs
```diff
@@ -0,0 +1,104 @@
...
+               position
+             )
+  end
+
+  test "getting exit data returns error when there is no utxo" do
+    assert {:error, :utxo_not_found} == Core.compose_output_exit([], Utxo.position(1_000, 1, 2))
+  end
+
+  test "return utxo when in blknum in utxos map" do
+    assert 7 ==
```
pnowosie: Uhh, this makes it really hard to reason about. I'm kindly asking to generate some `utxo-looking` structs just for test :)

apps/omg_watcher/test/omg_watcher/db/txoutput_test.exs
```diff
@@ -23,30 +23,6 @@ defmodule OMG.Watcher.DB.TxOutputTest do
   require Utxo
 
   @eth OMG.Eth.RootChain.eth_pseudo_address()
-
```
pnowosie: Once we still have a functionality in DB.TxOutput - let's keep the tests

apps/omg_watcher/lib/omg_watcher/api/core.ex
```diff
@@ -0,0 +1,86 @@
...
+
+      compose_output_exit(sorted_tx_bytes, decoded_utxo_pos)
+    else
+      {:error, :utxo_not_found}
+    end
+  end
+
+  @decorate measure_event()
+  def compose_output_exit(sorted_tx_bytes, Utxo.position(_blknum, txindex, _) = decoded_utxo_pos) do
+    if signed_tx = Enum.at(sorted_tx_bytes, txindex) do
```
InoMurko: why not use Elixir style pattern matching?
```
case Enum.at(sorted_tx_bytes, txindex) do
^signed_tx -> ...
```

apps/omg_watcher/lib/omg_watcher/api/core.ex
```diff
@@ -0,0 +1,86 @@
...
+      {:ok, %Transaction.Signed{sigs: sigs} = tx} = Transaction.Signed.decode(signed_tx)
+      proof = sorted_tx_bytes |> Block.inclusion_proof(txindex)
```
InoMurko: can we finally agree on consistent function calls?

apps/omg_watcher/lib/omg_watcher/api/core.ex
```diff
@@ -0,0 +1,86 @@
...
+
+      utxo_pos = decoded_utxo_pos |> Utxo.Position.encode()
+      sigs = Enum.join(sigs)
```
InoMurko: why is the first one piped and the second one isn't?

apps/omg_watcher/lib/omg_watcher/api/core.ex
```diff
@@ -0,0 +1,86 @@
...
+         sigs: sigs
+       }}
+    else
+      {:error, :utxo_not_found}
+    end
+  end
+
+  @decorate measure_event()
+  def get_deposit_utxo(utxos, Utxo.position(blknum, _, _)) do
+    with {:ok, utxos} <- utxos,
```
InoMurko: you can pattern match `utxos` in the function header, and don't have to write the with clause entirely

apps/omg_watcher/lib/omg_watcher/utxo_exit/core.ex
```diff
@@ -0,0 +1,96 @@
...
+      sorted_tx_bytes = Enum.sort_by(txs, & &1.txindex) |> Enum.map(& &1.txbytes)
+      compose_output_exit(sorted_tx_bytes, utxo_pos)
+    else
+      {:error, :utxo_not_found}
+    end
+  end
+
+  @decorate measure_event()
+  def compose_output_exit(sorted_tx_bytes, Utxo.position(_blknum, txindex, _) = utxo_pos) do
+    if signed_tx = Enum.at(sorted_tx_bytes, txindex) do
```
InoMurko: a proper use of pattern matching is using a `case` clause with `^` operator.

apps/omg_watcher/lib/omg_watcher/utxo_exit/core.ex
```diff
@@ -0,0 +1,96 @@
...
+               proof: binary
+             }}
+  def compose_deposit_exit(%{amount: amount, currency: currency, owner: owner}, utxo_pos) do
+    tx = Transaction.new([], [{owner, currency, amount}])
+    txs = [%Transaction.Signed{raw_tx: tx, sigs: []} |> Transaction.Signed.encode()]
+
+    {:ok,
+     %{
+       utxo_pos: Utxo.Position.encode(utxo_pos),
+       txbytes: Transaction.raw_txbytes(tx),
```
InoMurko:
```
txs = [%Transaction.Signed{raw_tx: tx, sigs: []} |> Transaction.Signed.encode()]
txbytes: Transaction.raw_txbytes(tx),
```
why the difference?

apps/omg_watcher/lib/omg_watcher/utxo_exit/core.ex
pthomalla: here I don't use pattern matching, then I don't need to use `^`.

InoMurko: what is it then?

apps/omg_watcher/lib/omg_watcher/utxo_exit/core.ex
pthomalla: I don't understand question.
Do you mean why raw_tx is different from txbytes?
Or why not this whay
```elixir
   [Transaction.Signed.encode(%Transaction.Signed{raw_tx: tx, sigs: []})]
```

InoMurko: No. I mean why is one usinga pipe and the other one doesn't. Both functions accept one argument and there's no function chaining.

apps/omg_watcher/lib/omg_watcher/utxo_exit/core.ex
pthomalla: check if is different from nil ant give name to use it lower

InoMurko: that's kinda confusing that you're making assignments in an if clause, no?

apps/omg_watcher/lib/omg_watcher/utxo_exit/core.ex
pthomalla: ok, the second one

apps/omg_watcher/lib/omg_watcher/utxo_exit/core.ex
pthomalla: That's why i changed it.
Although I'm a bit surprised because assignments in an if clause is very typical.
even some of these languages ​​add such a possibility to reduce the scope of the variable

pnowosie: I'm opting to leave the `if new_var_when_not_nil = smth()` kind of code pattern we already using elsewhere, and you can see in elixir/python code.

InoMurko: I've definitely never seen it before and it seems counter intuitive.

pthomalla: for me is quite intuitive :smile: 
in elixir-lang often occurs (170 matches in 83 files)

InoMurko: Cool! I'll adopt it then!

PR #793, opened 2019-06-18
Add `docker-compose` and `docker-compose.dev`

achiurizo:
:clipboard: first PoC https://github.com/omisego/elixir-omg/commit/31a85973a25b5a7432b0aebbaefafb91c54d0ff9
:clipboard: https://github.com/omisego/elixir-omg/issues/712

## Overview

This creates a new `docker-compose` that will allows us to run testing/development flows through docker and without having to install the dependencies locally and also have docker "hot/live reload" our code without rebuilding.

This also removes some duplication between our `docker-compose` files and encourages referencing it from `docker-compose.yml`:

```sh
docker-compose -f docker-compose.yml -f docker-compose.dev.yml build
```

## Changes

* Creates a new [`docker-compose.yml`](https://github.com/omisego/elixir-omg/pull/793/files#diff-0c4274d5850a4c83b7a70dfd428c8b28L14) that holds the common services/deps that each docker has. (e.g postgres, elixir-omg)
* Remove duplication of services like `postgres` [across the docker files](https://github.com/omisego/elixir-omg/pull/793/files#diff-0c4274d5850a4c83b7a70dfd428c8b28L14)
* Adds a [`docker-compose.dev.yml`](https://github.com/omisego/elixir-omg/pull/793/files#diff-f785228868530094a59a6dfaebc3a348R1) that enables "code reloading" without rebuilding
* ~Updates docker files to use `docker-compose` version `3.4`. There are some caveats to this change though:~
  - ~version 3.0 no longer allows [`condition`](https://docs.docker.com/compose/compose-file/#depends_on) when using `depends_on`~
  - ~Also open to backing out this change if we are not ready for this.~

## Testing

Startup a shell on the docker development instance:
```sh
docker-compose -f docker-compose.yml -f docker-compose.dev.yml run --rm --entrypoint bash elixir-omg
bash $ mix test
```

Alternatively, run it in one-line:

```sh
docker-compose -f docker-compose.yml -f docker-compose.dev.yml run --rm --entrypoint "mix test" elixir-omg
```

Run the full service via docker:

```sh
docker-compose -f docker-compose.yml -f docker-compose-full-services-non-mac.yml up
```


InoMurko:
Looks good!

achiurizo:
I'm going to break out the `3.x` `docker-compose` upgrade for a later PR to address the concerns around deprecated `condition` key.

Issue #805, opened 2019-06-25
API response should return OMG network version


pnowosie:
Standard response schema from the API (Watcher & ch-ch) consist of `version` field, set to `1.0`.
We agreed the `version` means `OMG network` version.

From the other side there is EIP712Domain structure - part of EIP-712 structural signing proposal, which contains the [version field](https://github.com/omisego/elixir-omg/blob/bdefa2c77661b3255fb467531f235d419a355da2/apps/omg/config/config.exs#L13) as well.

It could be beneficial that API response will present appropriate _network version_.  

However I'd like to put into discussion the fact that change in EIP712Domain requires contract redeployment (invalidates all previous signatures). 
This means that API response would not follow release tags (Samrong is now `v0.2.1`, API responses would show `version: "0.2"`)

What's your point on :point_up: @Pongch @kevsul @kasima @pdobacz :question:

kasima:
We'll match up the EIP-712 Domain version in the next Network Upgrade and we'll keep that matched to the minor version until we reach the major version.

In the meantime, let's merge #770 to `v0.2` branch **and** port it to `master`.

pnowosie:
@kasima I've created #/810 to `master` was `v0.2` cut yet or my change will eventually appear in `v0.2.x` :question:

kasima:
Please port it to `v0.2` branch (may be a fast-forward) and when a release happens, cut the `v0.2.2` tag.

PR #810, opened 2019-06-26
805 api response version master

pnowosie:
:clipboard: #805
## Overview

API response shows OMG network version (matched to the `major.minor` [semver](https://semver.org/))

## Changes

Changed standard response version field value.

## Testing

Unit tested

### DIFF
--- a/apps/omg_utils/lib/omg_utils/http_rpc/response.ex
+++ b/apps/omg_utils/lib/omg_utils/http_rpc/response.ex
@@ -72,7 +72,7 @@ defmodule OMG.Utils.HttpRPC.Response do
 
   defp to_response(data, result),
     do: %{
-      version: "1.0",
+      version: "0.2",
       success: result == :success,
       data: data
     }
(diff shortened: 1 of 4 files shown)


apps/omg_utils/lib/omg_utils/http_rpc/response.ex
```diff
@@ -72,7 +72,7 @@ defmodule OMG.Utils.HttpRPC.Response do
 
   defp to_response(data, result),
     do: %{
-      version: "1.0",
+      version: "0.2",
```
pdobacz: not entirely sure if that's what we want (i.e. whether we want to DRY and automate this), but you could grab the version from the top-level `mix.exs` file and use functions around https://hexdocs.pm/elixir/Version.html to pull major.minor out of that.

But maybe this is not such a good idea to couple the two so strongly :thinking: ?...

pnowosie: The idea is to couple this API reported version with EIP712 domain - which means OMG network version. This is separated from mix / application version.

Issue #767, opened 2019-06-07
Decommission AppSignal


achiurizo:
Follow up task once we have fully migrated off AppSignal. Dependent on issues like:
#549
#765 
#766 

move deploy marker functionality from appsignal to datadog

InoMurko:
https://github.com/omisego/elixir-omg/pull/851

Issue #765, opened 2019-06-07
APM with Datadog


achiurizo:
:clipboard: broken out from https://github.com/omisego/devops/issues/79

https://docs.datadoghq.com/tracing
https://www.datadoghq.com/apm

- [ ] Integrate monitoring with APM
- [ ] Add method integration for transaction traces from child chain and watcher services

PR #812, opened 2019-06-26
Chore: Replace AppSignal with Spandex/Datadog in OMG.Utils.Metrics

chrishunt:
:clipboard: https://github.com/omisego/elixir-omg/issues/765 https://github.com/omisego/elixir-omg/issues/767

## Overview

This PR replaces AppSignal with Spandex/Datadog in `OMG.Utils.Metrics`. We may want to scrap decorators in favor of [`telemetry`](https://github.com/beam-telemetry/telemetry), but using our exisiting decorators seemed like an easier first step to get things going (and was suggested by @pdobacz in https://github.com/omisego/elixir-omg/pull/707).

Opening this PR a bit early for feedback while I test. I'm sure things will be adjusted as we get it working. I wasn't really sure where to put the `deps` and `config` since we have so many different apps. These pieces might need to move around a bit as we test and I figure out what is going on here. We may make separate tracers for each application or remove the custom tracer completely (as described in the [spandex docs](https://github.com/spandex-project/spandex#setup-and-configuration))

After we merge https://github.com/omisego/elixir-omg/pull/775 and this PR, we can remove the AppSignal dependency (https://github.com/omisego/elixir-omg/issues/767).

## Changes

- Add `spandex`, `spandex_datadog`, `decorator` as dependencies
-  Add `OMG.Utils.Tracer`
- Replace AppSignal decorators with `Tracer.start_{trace|span}`

## Testing

I'm relying on existing tests and dialyzer right now - may write separate tests for `OMG.Utils.Metrics` as I test in development and try to put stuff in Datadog.


apps/omg/lib/omg/supervisor.ex
```diff
@@ -28,7 +28,15 @@ defmodule OMG.Supervisor do
     children = [
       {OMG.InternalEventBus, []},
       {OMG.EthereumClientMonitor, [alarm_module: Alarm]},
-      {OMG.EthereumHeight, []}
+      {OMG.EthereumHeight, []},
+      {SpandexDatadog.ApiServer,
```
chrishunt: Spandex uses this `ApiServer` to send requests to Datadog asynchronously. We have many supervisors and I'm still digging into where this should live, but tests pass when I put it here so it seemed like a good start. 😆

apps/omg_child_chain/config/config.exs
```diff
@@ -9,4 +9,9 @@ config :omg_child_chain,
   fee_specs_file_name: "fee_specs.json",
   ignore_fees: false
 
+config :omg_child_chain, OMG.Utils.Tracer,
```
chrishunt: We're providing a separate config for each app because the `service` name is different. This will be passed to Datadog and will help us see what part of the app is sending the trace.

apps/omg_eth/lib/omg_eth/root_chain.ex
```diff
@@ -141,7 +141,8 @@ defmodule OMG.Eth.RootChain do
   end
 
   @decorate measure_event()
-  @spec get_piggybacks(non_neg_integer, non_neg_integer, optional_addr_t) :: {:ok, [in_flight_exit_piggybacked_event]}
+  @spec get_piggybacks(non_neg_integer, non_neg_integer, optional_addr_t) ::
+          {:ok, [in_flight_exit_piggybacked_event]} | {:error, :failed_to_get_ethereum_events}
```
chrishunt: The `dialyzer` had an issue with this spec, so I updated it to include the `:error` result.

apps/omg_utils/lib/omg_utils/metrics.ex
```diff
@@ -33,18 +33,61 @@ defmodule OMG.Utils.Metrics do
...
       do: body,
-      else: Appsignal.Instrumentation.Decorators.transaction_event(event_group, body, context)
+      else: start_span(body, context)
+  end
+
+  defp start_trace(body, context) do
+    trace_name = trace_name(context)
+
+    quote do
+      _ = unquote(Tracer).start_trace(unquote(trace_name))
```
chrishunt: I decided to match `_` instead of `{:ok, _}` here because I don't think we want to change any behavior if the tracer returns an error result... we want to continue as normal.

apps/omg_watcher/lib/omg_watcher/api/transaction.ex
```diff
@@ -46,7 +46,7 @@ defmodule OMG.Watcher.API.Transaction do
   Length of the list is limited by `limit` argument
   """
   @decorate measure_event()
-  @spec get_transactions(Keyword.t()) :: list(%DB.Transaction{})
+  @spec get_transactions(Keyword.t()) :: Paginator.t()
```
chrishunt: I made this change to satisfy the dialyzer. It looks like `DB.Transaction.get_by_filters()` actually does return a `Paginator.t()`, so the change made sense. 🤷‍♂

dialyzer.ignore-warnings
```diff
@@ -266,4 +266,16 @@ lib/websockex.ex:19: Function handle_cast/2 only terminates with explicit except
...
+apps/omg_child_chain/lib/child_chain.ex:35: Type specification 'Elixir.OMG.ChildChain':submit(transaction::binary()) -> {'ok',#{'txhash':='Elixir.OMG.State.Transaction':tx_hash(),'blknum':=pos_integer(),'txindex':=non_neg_integer()}} | {'error',submit_error()} is a subtype of the success typing: 'Elixir.OMG.ChildChain':submit(binary()) -> {'error',_} | {'ok',#{'blknum':=_, _=>_}}
+apps/omg_child_chain/lib/child_chain.ex:49: Type specification 'Elixir.OMG.ChildChain':get_block(hash::binary()) -> {'ok',#{'hash':=binary(),'transactions':=[any()],'blknum':=integer()}} | {'error','not_found' | 'internal_error'} is a subtype of the success typing: 'Elixir.OMG.ChildChain':get_block(binary()) -> {'error',_} | {'ok',#{'blknum':=_, _=>_}}
```
chrishunt: Ignoring several "is a subtype of the success typing" as recommend by Ino. Let me know if that doesn't make sense here please.

apps/omg_utils/mix.exs
```diff
@@ -25,5 +25,12 @@ defmodule Utils.MixProject do
   defp elixirc_paths(:prod), do: ["lib"]
   defp elixirc_paths(_), do: ["lib", "test/support"]
 
-  defp deps, do: [{:appsignal, "~> 1.9"}]
+  defp deps do
+    [
+      {:appsignal, "~> 1.9"},
+      {:spandex, "~> 2.4"},
+      {:spandex_datadog, "~> 0.4"},
+      {:decorator, "~> 1.2"}
```
chrishunt: `decorator` is included as part of `appsignal`, but adding explicitly here as well for when we remove `appsignal` as a dependency.

pdobacz: hm, looks like you've already added the `deps` into the "top-level" `mix.exs`. I think it's enough and appropriate if the dependencies are used across the entire umbrella app. Did it not work if the `deps` were missing in `omg_utils/mix.exs`?

apps/omg_utils/lib/omg_utils/metrics.ex
```diff
@@ -33,18 +33,61 @@ defmodule OMG.Utils.Metrics do
   def measure_start(body, context) do
     # NOTE: the namespace and event group naming convention here is tentative.
     # It is possible we'll revert to standard coarser division into `web` and `background` namespaces Appsignal suggests
-    namespace = context.module |> Module.split() |> List.last() |> String.to_existing_atom()
-
-    if Enum.find(@discard_namespace_metrics, &match?(^&1, namespace)),
+    if Enum.find(@discard_namespace_metrics, &match?(^&1, namespace(context))),
       do: body,
-      else: Appsignal.Instrumentation.Decorators.transaction(namespace, body, context)
+      else: start_trace(body, context)
```
pdobacz: hm, I read my old comment and noticed there are some decorators (similar to those of AppSignal) available in `Spandex`. Did they not work? I known that the macros you did might be more inline with the `telemetry` plan, but just curious why not use the decorators of `Spandex`?

apps/omg_utils/mix.exs
chrishunt: When I removed `spandex` from this `mix.exs`, the dialyzer complained (see the circleci build [here](https://circleci.com/gh/omisego/elixir-omg/11307))

```
:0:unknown_function Function Spandex.continue_trace/3 does not exist.
:0:unknown_function Function Spandex.continue_trace_from_span/3 does not exist.
:0:unknown_function Function Spandex.current_context/1 does not exist.
:0:unknown_function Function Spandex.current_span/1 does not exist.
:0:unknown_function Function Spandex.current_span_id/1 does not exist.
:0:unknown_function Function Spandex.current_trace_id/1 does not exist.
:0:unknown_function Function Spandex.distributed_context/2 does not exist.
:0:unknown_function Function Spandex.finish_span/1 does not exist.
:0:unknown_function Function Spandex.finish_trace/1 does not exist.
:0:unknown_function Function Spandex.inject_context/3 does not exist.
:0:unknown_function Function Spandex.span_error/3 does not exist.
:0:unknown_function Function Spandex.start_span/2 does not exist.
:0:unknown_function Function Spandex.start_trace/2 does not exist.
:0:unknown_function Function Spandex.update_span/1 does not exist.
:0:unknown_function Function Spandex.update_top_span/1 does not exist.
:0:unknown_type Unknown type: Spandex.Tracer.opts/0.
```

It seemed like maybe spandex wasn't being included? When I re-added the deps here it fixed the dialyzer. It also appeared that we included `appsignal` in almost every single `mix.exs`, so I thought maybe it was normal to include something more than once.

apps/omg_utils/lib/omg_utils/metrics.ex
chrishunt: The decorators included with spandex produced a massive amount of `unmatched_return` dialyzer warnings (see build [here](https://circleci.com/gh/omisego/elixir-omg/11069?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)). 

Instead of ignore the warnings, it seemed like maybe a good idea to just write our own decorators since we're already doing that here to filter discarded namespace metrics. I copied the spandex decorator implementation (see it [here](https://github.com/spandex-project/spandex/blob/master/lib/decorators.ex)), but fixed the `unmatched_return` for our dialyzer warnings. As you mentioned, this also seemed useful since we want to put this in telemetry eventually.

apps/omg_utils/mix.exs
pdobacz: hm, when you mention that I now vaguely recall that there was something strange that caused us to include `appsignal` everywhere too, possibly dialyzer.

Normally it should suffice to include on the top level, but some deps have their quirks...

apps/omg/config/config.exs
```diff
@@ -13,4 +13,9 @@ config :omg, :eip_712_domain,
   version: "1",
   salt: "fad5c7f626d80f9256ef01929f3beb96e058b8b4b0e3fe52d84f054c0e2a7a83"
 
+config :omg, OMG.Utils.Tracer,
+  service: :omg,
+  adapter: SpandexDatadog.Adapter,
+  env: Atom.to_string(Mix.env())
```
achiurizo: Should this be using the new [`APP_ENV`](https://github.com/omisego/devops/pull/103) value? Following up from a conversation with Ino, [MIX_ENV will be deprecated and set to prod](https://github.com/omisego/devops/pull/101#discussion_r294955427)

Same with the other configurations using `MIX.env()` in this PR.

apps/omg/config/config.exs
```diff
@@ -14,4 +14,9 @@ config :omg, :eip_712_domain,
   version: "1",
   salt: "fad5c7f626d80f9256ef01929f3beb96e058b8b4b0e3fe52d84f054c0e2a7a83"
 
+config :omg, OMG.Utils.Tracer,
+  service: :omg,
+  adapter: SpandexDatadog.Adapter,
+  env: System.get_env("APP_ENV")
```
unnawut: The config files are evaluated at compile time so it will not be reflected for compiled code.

Please consider [`deferred_config`](https://github.com/mrluc/deferred_config) e.g. as used in https://github.com/omisego/elixir-omg/blob/master/apps/omg_child_chain_rpc/config/dev.exs#L11-L12

chrishunt: Great, thank you very much. I'll make that change. 👍

apps/omg/lib/omg/supervisor.ex
InoMurko: hmm... maybe omg_status isn't a bad place for this!

apps/omg_utils/mix.exs
```diff
@@ -25,5 +25,13 @@ defmodule Utils.MixProject do
   defp elixirc_paths(:prod), do: ["lib"]
   defp elixirc_paths(_), do: ["lib", "test/support"]
 
-  defp deps, do: [{:statix, "~> 1.1"}, {:appsignal, "~> 1.9"}]
+  defp deps do
```
InoMurko: maybe we should try keeping omg_utils thin and use omg_status for these things.

chrishunt: I added these dependencies here because metrics currently live at `OMG.Utils.Metrics` 

After reading the `OMG.Status` README, however, it does seem like that's a better place for it. 😃 

> Status is a umbrella application. It's purpose is to gather and send metrics

Are you proposing that we move `OMG.Utils.Metrics` to `OMG.Status.Metrics` ? If so, would it make sense to do that in another PR since it will be quite noisy with all the `use OMG.Utils.Metrics` macros throughout the code?

InoMurko: Yes. That should work (offloading to a separate pr) if it’s indeed part of this assignment/ticket.

InoMurko:
Is this ready for review and consequently merge?

chrishunt:
> @InoMurko  Is this ready for review and consequently merge?

It is now! Thanks for taking a look. Just cleaned up some conflicts and squashed.

apps/omg_utils/lib/omg_utils/metrics.ex
```diff
@@ -40,18 +40,61 @@ defmodule OMG.Utils.Metrics do
...
-    event_group = context.module |> Module.split() |> List.last() |> String.to_existing_atom()
-
-    if Enum.find(@discard_namespace_metrics, &match?(^&1, event_group)),
+    if Enum.find(@discard_namespace_metrics, &match?(^&1, namespace(context))),
       do: body,
-      else: Appsignal.Instrumentation.Decorators.transaction_event(event_group, body, context)
+      else: start_span(body, context)
+  end
+
+  defp start_trace(body, context) do
```
InoMurko: hm... is this the recommended way of building traces? why do we need to write these macros ourselves?

apps/omg_utils/mix.exs
```diff
@@ -25,5 +25,13 @@ defmodule Utils.MixProject do
...
+    [
+      {:appsignal, "~> 1.9"},
+      {:statix, "~> 1.1"},
+      {:spandex, "~> 2.4"},
+      {:spandex_datadog, "~> 0.4"},
```
InoMurko: I see `:spandex` has Plugs https://github.com/spandex-project/spandex#phoenix-plugs
what are we getting from `:spandex_datadog`?

PR #826, opened 2019-07-08
fix: statix not pickedup as dependency app in a release and wss env vars

InoMurko:
## Overview

Statix was not picked up in a dependecy tree when building a release.
Watcher image missing WS env. variables

## Changes

- ...

### DIFF
--- a/apps/omg_status/lib/omg_status/application.ex
+++ b/apps/omg_status/lib/omg_status/application.ex
@@ -31,6 +31,14 @@ defmodule OMG.Status.Application do
         []
       end
 
+    # if there's no SENTRY_DSN (local Watcher for example)
+    # don't attach Sentry as backend logger to report issues to
+    _ =
+      case System.get_env("SENTRY_DSN") do
+        nil -> :ok
+        _ -> {:ok, _} = Logger.add_backend(Sentry.LoggerBackend)
+      end
+
     Supervisor.start_link(children, strategy: :one_for_one, name: Status.Supervisor)
   end
 
--- a/docker-compose-watcher-non-mac.yml
+++ b/docker-compose-watcher-non-mac.yml
@@ -5,6 +5,7 @@ services:
     command: "full_local"
     environment:
       - ETHEREUM_RPC_URL=https://rinkeby.infura.io/v3/<your_api_key>
+      - ETHEREUM_WS_RPC_URL=wss://mainnet.infura.io/ws/v3/<your_api_key>
       - CHILD_CHAIN_URL=http://samrong.omg.network
       - ETHEREUM_NETWORK=RINKEBY
       - RINKEBY_CONTRACT_ADDRESS=0x740ecec4c0ee99c285945de8b44e9f5bfb71eea7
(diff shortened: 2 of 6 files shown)


apps/omg_status/lib/omg_status/application.ex
```diff
@@ -31,6 +31,14 @@ defmodule OMG.Status.Application do
...
       end
 
+    # if there's no SENTRY_DSN (local Watcher for example)
+    # don't attach Sentry as backend logger to report issues to
+    _ =
+      case System.get_env("SENTRY_DSN") do
+        nil -> :ok
+        _ -> {:ok, _} = Logger.add_backend(Sentry.LoggerBackend)
+      end
+
```
achiurizo: 🙏 ❤️  Thanks for the fix!

Also, is this the best place for us to load this? Why not [omg/application.ex](https://github.com/omisego/elixir-omg/blob/master/apps/omg/lib/omg/application.ex)?

docs/run_local_watcher.md
```diff
@@ -14,7 +14,7 @@ The `docker-compose` tooling in the root of `elixir-omg` allows users to run the
 
 2) From the root of the `elixir-omg` execute:
 
-- `docker-compose -f docker-compose.yml -f docker-compose-watcher-mac.yml up` (Mac)
-- `docker-compose -f docker-compose.yml -f docker-compose-watcher-non-mac.yml up` (Linux/Windows)
+- `docker-compose -f docker-compose.yml -f docker-compose-watcher-mac.yml up watcher postgres` (Mac)
+- `docker-compose -f docker-compose.yml -f docker-compose-watcher-non-mac.yml up watcher postgres` (Linux/Windows)
```
achiurizo: Is this just so that we don't load `elixir-omg` service?

InoMurko: yeah!

apps/omg_status/lib/omg_status/application.ex
InoMurko: the way umbrella applications are started is that the VM figures out what the dependency hierarchy is. Omg status is more or less at the top because it doesn't have any dependencies but it's a dependency to many - so it'll start as one of the first. And if you're adding a logger backend - you would want to add it as soon as possible so that you're able to catch exits that happen while applications are starting.

docker-compose-watcher-non-mac.yml
```diff
@@ -5,6 +5,7 @@ services:
     command: "full_local"
     environment:
       - ETHEREUM_RPC_URL=https://rinkeby.infura.io/v3/<your_api_key>
+      - ETHEREUM_WS_RPC_URL=wss://mainnet.infura.io/ws/v3/<your_api_key>
```
InoMurko: mainnet! oooups!

apps/omg_status/lib/omg_status/application.ex
InoMurko: there's another one... the macro endpoint hooks work independently of logger backend
```
use Sentry.Phoenix.Endpoint
```
need to check how to enable/disable these

jrhite:
LGTM

PR #828, opened 2019-07-08
Make coverage calculation encompass the entire umbrella app

### LINKED ISSUE #806: Fix test coverage (and/or) coverage calculation
Due to the way coverage is calculated in ExCoveralls/cover, we'll see drops when doing changes that don't affect coverage really. 

The trick is that if a tests covering a line in code of umbrella subapp A is put in tests of umbrella subapp B, coveralls will not "see" it.

For example, by splitting out `omg_watcher_rpc` in https://github.com/omisego/elixir-omg/pull/757, the functionality in `omg_watcher` that was tested via controller tests only appears to be not covered (tests live in `omg_watcher_rpc` not `omg_watcher`).

(The converse also holds, so lines which get covered in `omg_watcher_rpc` by tests living elsewhere, appear as not covered.)

Here is an issue that explores the problem: https://github.com/parroty/excoveralls/issues/23, with some follow up links. There's some workarounds proposed there.

Let's have this issue to either:
1/ alter the behavior of coverage calculation, so that the coverage takes all tests into account for the whole umbrella app
2/ rearrange (and/or) add tests to cover the individual subapps

I'm slightly in favor of (1/), if it doesn't end up being a rabbit hole/hack to pull off. I prefer convenience and thinness of tests to them being so stricly "bound" to the code they test. There's some threads in the issue linked above that indicate that there's consensus that the approach of (1/) is non-standard but acceptable

pdobacz:
Fixes #806, a proposition of a fix

TODO:
- [x] see why the `ecto.migrate` is needed now, and wasn't before
- [x] groom commit history to just a single commit
- [x] post something in the excoveralls issue

## Overview

`mix coveralls` via `mix test --cover` always calculates test coverage of an umbrella sub-app only by using that sub-apps own tests.

Since we have some code split across sub-apps (e.g. the `_rpc` bunch), some code only gets covered in foreign sub-apps of the umbrella.

C.f. this issue https://github.com/parroty/excoveralls/issues/23

## Changes

`excoveralls` has been updated to a derivative (rebase) of one of the forks proposed in the issue above.
Some adjustments were needed on our side to have both the (new) `mix coveralls --umbrella` and the usual `mix test` invocations work.

## Testing

existing tests, and an increase in test coverage as demonstrated by [coveralls.io](https://coveralls.io/github/omisego/elixir-omg?branch=806/fix_coverage3)


apps/omg_watcher/test/test_helper.exs
```diff
@@ -12,9 +12,11 @@
...
 
-ExUnit.configure(exclude: [integration: true])
+ExUnit.configure(exclude: [integration: true, property: true, wrappers: true])
 ExUnitFixtures.start()
 ExUnit.start()
 
 {:ok, _} = Application.ensure_all_started(:httpoison)
 {:ok, _} = Application.ensure_all_started(:fake_server)
+
+Mix.Task.run("ecto.migrate", ~w(--quiet))
```
InoMurko: why was this added now and wasn't needed beforehand?

pdobacz: bad as it may sound, I have not the faintest idea. Only thing I know is that I have seen other people including such migrations in their `test_helpers.exs` or in their manual processes, when I searched for the issue.

The issue was that sometimes (that is, unless one has done `MIX_ENV=test ecto.migrate` manually), tests would complain about things like tables not being there in the db. 

At first I thought that it makes sense for it to be there and I can just yolo the fact that it worked without before. On 2nd thought this deserves proper investigation...

pnowosie: I suspect I didn't work after #562, but nobody really drop tables from test database... but now nobody could be sure ;)

pdobacz: that's entirely possible, but how would these migrated test dbs emerge in the 1st place?

pdobacz: hm, I fiddled around with this, and it seems (locally, at least) this https://github.com/omisego/elixir-omg/pull/828/commits/a513ec255e9d1e9995834108a464f1ce12e0740f is also a viable alternative.

Weirdly enough, now, if I don't apply neither of these fixes, it complains about sth different:

```
     ** (EXIT from #PID<0.4439.0>) shutdown: failed to start child: OMG.Watcher.DB.Repo
         ** (EXIT) shutdown: failed to start child: DBConnection.Ownership.Manager
                 ** (MatchError) no match of right hand side value: {:error, {:noproc, {GenServer, :call, [DBConnection.Watcher, {:watch, DBConnection.ConnectionPool.Supervisor, {DBConnection.ConnectionPool.Pool, {#PID<0.4457.0>, #Reference<0.1063356787.2548695042.105969>, Postgrex.Protocol, [types: Postgrex.DefaultTypes, port: 5432, repo: OMG.Watcher.DB.Repo, telemetry_prefix: [:omg, :watcher, :db, :repo], otp_app: :omg_watcher, pool_size: 10, adapter: Ecto.Adapters.Postgres, timeout: 60000, connect_timeout: 60000, ownership_timeout: 180000, username: "omisego_dev", password: "omisego_dev", database: "omisego_test", hostname: "localhost"]}}}, :infinity]}}}
                     lib/db_connection/ownership/manager.ex:60: DBConnection.Ownership.Manager.init/1
                     (stdlib) gen_server.erl:374: :gen_server.init_it/2
                     (stdlib) gen_server.erl:342: :gen_server.init_it/6
                     (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
```

I'll poke around some more and come up with the least frightening option to distill a reviewable PR finally.

pdobacz: :man_facepalming: okay, so the `migrate` has always been there https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher/mix.exs#L56. It just isn't executed when doing the modified `coveralls --umbrella` task. Brawo ja.

Issue #832, opened 2019-07-08
use Integer instead of Float for TPS


achiurizo:
cc @jbunce who flagged this.

We should change our TPS to use int instead of floats: https://github.com/omisego/elixir-omg/blob/master/docs/perf_test_result_dumps.md

InoMurko:
done

PR #833, opened 2019-07-08
fix: use Kernel round instead of Float round

InoMurko:
https://github.com/omisego/elixir-omg/issues/832

## Overview

Kernel round instead of Float

## Testing
Standard testing.


InoMurko:
I have no idea. It seemed less effort to just do it instead of figuring out why - but @jbunce can give some insight why he thinks it matters.

jbunce:
Because you can't have part of a transaction. It either happened or it didn't.

pdobacz:
For the record, I think this:

> you can't have part of a transaction. It either happened or it didn't.

has nothing to do with the TPS measure being integer or not. `"tps":3937.23` doesn't imply you have `0.23` of a transaction.

PR #834, opened 2019-07-10
fix: include Sentry on condition and remove unnecessary logging

InoMurko:
Sentry issue running on local watcher.
## Overview

Adjust reporting on Sentry issues, logs on compilation and in runtime.

## Changes
- left some comment in the code already 
## Testing

Tests should continue working.

### DIFF
--- a/apps/omg_child_chain_rpc/lib/omg_child_chain_rpc/application.ex
+++ b/apps/omg_child_chain_rpc/lib/omg_child_chain_rpc/application.ex
@@ -13,9 +13,10 @@
 # limitations under the License.
 
 defmodule OMG.ChildChainRPC.Application do
-  # See https://hexdocs.pm/elixir/Application.html
-  # for more information on OTP Applications
   @moduledoc false
+
+  alias OMG.ChildChainRPC.Plugs.Health
+  alias OMG.ChildChainRPC.Web.Endpoint
 
   use Application
   require Logger
@@ -28,9 +29,11 @@ defmodule OMG.ChildChainRPC.Application do
     opts = [strategy: :one_for_one, name: OMG.ChildChainRPC.Supervisor]
 
     children = [
-      {OMG.ChildChainRPC.Plugs.Health, []},
-      {OMG.ChildChainRPC.Web.Endpoint, []}
+      {Health, []},
+      {Endpoint, []}
     ]
+
+    _ = Logger.warn("Is Sentry for OMG.ChildChainRPC.Web.Endpoint enabled: #{System.get_env("SENTRY_DSN") != nil}")
 
     Supervisor.start_link(children, opts)
   end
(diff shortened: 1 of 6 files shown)


apps/omg_child_chain_rpc/lib/omg_child_chain_rpc/application.ex
```diff
@@ -28,10 +29,12 @@ defmodule OMG.ChildChainRPC.Application do
     opts = [strategy: :one_for_one, name: OMG.ChildChainRPC.Supervisor]
 
     children = [
-      {OMG.ChildChainRPC.Plugs.Health, []},
-      {OMG.ChildChainRPC.Web.Endpoint, []}
+      {Health, []},
+      {Endpoint, []}
```
InoMurko: credo complaining

apps/omg_watcher_rpc/lib/web/endpoint.ex
```diff
@@ -15,7 +15,38 @@
 defmodule OMG.WatcherRPC.Web.Endpoint do
   use Phoenix.Endpoint, otp_app: :omg_watcher_rpc
   use Appsignal.Phoenix
-  use Sentry.Phoenix.Endpoint
+
+  ### we manually override call using https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/phoenix_endpoint.ex
+  ### because we don't want to decide if we want to use Sentry or not at compile time
+  def call(conn, opts) do
```
pnowosie: I'd prefer to take a different way to disable Setry on runtime. How about Sentry's `environment` as shown [here](https://github.com/getsentry/sentry-elixir/blob/1ccd934ae89a7baea7047a9747760d52a38c6b4a/lib/sentry.ex#L44)  
So we can set it by default that disables it and change with ENV variable, on env we want it.

InoMurko: That does not solve the problem of people running watcher from our build.

InoMurko: the container we publish - production watcher - that anyone can run is effectively tagged as _production_

InoMurko: also, the way deployments + configuration work
```
environment_name: System.get_env("APP_ENV"),
included_environments: [:dev, :prod, System.get_env("APP_ENV")],
```

pnowosie: I might be missing many details about how the realeases & containers work under the hood, so the author patience is required 😅

Once container is published, before we can run it we need to pass sentry’s DSN, right.
Let included_environments be [:blue_sky], and Sentry’s environment name points to env variable named Sentry_env.
When we pass along with DSN also Sentry_env=blue_sky, Sentry will know it’s ready to go.

Is it posible? As they put in ☝️ file, sentry’s environment_name isn’t limited to test | dev | prod.

PR #835, opened 2019-07-10
fix: configuration for Sentry

InoMurko:
Sentry issue running on local watcher.
## Overview

Adjust reporting on Sentry issues, logs on compilation and in runtime.

## Changes
- adjust included environments
## Testing

Tests should continue working.


apps/omg_status/lib/omg_status/application.ex
```diff
@@ -56,4 +50,14 @@ defmodule OMG.Status.Application do
       _ -> nil
     end
   end
+
+  defp configure_sentry do
+    app_env = System.get_env("APP_ENV")
+    sentry_dsn = System.get_env("SENTRY_DSN")
+
+    case {is_binary(app_env), is_binary(sentry_dsn)} do
+      {true, true} -> Application.put_env(:sentry, :included_environments, [sentry_dsn], persistent: true)
```
pnowosie: Shouldn't it be :question: 
```suggestion
      {true, true} -> Application.put_env(:sentry, :included_environments, [app_env], persistent: true)
```

PR #838, opened 2019-07-11
[Chore] Remove specdiffs from Dialyzer config

chrishunt:
## Overview

It sounds like we are unanimous on removing specdiffs, so this PR makes that happen. Related lines are also removed from `dialyzer.ignore-warnings`.

## Testing

```
$ mix dialyzer
```


InoMurko:
Could you also cleanup the dialyzer ignore file?

chrishunt:
> @InoMurko Could you also cleanup the dialyzer ignore file?

I think I already did that! Do you see anything else that should be removed?

InoMurko:
Oh sry! You’re right!

PR #840, opened 2019-07-12
dump thoughts on ALD in `elixir-omg` in spots around the codebase

pdobacz:
do not merge!

Abstract Layer Design comments/intro to `elixir-omg`

## Overview

I've dumped my current thoughts on places where ALD will impact `elixir-omg`. Not a comprehensive list yet, but anyway...

(not exactly the best idea, but I put those thoughts as comments into the code, prefixed with `# NOTE: ALD:`. I thought it would be useful, but now I see not necessarily so. It's throwaway copy so meh)

I put some very first drafty ideas on how the abstractness might be embraced, but that's definitely not super-thought through yet.

## Changes

N/A, throwaway comments

## Testing

N/A


pdobacz:
whoever might stumble onto this: this might be slightly outdated, as we're moving forward with some PoC ALD-refactors already. Will close when it becomes completely obsolete

PR #851, opened 2019-07-15
549 additional application metrics 2

InoMurko:

## Overview

Migration from AppSignal to Datadog with Spandex, Spandex Ecto and Spandex Phoenix with supporting lib VmStats.

## Changes

- Removed AppSignal completely
- Support to exclude :rocksdb as dependency, useful for local builds because it taking ages to compile.
- Swapped custom function annotations provided by Appsignal with SpandexDecorators and added example usage pattern for syncing procedure that happens in EthereumEventListener. Will be adding more as we proceed.
- to ensure telemetry handlers are attached they're part of startup procedure https://hexdocs.pm/elixir/Application.html#c:start_phase/3
- Removed Recorder process and introduced telemetry handlers (`*...Measure` modules - they should be understood as an extension (like Core modules) to the parent process module) for named processes. 
- removed `increment_metrics_counter` functions in controller and will later add them as a Plug.
- moved dependencies that we altered into OmiseGo repo
## Testing

No tests removed or features changed so same testing would apply.


apps/omg/lib/omg/state.ex
```diff
@@ -109,7 +101,15 @@ defmodule OMG.State do
                last_deposit_query_result,
                child_block_interval
              ) do
-        _ = Logger.info("Started State, height: #{height_query_result}, deposit height: #{last_deposit_query_result}")
+        _ =
+          Logger.info(
```
InoMurko: maybe this should be a metric? is it a relevant log entry?

InoMurko:
Some of the things we're able to gather now via APM.
Screen Shot 2019-07-24 at 23 04 02
Screen Shot 2019-07-24 at 23 03 31
Screen Shot 2019-07-24 at 23 02 13
Screen Shot 2019-07-24 at 23 01 57

apps/omg/lib/omg/state.ex
achiurizo: I think this makes sense as a metric :+1:

apps/omg/lib/omg/ethereum_event_listener/measure.ex
```diff
@@ -0,0 +1,69 @@
...
+
+  def handle_event([:trace, _], %{}, state, _config) do
+    # TODO change to compiler flags once we're running releases
+    is_child_chain =
+      Enum.any?(Application.started_applications(), fn
+        {:omg_child_chain, _, _} -> true
+        {_, _version, _desc} -> false
+      end)
+
+    if is_child_chain do
```
chrishunt: This bit is necessary even with the two different tracers? (`OMG.ChildChainRPC.Tracer` and `OMG.Watcher.Tracer`)

apps/omg/lib/omg/ethereum_event_listener/measure.ex
```diff
@@ -0,0 +1,69 @@
...
+      service = service(:child_chain, state.service_name)
+      Tracer.update_top_span(service: service, tags: [])
+    else
+      service = service(:watcher, state.service_name)
+      Tracer.update_top_span(service: service, tags: [])
+    end
+  end
+
+  defp service(app, service_name) do
+    :erlang.binary_to_atom(
```
chrishunt: We could possibly lean on some of the Elixir magic here:

```elixir
String.to_atom("#{app}_#{service_name}")
```

Or with some safety against creating random atoms:

```elixir
String.to_existing_atom("#{app}_#{service_name}")
```

apps/omg/lib/omg/ethereum_event_listener/measure.ex
```diff
@@ -0,0 +1,69 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.EthereumEventListener.Measure do
+  @moduledoc """
+  Counting business metrics sent to DataDog
```
chrishunt: Not sure why I'm noticed this or am pointing it out, but Datadog has a lowercase 'd' in dog. I think this comes from working at GitHub and seeing people write it Github all the time. 😆

apps/omg/lib/omg/root_chain_coordinator.ex
```diff
@@ -67,7 +66,7 @@ defmodule OMG.RootChainCoordinator do
   @doc """
   Gets all the current synced height for all the services checked in
   """
-  @decorate measure_event()
+
```
chrishunt: Extra line?

apps/omg/lib/omg/state.ex
```diff
@@ -206,7 +201,7 @@ defmodule OMG.State do
 
   Does its on persistence!
   """
-  @decorate measure_event()
+
```
chrishunt: Extra line?

apps/omg_db/lib/omg_db/leveldb/server.ex
```diff
@@ -50,17 +49,14 @@ defmodule OMG.DB.LevelDB.Server do
   def init(db_path: db_path, name: name) do
     # needed so that terminate callback is called on normal close
     Process.flag(:trap_exit, true)
-    table = create_stats_table(name)
+    ^name = create_stats_table(name)
```
chrishunt: Nice, today I learned about the pin operator, thanks! 😃

apps/omg_db/test/fixtures.exs
```diff
@@ -24,7 +24,8 @@ defmodule OMG.DB.Fixtures do
     db_path = Briefly.create!(directory: true)
     Application.put_env(:omg_db, :path, db_path, persistent: true)
 
-    :ok = OMG.DB.init()
+    _ = OMG.DB.init()
```
chrishunt: Don't care about `:ok` anymore?

apps/omg_status/lib/omg_status/metric/datadog.ex
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.Status.Metric.Datadog do
+  @moduledoc """
+  Datadog connection wrapper
+
```
chrishunt: Extra line?

apps/omg_status/lib/omg_status/metric/event.ex
```diff
@@ -0,0 +1,116 @@
...
+    :depositor,
+    :in_flight_exit,
+    :piggyback,
+    :exiter,
+    :convenience_deposit_processor,
+    :exit_processor,
+    :convenience_exit_processor,
+    :exit_finalizer,
+    :exit_challenger,
+    :in_flight_exit_processor,
```
chrishunt: Can we please sort these alphabetically?

apps/omg_watcher/lib/omg_watcher/api/transaction.ex
```diff
@@ -32,7 +32,7 @@ defmodule OMG.Watcher.API.Transaction do
   @doc """
   Retrieves a specific transaction by id
   """
-  @decorate measure_event()
+
```
chrishunt: Extra line? (I will stop commenting on these now, sorry)

apps/omg_watcher/lib/omg_watcher/utxo_exit/core.ex
```diff
@@ -21,10 +21,9 @@ defmodule OMG.Watcher.UtxoExit.Core do
   alias OMG.State.Transaction
   alias OMG.Utxo
   alias OMG.Watcher.DB
-  use OMG.Utils.Metrics
+  use Spandex.Decorators
```
chrishunt: Do we use these decorators in this file anywhere?

apps/omg_child_chain/lib/child_chain.ex
```diff
@@ -27,11 +27,10 @@ defmodule OMG.ChildChain do
   alias OMG.State
   alias OMG.State.Transaction
   use OMG.Utils.LoggerExt
-  use OMG.Utils.Metrics
+  use Spandex.Decorators
```
chrishunt: I commented once on this before (in another file), but are we decorating anything in this module? Does this need to be here?

apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
```diff
@@ -114,12 +113,22 @@ defmodule OMG.ChildChain.BlockQueue do
...
-      {:ok, _} = Recorder.start_link(%Recorder{name: __MODULE__.Recorder, parent: self()})
+      {:ok, _} =
+        :timer.send_interval(
+          Application.fetch_env!(:omg_child_chain, :metrics_collection_interval),
+          self(),
+          :send_metrics
+        )
 
-      _ = Logger.info("Started BlockQueue")
+      _ = Logger.info("Started #{inspect(__MODULE__)}")
```
chrishunt: Nice 👍

apps/omg/lib/omg/ethereum_event_listener/measure.ex
InoMurko: nice! I admit I didn't check!

apps/omg_child_chain/lib/child_chain.ex
InoMurko: nop! leftover!

apps/omg_watcher/lib/omg_watcher/api/transaction.ex
InoMurko: It appears like there's an extra line, but it's actually not. I think it has to do with GH web formatting of diffs?

apps/omg/lib/omg/ethereum_event_listener/measure.ex
InoMurko: Ha, yes, I've spend quite a bit of time here. You can specify a tracer by setting an argument to a decorator to a `span` or a `trace`.
`@decorate span(tracer: MyApp.OtherTracer)`
`@decorate trace(tracer: MyApp.OtherTracer)`
Macros in general get evaluated at compilation and `ethereum_event_listener` is being used by both Watcher and Child Chain. If we were using releases, we could do a pre-compile check that would allow us to figure out if we're doing a Watcher compilation or a Child Chain compilation by doing something like:
```
if Code.ensure_loaded?(OMG.Watcher) do
@decorate span(tracer: OMG.Watcher.Tracer)
else
@decorate span(tracer: OMG.ChildChain.Tracer)
end
```
and this would produce a Macro wrapper that injects code depending on whether a module is present at compilation. 

Currently, because we use the `launcher.py` script and `mix run` to launch the app this isn't possible, because erlang's VM always loads ALL modules into the VM.

apps/omg/lib/omg/ethereum_event_listener/measure.ex
InoMurko: Oh jeez. what was I thinking? TNX

apps/omg/lib/omg/ethereum_event_listener.ex
```diff
@@ -111,34 +108,36 @@ defmodule OMG.EthereumEventListener do
...
 
     {new_state, callbacks}
   end
 
+  @decorate span(service: :ethereum_event_listener, type: :backend, name: "maybe_update_event_cache/2")
   defp maybe_update_event_cache({:get_events, {from, to}, state_with_cache}, get_ethereum_events_callback) do
     {time, {:ok, new_events}} = :timer.tc(fn -> get_ethereum_events_callback.(from, to) end)
     time = round(time / 1000)
-
+    # TODO can we deprecate these logs?
```
pdobacz: I think yes, let's remove it!

apps/omg/lib/omg/state.ex
pdobacz: ok with adding a metric, but please keep the log too, I use it quite often and I'd not like to rely on a metric only.

apps/omg_db/test/fixtures.exs
pdobacz: Good catch! I'd say we should care. Leftover change perhaps?

apps/omg/lib/omg/ethereum_event_listener/core.ex
```diff
@@ -138,6 +144,7 @@ defmodule OMG.EthereumEventListener.Core do
 
     height_to_check_in = get_height_to_check_in(new_state)
     db_update = [{:put, update_key, height_to_check_in}]
+    :ok = :telemetry.execute([:process, __MODULE__], %{events: events}, state)
```
pdobacz: can we call this outside of `.Core` so that it remains side-effect free?

apps/omg/lib/omg/state.ex
```diff
@@ -39,7 +38,6 @@ defmodule OMG.State do
     GenServer.start_link(__MODULE__, :ok, name: __MODULE__)
   end
 
-  @decorate measure_event()
```
pdobacz: sometimes you replace these with `@decorate span` and sometimes you just remove, why?

apps/omg_db/lib/omg_db/leveldb/core.ex
```diff
@@ -18,6 +18,7 @@ defmodule OMG.DB.LevelDB.Core do
   """
 
   # adapter - testable, if we really really want to
+  use Spandex.Decorators
```
pdobacz: is this addition necessary? there's no other changes in the file

apps/omg/lib/omg/state.ex
InoMurko: because they need to be replace - but not so easily. Appsignal traces don't work the same as spandex ones. So I thought I'd do them in a later PR.

apps/omg/lib/omg/ethereum_event_listener/core.ex
InoMurko: hm... what causes the side effect?

pdobacz: Isn't talking to `:telemetry` one? 

Sorry if I'm having false assumptions here, but I'm guessing this talks to some dependency to do sth. `.Core.xyz` functions are written with the intent of remaining pure. 

Unless absolutely necessary we avoid doing calling external stuff in `.Core.xyz`, I think the only exception was made for `_ = Logger...` calls, but only in a few cases.

`events` and `state` are available outside, so the move seems to be an easy one.

InoMurko: This event gets forwarded to the dedicated Ethereum event listener handler located in module `OMG.EthereumEventListener.Measure`

I don't quite understand what you mean by _side effect_ though indeed the change isn't big, but would like to understand where `keeping .Core.xyz` pure comes from.

InoMurko:
A previous commit removes logging and replaces it with datadog trace
https://github.com/omisego/elixir-omg/pull/851/commits/dbcd522ff59a9d53e501f823b2ea1163c8d190ca#diff-5fcd99569a2ce28c62745940476ced06R233

looks liek this:

Screen Shot 2019-07-25 at 16 30 08

apps/omg/lib/omg/ethereum_event_listener/core.ex
pdobacz: > would like to understand where keeping .Core.xyz pure comes from.

Sure thing! It comes from the approach adopted, so segregate logic into such `Core` modules, with no side effects/side causes. These modules would hold pure functions (same output for same input etc.). Then they can be robustly/quickly tested in unit tests and also I think they are easier to reason about, in general, being just these reliable chunks of logic.

PR #859, opened 2019-07-16
rename statsd env to DD_HOSTNAME/DD_PORT

achiurizo:
:clipboard: Conversation with devops
:clipboard: https://github.com/omisego/devops/pull/116

## Overview

Changing our ENVVAR to instead to `DD`. The concern was that by naming it explicitly, this will help prevent confusion in the future in what this is pointing to for us.

This will still work with regular `statsd`, Open to suggestions.

## Changes

Changes `STATSD_HOST` -> `DD_HOSTNAME`
Changes `STATSD_PORT` -> `DD_PORT`

## Testing

N/A


config/config.exs
```diff
@@ -33,8 +33,8 @@ config :sentry,
   }
 
 config :statix,
-  host: {:system, "STATSD_HOST", "localhost"},
-  port: {:system, "STATSD_PORT", 8125, {String, :to_integer}}
+  host: {:system, "DOGSTATSD_HOST", "localhost"},
```
InoMurko: what if we kept the datadog standard? `DD_HOSTNAME`  https://docs.datadoghq.com/agent/docker/?tab=standard#global-options

Issue #634, opened 2019-04-19
Improve `mix test` experience: speed up and silence monitor logs


pdobacz:
TL;DR from slack:

Some `mix test` tests are slowish/loudish:

```
OMG.EthereumClientMonitorTest
2019-04-19 10:35:00.097 [warn] module=OMG.EthereumClientMonitor function=handle_event/2 ⋅Health check raised :ethereum_client_connection alarm.⋅
2019-04-19 10:35:00.097 [warn] module=OMG.EthereumClientMonitor function=handle_event/2 ⋅Health check established connection to the client. :ethereum_client_connection alarm clearead.⋅
  * test that alarm gets raised if there's no ethereum client running and cleared when it's running2019-04-19 10:35:00.597 [warn] module=OMG.EthereumClientMonitor function=handle_event/2 ⋅Health check raised :ethereum_client_connection alarm.⋅
2019-04-19 10:35:01.099 [warn] module=OMG.EthereumClientMonitor function=handle_event/2 ⋅Health check established connection to the client. :ethereum_client_connection alarm clearead.⋅
  * test that alarm gets raised if there's no ethereum client running and cleared when it's running (1009.6ms)
  * test that we don't overflow the message queue with timers when Eth client needs time to respond (4993.8ms)
  * test that alarm gets raised if there's no ethereum client running2019-04-19 10:35:06.601 [warn] module=OMG.EthereumClientMonitor function=handle_event/2 ⋅Health check raised :ethereum_client_connection alarm.⋅
```

1. let's speed up the 3 slowest monitor tests (5s, 1s, 500ms), proposition: tweak the waiting/interval times by configuring to lower values in `:test` env's config
    - those aren't `:integration` tests (`geth` is mocked out) so should be fast and don't belong with such tag
    - it's good to keep them in the rapid `mix test` run if possible
2. let's **selectively** silence the `warn` logs coming out of `EthereumClientMontiror`, in `:test` env only, so that the output is clean on passing tests.
    - this improves the experience and makes us focus on anything that gets outputed during tests
    - we don't want to elevate these logs to `info`, as they're signalling something's off but recoverable, so `warn` is justified

pdobacz:
UPDATE:

for `mix test --only integration` I'm also getting:

```
2019-04-19 14:01:50.105 [warn] module=OMG.RPC.Plugs.Health function=handle_event/2 ⋅Alarm boot_in_progress was raised.⋅
```

[...]

PR #860, opened 2019-07-17
chore: speed up and silence monitor logs

pthomalla:
:clipboard: #634 

## Overview
Speed up and fix the slowest monitor tests

## Changes
add helper function call_while, 
in ethereum_client_monitor_test.exs in steup_all start alsow Phoenix.PubSub.PG2

## Testing
simple mix test
to check that ther is less warning in apps/omg/test/omg/ethereum_client_monitor_test.exs :
 -comment @moduletag :capture_log
 -increse sleep time:
   - change call_while arity to 4 (`import OMG.TestHelper, only: [call_while: 4]`)
   - in call function call_while ad two argument time and sleep time (`call_while(&Alarm.all/0,,[], 5_000, 100)`)

### DIFF
--- a/apps/omg/test/omg/ethereum_client_monitor_test.exs
+++ b/apps/omg/test/omg/ethereum_client_monitor_test.exs
@@ -24,9 +24,21 @@ defmodule OMG.EthereumClientMonitorTest do
   alias OMG.Alert.AlarmHandler
   alias OMG.EthereumClientMonitor
 
+  import OMG.TestHelper, only: [call_while: 2]
+
+  @moduletag :capture_log
+
   setup_all do
+    _ = Application.put_env(:omg_child_chain, :eth_integration_module, EthereumClientMock)
     :ok = AlarmHandler.install()
     {:ok, _} = EthereumClientMock.start_link()
+
+    # OMG.EthereumClientMonitor use OMG.InternalEventBus
+    {:ok, _} =
+      Supervisor.start_link(
+        [{Phoenix.PubSub.PG2, [name: OMG.InternalEventBus]}],
+        strategy: :one_for_one
+      )
 
     on_exit(fn ->
       Application.put_env(:omg_child_chain, :eth_integration_module, nil)
@@ -34,7 +46,6 @@ defmodule OMG.EthereumClientMonitorTest do
   end
 
   setup do
-    _ = Application.put_env(:omg_child_chain, :eth_integration_module, EthereumClientMock)
     {:ok, {server_ref, websocket_url}} = WebSockexServerMock.start(self())
     {:ok, ethereum_client_monitor} = EthereumClientMonitor.start_link(alarm_module: Alarm, ws_url: websocket_url)
     Alarm.clear_all()
@@ -55,13 +66,13 @@ defmodule OMG.EthereumClientMonitorTest do
     WebSockexServerMock.shutdown(server_ref)
     true = is_pid(Process.whereis(EthereumClientMonitor))
 
+    node = Node.self()
+
     :ok =
-      pull_client_alarm(400, [
-        %{
-          details: %{node: Node.self(), reporter: EthereumClientMonitor},
-          id: :ethereum_client_connection
-        }
-      ])
+      call_while(
+        &Alarm.all/0,
+        &match?([%{details: %{node: ^node, reporter: EthereumClientMonitor}, id: :ethereum_client_connection}], &1)
+      )
   end
 
   test "that alarm gets raised if there's no ethereum client running and cleared when it's running", %{
@@ -78,19 +89,18 @@ defmodule OMG.EthereumClientMonitorTest do
     WebSockexServerMock.shutdown(server_ref)
 
     true = is_pid(Process.whereis(EthereumClientMonitor))
+    node = Node.self()
 
     :ok =
-      pull_client_alarm(300, [
-        %{
-          details: %{node: Node.self(), reporter: EthereumClientMonitor},
-          id: :ethereum_client_connection
-        }
-      ])
+      call_while(
+        &Alarm.all/0,
+        &match?([%{details: %{node: ^node, reporter: EthereumClientMonitor}, id: :ethereum_client_connection}], &1)
+      )
 
     :sys.replace_state(Process.whereis(EthereumClientMock), fn _ -> %{} end)
 
     {:ok, {_server_ref, ^websocket_url}} = WebSockexServerMock.start(self(), websocket_url)
-    :ok = pull_client_alarm(400, [])
+    :ok = call_while(&Alarm.all/0, &match?([], &1))
   end
 
   test "that we don't overflow the message queue with timers when Eth client needs time to respond", %{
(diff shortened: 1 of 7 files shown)


apps/omg/test/omg/ethereum_client_monitor_test.exs
```diff
@@ -55,13 +66,13 @@ defmodule OMG.EthereumClientMonitorTest do
...
+    node = Node.self()
+
     :ok =
-      pull_client_alarm(400, [
-        %{
-          details: %{node: Node.self(), reporter: EthereumClientMonitor},
-          id: :ethereum_client_connection
-        }
-      ])
+      call_while(
```
InoMurko: why exactly was this rewritten?

apps/omg/test/support/test_helper.ex
```diff
@@ -158,6 +158,23 @@ defmodule OMG.TestHelper do
     {:ok, full_path, file}
   end
 
+  def call_while(caller, matcher, time \\ 5_000, step \\ 10) do
```
InoMurko: this is actually a very unreadable implementation and I think we want to revert to the previous one.

InoMurko: this is actually a very unreadable implementation and I think we want to revert to the previous one.

InoMurko: to elaborate this: `It's harder to read code than to write it.`
and also `Always code as if the person who ends up maintaining your code is a violent psychopath who knows where you live.`

apps/omg/test/omg/ethereum_client_monitor_test.exs
pthomalla: because the previous implementation doesn't work :D
it's best seen here: https://github.com/omisego/elixir-omg/blob/master/apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs#L64-L72

pin value causes that if two objects are equal then it will be a match.
if `TestHelper.rpc_call(:post, "/block.get", %{})` return 
```
%{ 
 "data" => %{"code" => "operation:service_unavailable"},
 "error" ...
}
``` 
then in function pull_client_alarm never match
ps. function call with `%{ "data" => %{"code" => "operation:service_unavailable"}}`

InoMurko: Why would the endpoint return an `error`?

pthomalla: this return rpc_call
```elixir
%{
  "data" => %{
    "code" => "operation:service_unavailable",
    "description" => "The server is not ready to handle the request.",
    "object" => "error"
  },
  "success" => false,
  "version" => "0.2"
}
```
even if return 
```elixir
%{
  "data" => %{
    "code" => "operation:service_unavailable",
  },
  "success" => false,
  "version" => "0.2"
}
```
will not match

InoMurko: Yes you're right! The endpoints are returning a full error description and status. Fixed here: https://github.com/omisego/elixir-omg/pull/861

apps/omg/test/support/test_helper.ex
pthomalla: I know these quotes :D
I did not like the previous one more.
what about this
```elixir 
 def call_while_not_true(caller, matcher, time \\ 5_000, step \\ 10)

 def call_while_not_true(caller, matcher, time, step) when time <= 0 do
    :timeout
  end

  def call_while_not_true(caller, matcher, time, step) do
    if matcher.(caller.()) do
      :ok
    else
      Process.sleep(step)
      call_while_not_true(caller, matcher, time - step, step)
    end
  end
```

InoMurko: I mean, what is the `reason` for the rewrite? It's working and it's very easy to read and it's three weeks old - have the requirements changed? Your implementation is not even faster (because that's what the PR title is saying it's addressing).

pthomalla: `call_until` is better name than 'call_while_not_true',
`pull_client_alarmthis` causing an error that was not visible, I am not sure if it is so clear.

The changes definitely result in faster testing :D after repairing some test. 
not speed up function execution.

apps/omg/test/omg/integration/root_chain_coordinator_test.exs
```diff
@@ -29,6 +29,7 @@ defmodule OMG.RootChainCoordinatorTest do
 
   @moduletag :integration
   @moduletag :common
+  @moduletag :capture_log
```
pdobacz: would it be possible for the non-alarms-related tests not require this `moduletag`? Maybe the root cause of warnings can be (ideally) fixed or if not, the warnings silenced less globally?

pthomalla: I will dig deeper into integrative tests

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
```diff
@@ -26,15 +29,11 @@ defmodule OMG.ChildChainRPC.Plugs.HealthTest do
...
-          "data" => %{
-            "code" => "operation:service_unavailable"
-          }
-        },
-        fn -> TestHelper.rpc_call(:post, "/block.get", %{}) end
-      )
+      :ok =
+        call_while(
+          fn -> TestHelper.rpc_call(:post, "/block.get", %{}) end,
+          &match?(%{"data" => %{"code" => "operation:service_unavailable"}}, &1)
```
pnowosie: referencing this line elsewhere

apps/omg/test/omg/ethereum_client_monitor_test.exs
InoMurko: having fixed in #861 this is not needed and can be reverted.

apps/omg/test/omg/ethereum_client_monitor_test.exs
```diff
@@ -111,46 +121,31 @@ defmodule OMG.EthereumClientMonitorTest do
...
-      pull_client_alarm(100, [
-        %{
-          details: %{node: Node.self(), reporter: EthereumClientMonitor},
-          id: :ethereum_client_connection
-        }
-      ])
+    node = Node.self()
+
+    :ok =
+      call_while(
```
InoMurko: same here, this is an unnecessary rewrite (does not address any issues) and should be reverted

apps/omg_status/test/omg_status/integration/alarms_test.exs
```diff
@@ -43,16 +43,8 @@ defmodule OMG.Status.Alert.AlarmTest do
...
-    # sets memory check limit to 100%
-    # waits for next check so all previous alarm are cleared
-    # sets memory check limit to 1%
-    # see if memsup alarm was raised after next check
-    :memsup.set_sysmem_high_watermark(1)
-    Process.sleep(:memsup.get_check_interval() + 1000)
-    :memsup.set_sysmem_high_watermark(0.01)
-    Process.sleep(:memsup.get_check_interval() + 1000)
+    # memsup set alarm
+    :alarm_handler.set_alarm({:system_memory_high_watermark, []})
```
pthomalla: checked with using redbug what value are sent by memsup when call set_alarm
```
% memsup:set_alarm(system_memory_high_watermark, [])
  proc_lib:init_p_do_apply/3 
  gen_server:handle_msg/6 
  gen_server:try_dispatch/4 
```

PR #861, opened 2019-07-17
fix: matching endpoint return

InoMurko:
Correct maching on endpoint return

## Overview

PR from @pthomalla pointed out that the test wasn't matching on the correct endpoint return value.

## Testing

Same tests apply.

### DIFF
--- a/apps/omg/lib/omg/ethereum_height.ex
+++ b/apps/omg/lib/omg/ethereum_height.ex
@@ -51,7 +51,7 @@ defmodule OMG.EthereumHeight do
     case is_binary(value) do
       true ->
         ethereum_height = Encoding.int_from_hex(value)
-        _ = Logger.info("#{__MODULE__} got a newHeads event for new Ethereum height #{ethereum_height}.")
+        _ = Logger.debug("#{__MODULE__} got a newHeads event for new Ethereum height #{ethereum_height}.")
         {:noreply, ethereum_height}
 
       false ->
--- a/apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
+++ b/apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
@@ -26,15 +26,20 @@ defmodule OMG.ChildChainRPC.Plugs.HealthTest do
       :ok = :alarm_handler.clear_alarm(@alarm_2)
       :ok = :alarm_handler.set_alarm(@alarm_1)
 
-      pull_client_alarm(
-        300,
-        %{
-          "data" => %{
-            "code" => "operation:service_unavailable"
-          }
-        },
-        fn -> TestHelper.rpc_call(:post, "/block.get", %{}) end
-      )
+      :ok =
+        pull_client_alarm(
+          300,
+          %{
+            "data" => %{
+              "code" => "operation:service_unavailable",
+              "description" => "The server is not ready to handle the request.",
+              "object" => "error"
+            },
+            "success" => false,
+            "version" => "0.2"
+          },
+          fn -> TestHelper.rpc_call(:post, "/block.get", %{}) end
+        )
 
       :ok = :alarm_handler.clear_alarm(@alarm_1)
     end
@@ -61,15 +66,20 @@ defmodule OMG.ChildChainRPC.Plugs.HealthTest do
       :ok = :alarm_handler.set_alarm(@alarm_2)
       :ok = :alarm_handler.clear_alarm(@alarm_1)
 
-      pull_client_alarm(
-        300,
-        %{
-          "data" => %{
-            "code" => "operation:service_unavailable"
-          }
-        },
-        fn -> TestHelper.rpc_call(:post, "/block.get", %{}) end
-      )
+      :ok =
+        pull_client_alarm(
+          300,
+          %{
+            "data" => %{
+              "code" => "operation:service_unavailable",
+              "description" => "The server is not ready to handle the request.",
+              "object" => "error"
+            },
+            "success" => false,
+            "version" => "0.2"
+          },
+          fn -> TestHelper.rpc_call(:post, "/block.get", %{}) end
+        )
 
       :ok = :alarm_handler.clear_alarm(@alarm_2)
     end
--- a/apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
+++ b/apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
@@ -35,8 +35,7 @@ defmodule OMG.ChildChainRPC.Plugs.HealthTest do
               "description" => "The server is not ready to handle the request.",
               "object" => "error"
             },
-            "success" => false,
-            "version" => "0.2"
+            "success" => false
           },
           fn -> TestHelper.rpc_call(:post, "/block.get", %{}) end
         )
@@ -75,8 +74,7 @@ defmodule OMG.ChildChainRPC.Plugs.HealthTest do
               "description" => "The server is not ready to handle the request.",
               "object" => "error"
             },
-            "success" => false,
-            "version" => "0.2"
+            "success" => false
           },
           fn -> TestHelper.rpc_call(:post, "/block.get", %{}) end
         )
@@ -103,7 +101,10 @@ defmodule OMG.ChildChainRPC.Plugs.HealthTest do
   defp pull_client_alarm(0, _, _), do: :cant_match
 
   defp pull_client_alarm(n, match, fnn) do
-    case fnn.() do
+    endpoint_call_result = fnn.()
+    match = Map.put_new(match, "version", Map.get(endpoint_call_result, "version"))
+
+    case endpoint_call_result do
       ^match ->
         :ok
 
--- a/apps/omg/lib/omg/ethereum_client_monitor.ex
+++ b/apps/omg/lib/omg/ethereum_client_monitor.ex
@@ -123,7 +123,7 @@ defmodule OMG.EthereumClientMonitor do
     case is_binary(value) do
       true ->
         ethereum_height = Encoding.int_from_hex(value)
-        _ = Logger.info("Ethereum client monitor got a newHeads event for new Ethereum height #{ethereum_height}.")
+        _ = Logger.debug("Ethereum client monitor got a newHeads event for new Ethereum height #{ethereum_height}.")
         {:noreply, %{state | ethereum_height: ethereum_height}}
 
       false ->


apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
```diff
@@ -26,15 +26,20 @@ defmodule OMG.ChildChainRPC.Plugs.HealthTest do
...
+        pull_client_alarm(
+          300,
+          %{
+            "data" => %{
+              "code" => "operation:service_unavailable",
+              "description" => "The server is not ready to handle the request.",
+              "object" => "error"
+            },
+            "success" => false,
+            "version" => "0.2"
```
pthomalla: we should not catch the version

apps/omg/lib/omg/ethereum_height.ex
```diff
@@ -51,7 +51,7 @@ defmodule OMG.EthereumHeight do
     case is_binary(value) do
       true ->
         ethereum_height = Encoding.int_from_hex(value)
-        _ = Logger.info("#{__MODULE__} got a newHeads event for new Ethereum height #{ethereum_height}.")
+        _ = Logger.debug("#{__MODULE__} got a newHeads event for new Ethereum height #{ethereum_height}.")
```
pthomalla: :+1:

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
```diff
@@ -26,15 +26,19 @@ defmodule OMG.ChildChainRPC.Plugs.HealthTest do
...
-        },
-        fn -> TestHelper.rpc_call(:post, "/block.get", %{}) end
-      )
+      :ok =
+        pull_client_alarm(
+          300,
+          %{
+            "data" => %{
+              "code" => "operation:service_unavailable",
+              "description" => "The server is not ready to handle the request.",
```
pnowosie: We're too specific here, we'll need to change this test everytime we fix spelling or...

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
```diff
@@ -93,7 +101,10 @@ defmodule OMG.ChildChainRPC.Plugs.HealthTest do
   defp pull_client_alarm(0, _, _), do: :cant_match
 
   defp pull_client_alarm(n, match, fnn) do
-    case fnn.() do
+    endpoint_call_result = fnn.()
+    match = Map.put_new(match, "version", Map.get(endpoint_call_result, "version"))
```
pnowosie: Oh, we just dealing with version to be able to use: `case ... do ^match ->` and again paying to much attention to the response details.
I saw @pthomalla proposition, after all concerns with naming overall it sees to much cleaner and more importantly less prone to internal details of the response, e.g.: https://github.com/omisego/elixir-omg/pull/860/files#r304838025

pthomalla: It would be better to get the key out.

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
```diff
@@ -93,7 +101,10 @@ defmodule OMG.ChildChainRPC.Plugs.HealthTest do
   defp pull_client_alarm(0, _, _), do: :cant_match
 
   defp pull_client_alarm(n, match, fnn) do
```
pthomalla: the name of this function is also a bit confusing
match now also does not quite mean the same as the match in the elixir convention

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
InoMurko: what do you mean key out?

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
InoMurko: what's confusing about the name of the function? please, stick reviewing to what the PR is fixing. this went through a review already since it's in master.

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
InoMurko: do you mean key from the `match` variable? it's no there, that's why we're putting it there :)

InoMurko: why, the version comes as a response from the server? why would we want to match to the response?

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
InoMurko: But this is a *unit test*. That's exactly what the test is testing for!

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
InoMurko: @pnowosie again, this is a unit test. it tests a very specific response, because that's what it was written for. I want to test if I'm getting that *EXACT* response.

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
InoMurko: So if I accidentally change the description to "yolo what is this" and leave the code the same ("operation:service_unavailable"). Which test will cover for this bug?

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
InoMurko: also, take a deeper look, this is a _plug_ test. that's why this _plug_ was written for. for this exact purpose.

pnowosie: >  I want to test if I'm getting that EXACT response
Why? Isn't this plug handle also failure of `/submit` endpoint? Why then you've chosen `/block.get`.
IMHO we're testing here that `because alarm was raised no request is possible to fulfill`
being too specific we coupling exact error messages with `Health` plug

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
pnowosie: In that `yolo` case specific controller test will fail. C'mon Ino here you're testing `Health` plug not that controller response has specific format.

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
InoMurko: Because this plug kicks in before any controllers are being called. And because of that I don't need to test on all routes. 
`Health` plug produces that error message.
By definition, a *unit test*, is a coupling of the *test* and the *unit*.

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
InoMurko: No it will not. The `plug` kicks in before the controller and the plug is producing the response.
https://github.com/omisego/elixir-omg/blob/master/apps/omg_child_chain_rpc/lib/omg_child_chain_rpc/web/plugs/health.ex#L97

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
pnowosie: Ok, I missed that this's plug response, probably test title should not expose `/block.get` but say `all routes`.
Then I agree that **it's OK to specifically check error's code & description** (which is set by the plug).

Why then we require exact there will be `version` and all other fields in the response. Error response format is not `unit` we're testing here. Do you agree?

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
pnowosie: Reponded below :point_down:

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/plugs/health_test.exs
InoMurko: I simply followed how controller's are testing their response - https://github.com/omisego/elixir-omg/blob/master/apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/controllers/block_test.exs#L26-L38

pthomalla: and there we have matching no equality written in a strange way

pnowosie: Do it's just me or are we talking the same thing? The [example you mentioned](https://github.com/omisego/elixir-omg/blob/master/apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/controllers/block_test.exs#L26-L38).
Where are there `version` or `object => error` in the controller tests?

InoMurko: The version in this case doesn't really matter, because whatever is returned from the server is considered valid.
`object => error` is here https://github.com/omisego/elixir-omg/blob/master/apps/omg_child_chain_rpc/test/omg_child_chain_rpc/web/controllers/block_test.exs#L29

pnowosie: Don't you think then that function @pthomalla introduced was more intuitive. As he observed here we're checking **equality** not a match.

Also I meant we shouldn't be such specific in the controller tests.

InoMurko: No. I do not think his approach is more intuitive. I don't think it's making things better. It's actually making things worse by coupling :omg with :omg_child_chain_rpc and removing the test being self contained and it's not addressing any issues.

pnowosie: „Yet the wise men persisted in their resolution.” This is fine, I’m not going to burn more gas 🤓

✅

pthomalla: Address that this test was broken, that match is equality and that this name suggests that getting alarm is call ones.
As to the location of the function, We can move it to a more appropriate place.
#860

InoMurko: I don’t think I am able to add anything to this discussions. I’ve said everything I had.

PR #863, opened 2019-07-23
test: fix race in catching log in block_getter_test

pthomalla:
## Overview
Sometimes the log is printed before the capture_log is enabled, and test fails.
```
  * test transaction which is using already spent utxo from exit and happened after margin of slow validator(m_sv) causes to emit unchallenged_exit event2019-07-22 10:23:53.074 [warn] module=OMG.Watcher.Application function=set_cookie/1 ⋅Cookie not applied.⋅
2019-07-22 10:24:25.204 [warn] module=OMG.Watcher.BlockGetter.Core function=consider_exits/2 ⋅Chain invalid when taking exits into account, because of {:error, :unchallenged_exit}⋅
2019-07-22 10:24:25.204 [warn] module=OMG.Watcher.BlockGetter function=do_producer/1 ⋅Chain invalid when trying to download blocks, because of {:error, []}, won't try again⋅
  * test transaction which is using already spent utxo from exit and happened after margin of slow validator(m_sv) causes to emit unchallenged_exit event (48032.4ms)

  1) test transaction which is using already spent utxo from exit and happened after margin of slow validator(m_sv) causes to emit unchallenged_exit event (OMG.Watcher.Integration.BlockGetterTest)
     test/omg_watcher/integration/block_getter_test.exs:184
     Assertion with =~ failed
     code:  assert capture_log(fn -> IntegrationTest.wait_for_byzantine_events([%Event.UnchallengedExit{}.name()], @timeout) end) =~ inspect(:unchallenged_exit)
     left:  ""
     right: ":unchallenged_exit"
     stacktrace:
       test/omg_watcher/integration/block_getter_test.exs:224: (test)
```

## Changes
moved function causes log into capture_log 

## Testing
mix test --only integration test/omg_watcher/integration/block_getter_test.exs

### DIFF
--- a/apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
+++ b/apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
@@ -36,7 +36,6 @@ defmodule OMG.Watcher.Integration.BlockGetterTest do
   alias Watcher.TestHelper
 
   require Utxo
-  import ExUnit.CaptureLog
 
   @moduletag :integration
   @moduletag :watcher
@@ -148,10 +147,9 @@ defmodule OMG.Watcher.Integration.BlockGetterTest do
       different_hash
     )
 
-    {:ok, _txhash} = Eth.RootChain.submit_block(different_hash, 1, 20_000_000_000)
-
     # checking if both machines and humans learn about the byzantine condition
-    assert capture_log(fn ->
+    assert TestHelper.capture_log(fn ->
+             {:ok, _txhash} = Eth.RootChain.submit_block(different_hash, 1, 20_000_000_000)
              IntegrationTest.wait_for_byzantine_events([%Event.InvalidBlock{}.name], @timeout)
            end) =~ inspect({:error, :incorrect_hash})
   end
@@ -172,10 +170,10 @@ defmodule OMG.Watcher.Integration.BlockGetterTest do
     )
 
     invalid_block_hash = block_with_incorrect_transaction.hash
-    {:ok, _txhash} = Eth.RootChain.submit_block(invalid_block_hash, 1, 20_000_000_000)
 
     # checking if both machines and humans learn about the byzantine condition
-    assert capture_log(fn ->
+    assert TestHelper.capture_log(fn ->
+             {:ok, _txhash} = Eth.RootChain.submit_block(invalid_block_hash, 1, 20_000_000_000)
              IntegrationTest.wait_for_byzantine_events([%Event.InvalidBlock{}.name], @timeout)
            end) =~ inspect(:tx_execution)
   end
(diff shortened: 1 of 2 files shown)


apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
```diff
@@ -204,24 +203,24 @@ defmodule OMG.Watcher.Integration.BlockGetterTest do
...
-    # Here we're waiting for passing of margin of slow validator(m_sv)
-    exit_processor_sla_margin = Application.fetch_env!(:omg_watcher, :exit_processor_sla_margin)
-    Eth.DevHelpers.wait_for_root_chain_block(eth_height + exit_processor_sla_margin, @timeout)
-
-    # Here we're manually submitting invalid block to the root chain
-    {:ok, _} = OMG.Eth.RootChain.submit_block(bad_block_hash, 2, 1)
-
     # checking if both machines and humans learn about the byzantine condition
     assert capture_log(fn ->
+             {:ok, %{"status" => "0x1", "blockNumber" => eth_height}} =
```
pnowosie: This is unclear which of 2 function calls actually emit log event that we're expecting to capture?
Assuming from the change you're capturing to silence it from test output. Consider to add comment about which call we're checking log for and also we're silencing any other from appearing in the output.

pnowosie:
I noticed we have a flaky test, since it seems to be relevant to this :pr: could you please take a look.
Otherwise CI will make somebody else 🐨 

```
  * test fees in effect corrupted file does not make server crash (1128.5ms)
2019-07-23 08:15:32.403 [error] module=OMG.ChildChain.FeeServer function=update_fee_spec/0 ⋅Unable to update fees from file. Reason: {:error, %Jason.DecodeError{data: "[not a json]", position: 1, token: nil}}⋅

  1) test fees in effect corrupted file does not make server crash (OMG.ChildChain.Integration.FeeServerTest)
     test/omg_child_chain/integration/fee_server_test.exs:60
     Assertion with =~ failed
     code:  assert refresh_fees() =~ ~r"\\[error\\].*Unable to update fees"
     left:  ""
     right: ~r/\[error\].*Unable to update fees/
     stacktrace:
       test/omg_child_chain/integration/fee_server_test.exs:70: (test)
```

apps/omg_watcher/test/support/test_helper.ex
```diff
@@ -195,4 +196,28 @@ defmodule OMG.Watcher.TestHelper do
...
+      logs = CaptureLog.capture_log(fn -> function.() end)
+
+      case logs do
+        "" -> wait_for_log(max_waiting)
+        logs -> logs
+      end
+    end)
+  end
+
+  defp wait_for_log(max_waiting) when is_number(max_waiting) do
```
pnowosie:
```suggestion
  defp wait_for_log(timeout_ms, sleep_time_ms \\ 20) when is_number(timeout_ms) do
```

PR #864, opened 2019-07-23
Make coverage calculation encompass the entire umbrella app - tidied

### LINKED ISSUE #806: Fix test coverage (and/or) coverage calculation

pdobacz:
Closes #806. Supersedes #828, which due to a possible bug in GH doesn't handle the commits correctly after rebasing. We're keeping the #828 and branch intact for possible investigation.

This is also squashed etc etc

## Overview

`mix coveralls` via `mix test --cover` always calculates test coverage of an umbrella sub-app only by using that sub-apps own tests.

Since we have some code split across sub-apps (e.g. the `_rpc` bunch), some code only gets covered in foreign sub-apps of the umbrella.

C.f. this issue https://github.com/parroty/excoveralls/issues/23

## Changes

`excoveralls` has been updated to a derivative (rebase) of one of the forks proposed in the issue above.
Some adjustments were needed on our side to have both the (new) `mix coveralls --umbrella` and the usual `mix test` invocations work.

## Testing

existing tests, and an increase in test coverage as demonstrated by [coveralls.io](https://coveralls.io/github/omisego/elixir-omg?branch=806-fix_coverage_final)


InoMurko:
What a feel good PR this is! Yipeee @pdobacz

PR #865, opened 2019-07-23
fix: fix flaky test

pthomalla:
``` 
 * test fees in effect corrupted file does not make server crash (1128.5ms)
2019-07-23 08:15:32.403 [error] module=OMG.ChildChain.FeeServer function=update_fee_spec/0 ⋅Unable to update fees from file. Reason: {:error, %Jason.DecodeError{data: "[not a json]", position: 1, token: nil}}⋅

  1) test fees in effect corrupted file does not make server crash (OMG.ChildChain.Integration.FeeServerTest)
     test/omg_child_chain/integration/fee_server_test.exs:60
     Assertion with =~ failed
     code:  assert refresh_fees() =~ ~r"\\[error\\].*Unable to update fees"
     left:  ""
     right: ~r/\[error\].*Unable to update fees/
     stacktrace:
       test/omg_child_chain/integration/fee_server_test.exs:70: (test)
```

### DIFF
--- a/apps/omg_child_chain/test/omg_child_chain/integration/fee_server_test.exs
+++ b/apps/omg_child_chain/test/omg_child_chain/integration/fee_server_test.exs
@@ -66,8 +66,10 @@ defmodule OMG.ChildChain.Integration.FeeServerTest do
       assert {:ok, default_fees} == FeeServer.transaction_fees()
 
       # corrupt file, refresh, check fees did not change
-      overwrite_fee_file(file_name, "[not a json]")
-      assert refresh_fees() =~ ~r/\[error\].*Unable to update fees/
+      assert capture_log(fn ->
+               overwrite_fee_file(file_name, "[not a json]")
+               refresh_fees()
+             end) =~ ~r/\[error\].*Unable to update fees/
 
       assert {:ok, default_fees} == FeeServer.transaction_fees()
       assert server_alive?()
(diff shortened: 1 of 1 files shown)


apps/omg_child_chain/test/omg_child_chain/integration/fee_server_test.exs
```diff
@@ -136,9 +138,22 @@ defmodule OMG.ChildChain.Integration.FeeServerTest do
...
+      logs = capture_log(fn -> Process.send(pid, :update_fee_spec, []) end)
+
+      case logs do
+        "" -> wait_for_log()
+        logs -> logs
+      end
+    end)
+  end
+
+  defp wait_for_log do
```
pnowosie: Thinking of yours `call_until` as a test level helper function. This log handling case should be tricky enough that we could end with nice generic function that handles all wait_for cases.

apps/omg_child_chain/test/omg_child_chain/integration/fee_server_test.exs
```diff
@@ -66,8 +66,10 @@ defmodule OMG.ChildChain.Integration.FeeServerTest do
       assert {:ok, default_fees} == FeeServer.transaction_fees()
 
       # corrupt file, refresh, check fees did not change
-      overwrite_fee_file(file_name, "[not a json]")
-      assert refresh_fees() =~ ~r/\[error\].*Unable to update fees/
+      assert capture_log(fn ->
+               overwrite_fee_file(file_name, "[not a json]")
```
pnowosie: For me it's impossible we get log event before calling `refresh_fees()` therefor consider move `overwrite_fee_file(...)` outside capture_log lambda. What do you thnik?

pthomalla: forget about interval message :update_fee_spec :D 
1. overwrite_fee_file overwrites file
2. send_interval :update_fee_spec run next
3. refresh_fees()  have no logs becase file don't change

Issue #867, opened 2019-07-25
fix ecto.gen.migration task mix task


jrhite:
Currently this task does is broken.

1. If `mix ecto.gen.migration add_eth_txn_hash_column -r OMG.Watcher.Repo` is run from the base dir `elixir_omg` the following error will occur:

```
** (Mix) Cannot run task "ecto.gen.migration" from umbrella application
```

2. If run from `elixir-omg/apps/omg_watcher` then you will see an error about `OMG.Umbrella.MixProject.umbrella_version()` not existing. To hack/fix this you will need to go into the following child apps and hard-code the umbrella version to some random version...eg: '0.0.0'

```
apps/omg/mix.exs
apps/omg_child_chain/mix.exs
apps/omg_db/mix.exs
apps/omg_eth/mix.exs
apps/omg_status/mix.exs
apps/omg_utils/mix.exs
```

Finally, after doing this you'll be a bit closer, but when running `mix ecto.gen.migration add_eth_txn_hash_column` again from the `elixir-omg/apps/omg_watcher` dir you will see:

```
elixir-omg/apps/omg_watcher$ mix ecto.gen.migration add_eth_txn_hash_column
==> omg_eth
Compiling 19 files (.ex)
Generated omg_eth app
==> omg
Compiling 47 files (.ex)
Generated omg app
==> omg_watcher
** (Mix) The task "compile.phoenix" could not be found. Did you mean "compile.elixir"?
```

InoMurko:
Yeah, so I tried quickly figuring out what's with this ... this sorts it somewhat https://github.com/omisego/elixir-omg/pull/868

InoMurko:
The right command though would be `mix ecto.gen.migration <your migration file> -r OMG.Watcher.DB.Repo`

InoMurko:
Sorted in #868

PR #868, opened 2019-07-25
fix: phoenix compiler missing

InoMurko:
https://github.com/omisego/elixir-omg/issues/867

## Overview

Changes needed so that we're able to run `mix ecto.gen.migration add_eth_txn_hash_column -r OMG.Watcher.DB.Repo` from the `apps/omg_watcher` directory.

## Changes

- VERSION file in root directory where all the umbrella apps get the version string from
- Updated distillery to the latest stable version and made the breaking changes

## Testing

/


InoMurko:
https://github.com/omisego/elixir-omg/issues/867#issuecomment-515223633

apps/omg_watcher/mix.exs
```diff
@@ -37,6 +37,13 @@ defmodule OMG.Watcher.Mixfile do
       {:deferred_config, "~> 0.1.1"},
       {:telemetry, "~> 0.4.0"},
       {:spandex_ecto, "~> 0.6.0"},
+      # there's no apparent reason why libsecp256k1, spandex and distillery need to be included as dependencies
+      # to this umbrella application apart from mix ecto.gen.migration not working, so here they are, copied from
+      # the parent (main) mix.exs
+      {:libsecp256k1, git: "https://github.com/omisego/libsecp256k1.git", branch: "elixir-only", override: true},
+      {:spandex, "~> 2.4",
+       git: "https://github.com/omisego/spandex.git", branch: "fix_dialyzer_in_macro", override: true},
+      {:distillery, "~> 2.1", runtime: false},
```
jrhite: This is odd. What dependency does `mix ecto.gen.migration` have on these libs?

apps/omg_watcher/mix.exs
```diff
@@ -48,7 +55,8 @@ defmodule OMG.Watcher.Mixfile do
       # here only to leverage common test helpers and code
       {:fake_server, "~> 1.5", only: [:dev, :test], runtime: false},
       {:briefly, "~> 0.3.0", only: [:dev, :test], runtime: false},
-      {:omg_child_chain, in_umbrella: true, only: [:test], runtime: false}
+      {:omg_child_chain, in_umbrella: true, only: [:test], runtime: false},
+      {:phoenix, "~> 1.3", runtime: false}
```
jrhite: Just curious for me to learn, why does watcher have a phoenix dependency? It's not a webserver, or is it?

jrhite: Ah, nvm...for testing!

apps/omg_watcher/mix.exs
InoMurko: doesn't directly, what it tries to do is it's fetching dependencies of dependencies. so `:spandex_ecto` has a dependency on raw `:spandex`, but we've overriden that in the top level `mix.exs` with our fork... so that's what's hurting `:omg_watcher`

apps/omg_watcher/mix.exs
InoMurko: yep!

mix.exs
```diff
@@ -1,7 +1,7 @@
 defmodule OMG.Umbrella.MixProject do
   use Mix.Project
 
-  def umbrella_version, do: "0.2.0"
+  def umbrella_version, do: "#{String.trim(File.read!("VERSION"))}"
```
pdobacz: not needed anymore I think, it was only there to be able to modify all `version` entries in one place, which is done via `VERSION` now

InoMurko: It's now being used in https://github.com/omisego/elixir-omg/blob/master/Makefile#L116
and https://github.com/omisego/elixir-omg/blob/master/Makefile#L123 
so that releases can be built :)

PR #869, opened 2019-07-25
docs: add more details and rationale on the `unchallenged_exit` event

pdobacz:
Answers requests to document on our slack

## Overview

Provides details as in title

## Changes

N/A

## Testing

:eyes: :book: :scroll:

### DIFF
--- a/docs/api_specs/status_events_specs.md
+++ b/docs/api_specs/status_events_specs.md
@@ -37,6 +37,7 @@ Indicates that an invalid exit is occurring. It should be challenged.
 ```
 
 Indicates that an invalid exit is dangerously close to finalization and hasn't been challenged. User should exit.
+See docs on [`unchallenged_exit` condition](../exit_validation.md#unchallenged-exit-condition) for more details.
 
 
 #### `invalid_block`
@@ -84,7 +85,6 @@ The ChildChain is withholding a block whose hash has been published on the root 
 ```
 
 An in-flight exit of a non-canonical transaction has been started. It should be challenged.
-<aside class="warning"> Not Implemented Yet.</aside>
 
 Event details:
 
@@ -105,7 +105,6 @@ txbytes | Hex encoded string | The in-flight transaction that the event relates 
 ```
 
 A canonical in-flight exit has been challenged. The challenge should be responded to.
-<aside class="warning"> Not Implemented Yet.</aside>
 
 Event details:
 
@@ -160,7 +159,6 @@ available_inputs | Object array | The inputs (index and address) available to be
 ```
 
 An invalid piggyback is in process. Should be challenged.
-<aside class="warning"> Not Implemented Yet.</aside>
 
 Event details:
 
--- a/docs/exit_validation.md
+++ b/docs/exit_validation.md
@@ -4,15 +4,14 @@ This document describes the exit validation (processing) done by the Watcher in 
 **NOTE** not all of this is implemented yet.
 
 NOTE:
-* `exit_finality_margin` margin exit processor (in Ethereum blocks)
-* `SLA` Service Level Agreement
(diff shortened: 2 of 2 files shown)


docs/api_specs/status_events_specs.md
```diff
@@ -37,6 +37,7 @@ Indicates that an invalid exit is occurring. It should be challenged.
 ```
```
InoMurko: @pdobacz this document says `User should exit.`. What it really means is that the watcher should exit with all it's UTXOs? IS that what User means in this case?

docs/api_specs/status_events_specs.md
```diff
@@ -37,6 +37,7 @@ Indicates that an invalid exit is occurring. It should be challenged.
...
```
InoMurko: the document is saying that `invalid_ife_challenge` is not implemented yet, is this a blocker for mainnet? cc @Pongch

pdobacz: These are stale comments, everything that here said "not implemented" is implemented. Nice catch!

docs/api_specs/status_events_specs.md
pdobacz: The User here is the user of the Watcher. Watcher is understood as the app, not a role. 

In other words a "user" here is someone who holds value on the child chain, is interested in keeping that safe and uses the Watcher to that end

docs/exit_validation.md
```diff
@@ -51,6 +50,8 @@ This is acceptable as long as the delay doesn't exceed `scheduled finalization t
 
 Conditions 3 and 4 can be represented jointly, by **continuous** (every child block) validation of the exits.
 If any exit is invalid and its finalization is near, then actions listed under 3 and 4 should take place.
+Let's call this joint representation of 3 and 4 `unchallenged_exit`.
```
InoMurko: When the text says: `watcher must not allow spending and depositing` - is that prohibited on ChildChain as well? Or could a "hacker" exploit that assumption and allow spending/depositing from the watcher?

InoMurko: When the text says: `(optional) If finalization of an invalid exit occurs, the watcher must prompt an exit without guarantees of recovery of funds` this basically means FIFO? Whoever is first to exit can get _money back_, if you're last, maybe there won't be anything left to exit with from the contract?

pdobacz: This not allowing has nothing to do with the child chain. The Watcher protects its user, by not allowing certain actions under certain conditions. The child chain cannot possibly protect the user like that: it isn't trusted by the user.

Or, hmmm, maybe I don't get the question here... what would precisely such hacker do?

InoMurko: I'm just exploring the possibility of a hacker cloning our repository, modifying the code by *allowing* spending/depositing in certain conditions and how that would affect the operators child chain...

pdobacz: > When the text says: `(optional) If finalization of an invalid exit occurs, the watcher must prompt an exit without guarantees of recovery of funds` this basically means FIFO? 

FIFO is not 100% precise, as the Priority of exiting has richer logic. 

> Whoever is first to exit can get _money back_, if you're last, 

Again, not only "first to exit" but the exit priority counts here. It is impacted both by time exit started and age of utxo, see `tesuji_blockchain....` docs and others.

Having said that

> maybe there won't be anything left to exit with from the contract?

yep that can happen in this particular scenario (which we're avoiding by following protocol/Watching etc.)

pdobacz: > I'm exploring the possibility of a hacker cloning our repository, modifying the code by _allowing_ spending/depositing in certain conditions and how that would affect the operators child chain...

If someone uses such modified code, anything can happen. In particular one can turn this particular function of the Watcher off and whoever uses that is exposed. Assuming the operator is using unmodified child chain code, it is never affected by such Watcher's operation.

InoMurko:
This was a very good read!!!

PR #871, opened 2019-07-26
fix: Statsd connection can drop and it shouldn't be attached to the a…

InoMurko:
Statsd connection can drop and it shouldn't be attached to the application process. If it drops it'll take the application with it.

## Changes

- Process monitoring with alarm raising
- excluding Datadog when `DD_DISABLED=true`
- flag to optionally include ROCKSDB 

## Testing

### DIFF
--- a/apps/omg/config/config.exs
+++ b/apps/omg/config/config.exs
@@ -7,8 +7,7 @@ config :omg,
   ethereum_events_check_interval_ms: 500,
   coordinator_eth_height_check_interval_ms: 6_000,
   client_monitor_interval_ms: 4500,
-  metrics_collection_interval: 60_000,
-  ws_url: {:system, "ETHEREUM_WS_RPC_URL", "ws://localhost:8546/ws"}
+  metrics_collection_interval: 60_000
 
 config :omg, :eip_712_domain,
   name: "OMG Network",
--- a/apps/omg/config/config.exs
+++ b/apps/omg/config/config.exs
@@ -8,7 +8,7 @@ config :omg,
   coordinator_eth_height_check_interval_ms: 6_000,
   client_monitor_interval_ms: 4500,
   metrics_collection_interval: 60_000,
-  ws_url: System.get_env("ETHEREUM_WS_RPC_URL") || "ws://localhost:8546/ws"
+  ws_url: {:system, "ETHEREUM_WS_RPC_URL", "ws://localhost:8546/ws"}
 
 config :omg, :eip_712_domain,
   name: "OMG Network",
--- a/apps/omg/config/config.exs
+++ b/apps/omg/config/config.exs
@@ -7,7 +7,8 @@ config :omg,
   ethereum_events_check_interval_ms: 500,
   coordinator_eth_height_check_interval_ms: 6_000,
   client_monitor_interval_ms: 4500,
-  metrics_collection_interval: 60_000
+  metrics_collection_interval: 60_000,
+  ws_url: System.get_env("ETHEREUM_WS_RPC_URL") || "ws://localhost:8546/ws"
 
 config :omg, :eip_712_domain,
   name: "OMG Network",
--- a/apps/omg/lib/omg/ethereum_client_monitor.ex
+++ b/apps/omg/lib/omg/ethereum_client_monitor.ex
@@ -60,7 +60,7 @@ defmodule OMG.EthereumClientMonitor do
     state = %__MODULE__{
       alarm_module: alarm_module,
       ethereum_height: ethereum_height,
-      ws_url: Keyword.get(opts, :ws_url)
+      ws_url: Keyword.get(opts, :ws_url, Application.get_env(:omg, :ws_url))
     }
 
     _ = raise_clear(alarm_module, state.raised, ethereum_height)
--- a/apps/omg/lib/omg/ethereum_client_monitor.ex
+++ b/apps/omg/lib/omg/ethereum_client_monitor.ex
@@ -42,7 +42,7 @@ defmodule OMG.EthereumClientMonitor do
   defstruct interval: @default_interval,
             tref: nil,
             alarm_module: nil,
-            raised: true,
+            raised: false,
             ethereum_height: :error,
             ws_url: nil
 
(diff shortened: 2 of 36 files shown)


apps/omg/lib/omg/supervisor.ex
```diff
@@ -27,7 +27,7 @@ defmodule OMG.Supervisor do
   def init(:ok) do
     children = [
       {OMG.InternalEventBus, []},
-      {OMG.EthereumClientMonitor, [alarm_module: Alarm]},
+      {OMG.EthereumClientMonitor, [alarm_module: Alarm, ws_url: Application.get_env(:omg, :ws_url)]},
```
InoMurko: a good pattern is passing configuration top -> down.

apps/omg_db/mix.exs
```diff
@@ -44,4 +43,11 @@ defmodule OMG.DB.MixProject do
       {:omg_status, in_umbrella: true}
     ]
   end
+
+  defp rocksdb do
```
InoMurko: rocksdb is compiling FOR EVER.

InoMurko: `export EXCLUDE_ROCKSDB=1` and you don't need to

apps/omg_eth/lib/omg_eth/subscription_worker.ex
```diff
@@ -29,7 +29,7 @@ defmodule OMG.Eth.SubscriptionWorker do
   """
   @spec start_link(Keyword.t()) :: {:ok, pid()} | no_return()
   def start_link(opts) do
-    ws_url = Keyword.get(opts, :ws_url, Application.get_env(:omg_eth, :ws_url))
+    ws_url = Keyword.get(opts, :ws_url)
```
InoMurko: configuration top->down

apps/omg_status/lib/omg_status/metric/datadog.ex
```diff
@@ -23,4 +23,21 @@ defmodule OMG.Status.Metric.Datadog do
...
+
+  use GenServer
+  require Logger
+
+  def start, do: GenServer.start(__MODULE__, [], [])
+
+  def init(_opts) do
+    _ = Logger.info("Starting #{inspect(__MODULE__)} and connecting to Datadog.")
+    Process.flag(:trap_exit, true)
+    connect()
```
InoMurko: We plug the UDP connection port to a worker process that we can monitor separately.

apps/omg_status/lib/omg_status/metric/statsd_monitor.ex
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+  def handle_info(:connect, state) do
+    {pid, _ref} = monitor = Kernel.spawn_monitor(state.child_module, :start, [])
+    alive = Process.alive?(pid)
+    _ = raise_clear(state.alarm_module, state.raised, alive)
+    {:noreply, %{state | monitor: monitor}}
+  end
+
+  def handle_cast(:clear_alarm, state) do
```
InoMurko: we're subscribed to alarms and that keeps our internal state consistent!

apps/omg/config/config.exs
```diff
@@ -7,7 +7,8 @@ config :omg,
   ethereum_events_check_interval_ms: 500,
   coordinator_eth_height_check_interval_ms: 6_000,
   client_monitor_interval_ms: 4500,
-  metrics_collection_interval: 60_000
+  metrics_collection_interval: 60_000,
+  ws_url: System.get_env("ETHEREUM_WS_RPC_URL") || "ws://localhost:8546/ws"
```
InoMurko: this should be a tuple

apps/omg/lib/omg/ethereum_client_monitor.ex
```diff
@@ -42,7 +42,7 @@ defmodule OMG.EthereumClientMonitor do
   defstruct interval: @default_interval,
             tref: nil,
             alarm_module: nil,
-            raised: true,
+            raised: false,
```
achiurizo: 🤔 Do we have tests to cover this change?

InoMurko: same tests should continue working, but I can add more

apps/omg/config/config.exs
```diff
@@ -7,7 +7,8 @@ config :omg,
   ethereum_events_check_interval_ms: 500,
   coordinator_eth_height_check_interval_ms: 6_000,
   client_monitor_interval_ms: 4500,
-  metrics_collection_interval: 60_000
+  metrics_collection_interval: 60_000,
+  ws_url: {:system, "ETHEREUM_WS_RPC_URL", "ws://localhost:8546/ws"}
```
pdobacz: this config entry being moved from `:omg_eth` to `:omg` doesn't look right, this kind of configuration is strictly related to the connection to the eth node's APIs, hence `:omg_eth`. Configuring of this connection should be none of `:omg`s business.

InoMurko: Hmmm, yes this is confusing to me as well. The problematic part is how omg is coupled with omg_eth. On one point, omg_eth is a very weak interface towards ethereum clients, being standalone doesn’t make sense. Maybe we should consider making the integration stronger on omg_eth part and weaker on omg. That includes monitoring for example. 

Pulling application settings specific to omg_eth into omg was bothering me...

InoMurko: so maybe the solution is to put :omg_eth back as the holder of :ws_url, but then seriously consider moving `EthereumClientMonitor` and `EthereumHeight` into :omg_eth. WDYT @pdobacz ?
That would separate plasma related implementation into :omg, and non specific ethereum client integration into :omg_eth

pdobacz: > seriously consider moving EthereumClientMonitor and EthereumHeight into :omg_eth.

:+1: Based on my limited knowledge about these two, and their names, it sounds right to move them to `:omg_eth`. `:omg_eth` was intended to be super-minimalistic, just taking care of encodings, RPC etc. but those two might still fit, as long as they don't have any `omg`-specific logic involved.

apps/omg_eth/test/omg_eth/subscription_worker_test.exs
```diff
@@ -53,82 +53,4 @@ defmodule OMG.Eth.SubscriptionWorkerTest do
       end
     )
   end
-
-  test "that worker can subscribe to my own server via arguments" do
```
InoMurko: I have removed the test because we've changed how we pass in `ws url` so the test has no relevance anymore.

PR #873, opened 2019-07-29
feature: expose alarms via endpoint

InoMurko:
///

## Overview

Read the raised alarms on an Watcher endpoint _http://watcher/alarm.get_
ChildChain TBD.

## Changes

- Added a new route!

## Testing

Wrote a few tests that should cover all behaviors.


apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/alarm_test.exs
```diff
@@ -0,0 +1,55 @@
...
+defmodule OMG.WatcherRPC.Web.Controller.AlarmTest do
+  use ExUnitFixtures
+  use ExUnit.Case, async: false
+
+  use OMG.Fixtures
+  use OMG.Watcher.Fixtures
+  alias OMG.Watcher.TestHelper
+
+  setup %{} do
+    system_alarm = {:system_memory_high_watermark, []}
```
chrishunt: Since we have the same setup/tests already at the unit level in `API.AlarmTest`, what do you think about removing them here? We could still leave a test in here of course to make sure thing are working, but maybe just a single/simple test with one alert. What are your thoughts?

InoMurko: This is somehow ensuring that `OMG.WatcherRPC.Web.View.Alarm` and `OMG.Utils.HttpRPC.Response` can encode the alarms and properly return them as JSONs. I guess what I'm trying to say is that if I remove them from here, I would prefer to "move" them to `OMG.Utils.HttpRPC.ResponseTest` and check if they're being handled properly.

apps/omg_utils/lib/omg_utils/http_rpc/response.ex
```diff
@@ -54,6 +54,8 @@ defmodule OMG.Utils.HttpRPC.Response do
 
   def sanitize(bin) when is_binary(bin), do: Encoding.to_hex(bin)
   def sanitize({:skip_hex_encode, bin}), do: bin
+  def sanitize({{key, value}, _}), do: Map.put_new(%{}, key, value)
```
InoMurko: this, @chrishunt

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/alarm_test.exs
chrishunt: Ahhh, I see. Yeah that makes sense. I think moving to `HttpRPC.ResponseTest` sounds like a pretty good idea. It looks like we have some sanitation tests in there already and this would go well in there.

PR #883, opened 2019-08-06
Perf fix

InoMurko:


apps/omg_performance/lib/performance.ex
```diff
@@ -76,7 +76,10 @@ defmodule OMG.Performance do
         "PerfTest number of spenders: #{inspect(nspenders)}, number of tx to send per spender: #{inspect(ntx_to_send)}."
       )
 
-    DeferredConfig.populate(:omg_child_chain_rpc)
+    :ok = DeferredConfig.populate(:spandex_datadog)
+    :ok = DeferredConfig.populate(:statix)
+    :ok = DeferredConfig.populate(:omg_child_chain_rpc)
+    :ok = DeferredConfig.populate(:omg_watcher_rpc)
```
pdobacz: at 1st glance this (`omg_watcher_rpc`) should be unnecessary. If things break without it, then maybe the problem lies elsewhere? perf test shouldn't require the watcher in any form

docs/perf_test_result_dumps.md
```diff
@@ -5,7 +5,7 @@
 Command:
 
 ```
-mix run --no-start -e 'OMG.Performance.start_simple_perftest(8_000, 32, %{block_every_ms: 15_000})'
+DD_DISABLED=true mix run --no-start -e 'OMG.Performance.start_simple_perftest(8_000, 32, %{block_every_ms: 15_000})'
```
pdobacz: 1. in order for me to work, I needed `APP_ENV=localwatcher` too.
2. it seems we have this command copy-pasted many times here. Maybe let's remove the exact copies and rely on the "Command as above"
3. more general remark on the system environment being required to be set - wouldn't we rather put the values required in invocations like these into the defaults, as held in `config/dev.exs`.

InoMurko: 3. I'm doing that https://github.com/omisego/elixir-omg/pull/884 in progress.

InoMurko:
I would put this fix on hold for just a while (two open PR's should change the way we fix this).

PR #885, opened 2019-08-07
fix: repeated exits coming from kernel monitor, because it spawns a p…

InoMurko:
…rocess

//

## Overview

Swaping Kernel monitor for link.

## Changes

- test fixes
- genserver exposing start_link instead of kernel monitor

## Testing

/

### DIFF
--- a/apps/omg_status/test/omg_status/metric/datadog_test.exs
+++ b/apps/omg_status/test/omg_status/metric/datadog_test.exs
@@ -17,7 +17,17 @@ defmodule OMG.Status.Metric.DatadogTest do
   alias OMG.Status.Metric.Datadog
 
   setup do
-    %{datadog: Datadog.start()}
+    parent = self()
+
+    spawn(fn ->
+      {:ok, datadog_pid} = Datadog.start_link()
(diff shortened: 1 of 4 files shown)


apps/omg_status/test/omg_status/metric/datadog_test.exs
```diff
@@ -28,11 +38,10 @@ defmodule OMG.Status.Metric.DatadogTest do
         port = Port.open({:spawn, "cat"}, [:binary])
         true = Process.link(datadog_pid)
         true = Process.exit(port, :portkill)
+
         # we want to exit because the port forcefully closes
         # so this sleep shouldn't happen
         Process.sleep(10_000)
       end)
-
-    assert_receive {:trace, ^datadog_pid, :receive, {:EXIT, port, :portkill}}
```
pdobacz: removal of this assertion is a bit suspicious - there's no assertions left here. Also the name of the test indicates we want to test an exit signal.

InoMurko: Good catch! This one slipped!

PR #889, opened 2019-08-08
fix: provide a top-level coveralls.json for the new coveralls flow

pdobacz:
Spotted by @InoMurko

## Overview

PRovides a top level coveralls.json with exclude dirs - the ones sitting in the umbrella subapps aren't enough for the new `coveralls` task which works from the top-level

## Changes

N/A

## Testing

existing testes + :eyes: on the coveralls.io report

### DIFF
--- a/coveralls.json
+++ b/coveralls.json
@@ -1,27 +1,17 @@
 {
   "skip_files": [
     "apps/omg/test/support",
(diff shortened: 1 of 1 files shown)


InoMurko:
I'm writing tests for all `/release_tasks/*` so they could be removed from this list...

PR #898, opened 2019-08-16
DoS Watcher test for standard exit

pthomalla:
## Overview

As a security-concerned application provider, I would like to ensure the security-critical watcher will work on DOS condition so that I know my funds are secured

  first part generates proper exit: using watcher api to check available utxo
## Testing

This :pr: brings two new tests in Performance application:
* time response for asking for exit data - pretty fast, just ask Watcher for exit data,
* watcher catch all non_canonical_ife - modifies txs and starts IFE exits, then waits when Watcher's `in_flight_exit_processor` processor synchronize and detect these non-canonical exits.

:spiral_notepad: For the above tests we run Ch-ch and Watcher in `dev` env to not abuse other tests we have. Also we start Watcher app (1st time in tests) it's in fixture: `omg_watcher`.


apps/omg_performance/test/omg_performance/integration/bazantine_event_test.exs
```diff
@@ -0,0 +1,145 @@
...
+    ntx_toxend: #{ntx_to_send}
+    exits per dos user: #{length(spenders) * ntx_to_send}
+    total exits: #{length(spenders) * ntx_to_send * dos_users}
+    """)
+
+    # TODO find another way to wait for data generations, maybe smoke_test_statistics
+    Performance.start_extended_perftest(ntx_to_send, spenders, contract)
+    Process.sleep(4000)
+
+    statistics =
```
InoMurko: this absolutely cannot be written like this. do not write complicated piped chains.

apps/omg_performance/test/omg_performance/integration/bazantine_event_test.exs
```diff
@@ -0,0 +1,145 @@
...
+           Enum.reduce(exits_utxo, %{correct: 0, error: 0}, fn
+             %{"proof" => _}, stats -> Map.update(stats, :correct, 0, &(&1 + 1))
+             _, stats -> Map.update(stats, :error, 0, &(&1 + 1))
+           end)
+         end)
+         |> Enum.reduce(%{error: 0, correct: 0}, fn %{error: error, correct: correct}, statistics ->
+           Map.update(statistics, :error, 0, &(&1 + error))
+           |> Map.update(:correct, 0, &(&1 + correct))
+         end)}
+      end)
```
achiurizo: This is _really_ difficult to grok and could probably be broken up better to make it more clear what's happening in this assignment. Could we break this up/refactor this? 😄

apps/omg_performance/test/omg_performance/integration/bazantine_event_test.exs
```diff
@@ -0,0 +1,145 @@
...
+      Enum.map(1..dos_users, fn _ ->
+        Task.async(fn ->
+          stream_exit_data_blocks = generate_exit_info() |> Enum.map(&get_lazy_exit_data/1) |> Enum.to_list()
+          :timer.tc(fn -> stream_exit_data_blocks |> Enum.map(&Enum.to_list(&1)) end)
+        end)
+      end)
+      |> Enum.map(fn task ->
+        {time, exit_data_blocks} = Task.await(task, :infinity)
+
+        {time / 1_000,
```
pnowosie: Try to find std conversion function

apps/omg_performance/test/omg_performance/integration/bazantine_event_test.exs
```diff
@@ -0,0 +1,145 @@
...
+    {:ok, top_block} = RootChain.get_mined_child_block()
+    %{interval: interval, counter: trunc(top_block / interval)}
+  end
+
+  defp generate_exit_info_block(blknum) do
+    {:ok, {block_hash, _timestamp}} = RootChain.get_child_chain(blknum)
+    child_chain_url = Application.get_env(:omg_watcher, :child_chain_url)
+    {:ok, block} = Client.get_block(block_hash, child_chain_url)
+
+    Enum.shuffle(Enum.zip(block.transactions, Stream.iterate(0, &(&1 + 1))))
```
pnowosie: There is `Enum.with_index(...)` instead of `Enum.zip(block.transactions, Stream.iterate(...)))`
Also you can use `Stream` instead of `Enum` as you don't  need an intermediate result

apps/omg_performance/test/omg_performance/integration/bazantine_event_test.exs
```diff
@@ -0,0 +1,145 @@
...
+    |> Enum.map(fn {tx, nr} ->
```
pnowosie:
```suggestion
    |> Enum.map(fn {tx, index} ->
```

apps/omg_performance/test/omg_performance/integration/bazantine_event_test.exs
```diff
@@ -0,0 +1,145 @@
...
+  defp get_exit_data(%{utxo_pos: utxo_pos}) do
+    TestHelper.get_exit_data(utxo_pos)
+  rescue
+    error in MatchError -> error
+  end
+
+  defp generate_users(size, opts \\ [initial_funds: trunc(:math.pow(10, 18))]) do
+    async_generate_user = fn _ -> Task.async(fn -> generate_user(opts) end) end
+
+    Enum.chunk_every(1..size, 10)
```
pnowosie: Pipe more e.g. `1..size |> Enum.every()` and `chunk |> Enum.map()`
Also consider to use `Stream` when a result going to another transformation

apps/omg_performance/test/omg_performance/integration/bazantine_event_test.exs
```diff
@@ -0,0 +1,145 @@
...
+  alias OMG.State.Transaction
+  alias OMG.Utxo
+  alias OMG.Watcher.HttpRPC.Client
+  alias OMG.Watcher.TestHelper
+
+  @moduletag :integration
+  @moduletag timeout: 70_000
+
+  @tag fixtures: [:watcher, :child_chain, :contract]
+  test "time response for asking for exit data", %{contract: %{contract_addr: contract}} do
```
pdobacz: All other concerns aside, I'm missing a documented way to actually run this performance test, like we have with the old perf test. A test put in `omg_performance/test/` should only test that the perf test runs fine, and not be the way to invoke the perf test

apps/omg_performance/test/omg_performance/integration/bazantine_event_test.exs
pnowosie: `erlang:convert_time_unit(Time, FromUnit, ToUnit) -> ConvertedTime`

apps/omg_performance/test/omg_performance/integration/bazantine_event_test.exs
pthomalla: after add more workers, will expand the demo_05 and expand ByzantineEvents documentation

apps/omg_eth/lib/omg_eth/wait_for.ex
```diff
@@ -69,4 +69,18 @@ defmodule OMG.Eth.WaitFor do
       :error, {:badmatch, _} = _error -> repeat_until_ok(f)
     end
   end
+
+  def repeat_until_ok(f, acc) do
```
InoMurko: is there a limit the amount of times this would get called until it should time out?

InoMurko: Is this function used anywhere else then in `watcher_synchronize`? why not included it there and limit the fragmentation?

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -0,0 +1,121 @@
...
+
+  def start_dos_get_exits(dos_users, positions) do
+    Enum.map(1..dos_users, fn _ -> worker_dos_exit(positions) end)
+    |> Enum.map(&Task.await(&1, :infinity))
+  end
+
+  defp worker_dos_exit(exit_positions) do
+    worker = fn exit_positions ->
+      exit_positions
+      |> Enum.map(fn position ->
```
InoMurko: there's no chaining here so write the mapping like
`Enum.map(exit_positions, ...`

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -0,0 +1,121 @@
...
+        get_exit_data(position) |> valid_exit_data()
+      end)
+    end
+
+    Task.async(fn ->
+      exit_positions = exit_positions |> Enum.shuffle()
```
InoMurko: `Enum.shuffle(exit_positions)`

InoMurko:
The way you approached this with `Tasks` and anonymous functions I  think is not the best use case for `Tasks`. I would rather see a pool of GenServers that perform a set of operations in a clear and readable way. This will be in no way easy to debug. Also, please limit the fragmentation between tests and implementation modules, if there's no need. Don't write code for "future" use.

pnowosie:
From :pr: description: `"...watcher will work on DOS condition so..."`
can you please extend what kind of DOD is tested here: malicious agent is sending large number of standard exits (part of them are invalid)

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -0,0 +1,120 @@
...
+    worker = fn exit_positions ->
+      Enum.map(exit_positions, fn position ->
+        get_exit_data(position) |> valid_exit_data()
+      end)
+    end
+
+    Task.async(fn ->
+      exit_positions = Enum.shuffle(exit_positions)
+      {time, valid?} = :timer.tc(fn -> worker.(exit_positions) end)
+      %{time: time, correct: Enum.count(valid?, & &1), error: Enum.count(valid?, &(!&1))}
```
pnowosie:
```suggestion
      %{time: time, valids_count: Enum.count(valid?, & &1), errors_count: Enum.count(valid?, &(!&1))}
```

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -0,0 +1,120 @@
...
+    error -> error
+  end
+
+  def generate_users(size, opts \\ [initial_funds: trunc(:math.pow(10, 18))]) do
+    async_generate_user = fn _ -> Task.async(fn -> generate_user(opts) end) end
+
+    1..size
+    |> Enum.chunk_every(10)
+    |> Enum.map(fn chunk ->
+      Enum.map(chunk, async_generate_user)
```
pnowosie: with `chunk` piped it will be more readable (minor)

apps/omg_performance/test/omg_performance/integration/byzantine_event_test.exs
```diff
@@ -0,0 +1,66 @@
...
+
+  @tag fixtures: [:contract, :child_chain, :watcher]
+  test "time response for asking for exit data", %{contract: %{contract_addr: contract}} do
+    dos_users = 10
+    ntx_to_send = 100
+    spenders = ByzantineEvents.generate_users(2)
+    exit_per_dos = length(spenders) * ntx_to_send
+    total_exits = length(spenders) * ntx_to_send * dos_users
+
+    IO.puts("""
```
pnowosie: Do we need console output in tests? Hmm?

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -0,0 +1,120 @@
...
+
+    Stream.iterate(1, &(&1 + 1))
+    |> Stream.map(&get_block!(&1 * interval))
+  end
+
+  defp get_block!(blknum) do
+    child_chain_url = Application.get_env(:omg_watcher, :child_chain_url)
+
+    {:ok, block} =
+      Enum.find(
```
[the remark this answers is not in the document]
pnowosie: Also it looks like `Eth.WaitFor.repeat_until_ok` ...

apps/omg_performance/lib/omg_performance/mutation.ex
```diff
@@ -0,0 +1,168 @@
...
+        else: outputs
+
+    Transaction.Payment.new(input |> payment_input, outputs |> payment_output, metadata)
+  end
+
+  def random_binary(size) do
+    for _ <- 1..size, into: <<>>, do: <<:random.uniform(256)>>
+  end
+
+  def sign(raw_tx, recovered_tx, utxos, users) do
```
pnowosie:
```suggestion
  defp sign(raw_tx, recovered_tx, utxos, users) do
```

apps/omg_performance/lib/omg_performance/mutation.ex
```diff
@@ -0,0 +1,168 @@
...
+    map_owners =
+      [
+        utxos
```
pnowosie: Refactor to calculate this once above

apps/omg_utils/lib/omg_utils/http_rpc/client.ex
```diff
@@ -39,6 +39,20 @@ defmodule OMG.Watcher.HttpRPC.Client do
     |> decode_response()
   end
 
+  def get_exit_data(encoded_position, url) do
```
pnowosie: Refactor to `call(params, path, url)`

apps/omg_watcher/test/fixtures.exs
```diff
@@ -52,11 +52,18 @@ defmodule OMG.Watcher.Fixtures do
...
+    mix_env = Application.get_env(:omg_child_chain, :mix_env, to_string(Mix.env()))
 
     exexec_opts_for_mix = [
       stdout: :stream,
       cd: Application.fetch_env!(:omg_watcher, :umbrella_root_dir),
-      env: %{"MIX_ENV" => to_string(Mix.env())},
+      env: %{
+        "MIX_ENV" => mix_env,
+        "DD_DISABLED" => "true",
+        "APP_ENV" => "dev",
```
pnowosie: :question:  `"dev"`
```suggestion
        "APP_ENV" => mix_env,
```

apps/omg_utils/lib/omg_utils/http_rpc/response.ex
```diff
@@ -142,4 +138,13 @@ defmodule OMG.Utils.HttpRPC.Response do
         "watcher"
     end
   end
+  
+  defp get_vsn(service) when is_atom(service) do
+    with {:ok, vsn} <- :application.get_key(service, :vsn) do
+      vsn
+    else
+      _ ->
+        '0.0.0'
```
InoMurko: this is not the right approach.

pnowosie: Problem is that `:code.is_loaded(OMG.ChildChainRPC)` isn't right approach in all contexts we're running rpc servers, which includes:
* simple perf test where not full child_chain app is provisioned, but rather the pieces test need,
* ByzantineEventsTest - fixture `omg_watcher` which starts Watcher via Exexec

What's here just lowers the impact of bug introduced in `add_version`. Let's find a better solution in separate issue.

pnowosie:
Splitting to deliver standard exits part in #998 with a test under a load.

PR #900, opened 2019-08-20
feature: expose child chain alarms via alarm.get

InoMurko:
///

## Overview

Read the raised alarms on an Child Chain endpoint _http://child_chain/alarm.get_

## Changes

- Added a new route!
- Copied tests from Watcher
- Fixed a few other issues
- Decoupled Child Chain from Child Chain RPC so that coupling occurs in the same way as in Watcher. RPC->APP (this allowed me to remove slow dynamic module invocation `apply/3`)

## Testing

Wrote a few tests that should cover all behaviors.

### DIFF
--- a/apps/omg_child_chain/lib/omg_child_chain/monitor.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/monitor.ex
@@ -52,9 +52,8 @@ defmodule OMG.ChildChain.Monitor do
   def init([alarm_module, children_specs]) do
     install()
     Process.flag(:trap_exit, true)
-
+    _ = Logger.info("Starting #{__MODULE__}")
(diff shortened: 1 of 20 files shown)


apps/omg_child_chain/lib/omg_child_chain/monitor.ex
```diff
@@ -144,8 +143,8 @@ defmodule OMG.ChildChain.Monitor do
     alarms = alarm_module.all()
 
     alarms
-    |> Enum.find(fn x -> match?(%{id: :ethereum_client_connection}, x) end)
-    |> is_map()
+    |> Enum.find(fn x -> match?(:ethereum_client_connection, elem(x, 0)) end)
```
InoMurko: provided tests for these changes!

PR #902, opened 2019-08-21
Run integration ci

InoMurko:
Overview

`run mix test --only integration every week day at 15:15 UTC.`

## Changes

- Circle nightly flow

## Testing

/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -390,7 +390,7 @@ workflows:
   nightly:
     triggers:
       - schedule:
-          cron: "15 15 * * 1-5"
+          cron: "15 18 * * 1-5"
           filters:
             branches:
               only:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -203,6 +203,26 @@ jobs:
                 (retval=$? && curl -k https://coveralls.io/webhook?repo_token=$COVERALLS_REPO_TOKEN -d "payload[build_num]=$CIRCLE_WORKFLOW_WORKSPACE_ID&payload[status]=done" && exit $retval)
             fi
 
+  integration_tests:
```
pdobacz: let's add a comment, like:

```
# Slow, serial integration test, run nightly. Here to make sure the standard `mix test --only integration` works
```

Issue #908, opened 2019-08-22
Add deployment markers to Datadog and Sentry


InoMurko:
Datadog integration:
```
curl -X POST -H "Content-type: application/json" -d "{"title": "Starting Service", "text": "Starting Watcher with git SHA f2762839cf435086aee46c342baea6f4e1e095d8 on cluster Samrong", "alert_type": "info" }" "https://app.datadoghq.com/api/v1/events?api_key="
```

```
data that needs to be sent
git SHA, APP_ENV, HOSTNAME, ETHEREUM_NETWORK, APP_NAME
```

https://docs.sentry.io/api/releases/post-release-deploys/

InoMurko:
hey @jbunce upon talking with @achiurizo we think this really belongs in circle ci configuration and not in distllery (elixir) release hooks. 

If we include this as release hooks then every deployed instance will notify Datadog and Sentry which we think we don't really want...

jbunce:
Gents,

Ultimately it's up to you how you want to manage this. If this were an application I built I'd want to know exactly which code is running to troubleshoot production services. If you think you have the data you need via tooling in the other environments that mean you can skip this, than that's your prerogative.

InoMurko:
To be able to send only one event for deploy, that would need to come not from elixir, but deployment scripts.

InoMurko:
My part here is done. @jbunce if you have something then let's keep it open. Otherwise let's close it.

PR #912, opened 2019-08-26
908 deployment markers2

InoMurko:
https://github.com/omisego/elixir-omg/issues/908

## Overview

Adds detailed information about the GIT SHA the code was deployed from in every HTTP response.

## Changes

- version SHA gets written to VERSION file
- when a release gets created, the VSN (version 0.2.2+SHA) gets baked into the application vsn
- response building reads that VSN and includes it in response
- we don't want unintentional commits of VERSION file... added a precoomit hook that you can add to your hooks by running make init
## Testing

Updated tests to expect custom version node.


apps/omg_child_chain_rpc/lib/omg_child_chain_rpc/web/controllers/block.ex
```diff
@@ -20,14 +20,11 @@ defmodule OMG.ChildChainRPC.Web.Controller.Block do
   use OMG.ChildChainRPC.Web, :controller
 
   alias OMG.ChildChain
-  alias OMG.ChildChainRPC.Web.View
 
   def get_block(conn, params) do
     with {:ok, hash} <- expect(params, "hash", :hash),
          {:ok, block} <- ChildChain.get_block(hash) do
-      conn
```
InoMurko: I'm not saying this is better or worse, but at least now it's consistent with watcher's controllers.

apps/omg_db/mix.exs
```diff
@@ -50,4 +50,19 @@ defmodule OMG.DB.MixProject do
       _ -> []
     end
   end
+
+  defp version_and_git_revision_hash do
```
pdobacz: since this code grew a bit, maybe it's worth to DRY it out?

But maybe more importantly: I'm not super-convinced if this isn't going to be difficult to work with in practice - VERSION file is going to be always present in the git changesets, it will pop up as conflicting very often, and it still will be rather easy to push out a wrong one eventually.

Isn't there an alternative to:
  - leave `VERSION` as it is (just `0.2.2` etc.)
  - append the git sha later - e.g. fetch it during compilation and store it in the `config` and serve from there?

apps/omg/mix.exs
```diff
@@ -46,4 +46,19 @@ defmodule OMG.MixProject do
       {:plug_cowboy, "~> 1.0", only: [:dev, :test]}
     ]
   end
+
+  defp version_and_git_revision_hash do
+    {rev, _i} = System.cmd("git", ["rev-parse", "HEAD"])
+    sha = String.replace(rev, "\n", "") |> Kernel.binary_part(0, 7)
+    version = String.trim(File.read!("../../VERSION"))
+
+    updated_ver =
```
pdobacz: https://hexdocs.pm/elixir/Version.html may help here perhaps?

apps/omg/mix.exs
```diff
@@ -46,4 +46,19 @@ defmodule OMG.MixProject do
       {:plug_cowboy, "~> 1.0", only: [:dev, :test]}
     ]
   end
+
+  defp version_and_git_revision_hash do
+    {rev, _i} = System.cmd("git", ["rev-parse", "HEAD"])
+    sha = String.replace(rev, "\n", "") |> Kernel.binary_part(0, 7)
```
pdobacz: `git rev-parse --short=7 HEAD` can help make this smaller maybe?

InoMurko: yeah, but not by much :)
string.replace/3 is still needed.

apps/omg/mix.exs
InoMurko: Elixir 1.9.1 sadly...

apps/omg_db/mix.exs
InoMurko: I don't think there's any place where we could put it that would work with how releases fetch versions when they build applications. We had a similar solution before and it didn't work out. 

For the second part:
I was thinking the same and started out with a git pre-commit hook that always puts VERSION back from master. But you can still change it if your run it with `--no-verify`

Let me see what else I can do...

apps/omg/mix.exs
pdobacz: https://hexdocs.pm/elixir/1.8.0/Version.html :)

apps/omg_utils/lib/omg_utils/http_rpc/response.ex
```diff
@@ -16,6 +16,8 @@ defmodule OMG.Utils.HttpRPC.Response do
   Serializes the response into expected result/data format.
   """
   alias OMG.Utils.HttpRPC.Encoding
+  @version String.replace(elem(System.cmd("git", ["rev-parse", "--short=7", "HEAD"]), 0), "\n", "")
```
pdobacz: names clash here - this here is the `sha`, by version I take it we understand the entire semver `v0.2.2+12a356b` string

PR #916, opened 2019-08-29
fix: set tags programatically

InoMurko:
DD support has been rather unhelpful in mitigating our infrastructure->tags->datadog problem. 
Just to move us forwards adding this "hack"

## Overview

Adding tags to metrics depending on which application is running

## Changes

- Get the application mode
- Store it in process dictionary
- retrieve it, send it to DD

### DIFF
--- a/apps/omg_status/lib/omg_status/metric/vmstats_sink.ex
+++ b/apps/omg_status/lib/omg_status/metric/vmstats_sink.ex
@@ -29,9 +29,38 @@ defmodule OMG.Status.Metric.VmstatsSink do
 
   defp base_key, do: Application.get_env(:vmstats, :base_key)
 
-  def collect(:counter, key, value), do: _ = Datadog.set(key, value)
+  def collect(:counter, key, value), do: _ = Datadog.set(key, value, tags: ["application:#{get_application_mode()}"])
 
-  def collect(:gauge, key, value), do: _ = Datadog.gauge(key, value)
+  def collect(:gauge, key, value), do: _ = Datadog.gauge(key, value, tags: ["application:#{get_application_mode()}"])
 
-  def collect(:timing, key, value), do: _ = Datadog.timing(key, value)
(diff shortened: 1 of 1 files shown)


apps/omg_status/lib/omg_status/metric/vmstats_sink.ex
```diff
@@ -29,9 +29,38 @@ defmodule OMG.Status.Metric.VmstatsSink do
...
   defp base_key, do: Application.get_env(:vmstats, :base_key)
 
-  def collect(:counter, key, value), do: _ = Datadog.set(key, value)
+  def collect(:counter, key, value), do: _ = Datadog.set(key, value, tags: ["application:#{get_application_mode()}"])
 
-  def collect(:gauge, key, value), do: _ = Datadog.gauge(key, value)
+  def collect(:gauge, key, value), do: _ = Datadog.gauge(key, value, tags: ["application:#{get_application_mode()}"])
 
-  def collect(:timing, key, value), do: _ = Datadog.timing(key, value)
+  def collect(:timing, key, value), do: _ = Datadog.timing(key, value, tags: ["application:#{get_application_mode()}"])
```
achiurizo: I'm probably missing context, but why couldn't we just pick this value up from the environment? We already specify a value here at `ELIXIR_SERVICE` which returns `CHILDCHAIN` or `WATCHER`.

jbunce:
This will allow us to filter by `vmstats.*` metrics and the application name. But it kinda forces us into using distinct cluster environments for each deployment.

apps/omg_status/lib/omg_status/metric/vmstats_sink.ex
InoMurko: Because the elixir service environment variable is deprecated (makes no sense) if you’re running a release. Which we don’t yet. But it’s easier if I don’t rely on it.

InoMurko:
@jbunce do you want a change?

jbunce:
Let's roll with this for the time being

Issue #923, opened 2019-09-02
Perftest does not honour the `child_chain` value


jbunce:
Follows on from an internal discussion.

https://github.com/omisego/elixir-omg/blob/6df52caa9f7aee079cb9b7f78e63ee8d1d44a80d/apps/omg_performance/lib/omg_performance/sender_server.ex#L192 suspected as the cause.

```
iex(17)> OMG.Performance.start_extended_perftest(10_000, alices, contract_addr, %{destdir: ".", geth: System.get_env("ETHEREUM_RPC_URL"), child_chain: "http://childchain-samrong"})
2019-09-02 09:50:09.371 [warn] module=OMG.Watcher.HttpRPC.Adapter function=rpc_post/3 ⋅Child chain rpc post "http://localhost:9656/transaction.submit" failed with {:error, %HTTPoison.Error{id: nil, reason: :econnrefused}}⋅
```

pdobacz:
Thanks!

I'd vote to treat it as an opportunity to clean the way the URLs are delivered (there's two - ethereum rpc and child chain). Let's keep it as simple as possible - probably just rely on the configs and drop from the options delivered in the invocation. Nowadays both URLs can be given via ENV vars.

InoMurko:
Let's do that after we introduce releases? They already massively clean configs and how values are to be passed in.

- [ ] straight up fix `honour the `child_chain` value ` for perf tests
- [ ] cleanup configs after https://github.com/omisego/elixir-omg/pull/884

pdobacz:
should be fixed after #1106. Let's close and open a new issue if problem persists

PR #924, opened 2019-09-02
fix: pass child chain URL

InoMurko:
https://github.com/omisego/elixir-omg/issues/923

## Overview

Passing child chain url as an argument or as a env var was not honored  

## Changes
- 
## Testing

Test are already there. I just adapted them to new changes.


jbunce:
LGTM. Are there no tests for `OMG.Performance`?

InoMurko:
@jbunce yes they are. performance_test.exs

PR #926, opened 2019-09-03
Triage Sobelow findings

thec00n:
## Overview

This PR triages all sobelow findings in apps. 

## Remaining

I am not convinced that the usage of binary_to_term in the below instances is not a security issue. In all of the cases data from the db is deserialized. Even running into a DOS condition could have very negative consequences. Deserialization bugs can often lead to RCE, I am just thinking of JAVA and PHP. Can we explore alternatives? 

## References 
http://erlang.org/pipermail/erlang-questions/2001-July/003372.html (dated)
https://hexdocs.pm/sobelow/Sobelow.Misc.BinToTerm.html#content

```
[31mMisc.BinToTerm: Unsafe `binary_to_term` - High Confidence[0m
File: apps/omg_db/lib/omg_db/leveldb/core.ex
Line: 114
Function: decode_response:111
Variable: encoded

-----------------------------------------------

[31mMisc.BinToTerm: Unsafe `binary_to_term` - High Confidence[0m
File: apps/omg_db/lib/omg_db/rocksdb/core.ex
Line: 115
Function: decode_response:105
Variable: encoded

-----------------------------------------------

[31mMisc.BinToTerm: Unsafe `binary_to_term` - High Confidence[0m
File: apps/omg_db/lib/omg_db/rocksdb/core.ex
Line: 111
Function: decode_response:105
Variable: encoded

-----------------------------------------------
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,6 @@ env/
 
 # Docker
 src/
+
+# Sobelow
+.sobelow
--- a/apps/omg_child_chain/lib/omg_child_chain/application.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/application.ex
@@ -54,7 +54,7 @@ defmodule OMG.ChildChain.Application do
     end)
   end
 
-  # Only set once during bootup. cookie value retrieved from ENV. 
+  # Only set once during bootup. cookie value retrieved from ENV.
   # sobelow_skip ["DOS.StringToAtom"]
   defp set_cookie(cookie) when is_binary(cookie) do
     cookie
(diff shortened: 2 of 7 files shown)


InoMurko:
Does sobelow_skip come in file configuration as well? I prefer it there if it’s an option. Same as we have with dialyzer.

thec00n:
We ignore Config.HTTPS completely but for instance are interested in Traversal issues therefore we have to skip them on a function level.

InoMurko:
If we don’t trust decode_response, does that mean we don’t trust inserts as well? 
Then this needs to be fixed at the input, not output. 
(Though in my opinion it’s not needed).

PR #927, opened 2019-09-03
fix: safe option in binary to term

InoMurko:
binary_to_term has a `safe` switch for when you deserialize untrusted inputs (which is not the case in the places we're using it) but just so no one stores atoms in the DB ... here it is!

## Overview

Added a :safe argument to `binary_to_term/2`

## Changes

- http://erlang.org/doc/man/erlang.html#binary_to_term-2

## Testing

/


InoMurko:
Why is that? I presume we're doing what sobelow wants us to do and would emit warnings for this?

thec00n:
We want to suppress the warnings, since we have added the safe flag and we are going to review the input filtering so we can be sure that only sane input reaches the functions. Sobelow showing us no more errors allows us to define as part of the CI and fail the build if it finds sth new.

PR #932, opened 2019-09-04
updated testing docs

jrhite:
Updated testing docs regarding some gotchas...

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,11 +402,18 @@ mix deps.compile plasma_contracts
 
 You can setup the docker environment to run testing and development tasks:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -402,12 +402,20 @@ mix deps.compile plasma_contracts
 
 You can setup the docker environment to run testing and development tasks:
 
+(Note: If they exist, be sure to delete the `deps` and `_build` dirs from your host drive. Otherwise these dirs will
+get mapped into the docker container and can create build discrepancies and breakages.)
```
achiurizo: I think you can run `mix deps.clean` to remedy this too right?

jrhite: Ah yeah, good point a `mix clean && mix deps.clean` would do the trick. I'll add this.

PR #944, opened 2019-09-10
Functional Tests Posts on Slack on Failure

kendricktan:
##  Overview
Currently when the functional tests fail, it fails silently as the deployment happens _before_ the functional tests.

## Changes
`test_runner.py` now posts to slack on failure

## Testing
Manual testing - `python .circleci/test_runner.py`

### DIFF
--- a/.circleci/test_runner.py
+++ b/.circleci/test_runner.py
@@ -104,15 +104,18 @@ def start_workflow():
         check_job_result(envs['TEST_RUNNER_SERVICE'], job_id)
 
         # Post success to slack
+        success_text = """
+        Functional tests for elixir-omg completed successfully!\n{}
+        """.format(
+            envs['CIRCLE_BUILD_URL']
+        )
         for webhook in [
             envs['TEST_RUNNER_SLACK_WEBHOOK_01'],
             envs['TEST_RUNNER_SLACK_WEBHOOK_02']
         ]:
             if webhook is not None:
                 requests.post(webhook, json={
-                    'text': 'Functional tests for elixir-omg completed successfully!\n{}'.format(
-                        envs['CIRCLE_BUILD_URL']
-                    )
+                    'text': success_text
                 })
 
     except Exception as e:
--- a/.circleci/test_runner.py
+++ b/.circleci/test_runner.py
@@ -102,17 +102,23 @@ def start_workflow():
         logging.critical(
             f'Test runner service crashed, exception: {e}\ntraceback: {tb}')
 
+        # Formatted text
         circleci_build_url = envs['CIRCLE_BUILD_URL']
+        pretext = """
+        Test runner service crashed, exception: `{}`
+        CircleCI build results: {}
+        """.format(e, circleci_build_url)
         # Sends notification to the web slack hooks
-        for webhook in [envs['TEST_RUNNER_SLACK_WEBHOOK_01'], envs['TEST_RUNNER_SLACK_WEBHOOK_02']]:
+        for webhook in [
+            envs['TEST_RUNNER_SLACK_WEBHOOK_01'],
+            envs['TEST_RUNNER_SLACK_WEBHOOK_02']
+        ]:
             if webhook is not None:
                 requests.post(webhook, json={
                     'attachments': [
                         {
                             'title': 'Traceback',
-                            'pretext': 'Test runner service crashed, exception: `{}`\nCircleCI build results: {}'.format(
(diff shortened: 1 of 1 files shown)


.circleci/test_runner.py
```diff
@@ -3,6 +3,7 @@
 import os
 import sys
 import time
+import traceback
```
jbunce: Imports should be alphabetical

.circleci/test_runner.py
```diff
@@ -17,11 +18,8 @@ def create_job(test_runner: str) -> str:
             "cwd": "/home/omg/omg-js"
         }
     }
-    try:
-        request = requests.post(test_runner + '/job', json=payload)
-    except ConnectionError:
-        logging.critical('Could not connect to the test runner')
-        sys.exit(1)  # Return a non-zero exit code so CircleCI fails
+
+    request = requests.post(test_runner + '/job', json=payload)
```
jbunce: Why remove the try/except block?

.circleci/test_runner.py
```diff
@@ -82,11 +83,40 @@ def get_envs() -> dict:
...
+        check_job_result(envs['TEST_RUNNER_SERVICE'], job_id)
+
+    except Exception as e:
+        tb = traceback.format_exc()
+        logging.critical(
+            f'Test runner service crashed, exception: {e}\ntraceback: {tb}')
+
+        circleci_build_url = envs['CIRCLE_BUILD_URL']
+        # Sends notification to the web slack hooks
+        for webhook in [envs['TEST_RUNNER_SLACK_WEBHOOK_01'], envs['TEST_RUNNER_SLACK_WEBHOOK_02']]:
```
jbunce: Please run flake8 oon this. Way over 80 chars.

.circleci/test_runner.py
kendricktan: Each time there is a `try/catch` block it logs it and `sys.exit(1)`. There was a lot of code duplication for the posting to slack logic for each exception. 

And since the script is just going to exit and not execute anything anymore after each `try/catch`, it made more sense to just raise an exception there (with the relevant error code), and then have a `try/catch` around `start_workflow` (one of the main entrypoints), and handle the logic for exception handling and posting to slack from there - less code to maintain but still has the same logic

jbunce: With the `try/except` block you get the ability to add a log message so that any user checking the CircleCI logs can see why it failed.

.circleci/test_runner.py
```diff
@@ -82,11 +96,61 @@ def get_envs() -> dict:
...
+        for webhook in [
+            envs['TEST_RUNNER_SLACK_WEBHOOK_01'],
+            envs['TEST_RUNNER_SLACK_WEBHOOK_02']
+        ]:
+            if webhook is not None:
+                requests.post(webhook, json={
+                    'text': success_text
+                })
+
+    except Exception as e:
```
jbunce: Bare `Exception` should not be used, and the `try` statement should identify and catch one condition only.

.circleci/test_runner.py
```diff
@@ -73,7 +77,17 @@ def get_envs() -> dict:
     test_runner = os.getenv('TEST_RUNNER_SERVICE')
     if test_runner is None:
         logging.critical('Test runner service ENV missing')
-        sys.exit(1)  # Return a non-zero exit code so CircleCI fails
+        raise Exception('Test runner service ENV missing')
+
+    envs['TEST_RUNNER_SLACK_WEBHOOK_01'] = os.getenv(
```
jbunce: What's Webhook 01 and Webhook 02?

kendricktan: Its for #dev-plasma and #plasma-notifications I could make that into csv value if you prefer

.circleci/test_runner.py
kendricktan: I've chosen to remove the `try/catch` exception handling in the `create_job`, `check_job_completed`, `check_job_result`, ... and place them into the main function `start_workflow` and have that function act as the main function handler for all of them.

The rational is because the functions `check_job_completed`, `check_job_result`, etc are dealing with IO, and throws a bunch of IO related exceptions such as `TimeOutException`, `ConnectionException`, as well as the exception of the possibility of the job failing. Having a try/catch for each IO scenario in each function is overly verbose for a simple notifications script, imo and is overly complicated when our main goal is to surface information / notify the #dev-plasma team when the tests fail.

Issue #936, opened 2019-09-09
Efficient calculation of business metrics


pdobacz:
Larger history of the child chain leads the expensive recalculation of the metrics (`balance` and `unique_users`) to block the components processing transactions and requests for too long.

We need either a way to do the calculations much lighter or non-blocking.

Spotted in https://github.com/omisego/devops/issues/142

pdobacz:
First suggestion from @InoMurko :

> What I would propose is to:
> 
> 1. extract OMG.State storage (the one currently held in process state) into a behavior (look at OMG.DB). That would allow us to swap storage drivers.
> 2. Implement ETS storage on top of 1.
> 3. Keep the ETS table protected (writes are only permitted to ETS owner (omg.state) and let reads happen separately in a process just for metrics.

paulperegud:
Just a word of warning on moving things to ETS. It does not support atomic writes for more than one entry at a time. Thus it would expose intermediate states of UTXO set to readers (other than the privileged writer). Analyze carefully if this is acceptable.

InoMurko:
Thanks @paulperegud. I think that if we’re limiting ourselves to metrics in this particular use case we should be fine. The question is... are we seeing the whole picture?

InoMurko:
As discussed, there are two other options proposed by @paulperegud and @pdobacz:
Quote:

@pdobacz 
`
there were 2 quick fixes (that don't involve architecture digging) that we discussed briefly with @pepesza - 1) was to fire a Task.async to calculate,  sending the entire utxos Map (presumably better than calculation in OMG.State process) - 2) was... I don't remember :disappointed: @pepesza? that was your pitch
if not, I'm proposing a more subtle switch off: 3) if size(utxos)>threshold, dont: calculate_metrics()
`


I've tried the following experiment. I've created a map with 1M UTXOs and calculated the amount of copying between processes (using http://erlang.org/doc/efficiency_guide/processes.html#id67714):
:erts_debug.flat_size/1

```
:erts_debug.flat_size(large_map)
40720868
```

the problem I think is that process heap of (task is sort of a spawn/1) needs to constantly grow until it fits the large term. which is quite a lot.

`
The default initial heap size of 233 words is quite conservative to support Erlang systems with hundreds of thousands or even millions of processes. The garbage collector grows and shrinks the heap as needed.
`

pthomalla:
Enum.uniq is quite inefficiently written, 
We can still improve it a bit, which should speed up the counting of metrics.

```
[...]

PR #945, opened 2019-09-10
fix: move measurement into separate process

InoMurko:
https://github.com/omisego/elixir-omg/issues/936

## Overview

Lets try a solution and see how it behaves in the wild, metrics are available in Datadog.

## Changes

- Spawn a process that does a complete copy of OMG.State state and performs the necessary computation and dies. 

## Testing

/

### DIFF
--- a/apps/omg/lib/omg/state/measure.ex
+++ b/apps/omg/lib/omg/state/measure.ex
@@ -27,9 +27,14 @@ defmodule OMG.State.Measure do
   def supported_events, do: @supported_events
 
   def handle_event([:process, OMG.State], _, %Core{} = state, _config) do
-    Enum.each(MeasurementCalculation.calculate(state), fn
-      {key, value} -> _ = Datadog.gauge(name(key), value)
-      {key, value, metadata} -> _ = Datadog.gauge(name(key), value, tags: [metadata])
-    end)
+    execute = fn ->
+      Enum.each(MeasurementCalculation.calculate(state), fn
+        {key, value} -> _ = Datadog.gauge(name(key), value)
+        {key, value, metadata} -> _ = Datadog.gauge(name(key), value, tags: [metadata])
+      end)
+    end
+
+    _ = Task.start(execute)
+    :ok
   end
 end
--- a/apps/omg/lib/omg/state/measure.ex
+++ b/apps/omg/lib/omg/state/measure.ex
@@ -34,6 +34,8 @@ defmodule OMG.State.Measure do
       end)
     end
 
+    # TODO proper fix! this is a very hackish approach to get measurements off the back
+    # of OMG State
     _ = Task.start(execute)
     :ok
   end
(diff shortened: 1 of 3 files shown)


apps/omg/lib/omg/state/measure.ex
```diff
@@ -27,9 +27,14 @@ defmodule OMG.State.Measure do
...
-      {key, value, metadata} -> _ = Datadog.gauge(name(key), value, tags: [metadata])
-    end)
+    execute = fn ->
+      Enum.each(MeasurementCalculation.calculate(state), fn
+        {key, value} -> _ = Datadog.gauge(name(key), value)
+        {key, value, metadata} -> _ = Datadog.gauge(name(key), value, tags: [metadata])
+      end)
+    end
+
+    _ = Task.start(execute)
```
pdobacz: three things to consider:
1/ leave a comment here telling our future selves that we're sending a beefy term over, we acknowledge it and ask for forgiveness. Also one word why we're doing it like that (to not block the owner of `state`)
2/ pattern match `{:ok, _} = `. Not sure about this, but intuitively I think I'd prefer this to crash if task cannot be started. I'm assuming such crash would have us notified but wouldn't bring everything else down
3/ `start_link` maybe? for same reasons as 2/ - reasoning: previously if `.calculate` would fail, this process would crash.

paulperegud: > leave a comment here

+1

> I'm assuming such crash would have us notified but wouldn't bring everything else down

In this particular case it would bring down State process. Same goes for linking.

pdobacz: > In this particular case it would bring down State process. Same goes for linking.

ah doh, of course, brainfart on my part. It still might or might not make sense to crash that State in case of failure of either 2/ or 3/ - this is what used to be true for non-Task approach.

Leaving up to you, what's better.

apps/omg/lib/omg/state/measure.ex
```diff
@@ -27,9 +27,14 @@ defmodule OMG.State.Measure do
   def supported_events, do: @supported_events
 
   def handle_event([:process, OMG.State], _, %Core{} = state, _config) do
-    Enum.each(MeasurementCalculation.calculate(state), fn
-      {key, value} -> _ = Datadog.gauge(name(key), value)
-      {key, value, metadata} -> _ = Datadog.gauge(name(key), value, tags: [metadata])
-    end)
+    execute = fn ->
+      Enum.each(MeasurementCalculation.calculate(state), fn
```
paulperegud: This will copy all of state into other process, doubling memory usage. `Enum.each` and other `Enum` functions are notorious for doing `Enum.to_list` before doing any operations on maps. That would effectively triple memory usage. We will see a very nice spike on datadog every minute or so.

InoMurko: Yeah. I consider this a POC solution. Just to see how it'll behave!

PR #949, opened 2019-09-10
Use PlasmaFramework & friends to deposit, tx, start SE, finalize SE

### LINKED ISSUE #948: Integrate with PlasmaFramework (ALD) for deposits and txs
(from @achiurizo)

Should enable an `elixir-omg`-driven internal demo for deposits and txs (submitBlock only), sitting on top of the "new" `plasma-contracts` conforming to the current implementation of PlasmaFramework (ALD).

Dirty implementation to start off with & demo, to clean later.

pdobacz:
Fixes #948 

EDIT originally this was supposed to be deposits & txs only, but expanded this to also do start exit and finalize exit for at least minimum assurance that things make sense

## Overview

Switches the integration test to use PlasmaFramework.
(rather) Dirty, refactors pending after it's proven to work.

## Changes

- use solc 0.5
- pull in new contracts from a branch
- compile, link & deploy them, collect necessary contract addresses
- switch deposit stuff to new appropriate contract
- switch submitBlock stuff to new appropriate contract
- switch startExit and finalize for standard exit to new appropriate contract
- switch to new transaction data model an a couple of new semantics (see commit list)
- switched to a branch in `plasma_contracts` to use contract code that has been e2e tested using the "old" contract test suite (aka Python test suite)
- adapt to new contracts' API by **storing** exit_id instead of recreating it by calling `geth`

## Testing

Some tests will be skipped. Deposit, submitBlock, start SE & finalize SE related integration tests should pass.

Tests skipped are SE challenges, IFEs and EIP712 conformance


apps/omg/lib/omg/state/transaction.ex
```diff
@@ -48,26 +45,21 @@ defmodule OMG.State.Transaction do
 
   defmacro is_metadata(metadata) do
     quote do
-      unquote(metadata) == nil or (is_binary(unquote(metadata)) and byte_size(unquote(metadata)) == 32)
+      is_binary(unquote(metadata)) and byte_size(unquote(metadata)) == 32
```
InoMurko: wouldn't just byte_size == 32 be enough?

pdobacz: Ha, I always wondered and finally checked. For non-binaries it seems to throw, so I suppose it is enough

apps/omg/lib/omg/state/transaction/payment.ex
```diff
@@ -125,14 +109,27 @@ defmodule OMG.State.Transaction.Payment do
...
   rescue
     _ -> {:error, :malformed_inputs}
   end
 
+  # messy, see comments on the abstract output/input fixing this properly
+  defp from_new_rlp_input(binary_input) when is_binary(binary_input) do
+    binary_input
+    |> :binary.decode_unsigned(:big)
+    |> case do
+      0 ->
```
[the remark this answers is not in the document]
pdobacz: good catch, this is the leftover after the battle to get the integration running.

In fact I think I should remove the "gap" checking in inputs/outputs and the corresponding tests, this sounds like obsoleted in the new contracts

apps/omg_eth/lib/omg_eth/root_chain.ex
```diff
@@ -364,6 +372,22 @@ defmodule OMG.Eth.RootChain do
...
+  """
+  @spec maybe_fetch_addr!(%{atom => Eth.address()} | nil, atom) :: Eth.address()
+  def maybe_fetch_addr!(contract, name) do
+    contract[name] || from_hex(Application.fetch_env!(:omg_eth, :contract_addr)[name])
+  end
+
+  @doc """
+  Hexifies the entire contract map, assuming `contract_map` is a map of `%{atom => raw_binary_address}`
+  """
+  def contract_map_to_hex(contract_map),
```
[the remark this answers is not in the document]
pdobacz: I was trying to reflect the `to_hex` from `Encoding`, does this help?

apps/omg_eth/test/omg_eth/release_tasks/set_contract_test.exs
```diff
@@ -214,7 +217,7 @@ defmodule OMG.Eth.ReleaseTasks.SetContractTest do
 
   defp handle(conn) do
     body =
-      "{\"authority_addr\":\"authority_address_value\",\"contract_addr\":\"contract_address_value\",\"txhash_contract\":\"txhash_contract_value\"}"
+      "{\"authority_addr\":\"authority_address_value\",\"contract_addr\":#{@contract_addresses_json},\"txhash_contract\":\"txhash_contract_value\"}"
```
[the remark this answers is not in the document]
pdobacz: as in, for the outer json (`body` variable)? i suppose so

apps/omg_eth/test/support/deployer.ex
```diff
@@ -21,32 +21,91 @@ defmodule OMG.Eth.Deployer do
...
+        [
+          plasma_framework: pf_addr,
+          eth_vault: v1_addr,
+          erc20_vault: v2_addr,
+          output_guard_handler: ogh_addr,
+          spending_condition: sc_addr
+        ],
+        opts
+      ) do
+    args = [{:address, pf_addr}, {:address, v1_addr}, {:address, v2_addr}, {:address, ogh_addr}, {:address, sc_addr}]
```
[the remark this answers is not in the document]
pdobacz: yep, leftover missed, let's expand :+1:

apps/omg_watcher/lib/omg_watcher/exit_processor/core.ex
```diff
@@ -53,12 +53,13 @@ defmodule OMG.Watcher.ExitProcessor.Core do
...
   @type contract_piggyback_offset_t() :: 0..7
 
-  defstruct [:sla_margin, exits: %{}, in_flight_exits: %{}, competitors: %{}]
+  defstruct [:sla_margin, exits: %{}, in_flight_exits: %{}, exit_ids: %{}, competitors: %{}]
 
   @type t :: %__MODULE__{
           sla_margin: non_neg_integer(),
           exits: %{Utxo.Position.t() => ExitInfo.t()},
           in_flight_exits: %{Transaction.tx_hash() => InFlightExitInfo.t()},
+          exit_ids: %{non_neg_integer() => Utxo.Position.t() | Transaction.tx_hash()},
```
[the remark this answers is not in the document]
pdobacz: that's the intention. I'll leave a TODO to re-tackle when doing IFE support maybe. I'm not entirely sold on this mapping

apps/omg_watcher/lib/omg_watcher/exit_processor/exit_info.ex
```diff
@@ -25,27 +25,26 @@ defmodule OMG.Watcher.ExitProcessor.ExitInfo do
 
   require Utxo
 
-  @enforce_keys [:amount, :currency, :owner, :is_active, :eth_height]
+  @enforce_keys [:amount, :currency, :owner, :exit_id, :is_active, :eth_height]
```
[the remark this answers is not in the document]
pdobacz: I want all keys enforced always c.f. https://stackoverflow.com/a/41610049. Oh, what you might be missing is that `@enforce_keys` being there actually makes the keys enforced :), it isn't just for the `defstruct`

paulperegud:
One thing that has popped up in a call with @InoMurko and @achiurizo is lack of vault_id / token_standard_id in UTXO and output data structures.

InoMurko:
Hey @pdobacz I have a question to regards to UTXOs.
If we have multiple ERC20 currency support + ETH + ERC721. How do we differentiate UTXOs since they all live under the same key in RocksDB? I talked to @paulperegud and he mentioned we need to use a vault identifier... I don't see this here. Just wondering...

boolafish:
Wouldn't `outputType` be better identifier when thinking about output data structure? like what @pdobacz mentioned in https://github.com/omisego/plasma-contracts/issues/290

but I might be jumping in without right context so.....

pdobacz:
> lack of vault_id / token_standard_id in UTXO and output data structures.

Guys, this is not the PR you're looking for :). AFAIK, the output format you mention doesn't yet work in the ALD plasma framework contracts, of which the best indication is, that this passes CI :). This still uses the fixed `{owner, output, amount}` for output.

This PR integrates with the revamp of the contracts structure and typing of txs and some minor changes to the contracts that piggybacked on the ALD change.

The PR for the abstract/new/typed output structure is yet to surface (partly b/c the output's structure and abstract type is in flux).

> If we have multiple ERC20 currency support + ETH + ERC721. How do we differentiate UTXOs since they all live under the same key in RocksDB?

The way the UTXOs are keyed and looked up (on tx execution) will depend on the `input_pointer_type` that they use, i.e. either `output_id` or `utxo_pos`. That's agnostic of what the output represents (ERC/ETH/ERC721). The UTXOs will be differentiated by the value not key (unless we have a good reason to lookup by type in the RocksDB, in which case we'd need to provide more indexes - this sounds more like a requirement for the relational WatcherDB). Does this make sense?

But essentially the answer is the same - not in this PR, and this is not set in stone yet. This is all up to the abstract/new/typed output structure PR :point_up:

apps/omg/lib/omg/state/transaction.ex
pdobacz: ok I noticed there's many occurrences of this. I'd rather see this as a consistent change throughout the codebase in a separate PR.

InoMurko: makes sense! #cleanup

kasima:
Just got off a call with @pdobacz and @kendricktan to try to get this PR merged into `master` and deployed. 

The path forward:
- [ ] @kendricktan will write a new truffle migration to deploy the contracts in https://github.com/omisego/plasma-contracts/pull/204. And open a new PR based on the `150_python_tests_on_plasma_framework` branch. These migrations will only be needed until https://github.com/omisego/plasma-contracts/pull/204 is merged into `plasma-contracts/master`.
- [ ] @kendricktan work with @pdobacz to get this PR merged into `elixir-omg/master`, using the contracts deployed ☝️ 
- [ ] @pdobacz will work with @pik694 to get https://github.com/omisego/plasma-contracts/pull/204 merged into `plasma-contracts/master` within the next few days. If it takes longer than a few days, we will cherry-pick the needed changes into `plasma-contracts/master`.
- [ ] @pdobacz will change the plasma-contracts dependency in #1005 to a SHA in the `plasma-contracts/master` history before merging into `plasma-contracts/master`.
- [ ] :tada:


#### Background
We are trying to merge this to `master` and build a new **dev** environment. There are fixes in https://github.com/omisego/plasma-contracts/pull/204 that are needed for integration. 

_The goal is to minimize the time that the integration of elixir-omg is diverged from the work that is happening on `plasma-contracts/master`._ 

Rather than doing branch gymnastics now to get the changes we need into `plasma-contracts/master`, we decided to try to merge https://github.com/omisego/plasma-contracts/pull/204 as quickly as possible.

@pdobacz @kendricktan – please let me know if I'm off base on any of this. Update the plan, as necessary.

pdobacz:
an update/recap:

1. this PR is mergeable any moment the dev env is setup
2. https://github.com/omisego/plasma-contracts/pull/204 is mergable AFAIK, passes CI, only pending final CR.

kasima:
@kendricktan – ready for deployment into the new dev cluster once this merges?

InoMurko:
Hey @pdobacz this seems https://github.com/omisego/elixir-omg/issues/1017 like a blocker before we can deploy.

apps/omg_eth/lib/omg_eth/release_tasks/set_contract.ex
```diff
@@ -85,20 +86,25 @@ defmodule OMG.Eth.ReleaseTasks.SetContract do
...
 
-  defp update_configuration(txhash_contract, authority_address, contract_address, exit_period_seconds)
+  defp update_configuration(txhash_contract, authority_address, contract_addresses, exit_period_seconds)
        when is_binary(txhash_contract) and
-              is_binary(authority_address) and is_binary(contract_address) and is_integer(exit_period_seconds) do
+              is_binary(authority_address) and is_map(contract_addresses) and is_integer(exit_period_seconds) do
+    contract_addresses =
+      contract_addresses
+      |> Enum.into(%{}, fn {name, addr} -> {String.to_existing_atom(to_string(name)), String.downcase(addr)} end)
+
```
InoMurko: @pdobacz are you sure `to_existing_atom` is ok? the VM has just booted, it's likely no atoms exist at all.

```suggestion
Enum.into(contract_addresses, %{}, fn {name, addr} -> {String.atom(to_string(name)), String.downcase(addr)} end)
```

also, there's no need to check is_binary, is_binary, is_map if you're not intending to provide a valid exit/1 clause. pls follow the convention by validating input parameters and exit/1 the process with a descriptive message.

apps/omg_eth/lib/omg_eth/release_tasks/set_contract.ex
```diff
@@ -85,20 +86,25 @@ defmodule OMG.Eth.ReleaseTasks.SetContract do
 
     txhash_contract = get_env(network <> "_TXHASH_CONTRACT")
     authority_address = get_env(network <> "_AUTHORITY_ADDRESS")
-    contract_address = get_env(network <> "_CONTRACT_ADDRESS")
+    env_contract_addresses = get_env(network <> "_CONTRACT_ADDRESS")
+    contract_addresses = if env_contract_addresses, do: Jason.decode!(env_contract_addresses), else: nil
```
InoMurko: with the problems that devops has deploying this. pls let us move away from encoding JSON into the environment variables. please lets find another format to pass in the value. maybe comma separated values? 
or descriptive environment variables for each contract, like we had before? lets not break the pattern.

apps/omg_eth/lib/omg_eth/release_tasks/set_contract.ex
pdobacz: yeah, missing atoms seem to be the issue, and also an issue independent of how the system env variables are structured (packed in a JSON or separate).

pdobacz: on the latter part:

> there's no need to check is_binary, is_binary, is_map if you're not intending to provide a valid exit

but the checks `is_binary` weren't introduced in this PR. I was just conforming to the checks I already found here (by switching to an `is_map` check). Or am I missing a point here?

apps/omg_eth/lib/omg_eth/release_tasks/set_contract.ex
pdobacz: For future ref: sth similar proposed in `temp/948_fix_JSON_sys_env` branch, went with the latter option.

pdobacz:
superseded by https://github.com/omisego/elixir-omg/pull/1021 which includes:
  - @pdobacz@InoMurko fix: drop JSON from env approach … https://github.com/omisego/elixir-omg/pull/1021/commits/e685b8b193df7c52afe049b2cc1669c5ea2828eb
  - @InoMurko fix: contract exchanger bump fixes 8077248

PR #951, opened 2019-09-11
feature: deploy release

InoMurko:
https://github.com/omisego/elixir-omg/issues/579
Standardize application bootup #579 

## Overview

CircleCI should be able to push a new development deployment

## Changes

- Circleci steps + deployment script

## Testing

/

### DIFF
--- a/.circleci/ci_deploy.sh
+++ b/.circleci/ci_deploy.sh
@@ -1,0 +1,58 @@
+#!/bin/sh
+
+set -e
+
+echo_info() {
+    printf "\\033[0;34m%s\\033[0;0m\\n" "$1"
+}
+
+echo_warn() {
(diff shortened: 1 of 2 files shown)


jbunce:
Why is this required? I feel that defining this within the CirclCI configuration makes it easier to view the stages than in a bash script.

InoMurko:
I would disagree, since our circle config yml is getting bigger and bigger it’s painful to scroll and search + deploy and publish are two items you rarely need to look at.

kendricktan:
I'm for abstracting away the deployment process into a bash script. Yes it currently incurs some level of technical debt, but when compared to copying + pasting the same deployment boilerplate code with minor tweaks to the ${branch} var, its the lesser of two evils imo.

PR #965, opened 2019-09-13
Rm leveldb2

InoMurko:
As per Slack conversation, we're ready to move to RocksDB on next release cycle.

## Overview

rm leveldb everywhere

## Changes

/
## Testing
/


InoMurko:
Same thing as previous, just managed not to delete this one! :)

apps/omg_db/lib/omg_db/rocks_db.ex
```diff
@@ -11,143 +11,142 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-if Code.ensure_loaded?(:rocksdb) do
```
pdobacz: the diff is garbled here. What did here change is only the unindent and removal of the `if`?

Same question for `...core.ex` and `..server.ex` files

apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
```diff
@@ -79,7 +79,7 @@ defmodule OMG.Watcher.Integration.BlockGetterTest do
              %{"blknum" => ^block_nr}
            ] = TestHelper.get_utxos(alice.addr)
 
-    assert TestHelper.get_utxos(alice.addr) == TestHelper.get_exitable_utxos(alice.addr)
+    assert Enum.sort(TestHelper.get_utxos(alice.addr)) == Enum.sort(TestHelper.get_exitable_utxos(alice.addr))
```
pdobacz: let's Enum.sort in the endpoint (the `get_exitable_utxos` one, which is which became unsorted in RocksDB). I think `get_utxos` returns them sorted and so should `get_exitable_utxos`

mix.lock
```diff
@@ -28,7 +28,6 @@
   "excoveralls": {:git, "https://github.com/omisego/excoveralls.git", "23b97648ff5ed7b19d75364233bbf3e5fcb407ad", [ref: "23b97648ff5ed7b19d75364233bbf3e5fcb407ad"]},
   "exexec": {:git, "https://github.com/pthomalla/exexec.git", "213658a4ccb96d7a84ffddc642b2596c86acc5d0", [branch: "add_streams"]},
   "exjsx": {:hex, :exjsx, "4.0.0", "60548841e0212df401e38e63c0078ec57b33e7ea49b032c796ccad8cde794b5c", [:mix], [{:jsx, "~> 2.8.0", [hex: :jsx, repo: "hexpm", optional: false]}], "hexpm"},
-  "exleveldb": {:hex, :exleveldb, "0.14.0", "8e9353bbce38482d6971d254c6b98ceb50f3f179c94732b5d17db1be426fca18", [:mix], [{:eleveldb, "~> 2.2.20", [hex: :eleveldb, repo: "hexpm", optional: false]}], "hexpm"},
```
pdobacz: hmmm why didn't eleveldb go too?

apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
InoMurko: I've done this on rocksdb after iteration. It happens because of the way we append results at the end of the list!

apps/omg_db/lib/omg_db/rocks_db.ex
InoMurko: yes makes sense! I've now added the ensure loaded calls back and will remove them in a separate PR.

mix.lock
InoMurko: forgot they're separate! tnx!

apps/omg_db/lib/omg_db/rocks_db.ex
pdobacz: Ah, I was just asking if you only did the unindent.

If only the unindent, then all :+1: and no objections to getting rid of the `if` right away.

kasima:
🚀

Issue #966, opened 2019-09-13
Migrations fail on deployed development Watcher


InoMurko:
```
Running migration for OMG.Watcher.DB.Repo...
2019-09-13 15:25:19.212 [info] module=Ecto.Migration.Runner function=log/2 ⋅== Running 20190806111817 OMG.Watcher.DB.Repo.Migrations.AlterTxOutputsTableAddRootchainTxnHashDepositAndExitColumns.up/0 forward⋅
2019-09-13 15:25:19.214 [info] module=Ecto.Migration.Runner function=log/2 ⋅drop constraint txoutputs_creating_deposit_fkey from table txoutputs⋅
2019-09-13 15:25:19.216 [info] module=Ecto.Migration.Runner function=log/2 ⋅drop constraint txoutputs_spending_exit_fkey from table txoutputs⋅
2019-09-13 15:25:19.218 [info] module=Ecto.Migration.Runner function=log/2 ⋅drop constraint ethevents_pkey from table ethevents⋅
2019-09-13 15:25:19.220 [info] module=Ecto.Migration.Runner function=log/2 ⋅drop table ethevents⋅
2019-09-13 15:25:19.221 [info] module=Ecto.Migration.Runner function=log/2 ⋅create table ethevents⋅
2019-09-13 15:25:19.230 [info] module=Ecto.Migration.Runner function=log/2 ⋅create index ethevents_root_chain_txhash_index⋅
2019-09-13 15:25:19.234 [info] module=Ecto.Migration.Runner function=log/2 ⋅create index ethevents_log_index_index⋅
2019-09-13 15:25:19.238 [info] module=Ecto.Migration.Runner function=log/2 ⋅create index ethevents_root_chain_txhash_event_index⋅
2019-09-13 15:25:19.243 [info] module=Ecto.Migration.Runner function=log/2 ⋅execute "ALTER TABLE ethevents ALTER COLUMN inserted_at SET DEFAULT (now() at time zone 'utc');"⋅
2019-09-13 15:25:19.246 [info] module=Ecto.Migration.Runner function=log/2 ⋅execute "ALTER TABLE ethevents ALTER COLUMN updated_at SET DEFAULT (now() at time zone 'utc');"⋅
2019-09-13 15:25:19.247 [info] module=Ecto.Migration.Runner function=log/2 ⋅alter table txoutputs⋅
[...]

PR #968, opened 2019-09-15
relax timestamp not null constraints on txoutputs

jrhite:
Fix for #966


InoMurko:
How did we get into the state where `inserted_at` is null? If I'm looking at migrations, I see `inserted_at` should have been set `execute("ALTER TABLE txoutputs ALTER COLUMN updated_at SET DEFAULT (now() at time zone 'utc');")`

Is the problem somewhere else too?

jrhite:
No this problem doesn't exist anywhere but in the `txoutputs` table. The reason being, is I looked at our code and also confirmed with Pawel or Piotr (I forget who), the `ethevents` table was only written to and never read. Because of this I just dropped the table in my migration. So it and also the new join table `ethevents_txoutputs` have no rows. So no existing rows violate the `NOT NULL` constraint.

However, I did not drop the `txoutputs` table as we do read it. Thus adding the `NOT NULL` constraint breaks existing rows. The `SET DEFAULT` part only applies to newly inserted rows.

jrhite:
@InoMurko I just updated the PR as the original commit wasn't going to work. The migration hasn't even been applied, as it, itself is failing. Derp.

Additionally, ecto `timestamps` doesn't seem to support allowing these columns to be null. https://hexdocs.pm/ecto/Ecto.Schema.html#timestamps/1

I've updated the actual original migration fail to fix the problem instead of creating another migration on top of it.

apps/omg_watcher/priv/repo/migrations/20190806111817_alter_txoutputs_ethevents_make_many_to_many_relation.exs
```diff
@@ -62,13 +62,9 @@ defmodule OMG.Watcher.DB.Repo.Migrations.AlterTxOutputsTableAddRootchainTxnHashD
       remove(:creating_deposit)
       remove(:spending_exit)
 
-      timestamps([type: :utc_datetime])
+      timestamps(type: :utc_datetime, default: fragment("(now() at time zone 'utc')"), null: false)
```
pnowosie: Referring to [comment](https://github.com/omisego/elixir-omg/issues/966#issuecomment-531442169)
I think migration should handle the existing data by inserting `now()` to existing rows and then set `not null` constraint.
Therefore we'll have the same db on dev and prod. What do you think @jrhite :question:

InoMurko: I would just add that 
`I think migration should handle the existing data by inserting now()` 
now(), might not be a good idea, because we wouldn't know which are valid dates and which are not. So maybe `1 January 1970 00:00:00 UTC` then we at least know it's not a TRUE value.

jrhite: Yeah, postgres won't behave like this for existing rows, only newly inserted rows. If we are planning to wipe staging/dev again and I would like to slip the not null constraint back in. 😉

pnowosie: > Postgres doesn't know the correct the column values to insert for existing rows.
So that's why I suggested to: 
1. add a column without constraint, 
2. set some value in existing rows (if not now() then maybe epoch as @InoMurko said)
3. bring the constraint 
So we'd have the constraint and consistent DBs in each env.

> For example, for auditing purposes, or perhaps the UI wants to filter things based on created or updated date.
What's wrong with the above then? The existing rows would have the earliest date. How they will be sorted / audited with `NULL` values - why it's better

jrhite: Yes, I understand the process and steps. I can implement everything you say in this single migration.

There is no meaningful value to use, but I agree that if we're going to use any value the epoch is best.

PR #971, opened 2019-09-16
triggers for updating updated_at columns

jrhite:

## Overview

update `updated_at` columns when a row is updated

## Changes

create triggers for tables with `updated_at` columns so that they will be updated to `now()` when the row is updated

## Testing

local testing on postgres


InoMurko:
This, if I read it correctly, updates all historic entries with now()? Wouldn’t that confuyus what are real real inserted dates, and which were false to begin with?

jrhite:
`BEFORE UPDATE ON txoutputs FOR EACH ROW EXECUTE PROCEDURE update_updated_at_column()` translates to:

For each row that is about to be updated, run the procedure `update_updated_at_column()` on that row.

Within the procedure, `NEW` the row being updated. Set `NEW.updated_at` to `now()`.

InoMurko:
@jrhite can we close this? I think we can.

jrhite:
@InoMurko - this was never merged into master. Need to reopen and merge, no?

InoMurko:
Bur there are no changes?

jrhite:
@InoMurko 

Weird. This was just a `git mv` fail. Not sure what happened. I needed to change the file name timestamp to be ahead of the timestamp of the latest migration. Not sure why the new file ended up empty.

Fixed now. Please have a look.

unnawut:
Question: The [`Echo.Schema.timestamps/1`](https://hexdocs.pm/ecto/Ecto.Schema.html#timestamps/1) already present in those schemas aren't doing its job?

jrhite:
I will get a copy of dev db dump from @kendricktan then standup a watcher against dev childchain and test.

Will get back to you on the testing results.

Let me know if you'd like to see anything else regarding testing.

jrhite:
@unnawut 

Yes and also no. :smile: 

`timestamps` is working just fine and defaults are set correctly. But for the `updated_at`column you need a DB trigger to make sure the timestamp gets set correctly every time the column is updated, thus this PR.

unnawut:
It has been working fine for the ewallet. Plus just checked Ecto's code, it should be refreshing `update_at` on updates...

https://github.com/elixir-ecto/ecto/blob/v3.1.5/lib/ecto/schema.ex#L652-L655

```elixir
defmodule Ecto.Schema do
  defmacro timestamps(opts \\ []) do
      # ...

      autogen = timestamps[:autogenerate] || {Ecto.Schema, :__timestamps__, [type]}

      # ...

      if updated_at do
        Ecto.Schema.field(updated_at, type, [])
        Module.put_attribute(__MODULE__, :ecto_autoupdate, {[updated_at], autogen})
      end

      # ...
  end
end
```

which should then be picked up by https://github.com/elixir-ecto/ecto/blob/v3.1.5/lib/ecto/schema.ex#L522

```elixir
defmodule Ecto.Schema do
  # ...

  autoupdate = @ecto_autoupdate |> Enum.reverse

  # ...

  def __schema__(:autoupdate), do: unquote(Macro.escape(autoupdate))

  # ...
end
```

and triggered on update by https://github.com/elixir-ecto/ecto/blob/v3.1.5/lib/ecto/repo/schema.ex

```elixir
defmodule Ecto.Repo.Schema do
  # ...

  defp autogenerate_changes(schema, action, changes) do
    autogen_fields = action |> action_to_auto() |> schema.__schema__()
    # ...
  end

  defp action_to_auto(:update), do: :autoupdate
end
```

Can you give some examples where the code is not working as expected? Maybe I can have a look.

Main reason I'm raising this is because IMHO stored procedures can be hard to discover & hard to maintain as they'll be deeply buried in migration files, and the extra attention required for backups and restores.

jrhite:
Thanks for all the info @unnawut! This looks great.

> Main reason I'm raising this is because IMHO stored procedures can be hard to discover & hard to maintain as they'll be deeply buried in migration files, and the extra attention required for backups and restores.

I totally agree with this.

The code pointers are super helpful. Let me test out some things and check it out. We haven't used ecto timestamps in watcher's db until I recently started adding them, so it's been a learning experience.

Let me touch base in a bit and maybe we can drop this PR.

unnawut:
Noting down for the team's reference:

> Mainly I think it's because Repo.update_all is being called directly, and they work like raw queries so the autoupdate doesn't get triggered. 
>
> So I think there are 3 options here:
>
> 1. Go with stored procedures
> 2. Continue to use Repo.update_all but add updated_at to the list of attributes to be updated
> 3. Replace direct Repo.update_all calls into Ecto.Changeset
>
> It's a new direction to go down the stored procedures route so I want to make sure the team is aware/okay.

jrhite:
@unnawut 

I confirmed the behavior you described, and everything works just like you said. 😄 

A few things perhaps we can conclude...

1. No need for DB triggers (thus I will close this PR)
2. For timestamps, it looks like Ecto supports seconds or microseconds. I don't think seconds are granular enough, given we want regulatory compliance and auditability, it's better to use `:utc_datetime_usec` for timestamp fields.
3. There is only 1 place in watcher I saw that uses `update_all`. I think it's fine to use `update_all` if that is appropriate, passing in `updated_at` to the list of attributes to be updated as you said, we'll just want specific tests for `inserted_at` and `updated_at` fields.

Thoughts? 👍 or 👎

unnawut:
> No need for DB triggers (thus I will close this PR)

🎖🥇🍾🌟❤️

> For timestamps, it looks like Ecto supports seconds or microseconds. I don't think seconds are granular enough, given we want regulatory compliance and auditability, it's better to use :utc_datetime_usec for timestamp fields.

Sounds good! This was a breaking change in Ecto 3.0 (it used to be microseconds I think). There are a few places to update to make the changes complete. Searching https://github.com/omisego/ewallet/pull/771 's  code for `:naive_datetime_usec` should help. Though I don't know enough about elixir-omg to know if it has to be `:utc_datetime_usec` or `:naive_datetime_usec`. 😅

> There is only 1 place in watcher I saw that uses update_all. I think it's fine to use update_all if that is appropriate, passing in updated_at to the list of attributes to be updated as you said, we'll just want specific tests for inserted_at and updated_at fields.

Sounds good to me! Also just in case, it's possible to use Elixir's core module for datetime, or rely on the DB's clock using `Repo.update_all(set: [updated_at: fragment("NOW()")])`. Again I don't know enough whether which one is more appropriate.

PR #975, opened 2019-09-17
fix: tracer setup

InoMurko:
/

## Overview

As reported in the devops issue, Child Chain Phoenix has been taking Watcher's Datadog tracer causing crashes.

## Changes

- Set the correct tracer per application mode in multiple places.

## Testing
/


InoMurko:
I'm merging this into v0.2 @pdobacz is that ok?

VERSION
```diff
@@ -1 +1 @@
-0.2.2
+0.2.3
```
pdobacz: rebase on the HEAD of v0.2, this already got bumped

apps/omg_child_chain_rpc/lib/omg_child_chain_rpc/application.ex
```diff
@@ -22,7 +22,8 @@ defmodule OMG.ChildChainRPC.Application do
 
   def start(_type, _args) do
     DeferredConfig.populate(:omg_child_chain_rpc)
-
+    DeferredConfig.populate(:omg_watcher_rpc)
```
pdobacz: `omg_watcher_rpc` in chch?

and the same vice versa in `omg_watcher_rpc`. If those are intended, a comment would be v. useful

apps/omg_watcher_rpc/lib/application.ex
```diff
@@ -19,7 +19,8 @@ defmodule OMG.WatcherRPC.Application do
 
   def start(_type, _args) do
     DeferredConfig.populate(:omg_watcher_rpc)
-
+    DeferredConfig.populate(:omg_child_chain_rpc)
+    :ok = Application.put_env(:spandex_phoenix, :tracer, OMG.WatcherRPC.Tracer, persistent: true)
```
pdobacz: oh my, now I see, ouch. So `spandex_phoenix` is overwritten for every umbrella subapp, right?

If there isn't any other way to mend this, comments would be v. useful.

apps/omg_child_chain_rpc/lib/omg_child_chain_rpc/application.ex
InoMurko: yeah, that's kinda the issue we were seeing, tracers are global settings that don't support more then one phoenix application, so watcher's configuration comes in first and :spandex_datadog picks that one.

apps/omg_watcher_rpc/lib/application.ex
InoMurko: will do!

InoMurko:
> LGTM with some comments. Will you be cherry-picking this to `master` as well?

master does not have this problem, because of configuration providers that sort this out before any of the (dependency) applications boot.

PR #956, opened 2019-09-12
feature: run releases barebone

InoMurko:
HELLO
## Overview

This should allow you to run child chain and watcher baremetal.

## Changes

- makefile ftw
## Testing
 lemme know if it works

so you want to try baremetal ehm?
First, I suggest you stop any containers you might be running (check if there are any with `docker ps`) and stop `plasma-deployer`, `geth`, `watcher`, `childchain`and `postgres`. Or just do 
1. `docker stop $(docker ps -q)`.
We then go all in.
Run:
2. `make start-services` this starts geth, postgres and plasma-deployer
Beware, plasma-deployer can be flaky. 
3. Open another terminal window an run
`make start-child_chain`
4. Open another terminal window an run
`make start-watcher`
5. Run `make get-alarms` once 3. and 4. are done booting, you should see curl responses from both services.
6. Explore the makefile with `make list`,
7. Other baremetal commands
```
make stop-child_chain
make stop-watcher
make remote-child_chain (allows you to attach yourself to the running node)
make remote-watcher (allows you to attach yourself to the running node)
```
To push code changes:
```
make update-child_chain
make update-watcher
```

Watcher and Child Chain need exported environment variables, they're found bin `bin/variables`.

Thanks @T-Dnzt !

PR #979, opened 2019-09-18
multi-stage docker

jrhite:
## Overview

- Increase productivity by prebuilding all of our dependencies in an image. Starting from this prebuilt dependency image means we spend our time building our code and not building 3rd party code. Any time a dependency changes this image will be rebuilt.
  - `make docker-child_chain && make docker-watcher` are very expensive to do, because each make invocation ends up building the dependencies twice. A single dependency build is expensive on its own.
- Remove build artifacts that get left on the host `_build` and `deps`. This has caught a number of us resulting in build discrepancies and failures.
- Reduce image sizes via docker multi-stage builds

(Please note there is another PR out with similar goals at increasing build speeds and tidying up the build environment from @InoMurko. Have a look at it for reference of another perspective on this: https://github.com/omisego/elixir-omg/pull/956/files)

TL;DR on the files below. On a day-to-day basis developers will only need to care about `Dockerfile.omg-dev-env` and `Dockerfile.omg-service-env`. In general you’ll never need to touch or worry about the other Dockerfiles. In the end, only 2 of the Dockerfiles may end up living in the `elixir-omg` repository.

1. `Dockerfile.omg-erlang-env`
2. `Dockerfile.omg-elixir-env`
3. `Dockerfile.omg-dev-env`
4. `Dockerfile.omg-service-env`
5. `Dockerfile.omg-service-release`

The first 2 docker images were stolen from https://github.com/omisego-images/docker-elixir-omg/tree/master/builder. They build our foundational erlang/elixir images.

As a developer `Dockerfile.omg-dev-env` and `Dockerfile.omg-service-env` are what you will use the most.
- `Dockerfile.omg-dev-env` creates an elixir image with all the dependencies prebuilt and all the elixir tooling. Very fast as it only builds our code. Good for running tests and debugging code.
- `Dockerfile.omg-service-env` produces either a runnable child_chain or watcher image. It does not remove elixir or any other tooling so that you can run our services, step into the container and debug what you need to debug.

The last docker image was taken from `Dockerfile.childchain` and `Dockerfile.watcher` and combined into one. They have a few different dependencies, but there was a lot of duplicated code in them. Given that it doesn’t effect the production releases, I merged this into 1 file. You specify if you want to build child_chain or watcher via a `--build-arg` called `SERVICE`. This last image will definitely need buy-in from @InoMurko and @jbunce.

Unexpected bonus points! I either got really lucky or I broke something. 🤣  The release images created by this PR reduce our production images by 8%.

Current image sizes:
```
omisego/watcher                           169MB
omisego/child_chain                       163MB
```

Docker image sizes from this PR:
```
omg-service-release-watcher               156MB
omg-service-release-child_chain           151MB
```

Also just FYI and for fun! 😄 

The entire stack can be built from scratch with the command line below. This would obviously be tidied up in a nice script with functions for this or a Makefile, my preference being a script.

```
docker build -t omg-erlang-env --target omg-erlang-env - < docker/Dockerfile.omg-erlang-env \
  && docker build -t omg-elixir-env - < docker/Dockerfile.omg-elixir-env \
  && docker build -t omg-dev-env --target omg-dev-env -f docker/Dockerfile.omg-dev-env . \
  && docker build --build-arg SERVICE=child_chain --build-arg RELEASE_VERSION=0.2.2+ --build-arg SHA=`git rev-parse --short=7 HEAD` -t omg-service-env-child_chain -f docker/Dockerfile.omg-service-env . \
  && docker build --build-arg SERVICE=watcher --build-arg RELEASE_VERSION=0.2.2+ --build-arg SHA=`git rev-parse --short=7 HEAD` -t omg-service-env-watcher -f docker/Dockerfile.omg-service-env . \
  && docker build --build-arg SERVICE=child_chain --build-arg RELEASE_VERSION=0.2.2+ --build-arg PORT=9656 --build-arg SHA=`git rev-parse --short=7 HEAD` -t omg-service-release-child_chain -f docker/Dockerfile.omg-service-release . \
  && docker build --build-arg SERVICE=watcher --build-arg RELEASE_VERSION=0.2.2+ --build-arg PORT=7434 --build-arg SHA=`git rev-parse --short=7 HEAD` -t omg-service-release-watcher -f docker/Dockerfile.omg-service-release .
```

## Testing

Used the above command-line to build the system from scratch and bring up childchain and watcher and verify they were up and running correctly.

### DIFF
--- a/docker/Dockerfile.omg-service-release
+++ b/docker/Dockerfile.omg-service-release
@@ -1,0 +1,90 @@
+
+ARG ALPINE_VERSION=3.9
+
+ARG SERVICE
+
+FROM omg-service-env-$SERVICE as omg-service-env
+
+FROM alpine:$ALPINE_VERSION AS alpine
+
+FROM alpine AS omg-service-release
+
+LABEL maintainer="OmiseGO Team <omg@omise.co>"
+LABEL description="Official image for OmiseGO (Watcher) Plasma Network"
+
+ARG MIX_ENV=dev
+ARG PORT
+ARG RELEASE_VERSION
+
+ARG SHA
+
+# needed here again because cannot read the SERVICE ARG outside of the FROM context
+ARG SERVICE
+
+# USER directive is not being used here since privileges are dropped via
+# s6-setuigid in /entrypoint. s6-overlay is required to be run as root.
+ARG USER=$SERVICE
+ARG GROUP=$SERVICE
+ARG UID=10000
+ARG GID=10000
+ARG APP_DIR=/app
+
+ENV LANG=C.UTF-8
+
+## S6
+##
+
+ENV S6_VERSION="1.21.4.0"
+
+RUN set -xe \
+ && apk add --update --no-cache --virtual .fetch-deps \
+        curl \
+        ca-certificates \
+ && S6_DOWNLOAD_URL="https://github.com/just-containers/s6-overlay/releases/download/v${S6_VERSION}/s6-overlay-amd64.tar.gz" \
+ && S6_DOWNLOAD_SHA256="e903f138dea67e75afc0f61e79eba529212b311dc83accc1e18a449d58a2b10c" \
+ && curl -fsL -o s6-overlay.tar.gz ${S6_DOWNLOAD_URL} \
+ && echo "${S6_DOWNLOAD_SHA256}  s6-overlay.tar.gz" | sha256sum -c - \
+ && tar -xzC / -f s6-overlay.tar.gz \
+ && rm s6-overlay.tar.gz \
+ && apk del .fetch-deps
+
+## Application
+##
+
+#leveldb and rocksdb need libstdc++
+RUN apk add --update --no-cache --virtual .watcher-runtime \
+        bash \
+        libressl \
+        libressl-dev \
+        libstdc++ \
+        lksctp-tools
+
+#libsecp256k1
+RUN apk add --no-cache gmp-dev
+
+COPY --from=omg-service-env /entrypoint /entrypoint
+
+RUN set -xe \
+ && addgroup -g $GID $GROUP \
+ && adduser -D -h $APP_DIR -u $UID -G $GROUP $USER \
+ && chown ${UID}:${GID} $APP_DIR \
+ && chmod +x /entrypoint
+
+WORKDIR $APP_DIR
+
+COPY --from=omg-service-env ${APP_DIR}/_build/${MIX_ENV}/rel/${SERVICE}/releases/${RELEASE_VERSION}${SHA}/${SERVICE}.tar.gz $APP_DIR
+RUN set -xe \
+  && tar xzf ${APP_DIR}/${SERVICE}.tar.gz && chown -R ${UID}:${GID} $APP_DIR \
+  && rm ${APP_DIR}/${SERVICE}.tar.gz
+
+EXPOSE $PORT
+
+# These are ports required for clustering. The range is defined in vm.args
+# in inet_dist_listen_min and inet_dist_listen_max.
+#EXPOSE 4369 6900 6901 6902 6903 6904 6905 6906 6907 6908 6909
+# curl for healthchecks
+RUN apk add --no-cache curl
+
+ENTRYPOINT ["/init", "/entrypoint"]
+
+CMD ["foreground"]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,15 @@
+.circleci/
+.elixir_ls/
+.githooks/
+.github
+.idea/
+.vscode/
+
+_build/
+deps/
+
+docker/
+
+docker-compose.*
+
+elixir-omg.iml
--- a/docker/Dockerfile.omg-base-env
+++ b/docker/Dockerfile.omg-base-env
@@ -1,0 +1,109 @@
+
+# this file is an intermediate build file to build anc cache mix dependencies
+
+ARG ALPINE_VERSION=3.9
+
+FROM alpine:${ALPINE_VERSION} AS alpine
+
+FROM alpine AS omg-base-env
+
+LABEL maintainer="OmiseGO Team <omg@omise.co>"
+LABEL description="Builder image for OmiseGO elixir-omg"
+
+COPY --from=omg-elixir-env /usr/local/otp /usr/local/otp
+COPY --from=omg-elixir-env /usr/local/rebar3 /usr/local/rebar3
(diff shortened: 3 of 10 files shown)


sirn:
FWIW, the reason we separate `omisego-images` for builder image and `omisego` is because CircleCI do not cache builds, so multi-stage deploy will result in longer build time since all stages must be re-built.

It’s possible to configure CircleCI to cache the image (`docker save ...` / `docker load ...` / then `save_cache`) but we’ll have to dealt with cache expiration and such, which we get for free by separating builder image.

docker/Dockerfile.omg-service-release
```diff
@@ -0,0 +1,90 @@
...
+#leveldb and rocksdb need libstdc++
+RUN apk add --update --no-cache --virtual .watcher-runtime \
+        bash \
+        libressl \
+        libressl-dev \
+        libstdc++ \
+        lksctp-tools
+
+#libsecp256k1
+RUN apk add --no-cache gmp-dev
```
sirn: Any particular reason why this is not part of `apk add` in line 55? Also, please use `apk add --update` when using `apk` in a new build step because apk deletes older packages when newer one is released.

docker/Dockerfile.omg-service-release
```diff
@@ -0,0 +1,90 @@
...
+  && tar xzf ${APP_DIR}/${SERVICE}.tar.gz && chown -R ${UID}:${GID} $APP_DIR \
+  && rm ${APP_DIR}/${SERVICE}.tar.gz
+
+EXPOSE $PORT
+
+# These are ports required for clustering. The range is defined in vm.args
+# in inet_dist_listen_min and inet_dist_listen_max.
+#EXPOSE 4369 6900 6901 6902 6903 6904 6905 6906 6907 6908 6909
+# curl for healthchecks
+RUN apk add --no-cache curl
```
sirn: Better add this before `## S6` since `curl` is also required in `fetch-deps` and it should not change very often. apk’s `virtual` is pretty smart: if you do:

```
apk add --update curl
apk add --update --virtual .fetch-deps curl ca-certificates
apk del .fetch-deps
```

…will result in only `ca-certificates` being removed.

sirn: We might also need `ca-certificates` though if we ever connect to HTTPS (but I think healthcheck won’t do this?)

InoMurko:
Let me quickly respond to 
```
    make docker-child_chain && make docker-watcher are very expensive to do, because each make invocation ends up building the dependencies twice. A single dependency build is expensive on its own.

```
this is simply not true, because of the way mix works. The containers are mounting local storage, and the first pass puts build in `deps` and `_build`, the second one uses the already built dependencies.

docker/Dockerfile.omg-service-release
jrhite: This file was mostly copy and paste. Not sure why it's like that, but looks like it could be added to line 55.

jrhite:
@sirn the hope would be to also store an image of our built deps there in omisego-images too.

jrhite:
@InoMurko has something changed? I could have misread the build output and when I checked the Makefile I saw `make clean` twice once for childchain and watcher.

If I'm wrong here, my apologies and we can nix this as an issue.

docker/Dockerfile.omg-service-release
jrhite: Not sure. These files are basically copied. We can look into this more.

sirn:
@jrhite I actually tried that and it turns out to be quite... unpretty. See also [omisegoimages](https://hub.docker.com/u/omisegoimages) organization in Docker Hub :joy:

The above Docker image graveyard was actually the first iteration, and even with autobuild setup to trigger on parent image update (e.g. updating `erlang` image result in `elixir` also being updated and also `-runtime`, etc.) I found it pretty nightmare-ish to maintain, compare to the one builder image solution that's currently adopted.

It was kinda cool at first... until I spent the whole day trying to pulling my hair on why a certain bug still exists when building in local Docker when I actually fixed it in Docker Hub, and the CI works fine!? Turns out, I forgot to pull the intermediate image because they're using e.g. `:latest` tag. 

I tried to fix this by hard-coding tag in the image... suddenly updating `base` image means mass manual Dockerfile update to all images that depends on `base` image. It was not fun.

jrhite:
@sirn - I see I might have cause some confusion because of the structure of my PR (without more explanation).

I pulled the all the Dockerfiles apart and put them back together into a single directory (in separate files) to see all the puzzle pieces we are working with and present them to everyone.

If what you ran into was a pain, then I would be very happy to have the erlang, elixir and the project's deps all prebuilt and combined into 1 image, the builder. Currently we don't prebuild our deps into any image.

From the single builder image we would build our own code and only our own code. It's sounds like from your comment the ewallet team might be doing it this way already.

That would be my goal line. How does that sound?

This is still very much a WIP and RFC.

sirn:
@jrhite Hmm, I thought we're already doing that? e.g. the [builder](https://github.com/omisego-images/docker-elixir-omg/blob/61c6195626377e22e3ebcfb3fb66b19082c84a95/builder/Dockerfile) image in docker-elixir-omg repository. The approach we use in eWallet is what I like to call 2-stage builds, which can be summarized as:

```
$ docker run --rm -it -v $(PWD):/app omisegoimages/ewallet-builder -- mix release
# This is the first stage, using builder image to build ewallet.tar.gz
# - We can use volume mount for deps/_build if necessary to prevent mixing up build artifacts
# - Generates _build/prod/rel/ewallet/releases/VERSION/ewallet.tar.gz

$ docker build -t $(IMAGE_NAME) .
# This is the second stage, building against plain alpine:3.8 image
# - Copy & extract _build/prod/rel/ewallet/releases/VERSION/ewallet.tar.gz into base alpine:3.8 image
```

We did not use `mix release` as part of `RUN` because we want to cache `deps` and `_build` on filesystem itself rather than in Docker layer and want to use builder image for development. This approach allows us to setup development environment by, e.g. 

```
$ docker run -it -v $(PWD):/app omisegoimages/ewallet-builder -- bash
builder# make dev
builder# bin/ewallet foreground 
```

This allow us to have only one image for the whole development cycle :-) ewallet-builder is tagged according to the version of the ewallet itself. I know it's possible to run `mix release` as part of `RUN` step in `docker build`, but I choose this approach partly because the `docker build` approach for development involve sending PWD to docker build agent, which is slow. Beside, this approach also allows us to use code reloading facility, e.g. entr or Phoenix's code reloader if needed.

achiurizo:
I think this PR can now be closed correct? If i'm wrong please re-open :pray:

PR #991, opened 2019-09-21
fix: deploy specific tag

InoMurko:
/

## Overview

To avoid caching images by using :latest tag, use git sha of the build.

## Changes

- Get SHA
- Push image tagged with short git tag

## Testing

/


kendricktan:
LGTM

Issue #855, opened 2019-07-16
Complete security critical watcher under high load


Pongch:
NET-83

As a security-concerned application provider, I would like to ensure the security critical watcher will work on DOS condition, so that I know my funds are secured

**Value**
- ensure security functionalities are completed
- validate security-critical functionality of the watcher works under load

pnowosie:
I had an impression this task is about producing high number of exits on rootchain and test we were warned by watcher and could exit.
@Pongch, @pdobacz  can we refine requirement a little. Thanks.

pdobacz:
That was my understanding - how do the security functionalities behave when there are:
1/ lots of exits in general
2/ big blocks
3/ lots of invalid standard exits
4/ lots of valid/invalid IFEs, in all possible flavors

(1/) alone has been tested in `v0.1` times, now lets go for the (1/-4/). @Pongch's description is more general, but I think this boils down to that. If anything else is required, lets has it out.

Pongch:
⬆️  Yes, let's define all the cases together, feel free to add more if any is missing. Afterwards, we can coordinate and create tasks that would invoke the different conditions.

pthomalla:
steps to check how security-critical watcher will behavior on large number on different exits
1. generate history on chilld_chain (using performance test)
2. run workers that will generate different byzantine event and correct various types of exits

workers:
- generate proper exit: using watcher api to check available utxo
- generate invalid exit: random download block from child_chain, and use some transaction to generate exit (we can specify which utxo is speeded by size of value or using information from the correct exit)
- unchallenged exit: don't challenge part of invalid exit and check after sla margin if all appeared
- non canonical ife: random download block from child_chain
    randomly change trancaction (use mutation: change input/output order, remove ouput/input/metadata, add output/input/metadata, change value of utxo)
- piggyback available: make canonical ife  
- invalid piggyback: make canonical ife that output was spended and piggyback it

pdobacz:
@pthomalla Thanks, looking good! Just jumping in with a quick comment:

> unchallenged exit:

I think you can forget about this (or put it very last on the list, as optional). I anticipate no effects on performance/availability of the security critical features under that condition.

OTOH, make sure please include checking:
 - ability to generate proper IFE under load
 - invalid input piggybacking

pdobacz:
Closing as the work here done has become stale, considering the directions taken with the perf testing effort.

PR #998, opened 2019-09-24
855 dos watcher test for standard exit

pnowosie:
## Overview

As a security-concerned application provider, I would like to ensure the security-critical watcher will work on DOS condition so that I know my funds are secured

  first part generates proper exit: using watcher api to check available utxo
## Testing

This :pr: brings two new tests in Performance application:
* time response for asking for exit data - pretty fast, just ask Watcher for exit data,
* watcher catch all non_canonical_ife - modifies txs and starts IFE exits, then waits when Watcher's `in_flight_exit_processor` processor synchronize and detect these non-canonical exits.

:spiral_notepad: For the above tests we run Ch-ch and Watcher in `dev` env to not abuse other tests we have. Also we start Watcher app (1st time in tests) it's in fixture: `omg_watcher`.


InoMurko:
Also, seems like https://github.com/plataformatec/flow it's a better use case for this. You've been at the conference! :)

pnowosie:
@InoMurko 
>  ... thinking of a separate end2end test. ...

Absolutely. This test (regarding to standard exit data fetch), test only Watcher performance of composing exit data. So it's enough to sync a Watcher from any loaded network and then run the test on it.

Note also: test does not change the network or Watcher anyhow so no destruction can be made (if the Watcher withstand the load :smiley:).

pnowosie:
> Also, seems like https://github.com/plataformatec/flow it's a better use case

Possibly?! Please note that implementation part is done. My changes brings some refactorings and documentation to meet expected quality to :ship: it.

achiurizo:
> I’ve been thinking if a test like this actually belongs as a separate end2end test. The test would perform “load”, towards a complete standuped application. Wdyt @pnowosie, @achiurizo, @T-Dnzt ?

Yep 100% agree. ✨ These tests are cool, and I think they would be a great candidate to put into the [specs](https://github.com/omisego/specs) project and could help clean up `elixir-omg` tests as we move towards ALD.

achiurizo:
cc #855

pnowosie:
## TODO:
- [ ] Update the perftest commit sha to `after rebase` value :exclamation:

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -0,0 +1,112 @@
...
+
+  ```
+  [
+    %{time: 232000, corrects_count: 10, errors_count: 0},
+    %{time: 221500, corrects_count: 10, errors_count: 0},
+    %{time: 219900, corrects_count: 10, errors_count: 0},
+  ]
+  ```
+
+  where time is expressed in microseconds and the sum of `corrects_count + errors_count` should equal to
```
pdobacz: Let's retain the conventions from the original perftest: `time` => `span_ms` and give the time in miliseconds instead

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -0,0 +1,112 @@
...
+  """
+
+  alias OMG.Eth
+  alias OMG.Performance.ByzantineEvents.DoSExitWorker
+  alias OMG.Utils.HttpRPC.Client
+  alias OMG.Utils.HttpRPC.Encoding
+
+  @watcher_url Application.get_env(:byzantine_events, :watcher_url)
+
+  def start_dos_get_exits(positions, dos_users, url \\ @watcher_url) do
```
pdobacz: Document public functions, privatize whenever possible

apps/omg_performance/lib/omg_performance/byzantine_events/generators.ex
```diff
@@ -0,0 +1,103 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.Performance.ByzantineEvents.Generators do
+  @moduledoc """
+  Module to generate users,
```
pdobacz: formatting and English here, not sure what's meant

apps/omg_performance/lib/omg_performance/byzantine_events/generators.ex
```diff
@@ -0,0 +1,103 @@
...
+  alias OMG.Eth.RootChain
+  alias OMG.State.Transaction
+  alias OMG.Utils.HttpRPC.Client
+  alias OMG.Utxo
+
+  require Utxo
+
+  @child_chain_url Application.get_env(:byzantine_events, :child_chain_url)
+
+  def generate_users(size, opts \\ [initial_funds: trunc(:math.pow(10, 18))]) do
```
pdobacz: Document public functions, privatize whenever possible

apps/omg_performance/lib/omg_performance/sender_server.ex
```diff
@@ -144,6 +147,10 @@ defmodule OMG.Performance.SenderServer do
         _ = Logger.info("[#{inspect(seqnum)}]: Transaction submission will be retried, block is full.")
         :retry
 
+      {:error, {:client_error, %{"code" => "submit:utxo_not_found"}}} ->
+        _ = Logger.info("[#{inspect(seqnum)}]: Transaction submission will be retried, utxo not found.")
```
pdobacz: why is this necessary now? Shouldn't the sender crash & terminate if `utxo_not_found`?

apps/omg_performance/config/config.exs
```diff
@@ -2,4 +2,8 @@ use Mix.Config
 
 config :briefly, directory: ["/tmp/omisego"]
 
+config :byzantine_events,
```
pdobacz: wouldn't we like to use the existing config entries for this?

apps/omg_performance/lib/omg_performance/byzantine_events/generators.ex
```diff
@@ -0,0 +1,103 @@
...
+    async_generate_user = fn _ -> Task.async(fn -> generate_user(opts) end) end
+
+    1..size
+    |> Enum.chunk_every(10)
+    |> Enum.map(fn chunk ->
+      Enum.map(chunk, async_generate_user)
```
pdobacz: let's not nest piping operators, as this is controversial. Split a `defp` out from this

apps/omg_performance/test/omg_performance/integration/byzantine_event_test.exs
```diff
@@ -0,0 +1,80 @@
...
+
+  @tag fixtures: [:contract, :child_chain, :omg_watcher]
+  test "time response for asking for exit data", %{contract: %{contract_addr: contract}} do
+    dos_users = 3
+    ntx_to_send = 100
+    spenders = Generators.generate_users(4)
+    exit_per_dos = length(spenders) * ntx_to_send
+    total_exits = length(spenders) * ntx_to_send * dos_users
+
+    IO.puts("""
```
pdobacz: don't IO.puts in the test. Do some assertions that smoke test the perf test instead, for this to be useful

apps/omg_utils/lib/omg_utils/http_rpc/client.ex
```diff
@@ -32,23 +32,28 @@ defmodule OMG.Watcher.HttpRPC.Client do
...
   @spec get_block(binary(), binary()) :: response_t()
-  def get_block(hash, url) do
-    %{hash: Encoding.to_hex(hash)}
-    |> Adapter.rpc_post("block.get", url)
-    |> Adapter.get_response_body()
-    |> decode_response()
-  end
+  def get_block(hash, url), do: call(%{hash: Encoding.to_hex(hash)}, "block.get", url)
+
+  def get_exit_data(encoded_position, url), do: call(%{utxo_pos: encoded_position}, "utxo.get_exit_data", url)
```
pdobacz: these additions don't fit here :(. See the moduledoc for this. 

Secondly, I think moving this to the generic Utils module is also debatable (e.g. you're forcing the non-test modules to have access to Watcher's RPC too, e.g. `block_getter.ex`, which makes no sense).

Lastly, I think we have some `support/test_helper.ex` modules to talk to the Watcher RPC, can't we use that?

docs/perf_test_result_dumps.md
```diff
@@ -149,3 +149,39 @@ Command as above
   %{blknum: 5000, span_ms: 4059, tps: 2931, txs: 11897}
 ]
 ```
+
```
pdobacz: start a separate file with the dump of those tests, let's not mix

docs/demo_05.md
```diff
@@ -0,0 +1,27 @@
...
+
+In the elixir REPL continue with
+```elixir
+
+### PREPARATIONS
+
+alias OMG.Performance.ByzantineEvents
+alias OMG.Performance.ByzantineEvents.Generators
+ 
+dos_users = 5
```
pdobacz: I'm missing what "dos" means in this context. I think those are just `users` and `exits_per_user`, no?

docs/perf_test_result_dumps.md
```diff
@@ -149,3 +149,39 @@ Command as above
...
+  %{blknum: 1000, span_ms: 611, tps: 3905, txs: 2386},
+  %{blknum: 2000, span_ms: 5307, tps: 4107, txs: 21798},
+  %{blknum: 3000, span_ms: 6000, tps: 3962, txs: 23771},
+  %{blknum: 4000, span_ms: 6030, tps: 3674, txs: 22156},
+  %{blknum: 5000, span_ms: 6014, tps: 3460, txs: 20809},
+  %{blknum: 6000, span_ms: 2431, tps: 4352, txs: 10580}
+]
+```
+
+then pick one of senders at random, call
```
pdobacz: Let's have a single narrative of the example of the perf test invocation. Now we have three, each slightly different (here, demo_05 and in byzantine_events.ex).

Can we make it more consistent?

Maybe add a top-level function that allows you to kick off a parametrized test of this kind in a one-liner. Then you'd fire this function here, in the smoke test and in the docs? Not 100% sure though

apps/omg_performance/lib/omg_performance/sender_server.ex
pnowosie: Honestly I have no idea. I found out that:
1. [create_utxos_for_extended_perftest](https://github.com/omisego/elixir-omg/blob/c07675ba6c71799537e1b46965a31091838a4d04/apps/omg_performance/lib/performance.ex#L248) uses DepositHelper.deposit_to_child_chain/2
2. deposit_to_child_chain - waits until deposits are [recognized...](https://github.com/omisego/elixir-omg/blob/c07675ba6c71799537e1b46965a31091838a4d04/apps/omg/test/support/integration/deposit_helper.ex#L51)

But running the test as in demos with locally setup geth, ch-ch & watcher shows that `senders` starts before utxo appear in ch-ch.
I can create an issue to dig it further... (shall I?)

docs/demo_05.md
pnowosie: Little terminology clashing - I'll fix
Here we have spenders and `dos_users` describe how many threads will fetch exit data

InoMurko:
What if we extract this into a CLI that we can use towards deployments?

apps/omg_utils/lib/omg_utils/http_rpc/client.ex
pnowosie: Well in my opinion they fix in `HttpRpc.Client` the most. My reasoning is that it's a single rpc client that understands responses from both ch-ch and Watcher. 

Also please note that Watcher TestHelpers we use to test Watcher's controllers does not use http, shortcut the communication through Phoenix test features so nothing even listen on the network.

From `ByzantineEvents` we have no luck and we need to make a full http call.

> you're forcing the non-test modules to have access to Watcher's RPC too

Well `HttpRpc.Client` DRYs things a lot regarding making call and parse the response. Also providing just ch-ch's url the additional functions wouldn't work.

pdobacz:
> What if we extract this into a CLI that we can use towards deployments?

I like this, but let's take baby steps here, this one is open for too long anyway and, FWIW, at this point requires only some cleanup.

My primary objective here is to be able to check whether current implementation holds water, when there's a rich data set involved. In long term, I'm supportive of cleaning up and making it more convenient/automated and/or moving somewhere else, once those tests are figured out and functional.

InoMurko:
> > What if we extract this into a CLI that we can use towards deployments?
> 
> I like this, but let's take baby steps here, this one is open for too long anyway and, FWIW, at this point requires only some cleanup.
> 
> My primary objective here is to be able to check whether current implementation holds water, when there's a rich data set involved. In long term, I'm supportive of cleaning up and making it more convenient/automated and/or moving somewhere else, once those tests are figured out and functional.



> > What if we extract this into a CLI that we can use towards deployments?
> 
> I like this, but let's take baby steps here, this one is open for too long anyway and, FWIW, at this point requires only some cleanup.
> 
> My primary objective here is to be able to check whether current implementation holds water, when there's a rich data set involved. In long term, I'm supportive of cleaning up and making it more convenient/automated and/or moving somewhere else, once those tests are figured out and functional.

There's a couple of problems with how this PR was laid out:
- as @pnowosie already said, there's no network involved, so it does not really get us real understanding of DOS implications
- because of the way the child chain and watcher are standupped, this has little resembelance to  deployments
- code seems like it'll be really, really hard to maintain. a lot of bad practices we're stacked together in this PR

To me, the quality and assumptions that were made in this PR do not belong in master. Sorry to be so blunt, but if we we're starting to improve testing and ensure highest possible testing standards (cc @achiurizo ) this is how I feel :D

apps/omg_performance/lib/omg_performance/sender_server.ex
pdobacz: I've never seen this when doing the older perf tests (simple/extended alike).

My biggest worry is that ignoring this error could lead to false positives in test runs of these tests done by the CI.

As a minimum I'd bump the log level to warn to make it annoy us and post an issue, if there's still an issue

InoMurko:
When do we incorporate https://github.com/lexmag/elixir-style-guide ?

pnowosie:
I think we can say we explained all concerns in the call. I will summarize what we came here inline:

> There's a couple of problems with how this PR was laid out:
>     * as @pnowosie already said, there's no network involved, so it does not really get us real understanding of DOS implications

clarified, these tests focus on ability for the Watcher to provide security data under load, so network isn't involved (we pre-build the required data set before kicking off the test)

>     * because of the way the child chain and watcher are standupped, this has little resembelance to  deployments

As discussed, we ack this is imperfect, but we're lacking a better env to run such tests, for the time being. Happy to move over when this env surfaces

>     * code seems like it'll be really, really hard to maintain. a lot of bad practices we're stacked together in this PR

A lot has been tidied up + we're focussing on the minimal setup/OTP-infrastructure to get the test running, improvements pending as soon as we nail down how the tests scenarios should look like exactly

apps/omg_performance/config/config.exs
pdobacz: this?

pnowosie: So this is the first test that touches Watcher's RPC (and also ch-ch RPC), hence I think it's better to have all the urls together (e.g. somebody would like to just run exit perftest)

apps/omg_performance/lib/omg_performance/byzantine_events.ex
pdobacz: I see this got resolved but still the doc is out of date

apps/omg_performance/lib/omg_performance/byzantine_events/generators.ex
pdobacz: this?

apps/omg_performance/lib/performance.ex
```diff
@@ -44,12 +44,24 @@ defmodule OMG.Performance do
...
+
+  ## start_standard_exit_perftest runs test that fetches standard exit data from the Watcher
+  ## standard_exit_perftest is run on testnet make sure you followed instruction in `README.md` and `get`, `omg_childchain` & `omg_watcher` are running.
+
+  ```
+  mix run --no-start -e 'OMG.Performance.start_standard_exit_perftest([%{ addr: <<192, 206, 18, ...>>, priv: <<246, 22, 164, ...>>}], 3, "0xbc5f ...")'
+  ```
+
+  ## Parameters passed are: 1. list of senders, 2. number of users that fetching exits in parallel, 3. contract address
+  With default options, number of transactions sent to the network is 10 times the senders cound per each sender
```
pdobacz: count

docs/demo_05.md
pdobacz: this?

apps/omg_performance/lib/omg_performance/sender_server.ex
pdobacz: I resolved this (I think) in f3f4994911b6c16305bd484e2b200fa770676a01. The error handling is removed and tests pass

apps/omg_utils/lib/omg_utils/http_rpc/client.ex
pnowosie: I reverted the changes that moves HttpRPC.Client to OMG.Utils and Watcher RPC calls are made only from ByzantineEvents module.

apps/omg_watcher/lib/omg_watcher/http_rpc/adapter.ex
```diff
@@ -59,6 +56,10 @@ defmodule OMG.Watcher.HttpRPC.Adapter do
 
   def get_response_body(error), do: {:error, {:client_error, error}}
 
+  defp convert_keys_to_atoms(data) when is_list(data) do
+    data |> Enum.map(&convert_keys_to_atoms/1)
```
InoMurko: `Enum.map(data, &convert_keys_to_atoms/1)`

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -0,0 +1,146 @@
...
+  @spec get_exitable_utxos([%{addr: binary()}], watcher_url: binary()) :: [non_neg_integer()]
+  def get_exitable_utxos(entities, watcher_url \\ @watcher_url)
+
+  def get_exitable_utxos(addr, watcher_url) when is_binary(addr) do
+    {:ok, utxos} = WatcherClient.get_exitable_utxos(addr, watcher_url)
+    utxos
+  end
+
+  def get_exitable_utxos(%{addr: addr}, watcher_url) when is_binary(addr),
+    do: Encoding.to_hex(addr) |> get_exitable_utxos(watcher_url)
```
InoMurko: `get_exitable_utxos(Encoding.to_hex(addr), watcher_url)`

InoMurko: or if you want
```
addr
|> Encoding.to_hex() 
|> get_exitable_utxos(watcher_url)
```

apps/omg_performance/lib/omg_performance/byzantine_events/generators.ex
```diff
@@ -0,0 +1,128 @@
...
+
+  @doc """
+  Streams blocks from child chain rpc starting from the first block.
+  """
+  @spec stream_blocks(child_chain_url: binary()) :: [OMG.Block.t()]
+  def stream_blocks(child_chain_url \\ @child_chain_url) do
+    {:ok, interval} = RootChain.get_child_block_interval()
+
+    Stream.iterate(1, &(&1 + 1))
+    |> Stream.map(&get_block!(&1 * interval, child_chain_url))
```
InoMurko: please expand this

apps/omg_performance/lib/omg_performance/byzantine_events/generators.ex
```diff
@@ -0,0 +1,128 @@
...
+  defp to_utxo_position_list(block) do
+    Stream.with_index(block.transactions)
+    |> Stream.map(fn {tx, index} ->
+      recover_tx = Transaction.Recovered.recover_from!(tx)
+
+      Transaction.get_outputs(recover_tx)
+      |> Enum.with_index()
+      |> Enum.map(fn {_, oindex} ->
+        Utxo.position(block.number, index, oindex)
+        |> Utxo.Position.encode()
```
InoMurko:
```
Utxo.Position.encode(Utxo.position(block.number, index, oindex))
```
or (I prefer this one)
```
position = Utxo.position(block.number, index, oindex)
Utxo.Position.encode(position)
```

apps/omg_performance/lib/omg_performance/byzantine_events.ex
achiurizo:
```suggestion
    do: get_exitable_utxos(Encoding.to_hex(addr), watcher_url)
```

https://github.com/lexmag/elixir-style-guide#anonymous-pipeline

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -0,0 +1,146 @@
...
+  @doc """
+  For given utxo positions and given number of users start fetching exit data from Watcher.
+  User tasks are run asynchronously, each user receives the same positions list, shuffle its order
+  and call Watcher for exit data sequentially one position at a time.
+  """
+  @spec start_dos_get_exits([non_neg_integer()], [OMG.TestHelper.entity()], watcher_url: binary()) ::
+          stats_t()
+  def start_dos_get_exits(positions, dos_users, url \\ @watcher_url) do
+    1..dos_users
+    |> Enum.map(fn _ -> DoSExitWorker.get_exits_fun(positions, url) |> Task.async() end)
```
achiurizo: I think we should break this line up? Looks like the [pipe is unwarranted here](https://github.com/lexmag/elixir-style-guide#anonymous-pipeline). Also maybe we can use the other form here if you don't want to break this out into a named function?

```suggestion
    |> Enum.map(fn _ -> Task.async(DosExitWorker, :get_exits_fun, [positions, url]) end)
```

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -0,0 +1,146 @@
...
+
+  def watcher_synchronize(watcher_url \\ @watcher_url) do
+    Eth.WaitFor.repeat_until_ok(fn -> watcher_synchronized?(watcher_url) end)
+  end
+
+  def watcher_synchronize_service(expected_service, min_service_height, watcher_url \\ @watcher_url) do
+    Eth.WaitFor.repeat_until_ok(fn ->
+      with {:ok, %{services_synced_heights: services_synced_heights}} <- WatcherClient.get_status(watcher_url),
+           %{"height" => height} when height >= min_service_height <-
+             Enum.find(services_synced_heights, &match?(%{"service" => ^expected_service}, &1)) do
```
achiurizo: this is an amazing piece of code to parse 👀

apps/omg_performance/lib/omg_performance/byzantine_events.ex
pnowosie: Thanks for pointing me at it - I think function order is unfortunate here. Check now :)

pnowosie: I've refactored this further, here unfortunate fs order leads to bad design.

apps/omg_performance/lib/omg_performance/byzantine_events.ex
pnowosie: This is nice! Thank you ❗️

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -0,0 +1,146 @@
...
+    }
+  end
+
+  defp watcher_synchronized?(watcher_url) do
+    # Tricky part that deserves a note. This function is prepared to be called in `WaitFor.repeat_until_ok`.
+    # It repeatedly ask for Watcher's `/status.get` until:
+    #  1. last_mined_child_block_number == last_validated_child_block_number, so Watcher synced to last ch-ch block
+    #  2. last_validated_child_block_number > 0, and we expect there will be at least one block as it's called
+    # after perftest.
+    with {:ok,
```
InoMurko: with is not a good readable statement in my opinion
```suggestion
%{last_mined_child_block_number: last_mined_child_block_number,
    last_validated_child_block_number: last_validated_child_block_number} = WatcherClient.get_status(watcher_url)
case last_mined_child_block_number do
 ^last_validated_child_block_number when last_mined_child_block_number > 0
       _ = Logger.debug("Synced to blknum: #{last_validated_child_block_number}")
      {:ok, last_validated_child_block_number}
      _ -> :repeat
end
```
or
```suggestion
%{last_mined_child_block_number: last_mined_child_block_number,
    last_validated_child_block_number: last_validated_child_block_number} = WatcherClient.get_status(watcher_url)
case {last_mined_child_block_number, last_validated_child_block_number > 0} do
 {^last_validated_child_block_number, true}
       _ = Logger.debug("Synced to blknum: #{last_validated_child_block_number}")
      {:ok, last_validated_child_block_number}
      _ -> :repeat
end
```

apps/omg_performance/lib/omg_performance/byzantine_events/generators.ex
pnowosie: These is one of the examples we former proposition lacks readability and further introduce redundant variable (which also makes code harder to follow).

I'd opt to leave it as is and exceptionally break the rule.

InoMurko: let's stick to the convention we agreed we would use and see how it goes in the long run. We can still go back if we find it confusing later on!

apps/omg_performance/lib/omg_performance/byzantine_events.ex
pnowosie: 👍 like it

apps/omg_performance/lib/omg_performance/byzantine_events/generators.ex
pnowosie: What exactly you mean by expand here?

apps/omg_performance/lib/omg_performance/byzantine_events.ex
pnowosie: I think I push it even further ;)

apps/omg_performance/lib/omg_performance/byzantine_events.ex
pnowosie: Dead code spotted - removing. Thanks!

apps/omg_performance/lib/omg_performance/byzantine_events.ex
pnowosie: Looks easy but I was unable to make it work... refactored a little for readability.
Hope it's better now.

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -0,0 +1,146 @@
...
+          %{
+            last_mined_child_block_number: last_validated_child_block_number,
+            last_validated_child_block_number: last_validated_child_block_number
+          }}
+         when last_validated_child_block_number > 0 <- WatcherClient.get_status(watcher_url) do
```
InoMurko: is there a reason why you're avoiding case statements? it is difficult to notice that you're in fact expecting `last_mined_child_block_number == last_validated_child_block_number`.

correct this so that you use a pinned variable with a case statement if you want to use pattern matching.

apps/omg_performance/lib/omg_performance/byzantine_events/generators.ex
InoMurko: do not use single pipe statements
`Stream.map(Stream.iterate(1, &(&1 + 1)), &get_block!(&1 * interval, child_chain_url))`

apps/omg_performance/lib/omg_performance/byzantine_events/generators.ex
InoMurko: this is still unresolved

apps/omg_performance/lib/omg_performance/byzantine_events/generators.ex
```diff
@@ -0,0 +1,128 @@
...
+  defp generate_user(opts) do
+    user = OMG.TestHelper.generate_entity()
+    {:ok, _user} = Eth.DevHelpers.import_unlock_fund(user, opts)
+    user
+  end
+
+  defp get_block!(blknum, child_chain_url) do
+    {:ok, block} =
+      Eth.WaitFor.repeat_until_ok(fn ->
+        with {:ok, {block_hash, _timestamp}} <- RootChain.get_child_chain(blknum) do
```
InoMurko: developers were mentioning that the use of `with` statements is confusing because of these types. use `case` for single assigments.

apps/omg_performance/lib/omg_performance/http_rpc/watcher_client.ex
```diff
@@ -0,0 +1,61 @@
...
+  @spec get_exit_data(non_neg_integer(), binary()) :: OMG.Watcher.HttpRPC.Client.response_t()
+  def get_exit_data(encoded_position, url),
+    do: "utxo.get_exit_data" |> call(%{utxo_pos: encoded_position}, url) |> decode_response()
+
+  @doc """
+  Gets utxo for given address from Watcher's RPC
+  """
+  @spec get_exitable_utxos(OMG.Crypto.address_t(), binary()) :: OMG.Watcher.HttpRPC.Client.response_t()
+  def get_exitable_utxos(address, url),
+    do: "account.get_exitable_utxos" |> call(%{address: address}, url)
```
InoMurko: single pipe again

apps/omg_performance/lib/omg_performance/http_rpc/watcher_client.ex
```diff
@@ -0,0 +1,61 @@
...
+
+  defp call(path, params, url),
+    do: Adapter.rpc_post(params, path, url) |> Adapter.get_response_body()
```
InoMurko: single pipe

apps/omg_performance/lib/omg_performance/byzantine_events/generators.ex
```diff
@@ -0,0 +1,130 @@
...
+        end
+      end)
+
+    block
+  end
+
+  defp to_utxo_position_list(block) do
+    Stream.with_index(block.transactions)
+    |> Stream.map(fn {tx, index} ->
+      recover_tx = Transaction.Recovered.recover_from!(tx)
```
InoMurko: this is a rather complicated two pipe thingy. let's exract this anon. fun.

apps/omg_performance/lib/omg_performance/http_rpc/watcher_client.ex
```diff
@@ -0,0 +1,63 @@
...
+  @spec get_exit_data(non_neg_integer(), binary()) :: OMG.Watcher.HttpRPC.Client.response_t()
+  def get_exit_data(encoded_position, url),
+    do: call(%{utxo_pos: encoded_position}, "utxo.get_exit_data", url) |> decode_response()
+
+  @doc """
+  Gets utxo for given address from Watcher's RPC
+  """
+  @spec get_exitable_utxos(OMG.Crypto.address_t(), binary()) :: OMG.Watcher.HttpRPC.Client.response_t()
+  def get_exitable_utxos(address, url) when is_binary(address) and byte_size(address) == @address_bytes_size,
+    do: %{address: Encoding.to_hex(address)} |> call("account.get_exitable_utxos", url)
```
InoMurko: single pipe

apps/omg_performance/lib/performance.ex
```diff
@@ -139,6 +151,47 @@ defmodule OMG.Performance do
...
+          opts.exits_per_user
+        } exit data with #{exiting_users} users."
+      )
+
+    exit_positions = setup_standard_exit_perftest(opts)
+
+    statistics = ByzantineEvents.start_dos_get_exits(exit_positions, exiting_users)
+
+    %{
+      opts: Map.take(opts, [:spenders_count, :ntx_to_send, :exits_per_user]),
```
InoMurko: why do you need `Map.take`?

apps/omg_performance/lib/performance.ex
```diff
@@ -211,6 +264,28 @@ defmodule OMG.Performance do
...
+        %{addr: addr} ->
+          addr
+          |> ByzantineEvents.get_exitable_utxos()
+          |> Enum.map(& &1.utxo_pos)
+          |> Enum.shuffle()
+          |> Enum.take(opts.exits_per_user)
+
+        nil ->
+          ByzantineEvents.Generators.stream_utxo_positions()
+          |> Enum.take(opts.exits_per_user)
```
InoMurko: single pipe

apps/omg_performance/test/fixtures.exs
```diff
@@ -0,0 +1,104 @@
...
+      kill_group: true
+    ]
+
+    {:ok, _db_proc, _ref, [{:stream, db_out, _stream_server}]} =
+      Exexec.run_link(
+        "mix ecto.reset --no-start && mix run --no-start -e ':ok = OMG.DB.init()' --config #{config_file_path} 2>&1",
+        exexec_opts_for_mix
+      )
+
+    db_out |> Enum.each(&log_output("db_init_watcher", &1))
```
InoMurko: single pipe

apps/omg_performance/test/fixtures.exs
```diff
@@ -0,0 +1,104 @@
...
+
+    watcher_mix_cmd = "mix xomg.watcher.start --config #{config_file_path} 2>&1"
+
+    Logger.info("Starting watcher")
+
+    {:ok, watcher_proc, _ref, [{:stream, watcher_out, _stream_server}]} =
+      Exexec.run_link(watcher_mix_cmd, exexec_opts_for_mix)
+
+    fn ->
+      watcher_out |> Enum.each(&log_output("watcher", &1))
```
InoMurko: single pipe

apps/omg_performance/test/fixtures.exs
```diff
@@ -0,0 +1,104 @@
...
+          :normal ->
+            :ok
+
+          :shutdown ->
+            :ok
+
+          :noproc ->
+            :ok
+
+          other ->
```
InoMurko: no need to catch `other`, this will already produce an exception with a log

apps/omg_performance/lib/performance.ex
InoMurko: you cannot leave it unchanged because we agreed on specific rules we will use for the codebase.
I suggest you do:
```suggestion
          stream_utxo_positions = ByzantineEvents.Generators.stream_utxo_positions()
          Enum.take(stream_utxo_positions, opts.exits_per_user)
```

apps/omg_performance/test/fixtures.exs
```diff
@@ -0,0 +1,100 @@
...
+    Enum.each(db_out, &log_output("db_init_watcher", &1))
+
+    watcher_mix_cmd = "mix xomg.watcher.start --config #{config_file_path} 2>&1"
+
+    Logger.info("Starting watcher")
+
+    {:ok, watcher_proc, _ref, [{:stream, watcher_out, _stream_server}]} =
+      Exexec.run_link(watcher_mix_cmd, exexec_opts_for_mix)
+
+    fn ->
```
InoMurko:
```suggestion
        Task.async(fn -> Enum.each(watcher_out, &log_output("watcher", &1)) end)
```

apps/omg_performance/test/fixtures.exs
```diff
@@ -0,0 +1,100 @@
...
+            :ok
+
+          :shutdown ->
+            :ok
+
+          :noproc ->
+            :ok
+        end
+
+      File.rm(config_file_path)
```
InoMurko: maybe cleanup of files should be done first in case `stop_and_wait/1` crashes?

pnowosie: Not sure because:
1) can't wipe db when watcher is still alive
2) in a case of crash maybe we'd like to take a look into these files...

InoMurko: makes sense!

InoMurko:
good work!

Issue #901, opened 2019-08-21
Dependency conformance tests - test libs vs solidity implementations


pdobacz:
Taken from "OMG-394 - Exhaustive (possibly propchecky) tests of essential crypto and libs"

Some essential crypto-code we use in prod is tested quite thinly.

Let's make an exhaustive test suite which:
1/ generates some input data to the respective libs
2/ does Elixir calculations on it (invoked via some of our public APIs like `Transaction.hash`, `Transaction.encode`, `Block.hashed_txs_at` but effectively boiling down to dependency calls
3/ validates these calculations with Elixir code
4/ redos/validates these calculations against some public API of a `solc`-library (like `Merkle.sol` or `RLP.sol`) we use - deployed lightly to a `geth --dev` instance

test data generation can be manual, but this is perhaps a good (and much easier than state machine testing we did for `State.Core`) application of PropCheck

The candidates for such tests:
A/ `Block.hashed_txs_at` (invoking `merkle_tree_elixir`) vs `Merkle.sol` - correct proofs work and incorrect don't, for "all" imaginable combinations
B/ encoding/decoding https://github.com/omisego/security-issues/issues/2
C/ signatures

**Motivation** and parent issue is #853

EDIT - crossed RLP out
EDIT2 - uncrossed RLP back in (changed to generic encoding/decoding) :joy:

InoMurko:
If I understand correctly, you want to test if (for example) our dependency https://github.com/mana-ethereum/ex_rlp works as intended against `RLP.sol`?

InoMurko:
Would this go under https://github.com/omisego/elixir-omg/issues/853 @Pongch ?

pdobacz:
ah I forgot to mention #853, yes, adding to the desciption

InoMurko:
EX_RLP is running tests against Ethereum common tests https://github.com/ethereum/tests/tree/04940788157c88c75b014e5e6969a249f49db4c0 
not sure what coverage are we going to gain by adding tests against `RLP.sol`?

pdobacz:
> EX_RLP is running tests against Ethereum common tests

ah I wasn't aware of that (but kinda could've expected that). Thanks @InoMurko! that's one off the list or at least 1/2 off the list as:

That would leave `RLP.sol` to check - do we know whether it has the same coverage? @pik694 @boolafish I think you've mentioned touching RLP libs recently. What is the implementation of RLP that's going to be used in the long term for `plasma-contracts`?

[...]

PR #1005, opened 2019-09-25
integrate se challenges from ALD with elixir-omg

### LINKED ISSUE #986: Integrate with PlasmaFramework (ALD) for SE challenges
#949 integrates with ALD but leaves SE challenges out. The following needs to be done
 - [x] unskip and make pass the EIP712 signature conformance tests
 - [x] unremove the SE challenge integration test from `invalid_exit_test` (see TODO comment & blame)
 - [x] ~fix the reading of exit status on `new_exits` - just learned it's broken and will syncing old state with Watcher fail if challenges appear. This is because `exits` map in `PaymentExitGame` is now erased on processing. It is convenient to fix this when doing challenges. Will post details later (or will fix myself)~ EDIT: after closer inspection we're good, I must've been led astray with this. When we're reading the exit status of a processed exit, it returns zeros just fine and we can use this.
 - [x] from slack: `bytes exitingTx;` in `ChallengeStandardExitArgs` is new. We need to have this data when doing a SE challenge. Figure out the way to get it
 - [x] make the SE challenge integration test pass

Should land wherever #949 lands, `master` or feature branch

pdobacz:
fixes #986, builds upon #949 (which needs to merge to wherever we opt to merge it to)

## Overview

The integration test for SE challenges passes now.

## Changes

- integrate to contracts from rebased branch that has the spending condition registering finalized, this brought a couple of necessary modifications
- updated EIP712 hashing

## Testing

```
mix test --only integration test/omg_watcher/integration/invalid_exit_test.exs:68
```

among the usual tests.


pdobacz:
This still needs us to update the hashes in the Metamask compatibility test. For short term solution, would you be able to, @pnowosie? You'd probably be much faster to it than myself.

Also a long-term solution to test the compatibility (if we decide to keep this ofc) could be figured out. I had a bried idea to leverage [this utility from Metamask](https://github.com/MetaMask/eth-sig-util#typedsignaturehash-typeddata) maybe? Not sure if this was considered. It would require us to loop an `npm` package into the CI... :/ not nice, but if we might do it for `truffle` anyway in a near future, then maybe?

pnowosie:
> This still needs us to update the hashes in the Metamask compatibility test. For short term solution, would you be able to, @pnowosie? You'd probably be much faster to it than myself.

As you said long-term solution is to keep the tests outside. I discused with the omisego/omg-js and we agreed to combine tests there. (omisego/omg-js#130)

pdobacz:
OK, so I'll just :hocho: those tests out in this PR.

apps/omg/lib/omg/state/transaction/payment.ex
```diff
@@ -124,8 +124,10 @@ defmodule OMG.State.Transaction.Payment do
   end
 
   defp reconstruct_outputs(outputs_rlp) do
+    # TODO: ugly but to be cleaned up in the abstract output PR soon, so leaving as is
```
pdobacz: taking the liberty to cut a corner here. Handling this properly requires the abstract output type refactor, otherwise it's waste effort.

apps/omg/lib/omg/typed_data_hash.ex
```diff
@@ -47,6 +47,7 @@ defmodule OMG.TypedDataHash do
   @spec hash_transaction(Transaction.Payment.t()) :: Crypto.hash_t()
   def hash_transaction(%Transaction.Payment{} = raw_tx) do
```
InoMurko: this might be stupid but now that we're talking about specification docs so much. could you point me to where did you figure out that a payment transaction's hash consist of these fields? is there a doc or did you read solidity code?

apps/omg/lib/omg/typed_data_hash/config.ex
```diff
@@ -22,23 +22,31 @@ defmodule OMG.TypedDataHash.Config do
   alias OMG.TypedDataHash.Tools
 
   # Needed for test only to have value of address when `:contract_address` is not set
-  @fallback_ari_network_address "0x44de0ec539b8c4a4b530c78620fe8320167f2f74"
+  @fallback_contract_addr <<1::size(20)-unit(8)>>
```
InoMurko: could you help me understand why this is needed? I'm sensing there might be a better way of doing this...

apps/omg/test/omg/dependency_conformance/signature_test.exs
```diff
@@ -43,24 +40,41 @@ defmodule OMG.DependencyConformance.SignatureTest do
...
+    verify(contract, tx)
+  end
+
+  test "no inputs test", context do
+    contract = context[:contract]
+    tx = TestHelper.create_signed([], [{@alice, @eth, 100}])
+    verify(contract, tx)
+  end
+
+  test "no outputs test", context do
```
InoMurko: is this all `payment transaction` type?

apps/omg/test/omg/dependency_conformance/signature_test.exs
```diff
@@ -43,24 +40,41 @@ defmodule OMG.DependencyConformance.SignatureTest do
     {:ok, [addr | _]} = Ethereumex.HttpClient.eth_accounts()
 
     {:ok, _, signtest_addr} =
-      Eth.Deployer.create_new(OMG.Eth.Eip712SignatureWrapper, root_path, Eth.Encoding.from_hex(addr), [])
+      Eth.Deployer.create_new(OMG.Eth.PaymentEip712LibMock, root_path, Eth.Encoding.from_hex(addr), [])
```
InoMurko: first time seeing this... so is this deploying one special contract and you're identifying that contract by providing `OMG.Eth.PaymentEip712LibMock` module? why a module?

apps/omg/test/omg/state/core_test.exs
```diff
@@ -485,8 +485,8 @@ defmodule OMG.State.CoreTest do
 
     # precomputed fixed hash to check compliance with hashing algo
     assert block_hash ==
-             <<16, 139, 87, 31, 138, 148, 220, 238, 229, 44, 183, 103, 56, 33, 63, 158, 60, 71, 34, 132, 92, 211, 169,
-               221, 234, 110, 103, 233, 122, 84, 126, 180>>
+             <<218, 153, 13, 247, 52, 151, 31, 191, 14, 98, 96, 181, 113, 239, 1, 101, 78, 189, 159, 121, 97, 13, 89,
```
InoMurko: what made the signature change?

apps/omg/test/omg/state/transaction_test.exs
```diff
@@ -56,8 +56,8 @@ defmodule OMG.State.TransactionTest do
 
     test "raw transaction hash is invariant" do
       assert Transaction.raw_txhash(@transaction) ==
-               <<22, 144, 106, 111, 111, 175, 135, 22, 122, 230, 103, 220, 153, 161, 147, 239, 150, 205, 18, 226, 161,
-                 163, 127, 20, 163, 205, 24, 68, 0, 213, 103, 161>>
+               <<32, 195, 156, 58, 16, 182, 26, 126, 67, 146, 125, 172, 69, 109, 119, 124, 233, 106, 160, 244, 41, 62,
```
InoMurko: same as above, is that because a normal payment type transaction now consist of more fields then it did before ALD?

apps/omg/lib/omg/typed_data_hash/config.ex
pdobacz: yeah, I scratched my head a bit here too, finding a better way. So normally we want to use the address from the config. But if it's not there (e.g. in a unit test, but not in an integration test) we want just anything.

I tried to put the fallback addr into `config/test.exs` but something wasn't clicking.

I can revisit this now, having all the other stuff more or less behind.

apps/omg/test/omg/dependency_conformance/signature_test.exs
pdobacz: by inertia and convention I guess, but the convention broke anyway now.

If this is confusing, I'll poke around how to straighten out

apps/omg/test/omg/dependency_conformance/signature_test.exs
pdobacz: yep

apps/omg/test/omg/state/core_test.exs
pdobacz: Note that's a hash and it was the addition of `output_type` into the RLP encoding being the hash preimage that made it change.

apps/omg/test/omg/state/transaction_test.exs
pdobacz: yes, same reason and more specifically: addition of `output_type` field in the output

apps/omg/lib/omg/typed_data_hash.ex
pdobacz: https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/transactions/eip712Libs/PaymentEip712Lib.sol#L133 or https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/test/helpers/paymentEip712.js#L32

apps/omg/test/omg/dependency_conformance/signature_test.exs
```diff
@@ -43,24 +40,46 @@ defmodule OMG.DependencyConformance.SignatureTest do
...
-  test "signature test", context do
+  test "no outputs test", context do
+    contract = context[:contract]
+    tx = TestHelper.create_signed([{1, 0, 0, @alice}], [])
+    verify(contract, tx)
+  end
+
+  test "signature test - small tx", context do
+    contract = context[:contract]
+    tx = TestHelper.create_signed([{1, 0, 0, @alice}], [{@alice, @eth, 100}])
```
InoMurko: is there a way around using TestHelpers? I think they're somehow becoming a bucket for everything and make tests that should be self sufficient into a tree of jumps between modules and functions.

apps/omg/test/omg/dependency_conformance/signature_test.exs
```diff
@@ -87,21 +103,13 @@ defmodule OMG.DependencyConformance.SignatureTest do
...
-    {:ok, elixir_signer} =
-      tx
-      |> OMG.TypedDataHash.hash_struct()
-      |> OMG.Crypto.recover_address(signature)
+  defp verify(contract, %Transaction.Signed{raw_tx: tx}) do
+    {:ok, solidity_hash} =
+      Eth.call_contract(contract, "hashTx(address,bytes)", [contract, Transaction.raw_txbytes(tx)], [{:bytes, 32}])
 
-    assert solidity_signer == elixir_signer
+    assert solidity_hash == OMG.TypedDataHash.hash_struct(tx)
```
InoMurko: are these results not predictable? why do we need an external source to verify the results in what should be a unit test?

apps/omg/test/omg/state/transaction_test.exs
InoMurko: another noob question. why do we need `output_type` in the output hash?

apps/omg_eth/lib/omg_eth/root_chain.ex
```diff
@@ -294,7 +294,7 @@ defmodule OMG.Eth.RootChain do
   # TODO - missing description + could this be moved to a statefull process?
   @spec get_root_deployment_height(binary() | nil, optional_addr_t()) ::
           {:ok, integer()} | Ethereumex.HttpClient.error()
-  def get_root_deployment_height(txhash \\ nil, contract \\ nil) do
+  def get_root_deployment_height(txhash \\ nil, contract \\ %{}) do
```
InoMurko: if we're going down the route of replacing contracts with a map `%{}`, maybe a struct would be a better representation what's being expected?

apps/omg_eth/lib/omg_eth/root_chain.ex
```diff
@@ -479,7 +479,7 @@ defmodule OMG.Eth.RootChain do
 
   def decode_in_flight_exit_output_finalized(log) do
     non_indexed_keys = [:in_flight_exit_id, :output_index]
-    non_indexed_key_types = [{:uint, 192}, {:uint, 8}]
+    non_indexed_key_types = [{:uint, 160}, {:uint, 8}]
```
InoMurko: where in solidity did you find the change from uint 160 to 192?

apps/omg_watcher/test/omg_watcher/exit_processor/standard_exit_test.exs
```diff
@@ -104,6 +102,40 @@ defmodule OMG.Watcher.ExitProcessor.StandardExitTest do
...
+         %{alice: alice, processor_empty: processor} do
+      exiting_tx = TestHelper.create_recovered([], [{alice, @eth, 10}])
+      processor = processor |> start_se_from(exiting_tx, @utxo_pos_deposit)
+
+      recovered_spend = TestHelper.create_recovered([{@deposit_blknum, 0, 0, alice}], @eth, [{alice, 10}])
+      {txbytes, _alice_sig} = get_bytes_sig(recovered_spend)
+      {exiting_txbytes, _} = get_bytes_sig(exiting_tx)
+
+      assert {:ok, %{exiting_tx: ^exiting_txbytes, txbytes: ^txbytes}} =
+               %ExitProcessor.Request{
```
InoMurko: single pipe

apps/omg_watcher/test/omg_watcher/exit_processor/standard_exit_test.exs
```diff
@@ -104,6 +102,40 @@ defmodule OMG.Watcher.ExitProcessor.StandardExitTest do
   end
 
   describe "Core.create_challenge" do
+    test "returns a deposit exiting_tx as part of the challenge response",
+         %{alice: alice, processor_empty: processor} do
+      exiting_tx = TestHelper.create_recovered([], [{alice, @eth, 10}])
+      processor = processor |> start_se_from(exiting_tx, @utxo_pos_deposit)
```
InoMurko: single pipe

apps/omg_watcher/test/omg_watcher/integration/invalid_exit_test.exs
```diff
@@ -65,9 +65,34 @@ defmodule OMG.Watcher.Integration.InvalidExitTest do
     IntegrationTest.wait_for_byzantine_events([%Event.InvalidExit{}.name, %Event.InvalidExit{}.name], @timeout)
 
     # after the notification has been received, a challenged is composed and sent
-    _challenge = TestHelper.get_exit_challenge(first_tx_blknum, 0, 0)
-
-    # TODO: git blame here to bring back removed failing test code, when challenges are re-integrated with the contract
+    challenge = TestHelper.get_exit_challenge(first_tx_blknum, 0, 0)
+
+    assert {:ok, %{"status" => "0x1"}} =
+             OMG.Eth.RootChainHelper.challenge_exit(
```
InoMurko: single pipe

apps/omg/test/omg/dependency_conformance/signature_test.exs
pdobacz: perhaps something for a separate issue? Maybe let's propose a bunch of renames and split that particular module into submodules (entities, txs and fee-stuff separately), if it's really bugging you.

Note this PR doesn't really change the approach it just makes it consistent (use `create_signed` in all cases)

apps/omg/test/omg/dependency_conformance/signature_test.exs
pdobacz: not really, given the pace at which the EIP712 changes :). Also the point of this test is to test against a particular "reference" impl, vide our discussion on Merkling elsewhere (#901).

apps/omg/test/omg/state/transaction_test.exs
pdobacz: because a transaction referencing different `output_types` might mean a different thing and should thereby be included in the signed struct hash.

apps/omg_eth/lib/omg_eth/root_chain.ex
pdobacz: yeah, let's maybe tackle separately, this is the part that's not fully fleshed out yet (contract address management), but I think pieces of the puzzle are slowly falling into their places now.

apps/omg_eth/lib/omg_eth/root_chain.ex
pdobacz: (I hope I got this right, since IFEs aren't really tackled yet - separate issue).

This one I actually was just aware of (was following the debate on the topic of the change 192->160). But you can also see it in the diffs, wherever `exit_id`/`exitId` is mentioned.

apps/omg_watcher/test/omg_watcher/exit_processor/standard_exit_test.exs
pdobacz: not in this PR. Changing this just here would make it inconsistent with the rest of the file. Changing everywhere would obscure the PR with unrelated style changes, while I want this PR to only contain the changes relevant to/required by the change it's centered around.

Among other reasons behind this approach: your ability to review.

apps/omg_watcher/test/omg_watcher/exit_processor/standard_exit_test.exs
pdobacz: see other comment

apps/omg_watcher/test/omg_watcher/integration/invalid_exit_test.exs
pdobacz: same here, not in this PR. This test is the same as the test removed in the previous commit, which is the point I'm trying to make here. If it wasn't removed in the prior commit, we'd see this in the diff here (you can still see this, if you search for the right diff - against pre-ALD `master`).

PR #974, opened 2019-09-17
set inserted_at and updated_at null columns to epoch and add not null…

jrhite:
See the testing steps 2 and 3 below. 2 is pre-migration. Null values are still allowed. 3. is post-migration and notice the not null modifier. Finally see in step 4 the values are epoch times.

## Testing
1. Grab dump of dev DB locally and populate local DB:
`$ psql omisego_dev -h localhost -p 5433 -U omisego_dev < ~/Downloads/omisego_dev.dump`
2. Verify the txoutputs table is the table prior to migration
```
omisego_dev=# \d txoutputs;
                                   Table "public.txoutputs"
        Column        |              Type              |              Modifiers               
----------------------+--------------------------------+--------------------------------------
 blknum               | bigint                         | not null
 txindex              | integer                        | not null
 oindex               | integer                        | not null
 creating_txhash      | bytea                          | 
 spending_txhash      | bytea                          | 
 spending_tx_oindex   | integer                        | 
 owner                | bytea                          | not null
 amount               | numeric(81,0)                  | not null
 currency             | bytea                          | not null
 proof                | bytea                          | 
 child_chain_utxohash | bytea                          | 
 inserted_at          | timestamp(0) without time zone | default timezone('utc'::text, now())
 updated_at           | timestamp(0) without time zone | default timezone('utc'::text, now())
Indexes:
    "txoutputs_pkey" PRIMARY KEY, btree (blknum, txindex, oindex)
    "txoutputs_child_chain_utxohash_index" UNIQUE, btree (child_chain_utxohash)
    "txoutputs_creating_txhash_spending_txhash_index" btree (creating_txhash, spending_txhash)
    "txoutputs_owner_index" btree (owner)
    "txoutputs_spending_txhash_index" btree (spending_txhash)
Foreign-key constraints:
    "txoutputs_creating_txhash_fkey" FOREIGN KEY (creating_txhash) REFERENCES transactions(txhash)
    "txoutputs_spending_txhash_fkey" FOREIGN KEY (spending_txhash) REFERENCES transactions(txhash)
Referenced by:
    TABLE "ethevents_txoutputs" CONSTRAINT "ethevents_txoutputs_child_chain_utxohash_fkey" FOREIGN KEY (child_chain_utxohash) REFERENCES txoutputs(child_chain_utxohash) ON DELETE RESTRICT
```
3. Run the migration and verify the output
```
omisego_dev=# \d txoutputs;
                                       Table "public.txoutputs"
[...]

PR #1008, opened 2019-09-26
add timestamps to all table and fix tables that currently have timest…

jrhite:
## Overview

Add timestamps to all tables that support UTC times with time zone to microsecond granularity.

This is per general best practice. For more info please reference: https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp_.28without_time_zone.29_to_store_UTC_times

## Changes

- Fix the way migrations are done. Ecto's timestamp stuff is poorly named and works in a very unexpected way.
  - Migration files need to use `timestamps([type: :timestamptz, default: "1970-01-01T00:00:00Z"])` to get the correct column type. Note the `:timestamptz`type
  - The actual db model files need to use: `timestamps(type: :utc_datetime_usec)`
- New migration file added to fix tables already containing timestamps and also to correctly add timestamps to existing tables.
- Timestamps removed from join table `ethevents_txoutputs` because ecto has poor support for join tables with extra columns.
- Update all db model files to have: `timestamps(type: :utc_datetime_usec)`

## Testing

1. Start with database containing `blocks` and `transactions` rows that already exist, but have no `inserted_at` and `updated_at` columns.
2. Also start with the database containing `ethevents` and `txoutputs` rows that already exist, and also have `inserted_at` and `updated_at` columns.
3. Run the migration
4. Inspect the `blocks` and `transactions` tables to confirm NON NULL constraints were not violated and they contain the current default values of `1970-01-01T00:00:00Z`.
5. Create new transactions and blocks and confirm they contain correct UTC `inserted_at` and `updated_at` data equaling 'now'.
6. Run a 'fake' test that updates the `ethevents` table to confirm that the `updated_at` column gets updated when the row is updated.


apps/omg_watcher/lib/omg_watcher/db/eth_event.ex
```diff
@@ -44,7 +44,7 @@ defmodule OMG.Watcher.DB.EthEvent do
       join_keys: [root_chain_txhash_event: :root_chain_txhash_event, child_chain_utxohash: :child_chain_utxohash]
     )
 
-    timestamps(type: :utc_datetime)
+    timestamps(type: :utc_datetime_usec)
```
pnowosie: What is the reason for this change... last migration script will change it to `timestamptz`
Will the last script work on DB that have `utc_datetime` timestamps?

apps/omg_watcher/priv/repo/migrations/20190925132500_add_and_fix_timestamps.exs
```diff
@@ -0,0 +1,34 @@
+defmodule OMG.Watcher.Repo.Migrations.SetInsertedAtUpdatedAtToEpoch do
+  use Ecto.Migration
+
+  def change do
+    alter table(:ethevents) do
+      remove :inserted_at
+      remove :updated_at
+
+      timestamps([type: :timestamptz, default: "1970-01-01T00:00:00Z"])
```
pnowosie: If I'm correct `default` does not fill already empty (NULL) timestamps. For me `epoch` is OK for these, but for new records I'd prefer to have `now()` instead. WDYT @jrhite :question:

apps/omg_watcher/lib/omg_watcher/db/txoutput.ex
```diff
@@ -141,7 +141,9 @@ defmodule OMG.Watcher.DB.TxOutput do
       _ =
         DB.TxOutput
         |> where(blknum: ^blknum, txindex: ^txindex, oindex: ^oindex)
-        |> Repo.update_all(set: [spending_tx_oindex: spending_oindex, spending_txhash: spending_txhash])
+        |> Repo.update_all(
+          set: [spending_tx_oindex: spending_oindex, spending_txhash: spending_txhash, updated_at: DateTime.utc_now()]
```
pnowosie: It looks like ecto supports `timestamps` in combination with `changesets` - am I right?

apps/omg_watcher/lib/omg_watcher/db/eth_event.ex
jrhite: The last migration uses `utc_datetime` which is timezone unaware. Really bad naming on Ecto's part, in my opinion broken. If `utc_datetime`, then what is `naive_datetime`? Anser: also timezone unaware. 😄 

Also we want `utc_datetime_usec` as that is to the microsecond instead of `utc_datetime` which is only to the second.

apps/omg_watcher/priv/repo/migrations/20190925132500_add_and_fix_timestamps.exs
jrhite: `default` does fill pre-existing rows with the default value set above. Otherwise a NOT NULL violation would occur when the new column is added.

`default` does not affect new records. new records will be set to UTC `now`.

apps/omg_watcher/lib/omg_watcher/db/txoutput.ex
jrhite: Yes it does.

apps/omg_watcher/lib/omg_watcher/db/eth_event.ex
pnowosie: Ah sorry - there is no question - I realized you are tweaking Ecto.Schema not old migrations...
All good!

apps/omg_watcher/priv/repo/migrations/20190925132500_add_and_fix_timestamps.exs
pnowosie: I disagree here, I think it's exactly the opposite:
* default does affect only new records where column's value is not specified [ecto](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#add/3-options), [pg](https://www.postgresql.org/docs/9.3/ddl-default.html)
* default does NOT affect existing rows, so with combination with NOT NULL constraint would result of migration failure. [here is your fix that fixes NOT NULL problem](https://github.com/omisego/elixir-omg/pull/974/files#diff-96f62f10bc1471112c742df795e04389)

InoMurko: what if we write a test and confirm what the behavior is?

jrhite: Regarding point 1: Hmm...I could share my screen and demo it for you. You may be right that `default` affects new columns where the code explicitly does not set a value. But it certainly also has the effect of backfilling pre-existing rows that existed before the new non-nullable columns. The `timestamps()` columns are indeed managed by ecto, with the caveat that calls like `insert_all()` and `update_all()` will bypass ecto's autogenerated columns.

For point 2, yes this is correct, but that's for the 2 tables I updated that already had nullable `inserted_at` and `updated_at` columns. We wanted to make them non-nullable, this the fix you reference.

pnowosie: Yes my bad here.
What I quickly checked on raw postgres is when `NOT NULL` constraint is set during column addition (table exists) Default will fill pre-existing records with the value.
Inserting new records without specifying column's value will fail due to constraint.

I'm glad to learn / review what I thought I know :)

sirn:
Don’t we also need updating existing timestamps? Or there’s no existing timestamp as of this writing? :thinking:

jrhite:
@sirn none of elixir-omg's table have ecto's timestamp columns `inserted_at` and `updated_at`. 

There are some other timestamp like columns in the `transactions` and `blocks` table. But those are different.

apps/omg_watcher/priv/repo/migrations/20190925132500_add_and_fix_timestamps.exs
```diff
@@ -0,0 +1,36 @@
+defmodule OMG.Watcher.Repo.Migrations.AddAndFixTimestamps do
+  use Ecto.Migration
+
+  def change do
```
unnawut: Without myself having tested this, I want to note that this will probably fail on rollbacks as it would not have enough info to reconstruct `:inserted_at`, `updated_at` fields, etc. If rollback needs to be supported `change/0` here will need to be broken down to `up/0` and `down/0`.

unnawut:
Also, to test this, adding tests for `OMG.Watcher.DB.Repo.insert_all_chunked /3` 
and `OMG.Watcher.DB.TxOutput.spend_utxos/1` and assert for the updated `updated_at` would be nice.

apps/omg_watcher/lib/omg_watcher/db/block.ex
```diff
@@ -26,6 +26,8 @@ defmodule OMG.Watcher.DB.Block do
     field(:hash, :binary)
     field(:eth_height, :integer)
     field(:timestamp, :integer)
+
+    timestamps(type: :utc_datetime_usec)
```
unnawut: There's a way to apply a specific timestamp format to the entire system, by setting `migration_timestamps` as repo settings and `@timestamps_opts` in each schema:

http://blog.plataformatec.com.br/2018/10/a-sneak-peek-at-ecto-3-0-breaking-changes/

apps/omg_watcher/priv/repo/migrations/20190925132500_add_and_fix_timestamps.exs
jrhite: I don't think rollback needs to be supported. But even if we wanted to, when a column is deleted, there's no way to recover the data and thus the `inserted_at` and `updated_at` columns cannot be reconstructed (at least not with the previous data the columns contained).

I'm fully taking advantage of the fact that we can get away with this until mainnet. 🙂

apps/omg_watcher/lib/omg_watcher/db/block.ex
jrhite: Oooh, this would be nice!
```
config :my_app, MyApp.Repo,
  migration_timestamps: [type: :timestamptz]
```

I think we can get away with adding this now since I just started adding timestamps.

Good comment!

sirn: Since we already have `migration_timestamps` set, can’t we just leave this as simply `timestamps` (and probably define `@timestamps_opts [type: :utc_datetime_usec]` at very top?

apps/omg_watcher/test/omg_watcher/db/repo_test.exs
```diff
@@ -0,0 +1,44 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.Watcher.DB.RepoTest do
+  import Mox
+
+  use ExUnitFixtures
+  use OMG.Watcher.Case, async: true
+  use OMG.Fixtures
```
InoMurko: what is being imported and why?

apps/omg_watcher/test/omg_watcher/db/repo_test.exs
```diff
@@ -0,0 +1,44 @@
...
+
+  import Ecto.Query, only: [from: 2]
+
+  alias OMG.Watcher.DB
+
+  # Make sure mocks are verified when the test exits
+  setup :verify_on_exit!
+
+  @tag fixtures: [:phoenix_ecto_sandbox]
```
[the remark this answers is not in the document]
jrhite: to be honest, I'm not sure why. i was trying to just use the `watcher` fixture as I don't need all the extra stuff, but i kept seeing 'db unavailable` type errors until i used `phoenix_ecto_sandbox`. happy to update this to whatever works best though.

jrhite:
@InoMurko, any more questions/comments or are we good to go from your end?

InoMurko:
Ill see if this week we’ll have enough time to review this properly.

apps/omg_watcher/test/omg_watcher/db/repo_test.exs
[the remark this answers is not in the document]
jrhite: I just discovered why I needed this know. Our fixtures need some love. If I run `MIX_ENV=test mix test test/omg_watcher/db/repo_test.exs` and run it as a single test `phoenix_ecto_sandbox` is needed.

If I run something like `MIX_ENV=test mix test test/omg_watcher` then this will cause a failure  `Failed to start Ranch listener OMG.WatcherRPC.Web.Endpoint.HTTP in :ranch_tcp:listen([port: 7435]) for reason :eaddrinuse (address already in use)`

The fixture required by the test is dependent on how the test is run which is likely not what we want, but beyond the scope of this PR.

apps/omg_watcher/test/omg_watcher/db/txoutput_test.exs
```diff
@@ -13,12 +13,16 @@
 # limitations under the License.
 
 defmodule OMG.Watcher.DB.TxOutputTest do
+  import Mox
```
InoMurko: can you be more explicit in what you're importing?

jrhite:
@InoMurko yes I can do that. Shall we continue with Mox or not?

apps/omg_watcher/test/omg_watcher/db/txoutput_test.exs
jrhite: @InoMurko yes I can do that. Shall we continue with Mox or not?

apps/omg_watcher/lib/omg_watcher/db/repo.ex
```diff
@@ -36,6 +36,10 @@ defmodule OMG.Watcher.DB.Repo do
     chunk_size = @max_params_count |> div(entries |> hd |> fields_count)
 
     entries
+    |> Enum.map(fn entry ->
+      utc_now = OMG.Watcher.DateTime.utc_now()
+      Map.merge(entry, %{inserted_at: utc_now, updated_at: utc_now})
+    end)
```
unnawut: Is it worth adding a new `OMG.Watcher.DateTime` to do all the mocks if we could take one extra option instead? i.e.

```elixir
def insert_all_chunked(schema_or_source, entries, opts) do
  # ...
  {datetime, opts} = Keyword.pop(opts, :override_datetime, DateTime.utc_now())

  entries
  |> Enum.map(fn entry ->
    Map.merge(entry, %{inserted_at: datetime, updated_at: datetime})
  end)

  # ...
end
```

While arguably this changes the implementation to support a test, I think the mock introduces obscurity and complexity to this code by even larger & more hidden implementation in `OMG.Watcher.DateTime`.

jrhite: I don't understand how your code snippet this makes this code testable. Making a bare call to `DateTime.utc_now()` any where makes it impossible to control time. By wrapping it in a thin wrapper, we can either make the wrapper be a direct call-through to the native `DateTime` or we can direct it to a mock in our tests.

Also every time a test against this function runs it will potentially be a flaky test because the test_data will be random with every run (specifically DateTime.utc_now() will return a different value).

Can you explain a little bit more about your idea here?

Having said that I'm just about the mocking idea go and just write some flaky tests so this PR doesn't block anymore 😄 . If we're not ready to use mocks in our tests yet, then when can come back to this.

apps/omg_watcher_info/test/omg_watcher_info/db/txoutput_test.exs
```diff
@@ -43,4 +43,19 @@ defmodule OMG.WatcherInfo.DB.TxOutputTest do
     assert not is_nil(utxo)
     assert utxo.amount == big_amount
   end
+
+  @tag fixtures: [:initial_blocks]
+  test "spend_utxos sets timestamps correctly" do
```
unnawut: I think it should be possible to follow the same way as the test above: `DB.Block.insert_with_transactions/1` followed by `utxo = DB.TxOutput.get_by_position/1` to get the utxo.

Then pass that utxo into `DB.TxOutput.spend_utxos/1` then assert for the changed dates.

InoMurko:
@jrhite please consider writing tests without using fixtures, I think @unnawut can greatly help you with that.

apps/omg_watcher_info/test/omg_watcher_info/db/txoutput_test.exs
unnawut: 2nd thought if we're moving away from fixtures, then we can just directly pre-insert some utxos and use them in `spend_utxos/1`.

InoMurko: yes!!! if tests can be self contained, that would make test simple and readable! thats what we want and should strive for.

jrhite: Totally agree. Will update the PR and remove fixtures from tests.

unnawut:
> @jrhite please consider writing tests without using fixtures, I think @unnawut can greatly help you with that.

1. We can start with minimizing fixtures in new tests, i.e. use only:

    ```elixir
    @tag fixtures: [:phoenix_ecto_sandbox]
    ```

    Fixtures removal will go in a separate PR. E.g. see https://github.com/omisego/elixir-omg/commit/b6a230ff85112f74b4ded328e606b3c52fee4e7b for some work in progress

2. We can still introduce `ex_machina` in this PR, it should be compatible with `:phoenix_ecto_sandbox` fixture.

jrhite:
A very stale PR that I've picked up gain. Plan to have a mature PR out for this tomorrow. Want to get this done finally.

jrhite:
> Marking as change requested in case it gets accidentally merged from the 2 approvals.

👍   yeah it's been a while and could use a fresh look.

jrhite:
@unnawut just updating the ticket here with info from Slack and my follow up:

> My comment about DB.Block.insert_with_transactions/1 was before introducing ex_machina though.
> So now I would just go with ExMachina’s insert/2 and not even insert_all_chunked because we don’t really need “to avoid too many parameters error” that insert_all_chunked/3 was created for: https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher_info/lib/omg_watcher_info/db/repo.ex#L23

The idea you had for testing `insert_all_chunked` and `spend_utxos` was because they call `insert_all` and `update_all`, so we're testing that these functions did not forget to add the `inserted_at` and `updated_at` data, because Ecto's `insert_all` and `update_all` does not.

I think these are still valid tests.

Per your Slack comment, it may also be a good idea to add tests for `insert` and `update` directly on each model file (eg: DB.Block, etc..). This essentially just tests that each model file has correctly added `timestamps(type: :utc_datetime_usec)` to its schema.

Thoughts?

unnawut:
> The idea you had for testing `insert_all_chunked` and `spend_utxos` was because they call `insert_all` and `update_all`, so we're testing that these functions did not forget to add the `inserted_at` and `updated_at` data, because Ecto's `insert_all` and `update_all` does not.
> 
> I think these are still valid tests.

Nope I didn't imply that those tests are invalid. `insert_all_chunked` and `spend_utxos` should still be tested. So all good here. 👍 

---

Though the original discussion was about test data factories:

> @jrhite: We can talk about breaking the factories up or not if you'd like. My first attempt to exercise the insert_all_chunked code under test was be looking at how fixtures did it and by calling Block.insert_with_transactions. This quickly started going down a deep rabbit hole of signing transactions, 'recovering' transactions and on and on. With the separate factories a lot of complexity could be hidden in each factory as needed while also still providing simple map objects by just calling build. 

So I responded that:

> @unnawut: My comment about DB.Block.insert_with_transactions/1 was before introducing ex_machina though. So now I would just go with ExMachina’s insert/2 and not even insert_all_chunked because we don’t really need “to avoid too many parameters error” that insert_all_chunked/3 was created for: https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher_info/lib/omg_watcher_info/db/repo.ex#L23

Meaning that the factories shouldn't need to go through the complex transaction signing & recovering to build the test data because we don't need to do insert data through `DB.Block.insert_with_transactions/1` (which requires those signing & recovery) anymore.

But I see that there's no "signing transactions, 'recovering' transactions and on and on" in the factories anyway. So all good as well. 👍 

---

Per your Slack comment, it may also be a good idea to add tests for insert and update directly on each model file (eg: DB.Block, etc..). This essentially just tests that each model file has correctly added timestamps(type: :utc_datetime_usec) to its schema.

Yup sounds good!

unnawut:
Superceded by https://github.com/omisego/elixir-omg/pull/1008

jrhite:
Thanks @unnawut! 🙏 was, just about to close this. Minor clarification, the superceding PR is #1322

apps/omg_watcher_info/lib/omg_watcher_info/db/eth_event.ex
```diff
@@ -40,82 +41,131 @@ defmodule OMG.WatcherInfo.DB.EthEvent do
...
+          })
+        end
+      )
+      |> DB.Repo.transaction()
+
+    case result do
+      {:ok, _changes} -> :ok
+      {:error, _failed_operation, _failed_value, _changes_so_far} -> :error
     end
   end
```
jrhite: .

PR #1012, opened 2019-09-27
Tidy composing SEs and prevent timeouting of request to the Watcher

pdobacz:
related to the #855 effort. Driven by results from #998 

## Overview

There has been some dead code in the handling of composing data for standard exits in the Watcher.

Also it has been discovered, that when child chain blocks are larger, requests to the Watcher that ask for standard exit data tend to timeout. `Block.inclusion_proof` has received some no-brainer optimizations to remove the bottleneck.

In case of `Block.inclusion_proof`, for a block of 20K txs, the time to generate the proof dropped from the timeoutable 3.5s to 0.5s. After this optimization, the blocker is now the encoding of `raw_tx`, necessary to hash it. If necessary the further optimization would be to somehow not have to encode, but this is more involved.

## Changes

- removes dead code
- tidies the flows, adds some tests conforming to the new testing standards
- optimizes the use of `Block.inclusion_proof` which (according to `:fprof`) did a lot of unnecessary encoding/decoding what not
- `OMG.DB` can now get UTXOs by utxo_pos

## Testing

`mix test` - logic
`mix test --only wrappers test/omg_db` for the addition to `OMG.DB`
`mix test --only integration test/omg_watcher/integration/block_getter_test.exs:126` all working together

### DIFF
--- a/apps/omg_db/test/omg_db/db_test.exs
+++ b/apps/omg_db/test/omg_db/db_test.exs
@@ -78,6 +78,14 @@ defmodule OMG.DBTest do
     {:ok, ["xyz", "vxyz", "wvxyz"]} = OMG.DB.block_hashes([1, 2, 3], pid)
   end
 
+  test "utxo can be fetched by utxo position", %{db_pid: pid} do
+    index = 123
+    item = {{index, index, index}, %{test: :crypto.strong_rand_bytes(index)}}
+    db_writes = [{:put, :utxo, item}]
+    :ok = write(db_writes, pid)
+    assert {:ok, ^item} = DB.utxo({index, index, index}, pid)
+  end
+
   test "if multi reading exit infos returns writen results", %{db_dir: _dir, db_pid: pid} do
     db_writes = create_write(:exit_info, pid)
     {:ok, exits} = DB.exit_infos(pid)
--- a/apps/omg_watcher/lib/omg_watcher/utxo_exit/core.ex
+++ b/apps/omg_watcher/lib/omg_watcher/utxo_exit/core.ex
@@ -20,38 +20,22 @@ defmodule OMG.Watcher.UtxoExit.Core do
   alias OMG.Block
   alias OMG.State.Transaction
   alias OMG.Utxo
-  alias OMG.Watcher.DB
 
   require Utxo
 
-  @spec compose_output_exit(list(%DB.Transaction{}) | list(Transaction.Signed.tx_bytes()), Utxo.Position.t()) ::
-          {:error, :utxo_not_found}
-          | {:ok,
-             %{
-               utxo_pos: non_neg_integer(),
-               txbytes: binary,
-               proof: binary,
-               sigs: binary
-             }}
-  def compose_output_exit([%{txindex: _} | _] = txs, Utxo.position(_, txindex, _) = utxo_pos) do
-    if Enum.any?(txs, &match?(%{txindex: ^txindex}, &1)) do
-      sorted_tx_bytes = Enum.sort_by(txs, & &1.txindex) |> Enum.map(& &1.txbytes)
-      compose_output_exit(sorted_tx_bytes, utxo_pos)
-    else
-      {:error, :utxo_not_found}
-    end
-  end
+  def compose_output_exit(:not_found, _), do: {:error, :utxo_not_found}
 
-  def compose_output_exit(sorted_tx_bytes, Utxo.position(_blknum, txindex, _) = utxo_pos) do
-    if signed_tx = Enum.at(sorted_tx_bytes, txindex) do
-      %Transaction.Signed{sigs: sigs} = tx = Transaction.Signed.decode!(signed_tx)
+  def compose_output_exit(db_block, Utxo.position(blknum, txindex, _) = utxo_pos) do
+    %Block{transactions: sorted_txs_bytes, number: ^blknum} = block = Block.from_db_value(db_block)
+
+    if signed_tx_bytes = Enum.at(sorted_txs_bytes, txindex) do
+      signed_tx = Transaction.Signed.decode!(signed_tx_bytes)
 
       {:ok,
        %{
          utxo_pos: Utxo.Position.encode(utxo_pos),
-         txbytes: Transaction.raw_txbytes(tx),
-         proof: Block.inclusion_proof(sorted_tx_bytes, txindex),
-         sigs: Enum.join(sigs)
+         txbytes: Transaction.raw_txbytes(signed_tx),
+         proof: Block.inclusion_proof(block, txindex)
        }}
     else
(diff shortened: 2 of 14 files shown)


apps/omg_watcher/test/omg_watcher/utxo_exit/core_test.exs
```diff
@@ -25,82 +25,85 @@ defmodule OMG.Watcher.UtxoExit.CoreTest do
...
+        |> Block.to_db_value()
+
+      assert {:ok,
+              %{
+                proof: proof,
+                txbytes: ^tx_exit_raw_tx_bytes,
+                utxo_pos: ^encode_utxo
+              }} = Core.compose_block_standard_exit(block, position)
+
+      assert byte_size(proof) == 32 * 16
```
pnowosie: `32 * 16` - doesn't tell me anything :open_hands: , also `64 * 8` would be misleading as there are bytes not bits...
I'm opting to make it local var, e.g. `expected_signature_byte_size = 512`

mix.lock
```diff
@@ -66,7 +66,7 @@
   "spandex_ecto": {:hex, :spandex_ecto, "0.6.0", "feedda8456deb37a87ee698f1f779304b68e7e42afc7c1fe5b9f801a732b89d6", [:mix], [{:spandex, "~> 2.2", [hex: :spandex, repo: "hexpm", optional: false]}], "hexpm"},
   "spandex_phoenix": {:hex, :spandex_phoenix, "0.4.1", "d52e12f801fce4efba658ac4e98be18ea9903ecf9f37682b401656ed8d9c7abe", [:mix], [{:plug, "~> 1.3", [hex: :plug, repo: "hexpm", optional: false]}, {:spandex, "~> 2.2", [hex: :spandex, repo: "hexpm", optional: false]}], "hexpm"},
   "ssl_verify_fun": {:hex, :ssl_verify_fun, "1.1.4", "f0eafff810d2041e93f915ef59899c923f4568f4585904d010387ed74988e77b", [:make, :mix, :rebar3], [], "hexpm"},
-  "statix": {:git, "git@github.com:bleacherreport/statix.git", "f27c37bee69d1bbae89532049f6d3250b348f0ac", [branch: "otp-21.3.8.4-support"]},
+  "statix": {:git, "https://github.com/bleacherreport/statix.git", "f27c37bee69d1bbae89532049f6d3250b348f0ac", [branch: "otp-21.3.8.4-support"]},
```
pnowosie: Aha - I had the same issue :+1:

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/utxo_test.exs
```diff
@@ -57,14 +57,13 @@ defmodule OMG.WatcherRPC.Web.Controller.UtxoTest do
...
         %{amount: 100, creating_txhash: OMG.State.Transaction.raw_txhash(tx), currency: @eth, owner: bob.addr}}},
-      {:put, :block, %{number: 1000, hash: 0, transactions: [tx_encode]}}
+      {:put, :block, %{number: 1000, hash: <<>>, transactions: [tx_encode]}}
     ])
 
     %{
       "utxo_pos" => _utxo_pos,
       "txbytes" => _txbytes,
-      "proof" => proof,
-      "sigs" => _sigs
```
pnowosie: What happened to `sigs` - our swagger says it's returned :question: 
https://developer.omisego.co/elixir-omg/docs-ui/?url=master/security_critical_api_specs.yaml#/UTXO/utxo_get_exit_data

I consider it valuable to have 1 test that checks the response structure.

apps/omg_db/test/omg_db/db_test.exs
```diff
@@ -78,6 +78,14 @@ defmodule OMG.DBTest do
     {:ok, ["xyz", "vxyz", "wvxyz"]} = OMG.DB.block_hashes([1, 2, 3], pid)
   end
 
+  test "utxo can be fetched by utxo position", %{db_pid: pid} do
+    index = 123
+    item = {{index, index, index}, %{test: :crypto.strong_rand_bytes(index)}}
```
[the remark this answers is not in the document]
pdobacz: true, but since this is ubiquitous in db tests, I'll leave as is for consistency and self-contained PR. If necessary, let's have a separate issue/PR for this

apps/omg_watcher/lib/omg_watcher/utxo_exit/core.ex
```diff
@@ -20,62 +20,39 @@ defmodule OMG.Watcher.UtxoExit.Core do
...
+  def compose_block_standard_exit(:not_found, _), do: {:error, :utxo_not_found}
+
+  def compose_block_standard_exit(db_block, Utxo.position(blknum, txindex, _) = utxo_pos) do
+    %Block{transactions: sorted_txs_bytes, number: ^blknum} = block = Block.from_db_value(db_block)
 
-  def compose_output_exit(sorted_tx_bytes, Utxo.position(_blknum, txindex, _) = utxo_pos) do
-    if signed_tx = Enum.at(sorted_tx_bytes, txindex) do
-      %Transaction.Signed{sigs: sigs} = tx = Transaction.Signed.decode!(signed_tx)
+    if signed_tx_bytes = Enum.at(sorted_txs_bytes, txindex) do
+      signed_tx = Transaction.Signed.decode!(signed_tx_bytes)
```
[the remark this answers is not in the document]
pdobacz: nice catch. I'll fix to return `utxo_not_found` in this case and add a test

apps/omg_watcher/test/omg_watcher/utxo_exit/core_test.exs
pdobacz: it's 32 bytes times 16 (tree depth). It's the size of thep proof not signature. Also it's consistent with other occurrences of a similar test.

Not sure if adding an `expected_proof_byte_size = 32 * 16` is worth the extra line, this is exactly what the RHS of the assertion should be. I can add a `# hash byte size times tree depth` comment instead to clarify the missing bit

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/utxo_test.exs
pdobacz: the sigs are never used, if I'm not mistaken, submitting an exit doesn't require that. I think this is some ages-old leftovers. Good catch on the swagger though - looks like I have to delete there as well

apps/omg_watcher/test/omg_watcher/utxo_exit/core_test.exs
pnowosie: Ah true - I misget it with signature...

PR #1028, opened 2019-10-03
refactor: clean up OMG.Crypto tests

achiurizo:
## Overview

Adds unit tests for `OMG.Crypto` and additional coverage for functions like `OMG.Crypto.hash/1`. Also removes calls to functions that are only used sparingly in other modules.

## Changes

* Adds doctests for[ `OMG.Crypto.hash/1`](https://github.com/omisego/elixir-omg/pull/1028/commits/7473ab581487ad8283f742bdc17b2c0ccdc68006)
* consolidates `recover_address/2` and moves unpacking into `OMG.Signature` as another overloaded function and removes `unpack_signature` 487f60ea16445ff599f3b66b31876716a13f8bb0
* conform to style guide (extraneous piping) 9cd3f9cc2ac40af9ba4d285f04a77ed09225c67a
* move tests into unit test format and clarify description of test.

## Testing

```sh
mix test test/omg/crypto_test.exs
# And follow up with
mix test
```


apps/omg/lib/omg/signature.ex
```diff
@@ -42,6 +42,9 @@ defmodule OMG.Signature do
   This implements Eq.(208) of the Yellow Paper, adapted from https://stackoverflow.com/a/20000007
 
   """
+  def recover_public(hash, <<r::integer-size(256), s::integer-size(256), v::integer-size(8)>>, chain_id \\ nil),
+    do: recover_public(hash, v, r, s, chain_id)
```
T-Dnzt: I would be in favor of going multi-lines when it doesn't fit on one line, and enforce it as a styling guideline.

```suggestion
def recover_public(hash, <<r::integer-size(256), s::integer-size(256), v::integer-size(8)>>, chain_id \\ nil) do
  recover_public(hash, v, r, s, chain_id)
end
```

T-Dnzt: Hmm, don't think you can use my suggestion above to create the commit, lines are wrong (with multi-lines selection).

apps/omg/test/omg/crypto_test.exs
```diff
@@ -24,58 +25,72 @@ defmodule OMG.CryptoTest do
...
+    test "recovers address from an encoded transaction" do
+      {:ok, priv} = DevCrypto.generate_private_key()
+      {:ok, pub} = DevCrypto.generate_public_key(priv)
+      {:ok, address} = Crypto.generate_address(pub)
+
+      raw_tx = Transaction.Payment.new([{1000, 1, 0}], [])
+      signature = DevCrypto.signature(raw_tx, priv)
+      assert byte_size(signature) == 65
+
+      assert true ==
```
T-Dnzt: You don't really need `true` when asserting `true, just `assert expression` is enough.

apps/omg/test/omg/crypto_test.exs
```diff
@@ -24,58 +25,72 @@ defmodule OMG.CryptoTest do
...
+               |> (&match?({:ok, ^address}, &1)).()
 
-  test "digest sign, recover" do
-    {:ok, priv} = DevCrypto.generate_private_key()
-    {:ok, pub} = DevCrypto.generate_public_key(priv)
-    {:ok, address} = Crypto.generate_address(pub)
-    msg = :crypto.strong_rand_bytes(32)
-    sig = DevCrypto.signature_digest(msg, priv)
-    assert {:ok, ^address} = Crypto.recover_address(msg, sig)
+      assert false ==
```
T-Dnzt: You can use `refute` here (and get rid of the `false`), but that's up for discussion, depending on which style we want to go with.

apps/omg/test/support/test_helper.ex
```diff
@@ -61,8 +61,8 @@ defmodule OMG.TestHelper do
   def generate_entity do
     {:ok, priv} = DevCrypto.generate_private_key()
     {:ok, pub} = DevCrypto.generate_public_key(priv)
-    {:ok, addr} = Crypto.generate_address(pub)
-    %{priv: priv, addr: addr}
+    <<_::binary-size(12), address::binary-size(20)>> = Crypto.hash(pub)
```
pdobacz: hm, a leftover change? seems like `Crypto.generate_address(pub)` would still do the job here

apps/omg/lib/omg/crypto.ex
```diff
@@ -30,31 +30,27 @@ defmodule OMG.Crypto do
...
-  def recover_public(<<digest::binary-size(32)>>, <<packed_signature::binary-size(65)>>) do
-    {v, r, s} = unpack_signature(packed_signature)
-
-    with {:ok, _pub} = result <- Signature.recover_public(digest, v, r, s) do
-      result
+    with {:ok, pub} <- Signature.recover_public(digest, packed_signature) do
+      <<_::binary-size(12), address::binary-size(20)>> = hash(pub)
+      {:ok, address}
     else
       {:error, "Recovery id invalid 0-3"} -> {:error, :signature_corrupt}
```
InoMurko: why is this returning a binary error ? `"Recovery id invalid 0-3"`?

InoMurko: ahaa, I remember now, the dependency `:libsecp256k1` is a nif that can return
```
if (!get_recid(env, argv[3], &recid)) {
        return error_result(env, "Recovery id invalid 0-3");
    }
```

apps/omg/test/support/test_helper.ex
achiurizo: @pdobacz Ah good catch! Also removed some other duplication found :pray:

apps/omg/lib/omg/signature.ex
```diff
@@ -42,6 +42,10 @@ defmodule OMG.Signature do
   This implements Eq.(208) of the Yellow Paper, adapted from https://stackoverflow.com/a/20000007
 
   """
```
InoMurko: isn't this a great example for a doctest? https://elixir-lang.org/getting-started/mix-otp/docs-tests-and-with.html#doctests

Issue #1031, opened 2019-10-03
flaky byzantine events test


InoMurko:
```
  * test time response for asking for exit data (51352.0ms)


  1) test time response for asking for exit data (OMG.Performance.ByzantineEventsTest)

     test/omg_performance/integration/byzantine_event_test.exs:28

     Assertion with == failed

     code:  assert error_exits == 0

     left:  108

     right: 0

     stacktrace:

       test/omg_performance/integration/byzantine_event_test.exs:47: (test)



```

https://app.circleci.com/jobs/github/omisego/elixir-omg/20935/parallel-runs/0/steps/0-107

pdobacz:
yep, this happened to me locally, thx for reporting.

If this bites in the CI too, I have a quick fix. I suspect the cause is that the next phase of the test starts before the dust settles on the watcher after the preparation phase :(

InoMurko:
If they're impossible to isolate within `integration` tests maybe we introduce a new tag?

pdobacz:
I have poked around with the idea posted above and found that waiting around 2 seconds solves it.

For now, to bring the builds up, I'm going to patch with this, but then leave this issue for future investigation as it's quite interesting to see why this happens, esp. in case there's something serious going on.

The sleep I'm going to put around here https://github.com/omisego/elixir-omg/blob/master/apps/omg_performance/lib/omg_performance/byzantine_events.ex#L101

in a PR https://github.com/omisego/elixir-omg/pull/1034

PR #1034, opened 2019-10-04
fix: prevent flaky byzantine events test, dirty

pdobacz:
Workaround for #1031, but not a fix to the root cause yet.

## Overview

Sleeping a while makes it pass rather reliably, but I'm leaving up to #1031 to investigate why.

Note the intention here is just to allow us to CI reliably for the time being. Proper fix is still required.

## Changes

Allows the dust to settle after the watcher having been synchronized using a `Process.sleep`

## Testing

```
mix test --only integration test/omg_performance/integration/byzantine_event_test.exs
```

many times

### DIFF
--- a/apps/omg_performance/lib/omg_performance/byzantine_events.ex
+++ b/apps/omg_performance/lib/omg_performance/byzantine_events.ex
@@ -98,6 +98,10 @@ defmodule OMG.Performance.ByzantineEvents do
 
   def watcher_synchronize(watcher_url \\ @watcher_url) do
     Eth.WaitFor.repeat_until_ok(fn -> watcher_synchronized?(watcher_url) end)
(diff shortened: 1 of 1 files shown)


apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -98,6 +98,10 @@ defmodule OMG.Performance.ByzantineEvents do
 
   def watcher_synchronize(watcher_url \\ @watcher_url) do
     Eth.WaitFor.repeat_until_ok(fn -> watcher_synchronized?(watcher_url) end)
+    # NOTE: allowing some more time for the dust to settle on the synced watcher
+    # otherwise some of the freshest utxo's to exit will appear as missing on the Watcher
```
pgebal: utxo's -> utxos

Issue #844, opened 2019-07-15
Change in how tx data is modeled


Pongch:
NET-71

As a consumer of abstract layer contracts (operator and watcher), I would like to remodel my data to support new tx types, So that I could have future-proof tx upgrades in the future

**Risk**

- big refactor with many breaking changes
- potential for breaking changes in the future (mass exits incompatibility)

**Value**
- Future tx type support
- avoiding hard upgrade process

### Implementation

A rough breakdown of changes & progress is this:

 - [x] refactor child chain (`OMG.State.Core.exec` & `Transaction.Recovered.recover_from` flows), to be able to add new tx types easily #882
 - [x] experiment with settlement tx to validate the above
 - [x] refactor child chain (`OMG.State.Core.exec` & `Transaction.Recovered.recover_from` flows) to be able to add new output types easily #1036 
 - [x] experiment with ERC721 tx to validate the above (just a dirty [branch here](https://github.com/omisego/elixir-omg/tree/experimental/844_partial_erc721_support) for now; waiting for demand of a tidier integration/demo)
 - [x] experiment with MVP-based txs to validate the above ([branch here](https://github.com/omisego/elixir-omg/tree/844_experimental_tx_type_abstract_outputs), slightly outdated now, but seemed to hold water + require a bit of futher steps)
 - [x] refactor child chain to be able to modify/add input pointer types easily #1036 
 - [x] experiment with `output_id`-driven txs to validate the above ([branch here](https://github.com/omisego/elixir-omg/tree/844_draft_output_id), slightly outdated now, seemed to hold water)
 - [x] refactor watcher-related parts (double spend detection and others) to be able to add new tx types easily (probably still would require some effort on the informational api/postgres end)
 - [ ] ~experiment with different tx types to validate the above~ I'd opt to postpone this

pdobacz:
left some notes in the checkbox-list in the PR description

pdobacz:
@Pongch I'm revisiting this from the user-story PoV and wondering what would the acceptance criteria be here.

A lot has been discussed on this topic and still there's a couple of open points that could pivot the data model, e.g.:
 - MVP/MoreVP
 - some dex-stuff pending
 - approach to informational APIs on the Watcher
 - handling multiple ExitGames (which I think pertains more to the other ALD issues)

So far, what we got to is a lower the barrier of entry for new tx types with respect to the data model. To update the core functionality by introducing a new tx type, one would need to:
[...]

PR #1036, opened 2019-10-07
abstract output and fungible token output as the concrete implementation

pdobacz:
#844

## Overview

Introduces `OMG.Output` and `OMG.InputPointer` and their respective `.Protocol`s to abstract out behaviors related to various kinds of outputs and input pointers (i.e. how outputs are "pointed to" in transactions).

## Changes

- the mappings between the so-called markers of types (`<<1>>, <<2>>`) are now in the config files. Those effectively drive the scope of txs/outputs that are supported in the child chain and watcher
- `utxo_position` has been for now kept a tuple, but considering the current approach maybe a `defstruct` would be more suited, since the implementation of a `defprotocol` for a `Tuple` is awkward
- some tidies to the operations on the UTXO set have been done
- forfeited the optimization that filters out effectively unspendable UTXOs (like ones having zero amount). For simplicity sake
- changes on the Watcher-side to reflect on the outputs/input pointers being abstract (for example we can't rely on the input pointer to be a utxo position to derive the blknum where the output was created)

## Testing

`mix test` and `mix test --only integration test/omg_watcher/integration/block_getter_test.exs:126`


apps/omg/lib/omg/state/measurement_calculation.ex
```diff
@@ -35,13 +35,19 @@ defmodule OMG.State.MeasurementCalculation do
 
   defp unique_users(utxos) do
     utxos
-    |> Enum.map(fn {_, %OMG.Utxo{owner: owner}} -> owner end)
+    |> Enum.map(fn {_, %OMG.Utxo{output: output}} -> output end)
+    |> Enum.map(&Map.get(&1, :owner))
```
pgebal: `:owner` is not a part of Output protocol. As of now that's not a problem as we have only `FungibleMoreVPToken` struct. Just saying.

apps/omg/lib/omg/state/transaction.ex
```diff
@@ -120,3 +116,45 @@ defmodule OMG.State.Transaction do
...
+  def get_outputs(tx)
+
+  @doc """
+  Custom validation of the transaction with respect to its witnesses. Part of stateless validation routine
+  """
+  @spec valid?(t(), Transaction.Signed.t()) :: true | {:error, atom}
+  def valid?(tx, signed_tx)
+
+  @doc """
+  Custom stateful validity, based on pre-fetched subset of input UTXOs
```
pgebal: The meaning is 'checks if transaction can be applied to given input utxos`? If I'm right I would put that in the doc.

apps/omg/lib/omg/state/transaction/payment.ex
```diff
@@ -83,14 +74,8 @@ defmodule OMG.State.Transaction.Payment do
...
       when Transaction.is_metadata(metadata) and length(inputs) <= @max_inputs and length(outputs) <= @max_outputs do
-    inputs =
-      inputs
-      |> Enum.map(fn {blknum, txindex, oindex} -> %{blknum: blknum, txindex: txindex, oindex: oindex} end)
-
-    outputs =
-      outputs
-      |> Enum.map(fn {owner, currency, amount} -> %{owner: owner, currency: currency, amount: amount} end)
-
+    inputs = Enum.map(inputs, &new_input/1)
```
pgebal: I think `@doc` for this function is no longer valid. It does not add zero inputs / outputs anymore.

apps/omg/lib/omg/state/utxo_set.ex
```diff
@@ -62,12 +69,29 @@ defmodule OMG.State.UtxoSet do
...
 
   @doc """
-  Searches the UTXO set for a particular UTXO created with a `tx_hash` on `oindex` position.
+  Searches the UTXO set for a particular UTXO created with a `txhash` on `oindex` position.
 
   Current implementation is **expensive**
   """
-  def scan_for_matching_utxo(utxos, tx_hash, oindex) do
-    Enum.find(utxos, &match?({Utxo.position(_, _, ^oindex), %Utxo{creating_txhash: ^tx_hash}}, &1))
+  def find_matching_utxo(utxos, requested_txhash, oindex) do
```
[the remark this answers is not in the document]
pdobacz: added specs, more docs and the low-level unit tests required since the new approach.

apps/omg/test/omg/state/utxo_set_test.exs
```diff
@@ -0,0 +1,166 @@
...
+    test "will apply effects of new utxos being created", %{inputs: [input | _], outputs: [output | _]} do
+      utxo_map = %{input => %Utxo{output: output, creating_txhash: <<1>>}}
+
+      assert {:ok, [^output]} =
+               [] |> UtxoSet.init() |> UtxoSet.apply_effects([], utxo_map) |> UtxoSet.get_by_inputs([input])
+    end
+
+    test "will create first, spend second", %{inputs: [input | _], outputs: [output | _]} do
+      # this would not happen now, since `apply_effects/3` is called per tx, which cannot spend it's own input
+      # neverthe less, let's make sure this is catered for on this level too
```
pgebal: unnecessary white space neverthe less

PR #1039, opened 2019-10-10
refactor: clean TODOs - remove stale ones and fix easy ones

pdobacz:
## Overview

As title:
 - no need for padding with empty signatures
 - explain the test of one cornercase test in FeesTest
 - remove some completetly stale ones
 - prevent `performance_test.exs` from timeouting - they were just under the timeout threshold. It went over the threshold when going ALD

## Changes

...

## Testing

usual testing, `mix test` mainly

### DIFF
--- a/apps/omg_performance/test/omg_performance/integration/performance_test.exs
+++ b/apps/omg_performance/test/omg_performance/integration/performance_test.exs
@@ -35,7 +35,7 @@ defmodule OMG.PerformanceTest do
   end
 
   @tag fixtures: [:destdir]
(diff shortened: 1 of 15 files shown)


apps/omg_performance/test/omg_performance/integration/performance_test.exs
```diff
@@ -45,6 +45,7 @@ defmodule OMG.PerformanceTest do
   end
 
   @tag fixtures: [:destdir, :contract, :omg_child_chain, :alice, :bob]
+  @tag timeout: 120_000
   test "Smoke test - run start_extended_perf and see if it don't crash", %{
```
pik694: `see if it doesn't crash`

PR #1043, opened 2019-10-14
WIP: Refactor block queue

T-Dnzt:

## Overview

## Changes

- Change 1
- Change 2
- ...

## Testing


apps/omg_child_chain/lib/omg_child_chain/block_queue/block_queue_submitter.ex
```diff
@@ -0,0 +1,227 @@
...
+      ...>   %BlockSubmission{},
+      ...>   1
+      ...> )
+      :ok
+
+  """
+  @spec process_submit_result(submit_result_t(), BlockSubmission.t(), BlockSubmission.plasma_block_num()) ::
+          :ok | {:error, atom}
+  def process_submit_result(submit_result, submission, newest_mined_blknum) do
+    case submit_result do
```
InoMurko: this to me is still a weird implementation... we return `:ok` and hide the implementation detail, that it doesn't really matter if the block gets submitted here or not. in all but `{:ok, _}` tuple, the block will get resubmitted with adjusted gas price...

pdobacz: Not sure, what exactly do you find confusing here (this just interprets the block submission results from the eth node in the specific context we have), however, IIRC, the block is resubmitted with adjusted gas price not when submission fails but when it is doesn't get mined in the expected time-window. So it definitely can also happen on `{:ok, _}` in here. Gas price and re-submission seems to be out-of-scope of this sub-module.

EDIT: I just noticed the `@doc` looks incorrect (cc/ @T-Dnzt) and maybe this is what misled you?

T-Dnzt:
Whoops, forgot to give an update here. This was put on hold to focus on more stuff. Might get back to it later.

unnawut:
Master probably drifted so much away from this. Might want to close?

PR #1032, opened 2019-10-03
integrate with ALD IFE

pnowosie:
EDIT: pdobacz: for full PR description see below. The opening comment is Paweł's original draft PR's comment :point_down: 

Beginning of the journey to integrate ALD's inflight exits

Changes between old IFE events and ALD's ones -> check this [gist](https://gist.github.com/pnowosie/18c8cdf10fffa123f4aaeaa0ce203689)

I managed:
- [root_chain.ex](https://github.com/omisego/elixir-omg/pull/1032/commits/698b1d423397de876eb66a829b160b433b09c0f2) - which fetches lists of new events for both `inputs/outputs` and concatenate them
- started to change [inflight_exit_info](https://github.com/omisego/elixir-omg/pull/1032/commits/afd038e2e8150b8f54e0dfce68d89287692ead6a) where `output_index` is expressed as a 
tuple: `{:input | :output, oindex}`

PR #1050, opened 2019-10-17
fix: newest plasma contracts

InoMurko:
This updates plasma-deployer's contracts to the one in elixir-omg `7b8a2643568556c1d126749724666bc37edc8141`


unnawut:
Requires the image to pop up on https://hub.docker.com/r/omisego/plasma-deployer/tags

pdobacz:
yeah, as discussed - won't work just like that due to some breaking changes in the contrats. #1032 will bump the contracts sha and catch up, unless we need that faster then cherry-picking would be required. But #1032 is in review, so should land sometime soon if we pester the reviewer :)

docker-compose.yml
```diff
@@ -16,7 +16,7 @@ services:
       retries: 5
 
   plasma-deployer:
-    image: omisego/plasma-deployer:c7266b7
+    image: omisego/plasma-deployer:ff92de7
```
unnawut: No need to update since https://github.com/omisego/elixir-omg/pull/1054 already updated it to an even newer version

unnawut:
This PR is still useful to bump up plasma-contracts in `mix.lock`. Just need to remove the docker-compose change

InoMurko:
> This PR is still useful to bump up plasma-contracts in `mix.lock`. Just need to remove the docker-compose change

Currently the plasma-contracts above sha `7b8a2643568556c1d126749724666bc37edc8141` is not compatible with elixir-omg

PR #1055, opened 2019-10-18
feat: barebuild macos testing

sirn:
## Overview

Run macOS barebuild nightly

## Changes

- Setup barebuild for macOS in CircleCI
- Fix Ethereum installation in setup.sh 
- Some formatting and removing unused environment variables

## Testing

Nightly CI shouldn’t fail


.circleci/config.yml
```diff
@@ -83,6 +85,38 @@ jobs:
...
             mix test --trace --only integration
+  barebuild_macos:
+    executor: metal_macos
+    environment:
+      MIX_ENV: test
+    steps:
+      - checkout
+      - run: echo 'export PATH=~/.cargo/bin:$PATH' >> $BASH_ENV
+      - run: |
+          brew install postgres
```
InoMurko: wowow this is nice, could we do the same for the barebuild we do? maybe include this as part of `bin/setup` or is this because of `circle`?

sirn: I saw these stuff (well, part of it) commented out in `bin/setup`, do you want me to resurrect it? (or resurrect it and guard it with env `NO_SETUP_BREW` or something)

InoMurko: <3 yeah, I couldn't get it to work

InoMurko: and then I just used postgres and elixir containers :)

sirn: OK cool, I’ll update setup then.

.circleci/config.yml
```diff
@@ -83,6 +85,38 @@ jobs:
...
+          asdf plugin-add erlang https://github.com/asdf-vm/asdf-erlang.git
+          asdf install erlang 21.0
+          asdf install elixir 1.8.2
+          asdf global elixir 1.8.2
+          asdf global erlang 21.0
+      - run:
+          command: ./bin/setup
+          no_output_timeout: 2400
+      - run: make deps-elixir-omg
+      - run: ERLANG_ROCKSDB_BUILDOPTS='-j 2' make build-test
```
InoMurko: `ERLANG_ROCKSDB_BUILDOPTS='-j 2'` was used to limit the amount of cores the compiling of rocksdb should use. the compiler used to detect the underlying host cores (which was 32) and that blew up. not sure how many cores there is in baremetal...

jrhite:
Can we add a step that would start both CC and watcher, make a request to `alarm.get` for each and then stop the servers?

Things are building for me, but I still see core dumps on make when I try to start child_chain.

InoMurko:
That would be great for docker-compose up as well @jrhite

achiurizo:
> Can we add a step that would start both CC and watcher, make a request to alarm.get for each and then stop the servers?
> Things are building for me, but I still see core dumps on make when I try to start child_chain.

@jrhite can you open a separate PR for this? 

If this PR is ready and approved, let's get it merged in.

jrhite:
@achiurizo My comment is non-blocking, this can be merged. I'll create an issue.

sirn:
will split out extracting into bin/setup in other PR

PR #1058, opened 2019-10-21
refactor: clean up omg_eth usage of fixtures and use websocket mocks

achiurizo:
## Overview

Cleans up `omg_eth` to not use the fixtures no longer. the Subscription worker now instead uses mocks to test its behavior. 

**NOTE**: Child chain integration however still uses the `omg_eth` fixtures so we need to address that in another set of PRs to move forward with the complete removal of `omg_eth` fixtures.

## Changes

- change `subscription_worker_test` to use a `WebsocketexMock` instead 8ecc9aa
- Fix `docker-compose-fixture` port to map back to `8545` since ganache delivers ws via that port as well 
205605e

## Testing

```sh
mix test test/omg_eth/subscription_worker_test.exs --only common 
```


jrhite:
@achiurizo

Not sure if this is useful to you but checkout out https://github.com/omisego/elixir-omg/commit/fdd130b0dff5136887a1e1bbe686edb8e9ce754c

Mox to help with mocking.

achiurizo:
@jrhite thanks, this looks pretty cool will check it out.

PR #1059, opened 2019-10-21
refactor: Make test supporter less surprising

achiurizo:
## Overview

This renames our test modules to follow more expected conventions with the file structures they belong in.

After some discussion, going to pull these into a top level `Support` module as we continue to clean up these coupled test helpers.

## Changes

- Rename modules to match file directory paths
- Follower module naming convention, re: not plural
- use aliases to truncate long running module names.
- remove use of `import` for our own modules (still remaining items in code base)
- rename modules to not collide with `test_helper.exs` and give it a better name.

## Testing

```sh
mix test
```


apps/omg/test/omg/dependency_conformance/signature_test.exs
```diff
@@ -20,6 +20,7 @@ defmodule OMG.DependencyConformance.SignatureTest do
 
   alias OMG.DevCrypto
   alias OMG.Eth
+  alias OMG.Eth.Test.Support.DevNode
```
pdobacz: I did a quick scan of prior art like phoenix/ecto etc, that use the `test/support` dir structure like us.

I think including `Test` is quite nonstandard, `Support` gets included sometimes. If you want a qualifier here, I'd opt for just `Support`. If not that much then just let's drop both and try to have the actual module name actually used around (`Eth.DevNode`) meaningful?

achiurizo: @pdobacz resolved this by going with moving this to a top level `Support` module.

apps/omg_eth/lib/omg_eth/wait_for.ex
```diff
@@ -39,7 +39,7 @@ defmodule OMG.Eth.WaitFor do
   NOTE: `eth_receipt` takes txhash as raw decoded binary, like the rest of Eth APIs, but binaries in the receipt
   returned are in `0xhex-style`
 
-  This is low-level, consider using `|> Eth.DevHelpers.transact_sync!()` for eth-transactions' syncronicity in tests
+  This is low-level, consider using `|> Support.DevHelper.transact_sync!()` for eth-transactions' syncronicity in tests
```
InoMurko: this module should be moved into `support` as well, it's not used anywhere else.

apps/omg_eth/test/fixtures.exs
```diff
@@ -18,20 +18,23 @@ defmodule OMG.Eth.Fixtures do
   """
   use ExUnitFixtures.FixtureModule
 
-  alias OMG.Eth
+  alias OMG.Eth.Deployer
```
InoMurko: `OMG.Eth.Deployer` is also `test/support`

apps/omg_eth/test/fixtures.exs
```diff
@@ -18,20 +18,23 @@ defmodule OMG.Eth.Fixtures do
...
   alias OMG.Eth.Encoding
+  alias OMG.Eth.RootChain
+  alias OMG.Eth.RootChainHelper
```
InoMurko: same for `OMG.Eth.RootChainHelper`

apps/omg_eth/test/support/dev_geth.ex
```diff
@@ -64,7 +64,7 @@ defmodule OMG.Eth.DevGeth do
       if Application.get_env(:omg_eth, :node_logging_in_debug) do
         %Task{} =
           fn ->
-            geth_out |> Enum.each(&OMG.Eth.DevNode.default_logger/1)
+            geth_out |> Enum.each(&Support.DevNode.default_logger/1)
```
InoMurko:
```suggestion
            Enum.each(geth_out, &Support.DevNode.default_logger/1)
```

apps/omg_eth/test/support/dev_parity.ex
```diff
@@ -70,7 +71,7 @@ defmodule OMG.Eth.DevParity do
       if Application.get_env(:omg_eth, :node_logging_in_debug) do
         %Task{} =
           fn ->
-            parity_out |> Enum.each(&OMG.Eth.DevNode.default_logger/1)
+            parity_out |> Enum.each(&Support.DevNode.default_logger/1)
```
InoMurko:
```suggestion
           Enum.each(parity_out, &Support.DevNode.default_logger/1)
```

apps/omg_watcher/test/omg_watcher/integration/invalid_exit_test.exs
```diff
@@ -19,12 +19,14 @@ defmodule OMG.Watcher.Integration.InvalidExitTest do
...
   use Plug.Test
 
-  alias OMG.Utxo
-  require Utxo
   alias OMG.Eth
+  alias OMG.Utxo
   alias OMG.Watcher
-  alias OMG.Watcher.{Event, TestHelper}
   alias OMG.Watcher.Integration.TestHelper, as: IntegrationTest
+  alias OMG.Watcher.{Event, TestHelper}
```
InoMurko: this is the only occurrence of using `{}`

InoMurko:
```suggestion
  alias OMG.Watcher.Event
  alias OMG.Watcher.TestHelper
```

also, since it's support, you might as well move it around and rename it properly

docs/manual_service_startup.md
```diff
@@ -85,7 +85,7 @@ From the root dir of `elixir-omg`:
 mix compile
 mix run --no-start -e \
  '
-   contents = OMG.Eth.DevHelpers.prepare_env!() |> OMG.Eth.DevHelpers.create_conf_file()
+   contents = Support.DevHelper.prepare_env!() |> Support.DevHelper.create_conf_file()
```
InoMurko: this document could be removed completely, it's not relevant anymore.

apps/omg_watcher/test/fixtures.exs
```diff
@@ -29,7 +29,8 @@ defmodule OMG.Watcher.Fixtures do
   alias OMG.Crypto
   alias OMG.Watcher
   alias OMG.Watcher.DB
-  alias Watcher.TestHelper
+  alias OMG.Watcher.TestHelper
```
InoMurko: leftover?

achiurizo: unsure, but I'm not really _keen_ on a module named `TestHelper` given mix/elixir also generates a `test_helper.exs` that setups the tests. 🤔

apps/omg/test/omg/dependency_conformance/signature_test.exs
```diff
@@ -34,13 +36,12 @@ defmodule OMG.DependencyConformance.SignatureTest do
   @token TestHelper.generate_entity().addr
 
   setup_all do
-    {:ok, exit_fn} = Eth.DevNode.start()
+    {:ok, exit_fn} = DevNode.start()
```
pdobacz: hm, I looked at it again now, at the final result, and I still have some doubts. I think getting rid of `Eth` (or other component like `Watcher` obscures things a bit. `Eth.DevNode` was more explicit than `DevNode`; `DevNode` could be anything, it's not apparent it is an Ethereum node we're talking about. 

Same goes with `Deployer`, `DevHelper`. `Eth.Support.DevNode` would be explicit, but is lengthy.

Maybe let's try `defmodule Support.Eth.DevNode` and then `alias Support.Eth` and in-code use with `Eth.DevNode`? would this work?

docs/manual_service_startup.md
pnowosie: It's not true.  Also demos are based on manual startup.

pdobacz:
> IMHO removing `manual_service_startup` is premature

oh, I haven't noticed this. Yeah, agreed, this is the guide to run things on bare metal, which I suppose supported still, even I'm still sometimes using it (i.e. the commands inside) for convenience. Why remove?

apps/omg/test/omg/dependency_conformance/signature_test.exs
achiurizo: Well a few things:

> DevNode could be anything, it's not apparent it is an Ethereum node we're talking about.

What other node would we be talking about here?

> Same goes with Deployer, DevHelper. Eth.Support.DevNode would be explicit, but is lengthy.

* the code base is inconsistent with how we surface module usage.
  * Sometimes we fully alias the entire module,
  * sometimes alias just the top level
  * sometimes a portion of it
  * And other times it just doesn't alias at all. 
  * or we import it.
I'm focusing more on consistency here and assuming that we always alias the full module name. If we follow the above rule, it will be much more consistent to discover what the module is by looking at our alias declarations.
* The eventual goal is to remove all this code. As we move towards docker, the hope is we don't need to rely on these elixir code to prop up our services.

I would be open to `Support.Eth.Node` but we should still fully alias the module like so:

```elixir
alias Support.Eth.DevNode
# ...
DevNode.start()
```

pdobacz: > What other node would we be talking about here?

Node in a cluster, node in a tree. I'm talking about the experience of a first-time reader here. `Eth.DevNode` is pretty clear.

> the code base is inconsistent with how we surface module usage
> ...I'm focusing more on consistency here and assuming that we always alias the full module name

I don't think this little bit of consistency is worth the decrease of readability. E.g. turning `Utxo.Position` into `Position` will be detrimental, similar conflict would become of `Watcher.DB.Repo` and `Ecto.Repo` becoming `Repo` etc.

I think `Plug.Conn`, `Ecto.Repo` are commonly used non-fully-aliased module names out there.

Maybe let's settle for consistent use of a particular module throughout code, e.g. `Eth.DevNode` stays `Eth.DevNode` and `Support.Eth.DevNode` and `DevNode` are both avoided. WHat do you think?

> I would be open to Support.Eth.Node but we should still fully alias the module

That would be good, but I'd advise against forcing the full alias, see above

pdobacz:
I hear you on the "misleading" part. Still, it is useful to be able to spin up the services on bare metal using `mix`, for the purpose of debugging and experimentation. I suppose releases would not be enough, if you do something in the code and then restart the services. This is very useful to have.

If we perhaps rephrase the docs to be clear about that it is intended only for `elixir-omg` development process and is not the way the services should be started normally? We can leave all the explanations out and be explicit about what it is. What do you say?

pnowosie:
> On the other hand, using mix to startup the release is discouraged and is not supported and having documentation for that is misleading.

Honestly I'm totally missing the point in above statement. This is part of development documentation in code repository. In what part development tools such as `mix` are discourage to use with source code? What is misleading in documenting service startup steps in development environment? What does `release` mean in regards to development branch?

`manual service startup` is useful for developers and documents how to start the service from the codebase. What's wrong with it?

InoMurko:
> > On the other hand, using mix to startup the release is discouraged and is not supported and having documentation for that is misleading.
> 
> Honestly I'm totally missing the point in above statement. This is part of development documentation in code repository. In what part development tools such as `mix` are discouraged to use with source code? What is misleading in documenting service startup steps in development environment? What does `release` mean in regards to development branch?
> 
> `manual service startup` is useful for developers and documents how to start the service from the codebase. What's wrong with it?

It’s like this, if you provide documentation, you need to support it. Our repo has been suffering from excessive documentation that got outdated already. If your development cycle is to manually start supervision trees, it most definitely can be, but it is not something I would provide as part of official OmiseGO documentation. It serves no purpose. The development cycle are tests and end2end tests tbd. 

Every elixir and erlang developer understands the concept of applications because thats the architecture of beam systems. So what is the purpose of this doc? To educate developers on erlang and elixir? This isn’t the right place for that. 

These applications are not microservice components of a larger system (they could be-same as we’re forming watcher and child chain) but individually they don’t make sense.

pnowosie:
> Running services baremetal with releases is supported

I must be missing something, can you link doc with describe preparing a distillery release?

InoMurko:
> > Running services baremetal with releases is supported
> 
> I must be missing something, can you link doc with describe preparing a distillery release?

https://github.com/omisego/elixir-omg/blob/master/docs/deployment_configuration.md
+
Makefile ... ‘make help’

pnowosie:
> So what is the purpose of this doc?

In my opinion this doc describes the steps needed to start a service from the codebase. Unfortunately it requires not only `mix run` or even `mix xomg...` but some preliminaries like running geth, contract deployment and so. Therefore I think it is useful.

InoMurko:
> > So what is the purpose of this doc?
> 
> In my opinion this doc describes the steps needed to start a service from the codebase. Unfortunately it requires not only `mix run` or even `mix xomg...` but some preliminaries like running geth, contract deployment and so. Therefore I think it is useful.
> 
> Also this are required steps to run the demos... once we haven't yet rewritten them to use some other method.
> 
> > but it is not something I would provide as part of official OmiseGO documentation.
> 
> I agree and this is OK, because it isn't a part of the official OmiseGO documentation. It's documenting the codebase and is required to run demos locally.

The docs for this are already broken. If you want to rewrite them, go ahead and we might consider adding them back. Note that documentation here presumes one can set environment variables and they'll be read if you run things via `mix ...`. It's not the case because we use config providers that are assembled at the release startup.
Also, please double check what you're fighting for. The docs point to old contracts that we've abandoned and moved to ALD.

`In my opinion this doc describes the steps needed to start a service from the codebase. `
The doc describes how to start the underlying services in an unconventional. 

Which demos docs? We need to delete those as well.

pnowosie:
> If you want to rewrite them, go ahead and we might consider adding them back. 
> The docs point to old contracts that we've abandoned and moved to ALD.

I'll try to make it work for ALD.

> Which demos docs? We need to delete those as well.

I'll happily remove them once we have an alternative. Specs?

InoMurko:
> > If you want to rewrite them, go ahead and we might consider adding them back.
> > The docs point to old contracts that we've abandoned and moved to ALD.
> 
> I'll try to make it work for ALD.
> 
> > Which demos docs? We need to delete those as well.
> 
> I'll happily remove them once we have an alternative. Specs?

Specs repo is end 2 end testing. If we want to have demos on live systems we should implement mix tasks or custom commands.

apps/omg/test/omg/dependency_conformance/signature_test.exs
InoMurko: Actually, in elixir's and erlangs terms, a Node is already taken. A node is a beam VM instance.

pnowosie:
> The docs for this are already broken.

@InoMurko not so much. Regarding `manual_service_startup.md` only missing part is `prepare_env` step, because DevHelpers are not compiled now on `MIX_ENV=dev`

so it's sufficient to replace with:
`MIX_ENV=test mix run --no-start -e \ ... `

also it supports ALD contracts already - I've just run ch-ch following the steps.
So I'm opting to preserve the file with applying only above fix, it's still relevant.

What do you think?

InoMurko:
I don’t think the file is relevant or that it adds anything meaningful.

apps/omg/test/omg/dependency_conformance/signature_test.exs
achiurizo: > Maybe let's settle for consistent use of a particular module throughout code, e.g. Eth.DevNode stays Eth.DevNode and Support.Eth.DevNode and DevNode are both avoided. WHat do you think?

> Actually, in elixir's and erlangs terms, a Node is already taken. A node is a beam VM instance.

I think perhaps this is the problem, the use of the word `Node` was problematic here. I'm going to fix the merge conflicts then propose we just call this something else all together. (Perhaps something more verbose like EthereumService)

Again, the end goal is to remove all of this and move starting Ethereum outside of elixir.

achiurizo:
> @InoMurko not so much. Regarding manual_service_startup.md only missing part is prepare_env step, because DevHelpers are not compiled now on MIX_ENV=dev

@pnowosie doesn't this mean the doc is still broken, regardless if it's not so much? I think that's the point being made here. the documentation isn't being well maintained. if this PR had not brought this to attention, the document would have continued being broken. 

While I can appreciate the particular use-case and desire to do it this way, it's going to cause more problems in the long run as it exposes another way to build the project which is not being well maintained.

pnowosie:
@achiurizo you may consider to rescue the doc, it's still useful for PL team.
[here fix included](https://github.com/omisego/elixir-omg/commit/11d26f15b4b0e7c91ffc1d36ca8a3de9d6b56e55)

pnowosie:
> doesn't this mean the doc is still broken,

@achiurizo no, there was small fix required, it's OK [here fix included](https://github.com/omisego/elixir-omg/commit/11d26f15b4b0e7c91ffc1d36ca8a3de9d6b56e55)

> it's going to cause more problems in the long run as it exposes another way to build the project

As @pdobacz noticed it is "for the purpose of debugging and experimentation" on development environment. This is how we test things that require full setup. I'd be happy to retire the file once we'd have an alternative approach - but IMHO discussion with @InoMurko shows there is none.

achiurizo:
> Is it common Elixir practice to alias everything? I think the alias might make sense for a long namespace like OMG.Watcher.HttpRPC.Client (if we need to use it a lot), but seems less useful for something like Support.Deployer. It isn't painful to write the full module name and makes the code more readable and easier to search.

@chrishunt I think you make a good point here. I'm okay with that usage, but I think we need to be at _least_ consistent with how we decide to call out other modules.  If we decide to alias, we should fully alias the name. If there's one off cases where its un-necessary, sure let's just use the full qualified name in the code.

InoMurko:
> > doesn't this mean the doc is still broken,
> 
> @achiurizo no, there was small fix required, it's OK [here fix included](https://github.com/omisego/elixir-omg/commits/pnowosie/rescue-for-doc)
> 
> > it's going to cause more problems in the long run as it exposes another way to build the project
> 
> As @pdobacz noticed it is "for the purpose of debugging and experimentation" on development environment. This is how we test things that require full setup. I'd be happy to retire the file once we'd have an alternative approach - but IMHO discussion with @InoMurko shows there is none.

Ok. So let's go through the docs and check line by line how wrong they are.
```
This process is intended for users who wish to start services manually, perhaps
as part of a non-Docker deployment on a Linux host.
```
Our umbrella applications are not meant as a separate service you can deploy, because of the amount of coupling. Deploying them on a Linux host without releases is exactly what I've been working the last 6 months to get us away from. Remember `launcher.py`>

```
 1. (**Child chain server only**) Prepare the authority address and deploy `RootChain.sol`, see [Contracts section](#contracts). 
```
Please point to me where does trunk of plasma-contracts contain `RootChain.sol`?

```
 1. Produce a configuration file for `omg_eth` with the contract address, authority address and hash of contract-deploying transaction. 
```
Are you sure this still stands for truffle deploys and ALD?


```
- configuring with an environment variable `CHILD_CHAIN_URL=desired_childchain_url`
```
You cannot configure mix tasks with environment variables anymore because of config providers we use in releases.


```
```bash
MIX_ENV=test mix compile
MIX_ENV=test mix run --no-start -e \
 '
   contents = Support.DevHelper.prepare_env!() |> Support.DevHelper.create_conf_file()
   "~/config.exs" |> Path.expand() |> File.write!(contents)
 '
```
```

Wrong approach to deploy using releases or truffle.

```
#### Start it up!
```

The whole section start it up is against everything elixir community has been working against with distillery and releases.

InoMurko:
> > doesn't this mean the doc is still broken,
> 
> @achiurizo no, there was small fix required, it's OK [here fix included](https://github.com/omisego/elixir-omg/commits/pnowosie/rescue-for-doc)
> 
> > it's going to cause more problems in the long run as it exposes another way to build the project
> 
> As @pdobacz noticed it is "for the purpose of debugging and experimentation" on development environment. This is how we test things that require full setup. I'd be happy to retire the file once we'd have an alternative approach - but IMHO discussion with @InoMurko shows there is none.

You now need to decide. Is this for development or deployment? The documentation is clear, it’s meant for deployment to production. Look at the first paragraph. 

Also, please check the Makefile on root of the project. I don’t know how you can say that there’s no alternative approach since you can deploy a full local service with docker, docker compose, baremetal releases prod or dev.

achiurizo:
@pnowosie @InoMurko Considering there's quite a bit of discussion surrounding this document and I not wanting to prolong this PR any further, can we consider a separate PR where we decide if we want to re-introduce the document? 🙏

Issue #849, opened 2019-07-15
Multiple Childchain state storage


Pongch:
NET-62

As a child chain operator, I want to have redundancy in how my child chain's state is stored so that I can perform upgrade without any service interruptions.

**Risk**
- requiring a lot of engineering effort to be spent in rearchitecting the state storage

**Value**
- allowing for 0 downtime upgrade, hence better service availability. (Currently upgrading takes about 2-5 minutes of downtime)

- reducing risk of failure and loss of transactions

- able to keep production services running across cloud provider's zone

InoMurko:
Work needed to Scale Child Chain
1.	Block Queue OMG.ChildChain.BlockQueue (how we form a block before submission)
2.	Fees OMG.ChildChain.FeeServer (how much fees we're charging currently loads via file, unsalable)
3.	OMG.State.Core (centralized repo of the current state)

On the Child Chain's part there's a statefull procedure ~validation~ that happens when a transaction is submitted (http `post("/transaction.submit", Controller.Transaction, :submit)`).
```
@spec submit(transaction :: binary) ::
          {:ok, %{txhash: Transaction.tx_hash(), blknum: pos_integer, txindex: non_neg_integer}}
          | {:error, submit_error()}
  def submit(transaction) do
    with {:ok, recovered_tx} <- Transaction.Recovered.recover_from(transaction),
         {:ok, fees} <- FeeServer.transaction_fees(),
         fees = Fees.for_tx(recovered_tx, fees),
         {:ok, {tx_hash, blknum, tx_index}} <- State.exec(recovered_tx, fees) do
      {:ok, %{txhash: tx_hash, blknum: blknum, txindex: tx_index}}
    end
    |> result_with_logging()
  end
```
State.exec/2 is the nitty gritty. 
```
@doc """
  Checks (stateful validity) and executes a spend transaction. Assuming stateless validity!
  """
  def handle_call({:exec, tx, fees}, _from, state) do
    case Core.exec(state, tx, fees) do
      {:ok, tx_result, new_state} ->
        {:reply, {:ok, tx_result}, new_state}

      {tx_result, new_state} ->
        {:reply, tx_result, new_state}
    end
  end
```
 goes into
```
@spec can_apply_spend(state :: Core.t(), tx :: Transaction.Recovered.t(), fees :: Fees.fee_t()) ::
          true | {{:error, exec_error()}, Core.t()}
  def can_apply_spend(state, %Transaction.Recovered{} = tx, fees) do
[...]

PR #1066, opened 2019-10-24
Experiment with UTXO subsets to get closer to zero-downtime

pdobacz:
related to #849, could be a proposition/experiment of a 1st step towards.

**NOTE** this is super hacky and dirty. Done more as a past-time activity and satisfying curiosity.

## Overview

Using some out-of-internet time on the plane I experimented with child chain horizontal scaling using a "UTXO subset" approach.

In a nutshell, the approach makes `OMG.State` (the ledger) fetch UTXOs from DB on every transaction execution, instead of relying on what is held in the `OMG.State` process. Likewise, the UTXO set in the DB is updated on every transaction execution, instead of on every block forming.

### The Good

1. Performance doesn't drop too much (from ~1400 tps to ~1100-1200 on my laptop)
2. This should get us rid of the lengthy loading of UTXOs upon the start of Watcher/Child Chain. Among others it will alleviate considerably the problems with large UTXO set (not fully though!)
3. If Postgres-move comes, this here should be a friendly change
4. If we would be able to switch tx traffic from `chch-instance-1` to `chch-instance-2` on the block border (should be easy*), I think we can get close to **zero-downtime upgrades & fail-overs**. (~70% sure here). At least assuming a RocksDB (or any other kv store) database instance can be shared between those chch instances. If the latter is not possible, then anyway, this change here could be a first step refactor towards achieving this zero-downtime when Postgres is here
5. We still can leverage all the `OMG.State.Core` logic and its tests with close to no modification

### The Bad

1. Just like other horizontal-scaling movements, this could limit state-merklization-friendliness. (I posted details in https://github.com/omisego/elixir-omg/issues/849#issuecomment-545959636)
2. Makes the current UTXO set related metrics calculation useless (because the UTXO set in memory isn't all). Looks hard/awkward to solve without Postgres. We could try punting that somewhat, if we do those metrics just on the Watcher's `INFO` side, which already has all data in Postgres

-------------

*) "should be easy" - it's enough to start `BlockQueue` on the failover `chch-instance-2` only when we're sure that the failing `chch-instance-1` is done failing. Transaction would still be processed with no impact on the perceivable uptime

## Changes

- UTXO set in DB is updated on every tx, not on form block
- mempool txs are persisted in DB

## Testing

:eyes:
`mix test`
`mix test --only integration` works somewhat too, as long as invalid txs aren't involved.

### DIFF
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -21,6 +21,7 @@ defmodule OMG.State do
   alias OMG.DB
   alias OMG.Eth
   alias OMG.Fees
+  alias OMG.InputPointer
   alias OMG.State.Core
   alias OMG.State.Transaction
   alias OMG.State.Transaction.Validator
@@ -90,19 +91,17 @@ defmodule OMG.State do
     # Get data essential for the State and Blockgetter. And it takes a while. TODO - measure it!
     # Our approach is simply blocking the supervision boot tree
     # until we've processed history.
-    {:ok, utxos_query_result} = DB.utxos()
     {:ok, height_query_result} = DB.get_single_value(:child_top_block_number)
     {:ok, last_deposit_query_result} = DB.get_single_value(:last_deposit_child_blknum)
-    {:ok, [utxos_query_result, height_query_result, last_deposit_query_result], {:continue, :setup}}
+    {:ok, [height_query_result, last_deposit_query_result], {:continue, :setup}}
   end
 
-  def handle_continue(:setup, [utxos_query_result, height_query_result, last_deposit_query_result]) do
+  def handle_continue(:setup, [height_query_result, last_deposit_query_result]) do
     {:ok, child_block_interval} = Eth.RootChain.get_child_block_interval()
 
     {:ok, state} =
       with {:ok, _data} = result <-
              Core.extract_initial_state(
-               utxos_query_result,
                height_query_result,
                last_deposit_query_result,
                child_block_interval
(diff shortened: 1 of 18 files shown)


InoMurko:
How do you ensure consistency on concurrent execution?
I wonder, because thats actually our initial idea as well, but to ensure consistency we’ve chosen to use locks between transaction processing UTXOs.

jbunce:
Just a heads up that `At least assuming a RocksDB (or any other kv store) database instance can be shared between those chch instance` is an incorrect assumption. We're planning for a relational database of sorts, likely Postgres.

jbunce:
Can you multi-read with RocksDB?

InoMurko:
Yes, but not in our case. Read ONLY mode https://github.com/facebook/rocksdb/wiki/RocksDB-Basics#readonly-mode

pdobacz:
Thank you for the feedback :pray:!

@InoMurko 

> How do you ensure consistency on concurrent execution?
I wonder, because thats actually our initial idea as well, but to ensure consistency we’ve chosen to use locks between transaction processing UTXOs.

Hm, you're right, this strongly relies on having a `master node`/`backup nodes` approach, so one chch node must act as `master` (and have a long-lasting lock on the write access to the DB) and the rest kick in after only `master` is deprived of the lock.

Actually, on 2nd thought, this is still missing a couple of pieces for zero-downtime. At current state, the gain is getting rid of the eager UTXO set loading on start. :thinking: 

@jbunce 

> Just a heads up that At least assuming a RocksDB (or any other kv store) database instance can be shared between those chch instance is an incorrect assumption. We're planning for a relational database of sorts, likely Postgres.

yep, hence "any other kv store" there. Postgres can act as a reasonable KV I suppose, or other more feature-rich KVs can be put in RocksDB place, in case we want to separate the "make DB relational" from "make DB shared" issues.

Although I figured that one of the RocksDB advantages over LevelDB was a better support from infra-PoV. Maybe I over-interpreted that part.

Anyhow, the option is on the table.

InoMurko:
Reopen if needed!

PR #1071, opened 2019-10-27
feature: bump images and contracts

InoMurko:
## Overview

Bumping contracts and dependent images to stable or same versions (plasma-deployer) .
This makes contracts and elixir-omg running the same code (sha https://github.com/omisego/plasma-contracts/commit/ea36f5ff46ab72ec5c281fa0a3dffe3bcc83178b)

## Changes

Gather image tags from repos.

## Testing
/


.circleci/config.yml
```diff
@@ -19,12 +19,12 @@ executors:
 
   builder:
     docker:
-      - image: omisegoimages/elixir-omg-builder:dev-9a75a07
+      - image: omisegoimages/elixir-omg-builder:stable-20191024
```
jbunce: Can you use a git SHA in here please? This is more deterministic so we know which code we're using to build. How about:

`stable-9a75a07-20191024`

PR #1075, opened 2019-10-28
Fix bundles

InoMurko:
swaps swagger-combine with swagger-cli

## Overview
swaps swagger-combine with swagger-cli. I forgot to incl

## Changes

- Regenerate api descriptions to avoid `refs`, because generators are not able to digest them
- Rename `Watcher's` to `Watchers` because generated code doesn't digest `'`

## Testing

/

### DIFF
--- a/apps/omg_child_chain_rpc/priv/swagger/operator_api_specs.yaml
+++ b/apps/omg_child_chain_rpc/priv/swagger/operator_api_specs.yaml
@@ -2,12 +2,9 @@ openapi: 3.0.0
 info:
   version: 1.0.0
   title: Child chain API
-  description: >
+  description: |
     This is the documentation for child chain API.
-
-    Error codes are available in
-    [html](https://github.com/omisego/elixir-omg/blob/master/docs/api_specs/errors.md#error-codes-description)
-    format.
+    Error codes are available in [html](https://github.com/omisego/elixir-omg/blob/master/docs/api_specs/errors.md#error-codes-description) format.
   contact:
     name: OmiseGO
     email: omisego@omisego.co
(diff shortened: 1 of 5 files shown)


apps/omg_watcher_rpc/priv/swagger/informational_api_specs/swagger.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   version: '1.0.0'
-  title: Watcher's Informational API
+  title: Watchers Informational API
```
achiurizo:
```suggestion
  title: Watcher Informational API
```

Just make it simple and drop the 's'

apps/omg_child_chain_rpc/priv/swagger/operator_api_specs.yaml
```diff
@@ -90,16 +83,31 @@ paths:
...
+        '500':
           description: Returns an internal server error
           content:
             application/json:
               schema:
                 description: The response schema for an error
                 allOf:
-                  - *ref_0
+                  - description: The response schema for a successful operation
+                    type: object
```
chrishunt: It looks like our objects in the generated client code have questionable names like `inline_obect1`, `inline_object2`, `inline_object3`, ...etc. I wonder if there's something we can put in the spec to name them?

InoMurko: some have meaningful names, some don't ... need to diff and check!

chrishunt:
Thank you @InoMurko! Merging this so we can use in https://github.com/omisego/specs/issues/7

PR #1077, opened 2019-10-29
Self-contained plasma-deployer with ganache-cli and API server

kendricktan:
https://github.com/omisego/plasma-contracts/issues/400

## Overview

Adds an API-endpoint `plasma-contracts:8000/contracts` to retrieve the deployed contract addresses

## Changes

Edited `docker-compose-fixtures.yml` for deployment example

## Testing

```bash
docker-compose -f docker-compose-fixtures.yml up
```

### DIFF
--- a/docker-compose-fixtures.yml
+++ b/docker-compose-fixtures.yml
@@ -1,22 +1,31 @@
 version: "3.4"
 services:
+
   ganache:
     image: trufflesuite/ganache-cli:latest
     command: ganache-cli -d -e 100000 -m "myth like bonus scare over problem client lizard pioneer submit female collect"
     ports:
       - "8545:8545"
       - "8546:8545"
+
   plasma-contracts:
     build:
       context: .
       dockerfile: plasma-contract.Dockerfile
-    command: /bin/sh -c "cd /home/node/plasma-contracts/plasma_framework && npx truffle migrate --network local"
+    command:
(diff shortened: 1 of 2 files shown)


docker-compose-fixtures.yml
```diff
@@ -1,22 +1,31 @@
...
-    command: /bin/sh -c "cd /home/node/plasma-contracts/plasma_framework && npx truffle migrate --network local"
+    command:
+      - /bin/sh
+      - -c
+      - |
+          cd /home/node/plasma-contracts/plasma_framework
+          npx truffle migrate --network local
+          cd build
+          echo '{"contracts":' > db.json && cat outputs.json >> db.json && echo '}' >> db.json
+          npx json-server -w db.json -p 8000 -H 0.0.0.0
```
achiurizo: @kendricktan Why do we need this? This `docker-compose` is for the test suite and and recording cassettes. 

If there's a bigger use case, happy to merge this back into the primarily `docker-compose`.

kendricktan: @achiurizo its to act as the `CONTRACT_EXCHANGE_URL` as currently from my understanding there is currently no way to automatically get the deployed addresses https://github.com/omisego/elixir-omg/blob/9ede739e279a98ad95523e693e5972460dbe573e/docs/deployment_configuration.md

PR #1080, opened 2019-10-29
Add titles to request body schemas for swagger

achiurizo:
:clipboard: https://github.com/omisego/specs/pull/8

## Overview

This adds names to the swagger schema objects so we don't get generic titles like `InlineObject1`.

## Changes

Adds a `title:` key to all the `request_bodies.yaml`

## Testing

Generate the swagger:

```sh
make operator_api_specs
make informational_api_specs
make security_critical_api_specs
```

### DIFF
--- a/apps/omg_child_chain_rpc/priv/swagger/operator_api_specs/block/request_bodies.yaml
+++ b/apps/omg_child_chain_rpc/priv/swagger/operator_api_specs/block/request_bodies.yaml
@@ -4,7 +4,7 @@ GetBlockBodySchema:
   content:
     application/json:
       schema:
-        title: 'block_parameter'
+        title: 'GetBlockBodySchema'
         type: object
         properties:
           hash:
--- a/apps/omg_child_chain_rpc/priv/swagger/operator_api_specs/block/request_bodies.yaml
+++ b/apps/omg_child_chain_rpc/priv/swagger/operator_api_specs/block/request_bodies.yaml
@@ -4,6 +4,7 @@ GetBlockBodySchema:
   content:
     application/json:
       schema:
+        title: 'block_parameter'
         type: object
         properties:
           hash:
--- a/apps/omg_child_chain_rpc/priv/swagger/operator_api_specs/transaction/request_bodies.yaml
+++ b/apps/omg_child_chain_rpc/priv/swagger/operator_api_specs/transaction/request_bodies.yaml
@@ -4,6 +4,7 @@ TransactionSubmitBodySchema:
   content:
     application/json:
       schema:
+        title: 'transaction_parameter'
         type: object
         properties:
           transaction:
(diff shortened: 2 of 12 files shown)


apps/omg_child_chain_rpc/priv/swagger/operator_api_specs/block/request_bodies.yaml
```diff
@@ -4,6 +4,7 @@ GetBlockBodySchema:
   content:
     application/json:
       schema:
+        title: 'block_parameter'
```
pnowosie: It's not `parameter` but full body (here contains just one field).
Also: why not make it more predictable e.g. `get_block_body_schema` to follow the element key (`GetBlockBodySchema`) or even better keep it the same `title: 'GetBlockBodySchema'`

apps/omg_child_chain_rpc/priv/swagger/operator_api_specs/block/request_bodies.yaml
```diff
@@ -4,6 +4,7 @@ GetBlockBodySchema:
   content:
     application/json:
       schema:
+        title: 'GetBlockBodySchema'
```
pnowosie: Ehh, sorry about this - just noticed. This is only place we use apostrophes `'` - can you drop them, it's inconsistent to the rest of the file :cry:

achiurizo: @pnowosie but we use single quotes in the rest of the yaml definition? https://github.com/omisego/elixir-omg/pull/1080/files/5ff5f142338455b23b3bbaefe21bea72c97c64dc#diff-1e0e669f0f7d082e7fa14f9ad3dc427eL15

pnowosie: Well it looks like we use apostrophes in examples, but the rest goes without. Well if it does not make you a difference I'm for removing it there - it'll be simpler.

achiurizo:
@pnowosie what are you referring to exactly? Is it this? https://github.com/omisego/elixir-omg/pull/1080/files/5ff5f142338455b23b3bbaefe21bea72c97c64dc#diff-fc9e7275eab9fc4c76ea1a3ef6880a5dR35

These are auto-generated for what it's worth.

apps/omg_child_chain_rpc/priv/swagger/operator_api_specs/transaction/request_bodies.yaml
```diff
@@ -4,6 +4,7 @@ TransactionSubmitBodySchema:
   content:
     application/json:
       schema:
+        title: 'TransactionSubmitBodySchema'
         type: object
```
pnowosie: I mean this `type: object` is without apostrophes but it's a string as well - and also `description`

pnowosie:
> @pnowosie what are you referring to exactly? Is it this? ...
I mean: https://github.com/omisego/elixir-omg/pull/1080/files#r340127032

PR #1087, opened 2019-10-31
test docker-compose and release generation

InoMurko:
## Overview

This PR builds Watcher and Child Chain and invokes `docker-compose up -d` and proceeds with `make get-alarms` proving that all services booted correctly, which means, you're able to run `docker-compose up` locally and develop towards it.

## Changes

- Add circle CI step that supports docker-compose 2.1

## Testing

The test is the test itself.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -343,6 +343,17 @@ jobs:
 
       - run: mix dialyzer --format short --halt-exit-status
 
+  test_docker_compose:
+    machine:
+      image: circleci/classic:201808-01
+    steps:
+      - checkout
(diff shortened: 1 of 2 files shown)


InoMurko:
heeey so yeah, docker-compose can get into a mess of cache and stale images so if you encounter thigs like 
```
** (MatchError) no match of right hand side value: {:error, :root_chain_contract_not_available}
```

it probably means your ethereum network (geth) is not in sync with what contracts `plasma-deployer` is reporting. To start off from a clean slate `docker system prune -a` and try again!

PR #1093, opened 2019-10-31
 test baremetal release  #1087 

InoMurko:
## Overview

This PR builds Watcher and Child Chain on baremetal and invokes `make start-services` (geth, plasma-deployer, postgres) and proceeds with `make get-alarms` proving that all services booted correctly, which means, you're able to run elixir-omg baremetal locally and develop towards it.

## Changes

- run bin/setup to install dependencies
- Add circle CI step with machine executor

## Testing

The test is the test itself.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -352,8 +352,7 @@ jobs:
       - run: make docker-child_chain CHILD_CHAIN_IMAGE_NAME="omisego/child_chain"
       - run: make docker-watcher WATCHER_IMAGE_NAME="omisego/watcher"
       - run: docker-compose up -d
-      - run: sleep 60;
-      - run: make get-alarms
+      - run: sh .circleci/status.sh
 
   test_barebone_release:
     machine:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -158,6 +158,7 @@ jobs:
             - rel/
             - VERSION
             - .git
+            - Makefile
 
   lint:
     executor: builder_pg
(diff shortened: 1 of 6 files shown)


InoMurko:
After merge (if accepted) Ill make `test_barebone_release` and `test_docker_compose_release` run only on master branch.

apps/omg_watcher/test/fixtures.exs
```diff
@@ -274,4 +269,24 @@ defmodule OMG.Watcher.Fixtures do
     |> Enum.with_index()
     |> Enum.map(fn {recovered_tx, txindex} -> {blknum, txindex, recovered_tx.tx_hash, recovered_tx} end)
   end
+
+  defp ensure_web_started(module, function, args, 0), do: apply(module, function, args)
+
+  defp ensure_web_started(module, function, args, counter) do
+    {:ok, pid} = apply(module, function, args)
+    {:ok, pid}
```
achiurizo: just for my understanding, but we want to pattern match to ensure this does properly starts. But why pass the same result tuple down the chain again? Why not just the `pid` at this point? 

We also [pattern match](https://github.com/omisego/elixir-omg/pull/1093/files#diff-43636fce1bc9dcf250c274eddb21c7b9R150) for this result elsewhere so wondering if we should just simply pass it through and:

```elixir
defp ensure_web_started(module, function, args, counter), do: apply(module, function, args)
```

InoMurko: it's because I don't want to act like I do anything with the output of the `apply/3`. I want to return the raw result.

apps/omg_watcher/test/fixtures.exs
```diff
@@ -145,14 +145,7 @@ defmodule OMG.Watcher.Fixtures do
...
 
-    {:ok, pid} =
-      Supervisor.start_link(
-        [
-          %{id: OMG.WatcherRPC.Web.Endpoint, start: {OMG.WatcherRPC.Web.Endpoint, :start_link, []}, type: :supervisor}
-        ],
-        strategy: :one_for_one,
-        name: Watcher.Endpoint
-      )
+    {:ok, pid} = ensure_web_started(OMG.WatcherRPC.Web.Endpoint, :start_link, [], 100)
```
InoMurko: I just noticed that this sometimes brakes a build... so here's a fix for that.

.circleci/config.yml
```diff
@@ -350,9 +351,59 @@ jobs:
...
+          source ~/.bash_profile
+          asdf global erlang 21.0
+      - run:
+          command: |
+            echo 'export PATH=~/.cargo/bin:$PATH' >> $BASH_ENV
+            source ~/.bash_profile
+            asdf global elixir 1.8.2
+            mix local.rebar --force &&
+            mix local.hex --force &&
+            sudo killall dpkg || true &&
```
sirn: Hmm, we’re doing this to fix the lock thingy right?

Makefile
```diff
@@ -1,4 +1,5 @@
 MAKEFLAGS += --silent
+OVERRIDING_START=foreground
```
sirn: Probably need to be `OVERRIDING_START ?=` (with `?`) so it’s overridable across many makes (gmake/BSDmake)

.circleci/config.yml
InoMurko: yeee!

Makefile
InoMurko:
```suggestion
OVERRIDING_START ?= foreground
```
this @sirn ?

sirn: Yeap

Issue #1065, opened 2019-10-24
Child-chain, Watcher & Info API split


pnowosie:
As discussed on-site we want to split ch-ch, watcher and Watcher Informational API into separate repositories to simplify development of new features.

## Currently
Watcher shares a lot of code from ch-ch. Info API is build in Watcher (as a Watcher mode).
Service dependency: ch-ch & watcher depends on ethereum node. Watcher depends on ch-ch.
[pict](...)

### 1. Info API depends on Watcher
Pros: 
- Info API knows as much as the Watcher so transaction constructed on Info API data won't be rejected by Watcher (sync diff)
- Info API doesn't have to depend on ethereum node and contain event-listeners code. 

### 2. Info API depends on Ch-ch & Ethereum node
Pros:
- independence, can process ethereum events according to its rules
- data available faster

### 2+. Additional `tx.verify` watcher endpoint
As in pt 2 :point_up: but watcher exposes `tx.verify` endpoint that returns whether given tx is valid (inputs wasn't spent yet)
Proposed by @sirn. Info API can optionally consume this endpoint when used to transaction construction.

### Issues depending (anyhow) on this issue 
#1001, #1033, (... add more)

sirn:
Going to add that idea of 2+ is if someone want to run Info API without running watcher but want tx validity, they can just use `tx.verify` of public watcher address, e.g. `watcher.samrong.omg.network`

T-Dnzt:
Based on the current state of things, I'd like to recommend a new temporary option to get us off the ground ASAP and have everything we need for the upcoming milestones.

The idea is to do an early split of the informational API so that progress can be made on the frontend apps. It's not really a split though, mostly just copying the code and putting it into a new Elixir application, without changing anything in the `elixir-omg` codebase.

This new application would be acting as the backend for the frontend apps. It would need to be plugged directly on a read-only replica of a Watcher DB to get data/events about the child chain, and would listen to events from Ethereum as well. 

From those two sets of data, this app would store relevant information in its own Postgres database and compute all the user-facing data we want to expose in our end-user applications (for example, the average block time over the past X hours). 

Any of our frontend application would be able to get what they need via a web API exposed by this new informational API app.

---

Now, on why we would want to do that. 

[...]

PR #1096, opened 2019-11-03
fix: decouple info with security

InoMurko:
POC for https://github.com/omisego/elixir-omg/issues/1065
https://github.com/omisego/elixir-omg/issues/1065#issuecomment-549185608

### DIFF
--- a/apps/omg_watcher/lib/omg_watcher/poc_event_handler.ex
+++ b/apps/omg_watcher/lib/omg_watcher/poc_event_handler.ex
@@ -1,0 +1,25 @@
+defmodule OMG.Watcher.PocEventHandler do
+  use GenServer
+  alias OMG.Watcher.DB.EthEvent
+
+  def start_link(subscribe_to) do
+    GenServer.start_link(__MODULE__, subscribe_to, name: subscribe_to)
+  end
+
+  def init(subscribe_to) do
+    :ok = OMG.Bus.subscribe(Atom.to_string(subscribe_to), link: true)
+    {:ok, %{subscribe_to: subscribe_to}}
+  end
+
+  def handle_info(
+        {:internal_event_bus, :sync_height, data},
+        state
+      ) do
+    case state.subscribe_to do
+      "exit_processor" -> EthEvent.insert_exits!(data)
+      "depositor" -> EthEvent.insert_deposits!(data)
+    end
+
+    {:noreply, state}
+  end
+end
--- a/apps/omg/lib/omg/ethereum_event_listener.ex
+++ b/apps/omg/lib/omg/ethereum_event_listener.ex
@@ -119,6 +119,7 @@ defmodule OMG.EthereumEventListener do
       |> maybe_update_event_cache(callbacks.get_ethereum_events_callback)
       |> Core.get_events(sync_height)
 
+    :ok = OMG.Bus.direct_local_broadcast(Atom.to_string(core.service_name), {:sync_height, events})
     :ok = :telemetry.execute([:process, __MODULE__], %{events: events}, core)
 
     {:ok, db_updates_from_callback} =
--- a/apps/omg_watcher/lib/omg_watcher/sync_supervisor.ex
+++ b/apps/omg_watcher/lib/omg_watcher/sync_supervisor.ex
@@ -37,96 +37,118 @@ defmodule OMG.Watcher.SyncSupervisor do
   end
 
   defp children() do
+    convenience_children() ++
+      [
+        %{
+          id: OMG.Watcher.BlockGetter.Supervisor,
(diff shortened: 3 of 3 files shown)


jrhite:
@unnawut This is starting to get interesting, you may wanna have a look and we can discuss tomorrow.

unnawut:
@InoMurko From discussion with @T-Dnzt, how would this bus work for retrieving past events?

A listener downstream that consume & store all events, then serve that outside this `OMG.Bus` and `OMG.Bus.PubSub`? Or it'll be handled by `OMG.Bus.PubSub`.

InoMurko:
Not sure I fully grasp the dimension of your question because the answer rather simple:

security properties of watcher ensure all events are aligned to a certain block. When syncing stops (for example because of byzantine events) watcher continues synchronization from the last known block. If watcher syncs, you get all the events from the bus. 

Does that answer your question? If not, what product requirements are you talking about?

unnawut:
To sum up the discussion, we were talking two approaches indeed:

  1. A lean Informational API. Exposes as much raw data as possible. Provides no derived information. Likely only few endpoints with simple and flat response structure.

  2. A feature-rich informational API. Exposes as much useful data as possible. Performs aggregations and 3rd party sources to compile a complete response. Advanced filtering and conditions.

The lean approach is straightforward and probably not even need a split. And the `OMG.Bus` is suffice for the data flow within the umbrella app. So no questions there.

My question was specifically to the feature-rich approach with the Informational API split out to a new umbrella app. And so in this scenario we would need ...

  - A sync mechanism, including replay.
  - The sync mechanism would not reside in `OMG.Bus`.
  - The sync code would rely solely on the `OMG.Bus` to maintain strict separation of concerns as a data transporter.

Extra note: I take it that your primary concern now is to get the separation of concerns between security vs. info up via the `OMG.Bus`. Then the sync mechanism, be it DB inserts, DB replica, PubSub feed, options are all open, as long as it talks to the bus and only the bus.

apps/omg_watcher/lib/omg_watcher/poc_event_handler.ex
```diff
@@ -0,0 +1,25 @@
...
+  def init(subscribe_to) do
+    :ok = OMG.Bus.subscribe(Atom.to_string(subscribe_to), link: true)
+    {:ok, %{subscribe_to: subscribe_to}}
+  end
+
+  def handle_info(
+        {:internal_event_bus, :sync_height, data},
+        state
+      ) do
+    case state.subscribe_to do
```
unnawut: What do you think if I do this per module instead of branching here, e.g. `ProcessExitEventHandler` and `DepositEventHandler`?

InoMurko: @unnawut I completely agree. This is a POC just to show what can be done. not necessarily how!

jrhite:
Thanks for putting together these notes. I'm indeed in favor of a feature-rich informational API. As you said, if it's going to be lean, probably no reason to split it.

The feature-rich split will allow us to do pretty much anything business or product would like us to do. We could do DB aggregations in SQL rather than in code. It will give us the freedom to add tables as needed (eg: an ERC20 token table, reporting rollup tables). We could also build and perform our own business-analytics on our DB without having to worry about putting load on the watcher.

In addition to a feature-rich API, if we start the split off in the right way, we can get some really good tests written. Testable code is well-designed code. 😄

InoMurko:
I would just like to emphasize that splitting the codebase IS in our roadmap. 
- Childchain
- Watcher (no reason to further divide the codebase if it's much easier to support security and informational in the same codebase)
- Common stuff

My opinion is that it is much much harder to maintain stability with increased fragmentation. And increasing fragmentation WILL not solve code quality. If you can't address code quality in a mono repo, you sure won't in a fragmented codebase.

jrhite:
Glad to hear about the common lib. Didn't know about that part.

I know this is only anecdotal evidence, but I started working at Amazon around January 2000. Their entire codebase was a single C++ app that was so large it could literally not be built. When running it almost could not fit in memory either.

It was a nightmare. During my time they we refactored the code out to CustomerService, OrderService, etc...

That changed everything. Features of each service  could be released independently. The API boundaries between what defined OrderService and CustomerService became clearly defined and testable...

You've also got the whole Micro Services pattern these days. It has its merits and strengths too....but also some drawbacks.

My real-life experience in splitting things up into logical-grouped components is pretty compelling. Indeed that's exactly what classes and module are for.

Nothing is free though. There are tradeoffs. For example this does have a DevOps cost. Sorry Jake. 😉

InoMurko:
We’re hardly Amazon.

PR #1099, opened 2019-11-04
refactor: use the same naming for exit period seconds as plasma contracts

InoMurko:
Slack

## Overview

 exit period config inconsistency, thanks @unnawut 
`Incosistency between this code (elixir-omg) and plasma_contracts repository`

## Changes

- Find and replace

## Testing

/


unnawut:
cc: @kendricktan @jbunce in case EXIT_PERIOD_SECONDS is used on any deployed environments

unnawut:
Okay renaming them internally is fine as well :joy:

InoMurko:
> EXIT_PERIOD_SECONDS

yeah, pls let me know here before I merge.

pnowosie:
Can you please mention in description it's incosistency between this code and plasma_contracts repo.
Are we 100% sure it's the same thing in both repos - I remember that in `plasma_framework` `MIN_EXIT_PERIOD` describes exit period for IFEs...
(CC @boolafish @pdobacz )

InoMurko:
> Okay renaming them internally is fine as well 😂

Oh - was too fast :|

apps/omg_watcher/test/support/integration/test_helper.ex
```diff
@@ -76,9 +76,9 @@ defmodule OMG.Watcher.Integration.TestHelper do
   end
 
   def process_exits(vault_id, token, user) do
-    exit_period_ms = Application.fetch_env!(:omg_eth, :exit_period_seconds) * 1000
+    min_exit_period = Application.fetch_env!(:omg_eth, :min_exit_period) * 1000
```
boolafish: Do you really want to re-use the word `min_exit_period` for different unit ?

apps/omg_eth/lib/omg_eth/release_tasks/set_contract.ex
```diff
@@ -60,8 +60,7 @@ defmodule OMG.Eth.ReleaseTasks.SetContract do
           plasma_framework_tx_hash: txhash_contract
         } = Jason.decode!(body, keys: :atoms!)
 
-        exit_period_seconds =
-          validate_integer(get_env("EXIT_PERIOD_SECONDS"), Application.get_env(@app, :exit_period_seconds))
+        min_exit_period = validate_integer(get_env("MIN_EXIT_PERIOD"), Application.get_env(@app, :min_exit_period))
```
boolafish: From the contract side, only `development` is allowed to be overridden by env variable. As a result, if once the config here and the one on `production` in contract get out synced, this way of getting `min_exit_period` might cause issue as the real effective one would always be in contract.

Might consider to fetch the real value from contract to override the `update_configuration` part instead.

see: https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/migrations/config.js#L31

jrhite:
As mentioned, please use a name with units in it if it's not too much trouble. `MIN_EXIT_PERIOD` is not helpful. `MIN_EXIT_PERIOD_SECS` is the most descriptive. The contracts renaming might be more painful.

As a side note I really think we should adopt adding units to the end of any units-based variables as a coding practice.

InoMurko:
This now includes https://github.com/omisego/elixir-omg/issues/1101

unnawut:
I'll review this today

Issue #672, opened 2019-05-08
Watcher dies when BlockGetter restarts with a rich UTXO set


jbunce:
```
watcher-parity-0 watcher ** (stop) exited in: GenServer.call(OMG.DB.LevelDBServer, {:multi_update, [{:put, :last_convenience_deposit_processor_eth_height, 4321221}]}, 5000)
watcher-parity-0 watcher     ** (EXIT) time out
watcher-parity-0 watcher     (elixir) lib/gen_server.ex:989: GenServer.call/3
watcher-parity-0 watcher     (stdlib) timer.erl:166: :timer.tc/1
watcher-parity-0 watcher     (omg_db) lib/db.ex:33: OMG.DB.multi_update/2
watcher-parity-0 watcher     lib/ethereum_event_listener.ex:115: OMG.API.EthereumEventListener.sync_height/2
watcher-parity-0 watcher     lib/ethereum_event_listener.ex:90: OMG.API.EthereumEventListener.handle_info/2
watcher-parity-0 watcher     (stdlib) gen_server.erl:637: :gen_server.try_dispatch/4
watcher-parity-0 watcher     (stdlib) gen_server.erl:711: :gen_server.handle_msg/6
watcher-parity-0 watcher     (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
watcher-parity-0 watcher Last message: :sync⋅
```

This eventually kills the Watcher:

```
watcher-parity-0 watcher 2019-05-08 04:22:45.300 [info] module=OMG.API.EthereumEventListener function=init/1 ⋅Starting EthereumEventListener for challenges_responds_processor⋅
watcher-parity-0 watcher 2019-05-08 04:22:45.303 [info] module=OMG.API.EthereumEventListener function=init/1 ⋅Starting EthereumEventListener for ife_exit_finalizer⋅
watcher-parity-0 watcher 2019-05-08 04:22:45.306 [info] module=OMG.API.EthereumEventListener function=init/1 ⋅Starting EthereumEventListener for convenience_deposit_processor⋅
watcher-parity-0 watcher 2019-05-08 04:22:45.308 [info] module=OMG.API.EthereumEventListener function=init/1 ⋅Starting EthereumEventListener for piggyback_processor⋅
watcher-parity-0 watcher 2019-05-08 04:22:45.310 [info] module=OMG.API.EthereumEventListener function=init/1 ⋅Starting EthereumEventListener for exit_processor⋅
[...]

PR #1103, opened 2019-11-05
Allows to load utxo set lazily, so OMG.State initializes quickly

pnowosie:
:clipboard: Related to #672, https://github.com/omisego/devops/issues/142

## Overview

This is continuation of @pdobacz's work in #1066 in regards to utxo lazy loading.

### How it works 
During `OMG.State` startup no UTXOs are fetched from DB, which is no longer blocking significantly.
Then during each of 6 utxo-related operations (see below) UTXO set is extended with UTXOs from DB to ensure operation behavior hasn't been changed.

Transaction processing is populating the in-memory UTXO set and once block is formed newly created UTXO are inserted to DB, but are also kept in process State. Service restart looses all UTXO created by transactions processed in the current block.

### Affected operations
Here I describe what changes was needed to maintain the previous behavior of the utxo-related functionality

* `utxo_exists?(utxo_pos)` - because `utxo_pos` in question could be created before process restarts its fetched from DB and merged to the in-memory UTXO set. On the other hand `utxo_pos` could be spend in recent transaction (in current, not-formed block) so it yet doesn't exists in-memory, but still exists in DB. Therefore new in-memory set `recently_spent` was added to handle this case.
* `exec(tx, _)` - It works the same as previously, but checks whether `tx`'s inputs exists is done in the DB-extended UTXO set.  It also populates `recently_spent` collection of the UTXOs spent in the current block.
* `deposit()` - no changes
* `exit_utxos(...)` - because `exit` operation supports multiple inputs forms (uto_pos, encoded utxo_pos, ife exits infos ...) it was split into two operations. First `Core.get_exiting_utxos` translates any of the inputs into exiting utxo position list. Then these UTXOs are fetched from DB and merged with in-memory set. Then exit from utxos are preformed. It also (just in case) populates the `recently_spent` collection of the exiting utxos.
* `form_block(...)` - works as previously, but also clears the `recently_spent` collection (because db_updates are pushed to DB)
* `close_block(...)` - no changes, it uses `form_block` so as mentioned :point_up:   

## Testing

New test are added to demonstrate behavior of the extended state. Fixes to existing tests were necessary.  

More information related to performance and how the change anticipate the future work can be find in description to :pr: #1066

### DIFF
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -128,22 +128,10 @@ defmodule OMG.State do
   Checks (stateful validity) and executes a spend transaction. Assuming stateless validity!
   """
   def handle_call({:exec, tx, fees}, _from, state) do
-    utxos_query_result =
-      tx |> Transaction.get_inputs() |> Enum.reject(&Core.utxo_exists?(&1, state)) |> Enum.map(&utxo_from_db/1)
+    utxos_query_result = load_necessary_utxos_from_db(state, tx)
 
     state
-    # FIXME: put the above logic into a `Core.get_exec_db_queries` w/ unit tests
-    # FIXME: testy, testy
-    # FIXME: what if utxo is not found? it must be handled properly and tested
-    |> Core.with_utxos(utxos_query_result)
-    |> Core.exec(tx, fees)
-    # FIXME must write pending txs to disk every time an exec goes through. Form block must flush those
-    #       think how to read them back for full block accountability on fail-overs and upgrades
-    #       IDEA: write pending tx on every `exec` and keep it in state too. On every `form_block`,
-    #             check in-memory pending state:
-    #             If it's empty, then read from DB (it's a failover, so I've not been in the master)
-    #             If it isn't empty, just use it, (normal operation, I've been in the master)
-    # FIXME cleany, cleany
+    |> Core.exec_db_queries(tx, fees, utxos_query_result)
     |> case do
       {:ok, tx_result, %Core{utxo_db_updates: db_updates} = new_state} ->
         :ok = DB.multi_update(db_updates)
@@ -153,12 +141,6 @@ defmodule OMG.State do
       {tx_result, new_state} ->
         {:reply, tx_result, new_state}
     end
-  end
-
-  # FIXME: move
-  defp utxo_from_db(input_pointer) do
-    {:ok, utxo_kv} = DB.utxo(InputPointer.Protocol.to_db_key(input_pointer))
-    utxo_kv
   end
 
   @doc """
@@ -241,5 +223,18 @@ defmodule OMG.State do
   defp publish_block_to_event_bus(block, event_triggers) do
     :ok = OMG.Bus.broadcast("events", {:preprocess_emit_events, event_triggers})
     :ok = OMG.Bus.direct_local_broadcast("blocks", {:enqueue_block, block})
+  end
+
+  defp load_necessary_utxos_from_db(state, tx) do
+    tx
+    |> Transaction.get_inputs()
+    |> Enum.reject(&Core.utxo_exists?(&1, state))
+    |> Enum.map(&utxo_from_db/1)
+    |> Enum.reject(&(:utxo_not_found == &1))
+  end
+
+  defp utxo_from_db(input_pointer) do
+    with {:ok, utxo_kv} <- DB.utxo(InputPointer.Protocol.to_db_key(input_pointer)),
+         do: utxo_kv
   end
 end
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -21,6 +21,7 @@ defmodule OMG.State do
   alias OMG.DB
   alias OMG.Eth
   alias OMG.Fees
+  alias OMG.InputPointer
   alias OMG.State.Core
   alias OMG.State.Transaction
   alias OMG.State.Transaction.Validator
@@ -90,21 +91,16 @@ defmodule OMG.State do
     # Get data essential for the State and Blockgetter. And it takes a while. TODO - measure it!
     # Our approach is simply blocking the supervision boot tree
     # until we've processed history.
-    {:ok, utxos_query_result} = DB.utxos()
+    # TODO(pnowosie): Above comment?
     {:ok, height_query_result} = DB.get_single_value(:child_top_block_number)
-    {:ok, [utxos_query_result, height_query_result], {:continue, :setup}}
+    {:ok, [height_query_result], {:continue, :setup}}
   end
 
-  def handle_continue(:setup, [utxos_query_result, height_query_result]) do
+  def handle_continue(:setup, [height_query_result]) do
     {:ok, child_block_interval} = Eth.RootChain.get_child_block_interval()
 
     {:ok, state} =
-      with {:ok, _data} = result <-
-             Core.extract_initial_state(
-               utxos_query_result,
-               height_query_result,
-               child_block_interval
-             ) do
+      with {:ok, _data} = result <- Core.extract_initial_state(height_query_result, child_block_interval) do
         _ = Logger.info("Started #{inspect(__MODULE__)}, height: #{height_query_result}}")
 
         {:ok, _} =
@@ -132,13 +128,37 @@ defmodule OMG.State do
   Checks (stateful validity) and executes a spend transaction. Assuming stateless validity!
   """
   def handle_call({:exec, tx, fees}, _from, state) do
-    case Core.exec(state, tx, fees) do
-      {:ok, tx_result, new_state} ->
-        {:reply, {:ok, tx_result}, new_state}
+    utxos_query_result =
+      tx |> Transaction.get_inputs() |> Enum.reject(&Core.utxo_exists?(&1, state)) |> Enum.map(&utxo_from_db/1)
+
+    state
+    # FIXME: put the above logic into a `Core.get_exec_db_queries` w/ unit tests
+    # FIXME: testy, testy
+    # FIXME: what if utxo is not found? it must be handled properly and tested
+    |> Core.with_utxos(utxos_query_result)
+    |> Core.exec(tx, fees)
+    # FIXME must write pending txs to disk every time an exec goes through. Form block must flush those
+    #       think how to read them back for full block accountability on fail-overs and upgrades
+    #       IDEA: write pending tx on every `exec` and keep it in state too. On every `form_block`,
+    #             check in-memory pending state:
+    #             If it's empty, then read from DB (it's a failover, so I've not been in the master)
+    #             If it isn't empty, just use it, (normal operation, I've been in the master)
+    # FIXME cleany, cleany
+    |> case do
+      {:ok, tx_result, %Core{utxo_db_updates: db_updates} = new_state} ->
+        :ok = DB.multi_update(db_updates)
+        # FIXME move elsewhere?
+        {:reply, {:ok, tx_result}, %Core{new_state | utxo_db_updates: []}}
 
       {tx_result, new_state} ->
         {:reply, tx_result, new_state}
     end
+  end
+
+  # FIXME: move
+  defp utxo_from_db(input_pointer) do
+    {:ok, utxo_kv} = DB.utxo(InputPointer.Protocol.to_db_key(input_pointer))
+    utxo_kv
   end
 
   @doc """
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -128,15 +128,17 @@ defmodule OMG.State do
   Checks (stateful validity) and executes a spend transaction. Assuming stateless validity!
   """
   def handle_call({:exec, tx, fees}, _from, state) do
-    utxos_query_result = load_necessary_utxos_from_db(state, tx)
+    db_utxos =
+    tx
+    |> Transaction.get_inputs()
+    |> init_utxos_from_db()
 
     state
-    |> Core.exec_db_queries(tx, fees, utxos_query_result)
+    |> Core.with_utxos(db_utxos)
+    |> Core.exec(tx, fees)
     |> case do
-      {:ok, tx_result, %Core{utxo_db_updates: db_updates} = new_state} ->
-        :ok = DB.multi_update(db_updates)
-        # FIXME move elsewhere?
-        {:reply, {:ok, tx_result}, %Core{new_state | utxo_db_updates: []}}
+      {:ok, tx_result, new_state} ->
+        {:reply, {:ok, tx_result}, new_state}
 
       {tx_result, new_state} ->
         {:reply, tx_result, new_state}
@@ -167,7 +169,8 @@ defmodule OMG.State do
   Tells if utxo exists
   """
   def handle_call({:utxo_exists, utxo}, _from, state) do
-    {:reply, Core.utxo_exists?(utxo, state), state}
+    db_utxos = init_utxos_from_db([utxo])
+    {:reply, Core.utxo_exists?(utxo, state, db_utxos), state}
   end
 
   @doc """
@@ -225,12 +228,12 @@ defmodule OMG.State do
     :ok = OMG.Bus.direct_local_broadcast("blocks", {:enqueue_block, block})
   end
 
-  defp load_necessary_utxos_from_db(state, tx) do
-    tx
-    |> Transaction.get_inputs()
-    |> Enum.reject(&Core.utxo_exists?(&1, state))
+  @spec init_utxos_from_db(list(InputPointer.Protocol.t())) :: UtxoSet.t()
+  defp init_utxos_from_db(utxo_pos_list) do
+    utxo_pos_list
     |> Enum.map(&utxo_from_db/1)
     |> Enum.reject(&(:not_found == &1))
+    |> UtxoSet.init()
   end
 
   defp utxo_from_db(input_pointer) do
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -230,7 +230,7 @@ defmodule OMG.State do
     |> Transaction.get_inputs()
     |> Enum.reject(&Core.utxo_exists?(&1, state))
     |> Enum.map(&utxo_from_db/1)
-    |> Enum.reject(&(:utxo_not_found == &1))
+    |> Enum.reject(&(:not_found == &1))
   end
 
   defp utxo_from_db(input_pointer) do
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -242,10 +242,10 @@ defmodule OMG.State do
       {:ok, utxo_kv} ->
         utxo_kv
 
-      :not_found ->
+      response ->
+        # only unsuccessful, acceptable response is not_found, crash otherwise
+        :not_found = response
         nil
-
-        # error cases intentionally not handled here
     end
   end
 end
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -125,7 +125,7 @@ defmodule OMG.State do
     db_utxos =
       tx
       |> Transaction.get_inputs()
-      |> init_utxos_from_db(state)
+      |> fetch_utxos_from_db(state)
 
     state
     |> Core.with_utxos(db_utxos)
@@ -154,9 +154,9 @@ defmodule OMG.State do
   Exits (spends) utxos on child chain, explicitly signals all utxos that have already been spent
   """
   def handle_call({:exit_utxos, utxos}, _from, state) do
-    exiting_utxos = Core.get_exiting_utxo_positions(utxos, state)
+    exiting_utxos = Core.extract_exiting_utxo_positions(utxos, state)
 
-    db_utxos = init_utxos_from_db(exiting_utxos, state)
+    db_utxos = fetch_utxos_from_db(exiting_utxos, state)
     state = Core.with_utxos(state, db_utxos)
 
     {:ok, {db_updates, validities}, new_state} = Core.exit_utxos(exiting_utxos, state)
@@ -168,7 +168,7 @@ defmodule OMG.State do
   Tells if utxo exists
   """
   def handle_call({:utxo_exists, utxo}, _from, state) do
-    db_utxos = init_utxos_from_db([utxo], state)
+    db_utxos = fetch_utxos_from_db([utxo], state)
     new_state = Core.with_utxos(state, db_utxos)
 
     {:reply, Core.utxo_exists?(utxo, new_state), new_state}
@@ -229,8 +229,8 @@ defmodule OMG.State do
     :ok = OMG.Bus.direct_local_broadcast("blocks", {:enqueue_block, block})
   end
 
-  @spec init_utxos_from_db(list(InputPointer.Protocol.t()), Core.t()) :: UtxoSet.t()
-  defp init_utxos_from_db(utxo_pos_list, state) do
+  @spec fetch_utxos_from_db(list(InputPointer.Protocol.t()), Core.t()) :: UtxoSet.t()
+  defp fetch_utxos_from_db(utxo_pos_list, state) do
     utxo_pos_list
     |> Stream.reject(&Core.utxo_processed?(&1, state))
     |> Enum.map(&utxo_from_db/1)
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -161,7 +161,11 @@ defmodule OMG.State do
   Exits (spends) utxos on child chain, explicitly signals all utxos that have already been spent
   """
   def handle_call({:exit_utxos, utxos}, _from, state) do
-    {:ok, {db_updates, validities}, new_state} = Core.exit_utxos(utxos, state)
+    exiting_utxos = Core.get_exiting_utxos(utxos, state)
+
+    db_utxos = init_utxos_from_db(exiting_utxos)
+
+    {:ok, {db_updates, validities}, new_state} = Core.exit_utxos(exiting_utxos, state, db_utxos)
 
     {:reply, {:ok, db_updates, validities}, new_state}
   end
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -89,10 +89,7 @@ defmodule OMG.State do
   Start processing state using the database entries
   """
   def init(:ok) do
-    # Get data essential for the State and Blockgetter. And it takes a while. TODO - measure it!
-    # Our approach is simply blocking the supervision boot tree
-    # until we've processed history.
-    # TODO(pnowosie): Above comment?
+    # Get data essential for the State and Blockgetter.
     {:ok, height_query_result} = DB.get_single_value(:child_top_block_number)
     {:ok, [height_query_result], {:continue, :setup}}
   end
--- a/apps/omg/lib/omg/state.ex
+++ b/apps/omg/lib/omg/state.ex
@@ -29,8 +29,9 @@ defmodule OMG.State do
   alias OMG.Utxo
 
   use GenServer
-
   use OMG.Utils.LoggerExt
+
+  require Utxo
 
   @type exec_error :: Validator.exec_error()
 
@@ -153,7 +154,7 @@ defmodule OMG.State do
   Exits (spends) utxos on child chain, explicitly signals all utxos that have already been spent
   """
   def handle_call({:exit_utxos, utxos}, _from, state) do
-    exiting_utxos = Core.get_exiting_utxos(utxos, state)
+    exiting_utxos = Core.get_exiting_utxo_positions(utxos, state)
 
     db_utxos = init_utxos_from_db(exiting_utxos, state)
     state = Core.with_utxos(state, db_utxos)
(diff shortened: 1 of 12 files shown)


apps/omg/lib/omg/state.ex
```diff
@@ -90,21 +91,16 @@ defmodule OMG.State do
     # Get data essential for the State and Blockgetter. And it takes a while. TODO - measure it!
```
pdobacz: update `@doc` perhaps? now it's too vague, let's make it specific.

Also `State.Core` module doc and `extract_initial_state` docs require update (check if anything else, incl. higher level docs, `architecture.md` maybe?). Maybe `extract_initial_state` should just be `init` instead

apps/omg/lib/omg/state.ex
```diff
@@ -222,4 +224,17 @@ defmodule OMG.State do
     :ok = OMG.Bus.broadcast("events", {:preprocess_emit_events, event_triggers})
     :ok = OMG.Bus.direct_local_broadcast("blocks", {:enqueue_block, block})
   end
+
+  defp load_necessary_utxos_from_db(state, tx) do
+    tx
+    |> Transaction.get_inputs()
```
pdobacz: I think everything except `Enum.map(&utxo_from_db/1)` should go to `core.ex` and have tests

apps/omg/lib/omg/state/core.ex
```diff
@@ -99,31 +99,24 @@ defmodule OMG.State.Core do
...
-        _utxos_query_result,
-        :not_found,
-        _child_block_interval
-      ) do
+  def extract_initial_state(:not_found, _child_block_interval) do
     {:error, :top_block_number_not_found}
   end
 
+  @spec with_utxos(t(), UtxoSet.query_result_t()) :: t()
+  def with_utxos(%Core{utxos: utxos} = state, utxos_query_result),
```
pdobacz: `defp`? also `exec` would have to be `defp`

OTOH let's contrast this approach with the more explicit `def with_utxos`, which enriches the UTXOset then `exec` which remains untouched. This has the benefit of smaller impact on the API & tests.

apps/omg/lib/omg/state/core.ex
```diff
@@ -99,31 +99,24 @@ defmodule OMG.State.Core do
...
+    do: %{state | utxos: UtxoSet.merge_with_query_result(utxos, utxos_query_result)}
```
pdobacz: ah, there's a more serious problem I missed in my experiments. If there are 2 txs that spend the same UTXO in a single block, they'll pass (there should be appropriate tests for this being impossible BTW, so the `State.CoreTest` should be expanded a lot). The first spends the UTXOs, but the second will pull them back from the DB again.

Looks like one needs to commit the `utxo_db_updates` to the DB on every `exec` in this case too.

pnowosie: Ah, no - we couldn't commit the updates because if everything dies & restart before we form the block the state is corrupted.

pnowosie: :bulb: Good idea: add test to show :point_up: problem.

apps/omg/lib/omg/state/core.ex
pdobacz: after discussion: public `with_utxos` and public `exec` seem more elegant. 

Just let's treat `exec` and other UTXOSet-operating functions assume that the UTXOSet operated on is well prepared. `with_utxos` prepares it, and must caare about all the edge cases (like the `recently_spent` - case discussed)

Let's try this

apps/omg/test/omg/state/persistence_test.exs
```diff
@@ -198,7 +198,10 @@ defmodule OMG.State.PersistenceTest do
 
   defp persist_common(state, db_updates, db_pid) do
     assert :ok = OMG.DB.multi_update(db_updates, db_pid)
-    assert state == state_from(db_pid)
+    assert drop_recently_spent(state) == state_from(db_pid)
     state
   end
+
+  # NOTE: `recently_spent` is in-memory property of the %Core{}, which isn't persisted so it can't be recovered from db
+  defp drop_recently_spent(state), do: %Core{state | recently_spent: []}
```
pdobacz: eh, was going to comment to not reach into internals, but in fact this whole test is a bit silly right now. Let's stop and think what to do with it.

apps/omg_watcher/test/omg_watcher/exit_processor/core/state_interaction_test.exs
```diff
@@ -194,9 +196,11 @@ defmodule OMG.Watcher.ExitProcessor.Core.StateInteractionTest do
     {:ok, %{^ife_id1 => exiting_positions1, ^ife_id2 => exiting_positions2}} =
       Core.prepare_utxo_exits_for_in_flight_exit_finalizations(processor, finalizations)
 
+    empty_db_utxos = %{}
```
pdobacz: again, the `with_utxos` argument here.

apps/omg/lib/omg/state/core.ex
```diff
@@ -18,7 +18,7 @@ defmodule OMG.State.Core do
   All spend transactions, deposits and exits should sync on this for validity of moving funds.
   """
 
-  defstruct [:height, :utxos, pending_txs: [], tx_index: 0, utxo_db_updates: []]
+  defstruct [:height, utxos: %{}, pending_txs: [], tx_index: 0, utxo_db_updates: [], recently_spent: []]
```
pdobacz: Let's
1/ revisit the docs to see if there aren't any obsolete stuffs there. This includes `moduledoc` here `docs` here `architecture.md`
2/ let's also expand the `moduledoc` here. I think now it requires a section explaning how the state is managed - what is in db, what in mem, what the semantics are there. Also - what happens if this crashes and restarts.

This part, let's tackle last, after the flows are nailed down

apps/omg/lib/omg/state.ex
```diff
@@ -84,27 +86,14 @@ defmodule OMG.State do
   ### Server
 
   @doc """
-  Start processing state using the database entries
+  Initializes the state. UTXO set is not loaded now.
```
pdobacz: don't have a better place to put the comment: we need to do sth about the utxo-based metrics calculations :(. Out of ideas for now

InoMurko:
I'm trying to figure out what the purpose of `recently_spent` is. Can you help me with that? Wondering how we could simply this. 

Why are we loading UTXOs in the memory in the first place? Couldn't we maintain the list consistently in DB?

apps/omg/lib/omg/state.ex
pnowosie: I've created an isse for it #1148

pnowosie:
> Why are we loading UTXOs in the memory in the first place? Couldn't we maintain the list consistently in DB and avoid loading them altogether?

It works like this:
Let's assume the chain has history of transactions and `OMG.State` process gets restarted for whatever reason.

During `OMG.State`  startup no UTXO will be loaded into memory - they all persist in DB.
Then process starts to process new transactions, deposits, exits... and it builds also it's in-memory utxoset.
These in-memory set works as an cache-layer as we don't need to fetch UTXO from DB if we know it was processed by State (`utxo position` exists either in `utxos` or `recently_spent`)

> I'm trying to figure out what the purpose of `recently_spent` is. Can you help me with that? 

So first, please mind that right now no info is written to DB after transaction is executed on `State`.
So let say `State` got restarted and it's processing two transactions spending the same UTXO.
As the input exists only in DB and DB isn't updated after tx is processed - one could be able to double spend - and that's why we need to store info about the recent spends. This collection gets cleared after block is formed because at that time all db updates are applied.

InoMurko:
But why go through the difficulty of maintaining cache? Why do we need cache at all? I would say, ditch cache **completely**, fallback to strong consistency by using the DB on processing transactions. *The performance argument is irrelevant.*

apps/omg/lib/omg/state/core.ex
```diff
@@ -250,77 +262,79 @@ defmodule OMG.State.Core do
...
-  Checks if utxo exists
+  Checks whether utxo exists in UTXO set.
+  Note: state passed here is already extended with DB.
   """
   @spec utxo_exists?(Utxo.Position.t(), t()) :: boolean()
-  def utxo_exists?(Utxo.position(_blknum, _txindex, _oindex) = utxo_pos, %Core{utxos: utxos}),
-    do: UtxoSet.exists?(utxos, utxo_pos)
+  def utxo_exists?(Utxo.position(_blknum, _txindex, _oindex) = utxo_pos, %Core{utxos: utxos}) do
+    utxos
+    |> UtxoSet.exists?(utxo_pos)
```
InoMurko:
```suggestion
    |> UtxoSet.exists?(utxos, utxo_pos)
```

apps/omg/test/omg/state/core_test.exs
```diff
@@ -49,6 +49,143 @@ defmodule OMG.State.CoreTest do
...
+    end
+
+    @tag fixtures: [:alice, :bob, :state_alice_deposit]
+    test "form block drops recent spends", %{alice: alice, bob: bob, state_alice_deposit: state} do
+      tx = create_recovered([{1, 0, 0, alice}], @eth, [{bob, 10}])
+      expected_spends = MapSet.new([Utxo.position(1, 0, 0)])
+
+      {:ok, _, state1} =
+        state
+        |> Core.exec(tx, :no_fees_required)
```
InoMurko:
```suggestion
        Core.exec(state, tx, :no_fees_required)
```

pnowosie:
> I would say, ditch cache completely, fallback to strong consistency...

It would require to rewrite current behaviour completely which is risky. Currenlty all `State.Core` operations are done on processes state (I call it cache to draw an analogy) and therefore we can reuse a lot of existing tests. 

Also making all changes in DB requires to carefully consider all the cases, e.g. we would need to persist the mempool transactions...
What's more while this :pr: moves as closer to #849 it is here to address #672

apps/omg/lib/omg/state/core.ex
pnowosie: Extended @moduledoc - please see if it makes sense to you @pdobacz

pdobacz:
> > I would say, ditch cache completely, fallback to strong consistency...
> 
> It would require to rewrite current behaviour completely which is risky.

Yeah, we confronted a couple of approaches here, ditching cache being one (and tempting!), but then it seemed like a much bigger change. So instead this is taking baby-steps instead, though I admit - not perfect.

apps/omg/test/omg/state/core_test.exs
```diff
@@ -49,6 +49,132 @@ defmodule OMG.State.CoreTest do
...
+      |> success?()
+      # this is important - state is extended before applying tx
+      |> Core.with_utxos(db_utxos)
+      |> Core.exec(tx, :no_fees_required)
+      |> fail?(:utxo_not_found)
+    end
+
+    @tag fixtures: [:state_alice_deposit]
+    test "extending state with same utxos does not change it", %{state_alice_deposit: state} do
+      db_utxos = state.utxos
```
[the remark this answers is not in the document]
pdobacz: this

apps/omg/test/omg/state/core_test.exs
```diff
@@ -49,6 +49,132 @@ defmodule OMG.State.CoreTest do
...
+
+      state
+      |> Core.with_utxos(db_utxos)
+      |> same?(state)
+    end
+
+    @tag fixtures: [:alice, :bob, :state_empty]
+    test "extending state partially", %{alice: alice, bob: bob, state_empty: state} do
+      fst_utxo = make_utxos([{1000, 0, 0, alice, @eth, 6}])
```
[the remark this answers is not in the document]
pdobacz: this, pls, look throught the comments from that review, seems like some of them got missed.

apps/omg/test/omg/state/core_test.exs
```diff
@@ -49,6 +49,141 @@ defmodule OMG.State.CoreTest do
...
+    end
+
+    @tag fixtures: [:alice, :bob, :state_empty]
+    test "recent spends are stored regardless state extension", %{alice: alice, bob: bob, state_empty: state} do
+      init_state = do_deposit(state, alice, %{amount: 10, currency: @eth, blknum: 1})
+      tx1 = create_recovered([{1, 0, 0, alice}], @eth, [{bob, 10}])
+
+      {:ok, _, state1} = Core.exec(init_state, tx1, :no_fees_required)
+
+      %Core{recently_spent: spent1} = state1
```
pdobacz: now this is a bit awkward... why are we constantly adding cases where we test via internals? here and next `test`.

I'm guessing, you need to transform the tests into more behavior-driven - start with the name, the name of the test is coupled to internals. What is the behavior that we're after here?

pnowosie:
> + pinging about the previous ones

Appologizes, GH is fooling me hidding comments especially on second review or is it rebase effects?

apps/omg/lib/omg/state.ex
```diff
@@ -222,4 +224,17 @@ defmodule OMG.State do
...
+  @spec init_utxos_from_db(list(InputPointer.Protocol.t())) :: UtxoSet.t()
+  defp init_utxos_from_db(utxo_pos_list) do
+    utxo_pos_list
+    |> Enum.map(&utxo_from_db/1)
+    |> Enum.reject(&(:not_found == &1))
+    |> UtxoSet.init()
+  end
+
+  defp utxo_from_db(input_pointer) do
+    with {:ok, utxo_kv} <- DB.utxo(InputPointer.Protocol.to_db_key(input_pointer)),
```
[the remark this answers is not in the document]
pnowosie: As we talked - we need to handle :not_found. Now they're filtered out in UtxoSet.init

apps/omg/test/omg/state/core_test.exs
pnowosie: I drop these tests as this is an detail but we actually don't have to clear `recently_spent` on form_block

pnowosie:
@InoMurko FYI
> But why go through the difficulty of maintaining cache?

I created an :issue: #1156 that proposed next simple* steps to ditch the cache completely. 
*) it looks simple on paper :smile:

apps/omg/lib/omg/state/core.ex
```diff
@@ -16,9 +16,38 @@ defmodule OMG.State.Core do
...
+  on process startup.
+
+  During OMG.State startup no UTXOs are fetched from DB, which is no longer blocking significantly.
+  Then during each of 6 utxo-related operations (see below) UTXO set is extended with UTXOs from DB to ensure operation
+  behavior hasn't been changed.
+
+  Transaction processing is populating the in-memory UTXO set and once block is formed newly created UTXO are inserted
+  to DB, but are also kept in process State. Service restart looses all UTXO created by transactions processed as well
+  as mempool transactions therefore DB content stays block-by-block consistent.
+  the current block.
```
pdobacz: leftover?

apps/omg/lib/omg/state/core.ex
```diff
@@ -41,7 +70,10 @@ defmodule OMG.State.Core do
           tx_index: non_neg_integer(),
           # NOTE: that this list is being build reverse, in some cases it may matter. It is reversed just before
           #       it leaves this module in `form_block/3`
-          utxo_db_updates: list(db_update())
+          utxo_db_updates: list(db_update()),
+          # NOTE: because UTXO set is not loaded from DB entirely, we need to remember the UTXOs spent in already
+          # processed transaction before they get removed from DB on form_block.
+          recently_spent: MapSet.t(Utxo.Position.t())
```
pdobacz: shouldn't be `(InputPointer.Protocol.t)` here?

apps/omg/lib/omg/state/core.ex
```diff
@@ -96,34 +128,41 @@ defmodule OMG.State.Core do
...
-        _utxos_query_result,
-        :not_found,
-        _child_block_interval
-      ) do
+  def extract_initial_state(:not_found, _child_block_interval) do
     {:error, :top_block_number_not_found}
   end
 
+  @doc """
+  Tell whether utxo position was created or spend by current state.
```
pdobacz: `spent`

apps/omg/lib/omg/state/core.ex
```diff
@@ -96,34 +128,41 @@ defmodule OMG.State.Core do
...
+  It's an optimization as there is no need to ask DB of known utxos
```
pdobacz: drop this line, it is not an optimization, if you don't `or` with `recently_spent` you'll allow double spends.

apps/omg/test/omg/state/persistence_test.exs
```diff
@@ -33,172 +34,205 @@ defmodule OMG.State.PersistenceTest do
...
+    {:ok, _, _} = OMG.State.exit_utxos([Utxo.position(1, 0, 0)])
 
-    state
+    # exit db_updates won't get persisted yet, but alice tries to spent it immediately
+    assert :utxo_not_found == exec(create_recovered([{1, 0, 0, alice}], @eth, [{alice, 20}]))
+
+    # retry above with empty in-memory utxoset
+    :ok = restart_state()
+
+    {:ok, _, _} = OMG.State.exit_utxos([Utxo.position(1, 0, 0)])
```
pdobacz: hm, this shouldn't be necessary and could harm the test, if you weren't persisting

apps/omg/test/omg/state/persistence_test.exs
```diff
@@ -33,172 +34,205 @@ defmodule OMG.State.PersistenceTest do
...
+    # easiest to do this is to schedule another operation on `OMG.State` which is blocking
+    _ = OMG.State.utxo_exists?(Utxo.position(0, 0, 0))
+
+    :ok
   end
 
-  defp persist_exit_utxos(state, utxo_positions, db_pid) do
-    assert {:ok, {db_updates, _}, state} = utxo_positions |> Core.exit_utxos(state)
-    persist_common(state, db_updates, db_pid)
+  defp exec(:ok, tx), do: exec(tx)
```
pdobacz: oh, wait, in the previous `persistence_test` we were piping to transfer `state`, now we don't have anything to pipe through the sequence of steps. So you ended up extending the piped functions to handle the `:ok`'s, but you could just unpipe the whole thing :joy:. I feel a bit culpable for this old test, so if you like, I can take care of it.

oh, so I also had the thought to merge these tests into `state_test.exs` and tidy that up too, so that there is something resembling a proper test of `OMG.State`. But you can punt this on me too, just LMK

pnowosie: At first I unpiped the flow, but then I saw the pipe approach as  `assert the previous  operation succeeded` because we don't have the `:error` conterpart

apps/omg/lib/omg/state/core.ex
pnowosie: :+1:

apps/omg/lib/omg/state.ex
```diff
@@ -132,7 +122,15 @@ defmodule OMG.State do
...
   """
   def handle_call({:exec, tx, fees}, _from, state) do
-    case Core.exec(state, tx, fees) do
+    db_utxos =
+      tx
+      |> Transaction.get_inputs()
+      |> init_utxos_from_db(state)
+
+    state
+    |> Core.with_utxos(db_utxos)
```
InoMurko: this seems to be a bit confusing step. why is this step separate from processing a transaction?

apps/omg/test/omg/state/persistence_test.exs
pnowosie: IMHO the test is valid as it's race condition possible *persisting exit is done outside of  `OMG.State` process) 
Let say `OMG.State` will have two requests (messages in its inbox)
1. :exit_utxo([alice_utxo])
2. :exec(tx_from_alice_spending_utxo)

So the test shows that even on this edge case there will be no double spend
WDYT?

apps/omg/lib/omg/state.ex
```diff
@@ -156,7 +154,12 @@ defmodule OMG.State do
   Exits (spends) utxos on child chain, explicitly signals all utxos that have already been spent
   """
   def handle_call({:exit_utxos, utxos}, _from, state) do
-    {:ok, {db_updates, validities}, new_state} = Core.exit_utxos(utxos, state)
+    exiting_utxos = Core.get_exiting_utxo_positions(utxos, state)
+
+    db_utxos = init_utxos_from_db(exiting_utxos, state)
```
InoMurko: `init` prefix seems to be used a couple of times... that invalidates the meaning.

apps/omg/lib/omg/state/core.ex
```diff
@@ -250,75 +291,89 @@ defmodule OMG.State.Core do
   NOTE: It is done like this to accommodate different clients of this function as they can either be
   bare `EthereumEventListener` or `ExitProcessor`. Hence different forms it can get the exiting utxos delivered
   """
-  @spec exit_utxos(exiting_utxos :: exiting_utxos_t(), state :: t()) ::
-          {:ok, {[db_update], validities_t()}, new_state :: t()}
-  # empty list of whatever to bypass typing
-  def exit_utxos([], %Core{} = state), do: {:ok, {[], {[], []}}, state}
+  @spec get_exiting_utxo_positions(exiting_utxos_t(), t()) :: list(Utxo.Position.t())
+  def get_exiting_utxo_positions(exit_infos, state)
```
InoMurko: the correct naming (since there's no IO) would be *extract*

apps/omg/test/omg/state/persistence_test.exs
pdobacz: oh right, you're not persisting here, gotcha

apps/omg/lib/omg/state.ex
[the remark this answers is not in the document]
InoMurko: maybe a better way of handling this would be to let the developer know what output's you're able to handle
```
case DB.utxo(InputPointer.Protocol.to_db_key(input_pointer)) do
  :not_found -> :not_found
  {:ok, _utxo_kv} = result -> result
end
```
this means that if you get a value back you're not able to handle, you would *fail and crash early*

apps/omg/lib/omg/state/core.ex
```diff
@@ -16,9 +16,37 @@ defmodule OMG.State.Core do
...
+
+  We experienced long startup times on large UTXO set, which in some case caused timeouts and lethal `OMG.State`
+  restart loop. To mitigate this issue we introduced loading UTXO set on demand (see GH#1103) instead of full load
+  on process startup.
+
+  During OMG.State startup no UTXOs are fetched from DB, which is no longer blocking significantly.
+  Then during each of 6 utxo-related operations (see below) UTXO set is extended with UTXOs from DB to ensure operation
+  behavior hasn't been changed.
+
+  Transaction processing is populating the in-memory UTXO set and once block is formed newly created UTXO are inserted
```
InoMurko: side question:
```
Transaction processing is populating the in-memory UTXO set and once block is formed newly created UTXO are inserted to DB, but are also kept in process State.
```
what happens if an outage happens in between? Between block is formed - newly created UTXOs are inserted to DB?
1. What if you're not able to submit a block, but were able to persist to DB.
2. What if you're able to submit a block, but not able to persist to DB.

apps/omg/lib/omg/state/core.ex
```diff
@@ -250,75 +289,89 @@ defmodule OMG.State.Core do
...
     |> Enum.flat_map(fn %{call_data: %{in_flight_tx: tx_bytes}} ->
       {:ok, tx} = Transaction.decode(tx_bytes)
       Transaction.get_inputs(tx)
     end)
-    |> exit_utxos(state)
   end
 
   # list of IFE input piggybacked events (they're ignored)
-  def exit_utxos([%{tx_hash: _, omg_data: %{piggyback_type: :input}} | _] = piggybacks, state) do
+  def get_exiting_utxo_positions([%{tx_hash: _, omg_data: %{piggyback_type: :input}} | _] = piggybacks, %Core{}) do
```
InoMurko: useless matching on `%Core{}` ?

apps/omg/lib/omg/state/core.ex
```diff
@@ -250,75 +289,89 @@ defmodule OMG.State.Core do
...
+  """
+  @spec exit_utxos(exiting_utxos :: list(Utxo.Position.t()), state :: t()) ::
+          {:ok, {[db_update], validities_t()}, new_state :: t()}
+  def exit_utxos([], %Core{} = state), do: {:ok, {[], {[], []}}, state}
+
+  def exit_utxos(
+        [Utxo.position(_, _, _) | _] = exiting_utxos,
+        %Core{utxos: utxos, recently_spent: recently_spent} = state
+      ) do
     _ = Logger.info("Recognized exits #{inspect(exiting_utxos)}")
```
InoMurko: Dangerous logging if large number of exits!

apps/omg/lib/omg/state/core.ex
```diff
@@ -250,75 +289,89 @@ defmodule OMG.State.Core do
...
-    |> exit_utxos(state)
   end
 
   # list of utxo positions (decoded)
-  def exit_utxos([Utxo.position(_, _, _) | _] = exiting_utxos, %Core{utxos: utxos} = state) do
+  def get_exiting_utxo_positions([Utxo.position(_, _, _) | _] = exiting_utxos, %Core{}), do: exiting_utxos
+
+  @doc """
+  Spends exited utxos.
+  Note: state passed here is already extended with DB.
```
InoMurko: the reason why you had to leave this comment is because you introduced `Core.with_utxos/2` which makes it harder to reason about what it does with state!

apps/omg/test/omg/state/persistence_test.exs
```diff
@@ -33,172 +34,205 @@ defmodule OMG.State.PersistenceTest do
...
-    {:ok, {_, _, db_updates}, state} = Core.form_block(@interval, state)
-    persist_common(state, db_updates, db_pid)
+  defp persist_form(:ok), do: persist_form()
+
+  defp persist_form() do
+    OMG.State.form_block()
+
+    # because `form_block` is non-blocking operation we need to wait it finishes
+    # easiest to do this is to schedule another operation on `OMG.State` which is blocking
+    _ = OMG.State.utxo_exists?(Utxo.position(0, 0, 0))
```
InoMurko: I don't understand this. Can you please elaborate what `persist_form` does?

apps/omg/test/omg/state/persistence_test.exs
```diff
@@ -33,172 +34,205 @@ defmodule OMG.State.PersistenceTest do
...
+
+    :ok = OMG.DB.multi_update(db_updates)
   end
 
-  defp persisted_standard_exitable_utxos(address, db_pid) do
-    {:ok, utxos_query_result} = OMG.DB.utxos(db_pid)
-    Core.standard_exitable_utxos(utxos_query_result, address.addr)
+  defp make_deposits(list) do
+    list
+    |> Enum.map(fn %{owner: owner, currency: currency, amount: amount, blknum: blknum} ->
```
InoMurko:
```suggestion
    Enum.map(list, fn %{owner: owner, currency: currency, amount: amount, blknum: blknum} ->
```

apps/omg/test/omg/state/persistence_test.exs
```diff
@@ -33,172 +34,205 @@ defmodule OMG.State.PersistenceTest do
...
+    end)
   end
 
-  defp persist_common(state, db_updates, db_pid) do
-    assert :ok = OMG.DB.multi_update(db_updates, db_pid)
-    assert state == state_from(db_pid)
-    state
+  defp restart_state() do
+    GenServer.stop(OMG.State)
+    WaitFor.ok(fn -> if(GenServer.whereis(OMG.State), do: :ok) end)
```
InoMurko: `stop/1,2` is a synchronous operation.
https://hexdocs.pm/elixir/GenServer.html#stop/3

apps/omg/test/omg/state/utxo_set_test.exs
```diff
@@ -63,6 +63,16 @@ defmodule OMG.State.UtxoSetTest do
                |> UtxoSet.init()
                |> UtxoSet.get_by_inputs(inputs)
     end
+
+    test "ignores db query possible not_founds",
```
InoMurko: I'm having difficulties understanding the purpose of this test... a better description would help.

apps/omg/test/omg/state_test.exs
```diff
@@ -51,7 +51,8 @@ defmodule OMG.StateTest do
   @tag fixtures: [:alice, :standalone_state_server]
   test "can execute various calls on OMG.State, one happy path only", %{alice: alice} do
```
InoMurko: I'm having difficulties understanding the purpose of this test... can't see why your changes would affect the test.

docs/architecture.md
```diff
@@ -103,7 +103,7 @@ Each instance of either Child Chain Server or Watcher should have it's own insta
 
 Database necessary to properly ensure validity and availability of blocks and transactions
 
-- it is read by `OMG.State` to discover the UTXO set on restart
+- it is read by `OMG.State`  on restart to discover the current block number and Ethereum height, UTXO set is lazily loaded
```
InoMurko: this could be formulated better. Also... this document is pretty stale. might be better if we retire this one until we have something more firm.

apps/omg/test/omg/state/persistence_test.exs
pnowosie: That's correct - `WaitFor` is here to assume the supervisor bring the process back

InoMurko: Why the assert? 
`assert OMG.State |> GenServer.whereis() |> Kernel.is_pid()`

pnowosie: To double check :smile: 
I'll drop the `assert` line

apps/omg/test/omg/state_test.exs
pnowosie: Ouch left off from initial branch - reverted.
Thx!

apps/omg/lib/omg/state.ex
pnowosie: Well extending the state of the utxos from DB is in several places, not just for tx processing.

apps/omg/lib/omg/state/core.ex
pnowosie: Case 2. is impossible. I'll describe case 1:

There are 2 processes engaged to block publishing
* OMG.State - check `form_block`
* OMG.ChildChain.BlockQueue - it check what gets mined on Ethereum and retries if needed

All _artifacts_ (mempool txs, new utxo, spends, block) from the OMG.State form_block is inserted to DB in single transaction (`multi_update`).
Then it publishes the event that new block is available - BQ is listening on these events.
If the outage blocks the event :point_up: propagation, BQ on init will check what gets mined on Ethereum and what block we have in DB and resend the missings.

Am I answered your concern?

PR #1106, opened 2019-11-05
:omg_performance overhaul, cleanup and add SE challenges testing

pdobacz:
Part of #855 - insofar as SEs and SE challenges go.

(SE - Standard Exit)

## Overview

Doing proper SE starting and challenges measurements required a significant overhaul of the perf test (a simplifying one). Also a lot of cascading cleanups were done. I see the diff reads reasonably well (the split into commits turned out quite well too, so review-by-commit is an option), but if that's too much to handle, I can try to split into smaller PR chunks (but the individual partial results will be uglier then).

Objectives were:
 - simplify & document existing machinery to ease migrating to `specs` (if/when that comes)
 - run the challenge tests easily
 - not overdo, b/c we're sun-setting this someday.

## Changes

1. `OMG.Performance.Byzantine` now has a new look and feel. It's expected to be run in short scripts from the `iex` shell. The stages can be mix and matched as the tester pleases. The "single, big, test command" approach was abandoned, in favor of composable scripts
2. `OMG.Performance.Byzantine` allows one to evaulate the behavior of processing new exits, querying the `byzantine_events`, getting challenge data and processing challenges
3. Lots of YAGNI/leftover stuff was removed like: concurrent Watcher-querying, superfluous defensive coding (like measuring totally borked requests), demos
4. Massive cleanups in the perftests and beyond (cleanups cascaded into more cleanups being required) and documentation. See commit list for details, more or less these: `omg_performance` dialyzing, watcher/chch fixtures, superfluous `ensure_all_started`s (O_o), homemade `Task.async_stream`, options handling, `WaitFor` stuff, chch/watcher/eth RPC urls handling
5. `ganache` support for tests was thrown in (to be able to do all the refactors) **EDIT** in fact, I'll open a separate PR with the `ganache` stuff, so feel free to ignore that part in the review. I'll still include the commits here, as they're quite useful for testing

## Testing

Existing tests, `mix test --include integration test/omg_performance` in particular.

Those tests can be run much faster using `ganache`. Put:
```
  eth_node: :ganache,
  run_test_eth_dev_node: false
```

into your `omg_eth/config/test.exs`, `docker-compose -f docker-compose-fixtures.yml up ganache` and enjoy faster tests.

Also one can give the `iex` perf testing scripts a spin (Example usage should be documented enough in the modules).


apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -163,8 +137,32 @@ defmodule OMG.Performance.ByzantineEvents do
...
+  # this allows one to map a contract-transacting function over a collection nicely.
+  # It initiates all the transactions concurrently. Then it waits on all of them to mine successfully.
+  # Returns the last receipt result, so you can synchronize on the block number returned (and the entire bundle of txs)
+  defp map_contract_transaction(enumberable, transaction_function) do
+    enumberable
+    |> Enum.map(transaction_function)
+    # NOTE: infinity doesn't work, hence the large number
+    |> Task.async_stream(&Support.DevHelper.transact_sync!/1, max_concurrency: 10_000)
+    |> Enum.map(fn {:ok, result} -> result end)
+    |> List.last()
```
pnowosie: It isn't obvious to me that the particular last transaction in the list completes the last.
Is it possible that but-last tx completes as last one - but we'll be waiting "little" shorter (?)

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -14,130 +14,248 @@
...
   @moduledoc """
-  OMG network child chain server byzantine event test entrypoint. Setup and runs performance byzantine tests.
+  OMG network child chain server byzantine event test entrypoint. Runs performance byzantine tests.
 
-  # Usage
+  ## Usage
 
-  See functions in this module for options available
+  To setup, once you have your Ethereum node and a child chain running, from a configured `iex -S mix run --no-start`
+  shell do:
```
pnowosie: Do you mean repl here? I believe it's common term
```suggestion
  REPL do:
```

Please replace other occurences :point_down:

apps/omg_performance/test/omg_performance/integration/performance/simple_perftest_test.exs
```diff
@@ -13,11 +13,15 @@
...
+  @moduledoc """
+  Simple smoke testing of the performance test
+  """
+
   use ExUnit.Case, async: false
 
   import ExUnit.CaptureIO
 
-  alias OMG.Performance
+  use OMG.Performance
```
pnowosie: I like it :+1:

dialyzer.ignore-warnings
```diff
@@ -1,4 +1,3 @@
-apps/omg_performance/
```
pnowosie: :+1:

apps/omg_eth/test/support/wait_for.ex
```diff
@@ -52,18 +51,25 @@ defmodule Support.WaitFor do
...
     end
 
+    WaitFor.ok(f, timeout)
+  end
+
+  # Repeats f until f returns {:ok, ...}, :ok OR exception is raised (see :erlang.exit, :erlang.error) OR timeout
+  # after `timeout` milliseconds specified
+  #
+  # Simple throws and :badmatch are treated as signals to repeat
+  def ok(f, timeout \\ 5_000) do
```
pnowosie: Aha - I finally got it :exclamation: 
This is super clever - much better, comparing to the prev implementation

apps/omg_performance/test/omg_performance/integration/byzantine_event_test.exs
```diff
@@ -35,7 +35,7 @@ defmodule OMG.Performance.ByzantineEventsTest do
     {:ok, %{destdir: destdir}}
   end
 
-  @tag fixtures: [:perf_test, :child_chain, :omg_watcher]
+  @tag fixtures: [:perf_test, :mix_based_child_chain, :mix_based_watcher]
```
pnowosie: I :heart: new names!

apps/omg_performance/lib/performance.ex
```diff
@@ -71,6 +71,42 @@ defmodule OMG.Performance do
 
   @eth OMG.Eth.RootChain.eth_pseudo_address()
 
+  # FIXME docs
+  # FIXME map to keyword for opts everywhere
```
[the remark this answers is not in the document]
pdobacz: more idiomatic + syntax sugar `f(arg1, arg2, pos_arg1: 5)` (no brackets)

apps/omg_performance/lib/omg_performance/byzantine_events.ex
pdobacz: but, but https://hexdocs.pm/iex/IEx.html says shell and it's more common according to google :P

apps/omg_performance/lib/omg_performance/byzantine_events.ex
pdobacz: ha, I stopped at this too. But they will actually all complete (`Task.async_stream` isn't that async :) ).

What's more, it will also return the last transaction sent out - note that the actual sending of transactions is synchronous and sequentioal (`Enum.map(transaction_function)`). Only the "waiting" and "receipt getting" is `Task.async_stream`, and there, order is the same as with inputs :sunglasses:

dialyzer.ignore-warnings
pdobacz: that was surprisingly easy to get passing actually. I think a leftover from the `over/underspecs` episode

apps/omg_eth/test/support/wait_for.ex
pdobacz: :) I didn't really change anything, just now the timeout is included, doesn't need to be provided by caller. All cases did the same thing always :).

pnowosie: but previously we had some expectations of the return `:repeat` :arrow_forward: ward: it's cleaner now

PR #1107, opened 2019-11-05
Add ganache support for convenience/new integration tests

pdobacz:
Cherry picked from #1106 

## Overview

Adds support to run integration tests (and the chch/watcher as a whole) against `ganache-cli`. Should be useful immediately (was when I was doing #1106) and as a baby step towards the new integration testing.

Contrary to `geth`/`parity` it doesn't auto-start `ganache` when doing integration tests, bring your own RPC. There's little point in providing this auto-start, facing impending revamp of integration tests.

### Caveats of `ganache`:

- BlockQueue doesn't like the way `ganache` handles submitted transactions. This looks like a discrepancy wrt. other nodes. Submitted transactions here contribute to the addresses nonce, while for other nodes, only mined transactions do. This is punted now just by turning what would be an error into a warnings, logging and moving on (only if `eth_node: :ganache`!).
- it's slow (`eth_call`s we do turns out ~50 times slower), so things that work fine with `geth`/`parity` might be unstable with `ganache`. I'll file a separate issue with the details.

## Changes

- introduce a `ganache` setting to `eth_node`, document
- introduce some handling functions to make it run and test
- make the docker-compose invocation of `ganache-cli` mine blocks on a cadence (10blocks/s)
- patch up `BlockQueue` to cater for one peculiarity of `ganache`
- this in turn resulted in woes from `credo` so the function there got a slight refactor

## Testing

Put:
```
  eth_node: :ganache,
  run_test_eth_dev_node: false
```
into your `omg_eth/config/test.exs`, `docker-compose -f docker-compose-fixtures.yml up ganache` and 
`mix test --only integration`.

One of the tests is flaky with this, but let's punt it maybe? CI continues on `geth` for now.

### DIFF
--- a/apps/omg_eth/test/support/dev_helper.ex
+++ b/apps/omg_eth/test/support/dev_helper.ex
@@ -81,9 +81,7 @@ defmodule Support.DevHelper do
   import priv key->unlock->fund with lots of ether on that account
   """
   def import_unlock_fund(%{priv: account_priv}, opts \\ []) do
-    account_priv_enc = Base.encode16(account_priv)
-
-    {:ok, account_enc} = create_account_from_secret(backend(), account_priv_enc, @passphrase)
+    {:ok, account_enc} = create_account_from_secret(backend(), account_priv, @passphrase)
     {:ok, _} = fund_address_from_faucet(account_enc, opts)
 
     {:ok, from_hex(account_enc)}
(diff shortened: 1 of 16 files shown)


apps/omg_eth/test/support/dev_helper.ex
```diff
@@ -154,13 +152,19 @@ defmodule Support.DevHelper do
...
-  def create_account_from_secret(:geth, secret, passphrase) do
-    {:ok, _} = Ethereumex.HttpClient.request("personal_importRawKey", [secret, passphrase], [])
+  def create_account_from_secret(:ganache, account_priv, passphrase) do
+    account_priv_enc = Eth.Encoding.to_hex(account_priv)
+    {:ok, _} = Ethereumex.HttpClient.request("personal_importRawKey", [account_priv_enc, passphrase], [])
+  end
+
+  def create_account_from_secret(:geth, account_priv, passphrase) do
+    account_priv_enc = Base.encode16(account_priv)
+    {:ok, _} = Ethereumex.HttpClient.request("personal_importRawKey", [account_priv_enc, passphrase], [])
```
achiurizo: Could we refactor this a bit?

```elixir
 def create_account_from_secret(:ganache, secret, passphrase),.                                                                                                                      
   do: foo("personal_importRawKey", Eth.Encoding.to_hex(secret), passphrase)
                                                                                                                                                                                   
def create_account_from_secret(:geth, secret, passphrase),.                                                                                                                        
  do: foo("personal_importRawKey", Base.encode16(account_priv), passphrase)
                                                                                  
def create_account_from_secret(:parity, secret, passphrase) when byte_size(secret) == 64,
  do: foo("parity_newAccountFromSecret", Eth.Encoding.to_hex(secret), passphrase)
 
defp create_account_from_secret(method_name, secret, passphrase) do
  {:ok, _} = Ethereumex.HttpClient.request(method_name, [secret, passphrase], [])
end
```

Issue #1108, opened 2019-11-06
Update SE vs IFE interaction doc


boolafish:
We change the mechanism between SE and IFE interaction quite heavily during ALD.
Should upgrade the document accordingly.

https://github.com/omisego/elixir-omg/blob/master/docs/standard_vs_in_flight_exits_interaction.md

PR #1118, opened 2019-11-08
docs: update SE<>IFE interaction doc to new mechanism

boolafish:
### Note
Update the SE<>IFE document. Pretty much a re-write and then put a reference of old mechanism using `0.2` branch.

closes: #1108

### DIFF
--- a/docs/standard_vs_in_flight_exits_interaction.md
+++ b/docs/standard_vs_in_flight_exits_interaction.md
@@ -1,103 +1,44 @@
 # Standard vs In-flight Exits
 
-Since both Standard exit (aka SE) and In-flight exit (aka IFE) provide an exit opportunity, it is possible to effectively double-spend.
-To prevent such possibility, and to preserve the rule that honest owner can ignore IFE provided that he enjoys data availability, we need a set of rules.
-We derive this set of rules by analyzing different scenarios, depending on order of actions taken by calling the MoreVP root chain contract.
-Every scenario revolves around a conflict of exit actions (a double-spend attempt).
-In each we explain possible ways of preventing the double spend.
+Since both Standard exit (aka SE) and In-flight exit (aka IFE) provide an exit opportunity, it is possible to effectively double-spend. To prevent such possibility, whenever an output is finalized/withdrawn from the Plasma M(ore)VP framework, we flag the output as spent to the `PlasmaFramework`. For all exit game contract implementations, they should ignore those already-spent-outputs during `processExit`. In other words, though there can be multiple exits for the same output started concurrently, only the first that get processed can really exit the output. All the others would just not process and ignore the output if it is already flagged.
 
 #### Glossary
 
 * `IFE` - in-flight exit, active unless stated otherwise;
 * `SE` - standard exit, active unless stated otherwise;
-* `SE on (in|out)put` - standard exit on utxo mentioned in in-flight exit as input or output;
-* `->` denotes order of events;
-* `=>` denotes proposed solution / action;
-* `contract:` solution is contained to change in the contract;
+* `utxoPos` - output position inside Plasma chain. It is a combination of `blockNumber`, `txIndex` and `outputIndex` that shows where the output is located in the plasma chain;
+* `OutputId` - global output identifier used in Plasma Framework, see description below;
 
-### Possible scenarios
+#### OutputId
+To be able to flag all finalized outputs, we would need a global schema for all outputs in our Plasma Framework. In current implementation, the global identifier for output is called `OutputId`. The schema is as followed:
 
-1. Mallory does SE on input -> Mallory submits IFE  
-    => (IMPLEMENTED) contract: IFE challenges SE (~ +30k gas)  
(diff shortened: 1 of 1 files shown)


boolafish:
Do we want to go through old all possible scenario like
```
 Mallory does SE on input -> Mallory submits IFE  
```

Or this is clear enough to state one output can only exit once as it is flagged?

paulperegud:
> Do we want to go through old all possible scenario like
> 
> ```
>  Mallory does SE on input -> Mallory submits IFE  
> ```
> 
> Or this is clear enough to state one output can only exit once as it is flagged?

Yeah, with explicit flagging of out output we no longer have a complicated game. Thus lets just drop alice-bobbing from the doc.

pdobacz:
hm, late to the party but... does it mean that we never ever have to challenge SEs with IFEs spending the SE-exiting outputs as their inputs?

This was the condition that the Watcher watched for in the previous design. Does the flagging mechanism make it optional? (as in, the invalid SE will never actually exit, so the only reason to watch is to challenge the SE and profit).

I had this on my "catch-up" list, but actually I have now stumbled upon some trouble related to the SE vs IFE challenging. If I could just :scissors: it out, that would solve it.

boolafish:
We need to challenge, because IFE can happen first and then SE

paulperegud:
> We need to challenge, because IFE can happen first and then SE

But if both IFE is marking it's inputs and SE is marking the output, only one of them can exit. Thus the challenge seems to be mostly to grab the bond. @boolafish am I missing something? SE was always marking the output as spent, BTW.

boolafish:
oh, wow....did not realize flagging *all* inputs For IFE would have such good effect automatically :)

But on the other hand, if we are to automatically collect IFE tx into chain, this might be sth we would like to consider.

pdobacz:
> We need to challenge, because IFE can happen first and then SE

ok so let's double-triple check. The docs about challenging have been removed and it seems like no challenging should be done. But if there's doubt, let's clarify to be 100% sure

> But on the other hand, if we are to automatically collect IFE tx into chain

hmmm, can you think of a scenario where things would break?

boolafish:
> hmmm, can you think of a scenario where things would break?

Both SE and IFE showed. Child chain include the IFE tx to plasma block. However, it end up the SE is not challenged so SE succeed and IFE failed. Now a user can SE the output of the IFE tx as that is included to plasma block now.

pdobacz:
> > hmmm, can you think of a scenario where things would break?
> 
> Both SE and IFE showed. Child chain  include the IFE tx to plasma block

but then the SE will be challenge with the IFE tx, as it now is included in the block. It _should_ be the same as if the IFE sender submitted/included the tx themselves, just before IFEing.

But I agree, there might be something lurking there. Would you be willing spend a moment and analyze, whether "SE by IFE" challenging can in fact be dropped altogether, from a security-only-concerned Watcher? The change in the document implies that.

It would be quite convenient for me to just drop it, but I'm erring on the side of caution here.

Oh, I just thought about a different concern - in a situation, when `processExit` is "omitted" due to having been flagged as exited earlier - does the bond get returned anyway. My issue is here, that if we don't challenge the SE then, despite not being able to double-spend, the attempt is left unpunished, which might be a grieving vector.

I'll think about this a bit too.

pdobacz:
OK I spent some time on it and have a couple of loosely coupled comments:

1. 

> It would be quite convenient for me to just drop it

On 2nd thought, dropping doesn't help me with what I'm struggling with, so there's no motivation from that end. I think the Watcher will keep detecting such invalid SEs and allowing to challenge.

2. Where can I find the complete write up of the current behavior of flagging? The doc in question only references a post with an idea and some discussion, this doesn't provide much detail. These need to be answered precisely:
  - what is flagged on IFE start
  - what is flagged on IFE input piggyback
  - what is flagged on IFE output piggyback
  - what is flagged on IFE finalization, canonical/non-canonical, piggybacked/non-piggybacked inputs/outputs
  - what are the conditions checked on IFE finalization, for inputs/outputs, for canonical/non-cannonical
  - what is flagged on SE start
  - what is flagged on SE process
  - what are the conditions checked on SE finalization
  - what happens to bonds in all cases, where flagging "blocks" exiting

If such write up exists, this should be linked from the document here.

3. I thought I had a scenario where sth breaks, but not sure now, need 2 first.

boolafish:
Currently it is only flagged when `processExit` (both SE and IFE). The doc here has the description of that. (However, seems like need to rephrase to be more clear as you as a reader did not get that.) Another PR of mine working on doc might touch a bit as well: https://github.com/omisego/plasma-contracts/blob/ec3b2043363fa67800e581b2f8cf55e91c2cdb38/plasma_framework/docs/design/payment-game-implementation-v1.md

pdobacz:
yeah, I think the docs/specs should hold the current design of the mechanics and (if necessary) explanation why it works, in its entirety. Just state what and when is flagged and when flags are checked and that's it.

All the links to prior discussions/PRs should only be used to provide historical background. But one should not need to follow those links to understand how and why things work.

(sorry to be late with those comments, since I'm opening a bug report now, I'll include a request to produce reference docs there in, and link to from `elixir-omg`)

PR #1124, opened 2019-11-11
Retry plasma-contracts deployment in docker-compose

unnawut:
## Overview

This fixes the issue where `docker-compose up` sometimes failing to start `childchain` complaining that the `plasma-contract` container is unhealthy.

## Details

A few reports have been made before regarding `docker-compose up` failing and working after the 2nd try.

```
Jacob 10:47
I just pulled the latest master to a fresh branch and wanted to bring up the cluster. docker-start-cluster croaks with:

2019-11-11 03:39:37.426 [info] module=application_controller function=info_exited/3 ⋅Application omg_child_chain exited: OMG.ChildChain.Application.start(:normal, []) returned an error: an exception was raised:
    ** (MatchError) no match of right hand side value: {:error, :root_chain_contract_not_available}
        (omg_child_chain) lib/omg_child_chain/supervisor.ex:35: OMG.ChildChain.Supervisor.init/1
        (stdlib) supervisor.erl:295: :supervisor.init/1
        (stdlib) gen_server.erl:374: :gen_server.init_it/2
        (stdlib) gen_server.erl:342: :gen_server.init_it/6
        (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3⋅
```

```
jarindrian the piggybacker 11:02
anyone able to run docker-compose on elixir-omg successfully ?

2019-10-31 04:00:59.817 [info] module=application_controller function=info_exited/3 ⋅Application omg_child_chain exited: OMG.ChildChain.Application.start(:normal, []) returned an error: an exception was raised:
    ** (MatchError) no match of right hand side value: {:error, :root_chain_contract_not_available}
        (omg_child_chain) lib/omg_child_chain/supervisor.ex:35: OMG.ChildChain.Supervisor.init/1
        (stdlib) supervisor.erl:295: :supervisor.init/1
        (stdlib) gen_server.erl:374: :gen_server.init_it/2
        (stdlib) gen_server.erl:342: :gen_server.init_it/6
        (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3⋅

nicholas 13:03
can confirm with @jarindrian the piggybacker i get the same error when trying to run latest master on elixir-omg. can’t bring up the watcher container, logs show same error
```

A closer look at `plasma-contracts` container logs, reveals this:

```
   Deploying 'Migrations'
   ----------------------

Error: Error: Error:  *** Deployment Failed ***

"Migrations" exceeded the block limit (with a gas value you set).
   * Block limit:  0x2019a35
   * Gas sent:     6721975
   * Try:
      + Sending less gas.
      + Setting a higher network block limit if you are on a
        private network or test client (like ganache).

    at Object.run (/home/node/plasma-contracts/plasma_framework/node_modules/truffle/build/webpack:/packages/migrate/index.js:96:1)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
Truffle v5.0.38 (core: 5.0.38)
Node v8.16.2
cat: can't open 'outputs.json': No such file or directory
npx: installed 237 in 20.965s

  \{^_^}/ hi!

  Loading db.json
SyntaxError: Malformed JSON in file: db.json
Unexpected end of JSON input
    at FileAsync.parse [as deserialize] (<anonymous>)
    at /root/.npm/_npx/33/lib/node_modules/json-server/node_modules/lowdb/adapters/FileAsync.js:41:35
    at <anonymous>

# Followed by logs repeating the steps which hint a container restart
```

Because truffle fails to deploy the contracts due to block gas limit, it quickly trickles down the container and causes the container to exit. 

And since `childchain` depends on `plasma-contracts`, instead of waiting while the contracts container has status `Up (health: starting)`, it sees the container's exit and consider this as unhealthy immediately.

Then, since `plasma-contracts` docker-compose definition contains `restart: always` it attempts to restart the container. The contracts redeployed (with now the limit is raised), `docker-compose up` is then reported to work.

## Changes

- Add for loop to retry contracts deployment up to 5 times without making the container restart
- Lower `plasma-contracts` healthcheck value now that we have the retry

Extra stuff. Not related to the issue but trivial enough:
- Pin plasma-contracts docker-compose image to stable instead of dev (it's the same image)
- Fix Makefile typo
- Add missing `PRIVATE_KEY` env var in docker-compose.infura.yml
- Replace string placeholders with env vars in docker-compose.infura.yml

## Testing

`docker-compose up` should experience less or no random failures.


boolafish:
No the issue is outside truffles’ control actually. I think this is the same one me and Ino experienced which is we cannot start Geth from the certain gas block limit directly, it took some time for Geth to move to that.

From truffle we could decrease the default gas used to deploy a bit but our PaymentExitGame is using some 6m gas to deploy as well so probably won’t help too much now

unnawut:
> we cannot start Geth from the certain gas block limit directly, it took some time for Geth to move to that.

Can confirm the same thing still happened to me while testing this PR. Even with `--miner.gastarget 7500000` already set, the actual gas limit is much lower at 6m something which is not enough to deploy the contracts.

unnawut:
@boolafish Just to make sure, were you requesting a change to this PR or are we good?

boolafish:
lol, I am just replying to @jrhite 

> Surprised truffle doesn't have a mechanism to support this directly.

PR #1126, opened 2019-11-12
931 blocks ethevents db

jrhite:
Add new `root_chain_txhash_event` column to `blocks` table so that the eth txn hash of a block is recorded.

Do the DB migration first in this PR, and then follow it by another PR that will contain the code changes to store the eth txn hash value in the blocks table.

For testing, ran the migration locally.


unnawut:
1. Doesn't `SET root_chain_txhash_event = '\\000';` cause fk constraint error?

2. Would this work on new inserts? Since the current code will not pass any value for `:root_chain_txhash_event`, so I'm assuming the insert value defaults to null and we have already have NOT NULL.

3. Not sure of the practice in `elixir-omg`, do we do PRs for db migrations if it's not a feature or a fix by itself? I guess this one cc: @T-Dnzt @achiurizo

jrhite:
> 1. Doesn't `SET root_chain_txhash_event = '\\000';` cause fk constraint error?
> 2. Would this work on new inserts? Since the current code will not pass any value for `:root_chain_txhash_event`, so I'm assuming the insert value defaults to null and we have already have NOT NULL.
> 3. Not sure of the practice in `elixir-omg`, do we do PRs for db migrations if it's not a feature or a fix by itself? I guess this one cc: @T-Dnzt @achiurizo

Yeah, all your points are correct. 😄 Was trying to break it up into migration, then code, but that's not going to work. Let me add the elixir code for this to to this PR. It shouldn't be much.

About backfilling...this should only be weird for our local/dev/staging environments... The column would ideally be non-nullable though. I'll think about an appropriate back-fill value that won't be an FK-constraint violation...Maybe I can grab the first ethevents row and backfill it with that as an FK. If there are no ethevents rows, then the assumption is there will also be no blocks rows.

unnawut:
I keep getting notifications to review this PR so I'm going to remove myself as reviewer for now. Feel free to add me back later 🙇

InoMurko:
@jrhite I'll close this and you re-open if you need to.

jrhite:
Still a legit ticket, but currently blocked. Reopening for it to stay on the project board for planning purposes.

unnawut:
Given now we're handling bugs (although indirectly) from code and DB fields that are not being used, I'll close this DB-only issue so we don't need to maintain what we don't yet use.

Feel free to submit a new issue when a user story arises and is promoted into an engineering cycle!

PR #1135, opened 2019-11-13
Add a fuzzer for solidity `RLPReader.sol`

paulperegud:
It attempts to find two encodings that produce the same structure when unpacked.

To run it run `mix test --only fuzz`

It finds the same bugs as we've found in ex_rlp.


boolafish:
Would like to confirm what is the goal:
1. Do we want to compare elixir result vs solidity result?
    1. If no, probably it is better to just move the test under truffle and have it on the CI here.
    1. If yes (which I think should be the case), a minor comment would be this probably belongs under `mock/` folder as this is not a production contract code more of a test code. You might want a specific folder for that (eg. `conformance_test`?) Or you can have some special folder for that even not under `mock`, just remember to add the folder to `.solcover.js`

paulperegud:
>        1. If no, probably it is better to just move the test under truffle and have it on the CI here.

Moving to truffle test suite would mean that test itself needs to be re-implemented using one of quickcheck implementations for JS. I've looked into it and [FastCheck](https://github.com/dubzzz/fast-check) looks fitting. It has no generator for binary type, but that can be easily simulated using generator for hex strings and converting result to binary.

pdobacz:
un-requesting review from myself, since I'm assigned anyway. I will see this through this way or another. Thx!

InoMurko:
If this is still needed, please reopen. Closing for now. KTNX

pdobacz:
I might reopen, after I (we) wrap more head around how to structure those kinds of tests. Thx for grooming.

Issue #1141, opened 2019-11-15
OMG.ChildChain.SyncSupervisor can die quietly


pdobacz:
If `OMG.ChildChain.SyncSupervisor` dies (e.g. when a child process failed to restart after crashing), it dies silently, leaving `omg_child_chain` app damaged. There's no log entry, and no alarms:

```
iex(4)> OMG.ChildChain.get_alarms()
{:ok, [{{:disk_almost_full, '/snap/core/7396'}, []}]}
```
(disregard the `disk_almost_full` it's noise, there's no alarms)

### Steps to reproduce

Run a child chain instance, do `:observer.start()`, go to "Applications", select 
 `OMG.ChildChain.SyncSupervisor` and kill.

### Notes

 - `OMG.Watcher.SyncSupervisor` is a similar, except it does produce a related (but not explicit enough) log message about `BlockGetter` going down. I think same fix should apply there

InoMurko:
hey @pdobacz have you had another thought about this?
I would simply do what we've done for ethereum client monitoring.
If `OMG.ChildChain.Monitor` or `OMG.Watcher.Monitor` receive an EXIT message in
```
# there's a supervisor below us that did the needed restarts for it's children
  # so we just ignore the exit from the supervisor, if the alarm clears, we restart it
  def handle_info({:EXIT, _from, _reason}, state) do
    {:noreply, state}
  end
``` 
raise an alarm, `unhealthy_node`, and `Logger.error` the exit. And add the alarm type to the `Health` plug of each RPC applications, so that we cut the wire for new requests.

Now this really is unrecoverable error, that would likely happen if `EthereumEventListeners` or any if it's dependencies had a bug. Do we agree on that?

pdobacz:
> hey @pdobacz have you had another thought about this?

not in depth, no :(. So to regurgitate, the thing that I guess still bugs me is this:

We raise `alarm` on an unrecoverable error in code ran under the `Monitor`, while we crash the VM on a similar error in code ran outside this "protective umbrella". Where's the line drawn here?

I still have a hunch that we should fail louder, whenever broken code is executed. With an alarm, we're relying on a number of moving pieces (alarms themselves, `get_alarm` endpoint, datadog) to notify us about the condition, which might be anything, including cases when we don't want continue running at all.

InoMurko:
First batch done here: https://github.com/omisego/elixir-omg/pull/1027
Second part comes with events going to Datadog.

PR #1142, opened 2019-11-15
fix: fix, simplify and tidy exiting from piggyback events in OMG.State

pdobacz:
Found when #855 

## Overview

When doing ALD compatibility, I managed to introduce a bug, where `input` piggybacks would crash the child chain (a bit silently, b/c #1141, hence our integration tests didn't pick this up)

## Changes

- fix bug - now child chain will completetly ignore input piggybacking events (like it did before). The bug manifests with this being thrown from the child chain server, when one piggybacks on input:
```
(omg) lib/omg/state/core.ex:387: OMG.State.Core.find_utxo_matching_piggyback(%{eth_height: 576, event_signature: "InFlightExitInputPiggybacked(address,bytes32,uint16)".........
...
        ** (FunctionClauseError) no function clause matching in OMG.State.Core.find_utxo_matching_piggyback/2
...
```
- added an explicit `OMG.State.Core` test for this
- added a more explicit `ChildChain` integration test for this to be extra sure

## Testing

`mix test`

### DIFF
--- a/apps/omg_child_chain/test/omg_child_chain/integration/happy_path_test.exs
+++ b/apps/omg_child_chain/test/omg_child_chain/integration/happy_path_test.exs
@@ -209,6 +209,41 @@ defmodule OMG.ChildChain.Integration.HappyPathTest do
     assert {:ok, _} = submit_transaction(second_output_spend)
   end
 
+  @tag fixtures: [:alice, :in_beam_child_chain, :alice_deposits]
+  test "check in-flight exit input piggybacking is ignored by the child chain",
+       %{alice: alice, alice_deposits: {deposit_blknum, _}} do
+    # create transaction, submit, wait for block publication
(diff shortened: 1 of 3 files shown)


apps/omg_child_chain/test/omg_child_chain/integration/happy_path_test.exs
```diff
@@ -209,6 +209,41 @@ defmodule OMG.ChildChain.Integration.HappyPathTest do
     assert {:ok, _} = submit_transaction(second_output_spend)
   end
 
+  @tag fixtures: [:alice, :in_beam_child_chain, :alice_deposits]
+  test "check in-flight exit input piggybacking doesn't crash",
```
InoMurko: What does this test actually test? Just not crashing is a bit obscure definition of a test. Under which conditions?

pdobacz: ah, absolutely right, there's a better way to word it, fixing...

pdobacz:
CC @jarindr

Issue #1128, opened 2019-11-12
API ` /in_flight_exit.get_data` crashes causing high amount of logging


jbunce:
```
watcher-0 watcher 2019-11-12 05:32:11.728 [info] module=Phoenix.Logger function=phoenix_error_rendered/4 request_id=FdZTUg-b4_1fYRkABSMB trace_id=6798762404176944400 span_id=3860092244967034233 ⋅Converted error {:case_clause, {:error, :no_deposit_for_given_blknum}} to 500 response⋅
watcher-0 watcher 2019-11-12 05:32:11.730 [error] ⋅#PID<0.24608.6> running OMG.WatcherRPC.Web.Endpoint (cowboy_protocol) terminated
watcher-0 watcher Server: audit-e3c15df-watcher-rinkeby-01.omg.network:80 (http)
watcher-0 watcher Request: POST /in_flight_exit.get_data
watcher-0 watcher ** (exit) an exception was raised:
watcher-0 watcher     ** (CaseClauseError) no case clause matching: {:error, :no_deposit_for_given_blknum}
watcher-0 watcher         (omg_watcher) lib/omg_watcher/api/in_flight_exit.ex:102: OMG.Watcher.API.InFlightExit.find_single_input_data/2
watcher-0 watcher         (elixir) lib/enum.ex:3317: Enumerable.List.reduce/3
watcher-0 watcher         (elixir) lib/enum.ex:1990: Enum.reduce_while/3
watcher-0 watcher         (omg_watcher) lib/omg_watcher/api/in_flight_exit.ex:40: OMG.Watcher.API.InFlightExit.get_in_flight_exit/1
watcher-0 watcher         (omg_watcher_rpc) lib/web/controllers/in_flight_exit.ex:61: OMG.WatcherRPC.Web.Controller.InFlightExit.handle_txbytes_based_request/4
watcher-0 watcher         (omg_watcher_rpc) lib/web/controllers/in_flight_exit.ex:15: OMG.WatcherRPC.Web.Controller.InFlightExit.action/2
watcher-0 watcher         (omg_watcher_rpc) lib/web/controllers/in_flight_exit.ex:15: OMG.WatcherRPC.Web.Controller.InFlightExit.phoenix_controller_pipeline/2
watcher-0 watcher         (omg_watcher_rpc) lib/web/endpoint.ex:15: OMG.WatcherRPC.Web.Endpoint.instrument/4⋅
[...]

PR #1149, opened 2019-11-19
Allow perftesting all kinds of exits, as handled by the Watcher

### LINKED ISSUE #855: Complete security critical watcher under high load

pdobacz:
closes #855 

## Overview

This PR prepares the perf/load-testing tools to use when testing scenarios in #855.

## Changes

- can use scripts start valid/invalid IFEs, challenge them, challenge them invalidly, piggyback validly/invalidly, detect all this and react 
- tried to be as documented & readable to ease future transition into different test harness (`specs`?), and also not overdo it
- needed to make some cleanups around the `elixir-omg` Watcher API client(s), unfortunately, for separation of concerns' sake, some code got copied over, but I'm prefering decoupling to DRY here. Can discuss this, but OTOH, I don't want to over-talk/over-do this
- improved the use of `Task.async_stream` further so now it doesn't crash `iex` when ethereum txs fail to mine!
- introduced `throttle_ms` option to the transaction-sending perftest, allows to put mild, continuous load on the system

## Testing

```
mix test --include integration test/omg_performance
```

and/or try running the perftests as in `ByzantineEvents` module's docs

### DIFF
--- a/apps/omg_performance/lib/omg_performance/generators.ex
+++ b/apps/omg_performance/lib/omg_performance/generators.ex
@@ -61,9 +61,9 @@ defmodule OMG.Performance.Generators do
 
   Options:
     - :use_blocks - if not nil, will use this as the stream of blocks, otherwise streams from child chain rpc
-    - :no_deposit_spends - if true, will limit only to transactions that do ont spend deposits (see function with the
+    - :no_deposit_spends - if true, will limit only to transactions that do not spend deposits (see function with the
       same name for explanation)
-    - :sent_by - if not nil, will limit to txs sent
+    - :sent_by - if not nil, will limit to txs sent by this particular address
     - :take - if not nil, will limit to this many results
   """
   @spec stream_transactions([OMG.Block.t()]) :: [binary()]
--- a/apps/omg_performance/lib/omg_performance/generators.ex
+++ b/apps/omg_performance/lib/omg_performance/generators.ex
@@ -88,7 +88,8 @@ defmodule OMG.Performance.Generators do
     if opts[:take], do: Enum.take(transactions, opts[:take]), else: transactions
   end
 
-  # FIXME: whoops, we cannot open IFEs from included txs spending deposits. When fixed, remove this filter
+  # NOTE: whoops, we cannot open IFEs from included txs spending deposits. When fixed, remove this filter
+  #       https://github.com/omisego/elixir-omg/issues/1128
   defp no_deposit_spends?(txbytes) do
     txbytes
     |> Transaction.Signed.decode!()
--- a/apps/omg_performance/lib/omg_performance/generators.ex
+++ b/apps/omg_performance/lib/omg_performance/generators.ex
@@ -61,6 +61,9 @@ defmodule OMG.Performance.Generators do
 
   Options:
     - :use_blocks - if not nil, will use this as the stream of blocks, otherwise streams from child chain rpc
+    - :no_deposit_spends - if true, will limit only to transactions that do ont spend deposits (see function with the
+      same name for explanation)
+    - :sent_by - if not nil, will limit to txs sent
     - :take - if not nil, will limit to this many results
   """
   @spec stream_transactions([OMG.Block.t()]) :: [binary()]
@@ -70,7 +73,28 @@ defmodule OMG.Performance.Generators do
       |> if(do: opts[:use_blocks], else: stream_blocks())
       |> Stream.flat_map(& &1.transactions)
 
+    transactions =
+      if(opts[:no_deposit_spends],
+        do: Stream.filter(transactions, &no_deposit_spends?/1),
+        else: transactions
+      )
+
+    transactions =
+      if(opts[:sent_by],
+        do: Stream.filter(transactions, &is_sent_by?(&1, opts[:sent_by])),
+        else: transactions
+      )
+
     if opts[:take], do: Enum.take(transactions, opts[:take]), else: transactions
+  end
+
+  # FIXME: whoops, we cannot open IFEs from included txs spending deposits. When fixed, remove this filter
+  defp no_deposit_spends?(txbytes) do
+    txbytes
+    |> Transaction.Signed.decode!()
+    |> Transaction.get_inputs()
+    |> Enum.any?(&Utxo.Position.is_deposit?/1)
+    |> Kernel.not()
   end
 
   @doc """
@@ -102,6 +126,18 @@ defmodule OMG.Performance.Generators do
     # interval <= blknum <= mined_block
     blknum = :rand.uniform(div(mined_block, interval)) * interval
     get_block!(blknum, child_chain_url)
+  end
+
+  defp is_sent_by?(txbytes, address) do
+    %Transaction.Recovered{witnesses: witnesses} = tx = Transaction.Recovered.recover_from!(txbytes)
+    # runtime assertion if we're not using this for non-Payment txs. The next line might not work if witnesses are
+    # not addresses
+    %Transaction.Payment{} = tx.signed_tx.raw_tx
+
+    witnesses
+    |> Map.values()
+    |> Enum.uniq()
+    |> Kernel.==([address])
   end
 
   defp generate_user(opts) do
--- a/apps/omg_performance/lib/omg_performance/http_rpc/watcher_client.ex
+++ b/apps/omg_performance/lib/omg_performance/http_rpc/watcher_client.ex
@@ -14,7 +14,15 @@
 
 defmodule OMG.Performance.HttpRPC.WatcherClient do
   @moduledoc """
-  Provides access to Watcher's RPC API
+  Provides access to Watcher's RPC API, as required by the `OMG.Performance` tool
+
+
+  TODO: This module includes a big fat copy paster from `Support.WatcherHelper` - do sth about this later
+  We need an exact set of functionalities from the `omg_performance`'s Watcher client, but they're executed
+  differently, because in here we're calling the Watcher's API on a slightly different level (full HTTP stack)
+  Differences:
+   - atoms not strings as keys
+   - don't have the many helpers to assert `success?`/`no_success?` etc., success is always expected
   """
 
   alias OMG.Utils.HttpRPC.ClientAdapter
@@ -23,9 +31,9 @@ defmodule OMG.Performance.HttpRPC.WatcherClient do
 
   require Utxo
 
-  #
-  # TODO: here begins a big fat copy paster from `watcher_helper.ex` - do sth about this later
-  #
+  @doc """
+  Copied from `Support.WatcherHelper`, see moduledoc
+  """
   def get_balance(address, token) do
     encoded_token = Encoding.to_hex(token)
 
@@ -35,42 +43,66 @@ defmodule OMG.Performance.HttpRPC.WatcherClient do
     |> Map.get("amount")
   end
 
+  @doc """
+  Copied from `Support.WatcherHelper`, see moduledoc
+  """
   def get_utxos(address) do
-    success?("/account.get_utxos", %{:address => Encoding.to_hex(address)})
+    call("/account.get_utxos", %{:address => Encoding.to_hex(address)})
   end
 
+  @doc """
+  Copied from `Support.WatcherHelper`, see moduledoc
+  """
   def get_exitable_utxos(address) do
-    success?("/account.get_exitable_utxos", %{:address => Encoding.to_hex(address)})
+    call("/account.get_exitable_utxos", %{:address => Encoding.to_hex(address)})
   end
 
+  @doc """
+  Copied from `Support.WatcherHelper`, see moduledoc
+  """
   def get_balance(address) do
-    success?("/account.get_balance", %{:address => Encoding.to_hex(address)})
+    call("/account.get_balance", %{:address => Encoding.to_hex(address)})
   end
 
+  @doc """
+  Copied from `Support.WatcherHelper`, see moduledoc
+  """
   def get_exit_data(blknum, txindex, oindex),
     do: get_exit_data(Utxo.Position.encode(Utxo.position(blknum, txindex, oindex)))
 
+  @doc """
+  Copied from `Support.WatcherHelper`, see moduledoc
+  """
   def get_exit_data(encoded_position) do
-    data = success?("utxo.get_exit_data", %{utxo_pos: encoded_position})
-    decode_response(data, [:txbytes, :proof])
+    data = call("utxo.get_exit_data", %{utxo_pos: encoded_position})
+    ClientAdapter.decode16(data, [:txbytes, :proof])
   end
 
+  @doc """
+  Copied from `Support.WatcherHelper`, see moduledoc
+  """
   def get_exit_challenge(utxo_pos) do
-    data = success?("utxo.get_challenge_data", %{utxo_pos: utxo_pos})
+    data = call("utxo.get_challenge_data", %{utxo_pos: utxo_pos})
 
-    decode_response(data, [:exiting_tx, :txbytes, :sig])
+    ClientAdapter.decode16(data, [:exiting_tx, :txbytes, :sig])
   end
 
+  @doc """
+  Copied from `Support.WatcherHelper`, see moduledoc
+  """
   def get_in_flight_exit(transaction) do
-    exit_data = success?("in_flight_exit.get_data", %{txbytes: Encoding.to_hex(transaction)})
+    exit_data = call("in_flight_exit.get_data", %{txbytes: Encoding.to_hex(transaction)})
 
-    decode_response(exit_data, [:in_flight_tx, :input_txs, :input_txs_inclusion_proofs, :in_flight_tx_sigs])
+    ClientAdapter.decode16(exit_data, [:in_flight_tx, :input_txs, :input_txs_inclusion_proofs, :in_flight_tx_sigs])
   end
 
+  @doc """
+  Copied from `Support.WatcherHelper`, see moduledoc
+  """
   def get_in_flight_exit_competitors(transaction) do
-    competitor_data = success?("in_flight_exit.get_competitor", %{txbytes: Encoding.to_hex(transaction)})
+    competitor_data = call("in_flight_exit.get_competitor", %{txbytes: Encoding.to_hex(transaction)})
 
-    decode_response(competitor_data, [
+    ClientAdapter.decode16(competitor_data, [
       :in_flight_txbytes,
       :competing_txbytes,
       :competing_sig,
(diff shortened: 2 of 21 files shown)


apps/omg_utils/lib/omg_utils/http_rpc/client_adapter.ex
```diff
@@ -56,12 +58,44 @@ defmodule OMG.Watcher.HttpRPC.Adapter do
...
+        bin_list =
+          value
+          |> Enum.map(&Encoding.from_hex/1)
+          |> Enum.map(fn {:ok, bin} -> bin end)
+
+        {key, bin_list}
+    end
+  end
+
+  defp decode_binary!(value) do
```
pnowosie: Name is suggesting it's something else than `decode16` but it's not...
Thinking about "matching" name: `decode16_value!` :question:

apps/omg_utils/test/omg_utils/http_rpc/client_adapter_test.exs
```diff
@@ -0,0 +1,87 @@
...
+
+    test "is safe and fails when asked to decode anything else than hex-encoded values" do
+      expected = %{
+        "not_bin1" => 0,
+        "not_bin2" => :atom,
+        "not_hex" => "string",
+        "not_value" => nil
+      }
+
+      assert_raise CaseClauseError, fn ->
```
pnowosie: I wonder how easy is to connect `CaseClauseError` with `try to decode incorrect value`
If it isn't smth that requires serious digging - then never mind

apps/omg_performance/lib/omg_performance/generators.ex
```diff
@@ -61,6 +61,9 @@ defmodule OMG.Performance.Generators do
 
   Options:
     - :use_blocks - if not nil, will use this as the stream of blocks, otherwise streams from child chain rpc
+    - :no_deposit_spends - if true, will limit only to transactions that do ont spend deposits (see function with the
```
InoMurko:
```suggestion
    - :no_deposit_spends - if true, will limit only to transactions that do not spend deposits (see function with the
```

apps/omg_performance/lib/omg_performance/generators.ex
```diff
@@ -70,9 +73,31 @@ defmodule OMG.Performance.Generators do
...
+    transactions =
+      if(opts[:sent_by],
+        do: Stream.filter(transactions, &is_sent_by?(&1, opts[:sent_by])),
+        else: transactions
+      )
+
     if opts[:take], do: Enum.take(transactions, opts[:take]), else: transactions
   end
 
+  # NOTE: whoops, we cannot open IFEs from included txs spending deposits. When fixed, remove this filter
```
InoMurko: I'm missing some background buy I'm having troubles understanding what this means and what the function does... I presume you don't want to spend UTXOs that were just deposited?
and you cannot IFE just deposited UTXOs?

apps/omg_performance/lib/omg_performance/generators.ex
```diff
@@ -61,6 +61,9 @@ defmodule OMG.Performance.Generators do
 
   Options:
     - :use_blocks - if not nil, will use this as the stream of blocks, otherwise streams from child chain rpc
+    - :no_deposit_spends - if true, will limit only to transactions that do ont spend deposits (see function with the
+      same name for explanation)
+    - :sent_by - if not nil, will limit to txs sent
```
InoMurko: hm... elaborate what this flag does?

apps/omg_performance/lib/omg_performance/http_rpc/watcher_client.ex
```diff
@@ -14,72 +14,167 @@
 
 defmodule OMG.Performance.HttpRPC.WatcherClient do
   @moduledoc """
-  Provides access to Watcher's RPC API
+  Provides access to Watcher's RPC API, as required by the `OMG.Performance` tool
+
+
+  TODO: This module includes a big fat copy paster from `Support.WatcherHelper` - do sth about this later
```
InoMurko: no need to couple them, have a separate implementation if the requirements are different.

apps/omg_performance/lib/omg_performance/http_rpc/watcher_client.ex
```diff
@@ -14,72 +14,167 @@
 
 defmodule OMG.Performance.HttpRPC.WatcherClient do
```
InoMurko: too bad this is implemented this way. you get a client implementation in elixir for free if you'd use swagger API definition!

apps/omg_performance/test/omg_performance/byzantine_event_test.exs
```diff
@@ -47,6 +47,17 @@ defmodule OMG.Performance.ByzantineEventsTest do
     |> ByzantineEvents.get_many_standard_exits()
   end
 
+  @tag fixtures: [:perf_test, :mix_based_child_chain, :mix_based_watcher]
+  test "can provide timing of response when asking for IFE data", %{perf_test: {:ok, %{destdir: destdir}}} do
```
InoMurko: what does the test trying to prove? 
`can provide timing of response when asking for IFE data`

apps/omg_performance/test/omg_performance/byzantine_event_test.exs
```diff
@@ -62,7 +73,26 @@ defmodule OMG.Performance.ByzantineEventsTest do
 
     :ok = ByzantineEvents.watcher_synchronize(root_chain_height: last_exit_height)
     # assert that we can call this testing function reliably and that there are no invalid exits
-    assert [] = ByzantineEvents.get_byzantine_events("invalid_exit")
+    assert [] = ByzantineEvents.get_byzantine_events()
+  end
+
+  @tag fixtures: [:perf_test, :mix_based_child_chain, :mix_based_watcher]
+  test "can provide timing of status.get under many valid IFEs for included txs",
```
InoMurko: what does the test trying to prove? 
`can provide timing of status.get under many valid IFEs for included txs`

apps/omg_performance/test/omg_performance/byzantine_event_test.exs
```diff
@@ -83,6 +113,144 @@ defmodule OMG.Performance.ByzantineEventsTest do
     assert Enum.count(ByzantineEvents.get_byzantine_events("invalid_exit")) > 10
   end
 
+  @tag fixtures: [:perf_test, :mix_based_child_chain, :mix_based_watcher]
```
InoMurko: these tests bellow need a bit more data for me ... I can't digest them because I don't know what the plan is and what they test.

apps/omg_performance/test/omg_performance/byzantine_events/mutations_test.exs
```diff
@@ -0,0 +1,45 @@
...
+  @eth OMG.Eth.RootChain.eth_pseudo_address()
+
+  setup do
+    alice = OMG.TestHelper.generate_entity()
+    {:ok, %{alice: alice}}
+  end
+
+  describe "mutate_txs/2" do
+    test "should produce a different signed transaction every time", %{alice: alice} do
+      tx = OMG.TestHelper.create_encoded([{1000, 0, 0, alice}], @eth, [{alice, 1}])
```
InoMurko: gaahh I hate these helper functions. they're really good at hiding what they're doing. bad naming doesn't help either!

apps/omg_utils/lib/omg_utils/http_rpc/client_adapter.ex
```diff
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
InoMurko: same comment as above. this is a really wasteful implementation considering we have full swagger specifications we could utilize. that's what they're for!

apps/omg_utils/lib/omg_utils/http_rpc/client_adapter.ex
```diff
@@ -56,12 +58,44 @@ defmodule OMG.Watcher.HttpRPC.Adapter do
 
   def get_response_body(error), do: {:error, {:client_error, error}}
 
+  @doc """
+  Decodes specified keys in map from hex to binary
+  """
+  @spec decode16(map(), list()) :: map()
```
InoMurko: where is the requirement for this coming from?

apps/omg_performance/lib/omg_performance/generators.ex
pdobacz: whoops, part of the sentence chopped off

apps/omg_performance/lib/omg_performance/generators.ex
pdobacz: the latter. There's a bug #1128 as mentioned here, that prevents one from doing IFE from an tx spending a spent deposit.

So the function just cuts those failing txs out from the batch used. When #1128 is fixed this can be removed (I'll make a note to not forget about this)

apps/omg_performance/lib/omg_performance/http_rpc/watcher_client.ex
pdobacz: I'll change `TODO` to `NOTE` then, if >1 person is comfortable with this :)

apps/omg_performance/lib/omg_performance/http_rpc/watcher_client.ex
pdobacz: yyyyyeahh, considering this already exists, would you say it's better to try to swap out for the swagger client or meh? How much effort do you think it'd take?

apps/omg_performance/lib/performance.ex
```diff
@@ -23,11 +23,17 @@ defmodule OMG.Performance do
...
       alias OMG.Performance.ExtendedPerftest
       alias OMG.Performance.Generators
+      alias OMG.Performance.HttpRPC.WatcherClient
 
       import Performance, only: [timeit: 1]
       require Performance
 
       use OMG.Utils.LoggerExt
+
+      {:ok, _} = Application.ensure_all_started(:briefly)
```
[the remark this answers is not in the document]
pdobacz: yeah, I tried that, it just wouldn't start for some reason. I tried a lot of stuffs, but got tired so there it is. If anyone manages to start `:briefly` in tests without any occurrence of `Application.ensure_all_started(:briefly)` in the code, LMK, I might have missed sth obvious

apps/omg_performance/lib/performance.ex
```diff
@@ -79,24 +82,24 @@ defmodule OMG.Performance do
     watcher_url =
       System.get_env("WATCHER_URL") || Application.get_env(:omg_performance, :watcher_url, "http://localhost:7434")
 
+    # Needed here to have some value of address when `:contract_address` is not set explicitly
+    # required by the EIP-712 struct hash code
+    contract_addr =
+      Application.get_env(:omg_eth, :contract_addr, %{plasma_framework: "0x0000000000000000000000000000000000000001"})
```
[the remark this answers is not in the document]
pdobacz: no, that was a case of test values trickling into production code so I got rid of this at one point.

apps/omg_performance/test/omg_performance/byzantine_event_test.exs
pdobacz: so the tooling under test is aimed to (among others) provide timing of stuffs (you execute a command and measure time it takes, somehow). So if code in this test works, you're able to do the "provide timing" part, that's where the title is coming from. Considering this, if this wording is confusing to you, would something simpler like `can ask for many IFE data` work better? This alternative is more in line with what the test actually runs, less in line with how the tested tool is intended to be used.

apps/omg_performance/test/omg_performance/byzantine_event_test.exs
pdobacz: see earlier comment, if we decide on the wording, I'll reword consistently

apps/omg_performance/test/omg_performance/byzantine_event_test.exs
pdobacz: falls under an earlier comment, let's agree on how to describe the tests there and I'll update all.

apps/omg_performance/test/omg_performance/byzantine_events/mutations_test.exs
pdobacz: would `OMG.TransactionHelper.create_encoded` help? Mind that their intention is to provide a short-hand way to create transactions needed to test various things, so the syntax is concise by intent.

anyway, I'd suggest taking this to a separate PR.

apps/omg_utils/lib/omg_utils/http_rpc/client_adapter.ex
pdobacz: let's take it to that other comment's thread (and let's be constructive about it. I actually don't know whether swagger specs were in place, and decided they're staying, when this `Client` was first added, so this still existing might be a case of inertia. No need to judge, if we can improve this, we will.)

apps/omg_utils/lib/omg_utils/http_rpc/client_adapter.ex
pdobacz: We need to deep-decode hex values from the json responses that the RPC clients are giving us. Also note that this code is being moved.

apps/omg_utils/test/omg_utils/http_rpc/client_adapter_test.exs
pdobacz: the connection is, that it is expected that such clauses aren't implemented at all and throw. I think anything more elaborate would require digging. Currently - it is what it is - unimplemented case.

InoMurko:
@achiurizo @pdobacz  what should we do with this test? Thought it tests all the valid things we absolutelly need to do, and kudos for that @pdobacz ! But... I don't like the approach. It's too involved with the core implementation. In my taste, performance "tools" or tests, need to be run against a deployed environment that looks like a deployment so that you can extrapolate the results. Mimicking childchain and watcher by spawning them in the background `:mix_based_child_chain` and `:mix_based_watcher` does not resemble a deployed environment, making the results of the tests... invalid.

apps/omg_performance/lib/omg_performance/http_rpc/watcher_client.ex
InoMurko: It's not too difficult (look at `priv/`) but I'm not sure how to get this into omg_perfromance...

pdobacz:
@InoMurko 

> In my taste, performance "tools" or tests, need to be run against a deployed environment that looks like a deployment so that you can extrapolate the results. Mimicking childchain and watcher by spawning them in the background `:mix_based_child_chain` and `:mix_based_watcher` does not resemble a deployed environment, making the results of the tests... invalid.

Just to clarify this one quickly: it is **only** the test of the test that runs against the stuff spawned in the background. When you  actually test to see how it performs, you point it at whatever you like (by specifying urls and contract addresses)

apps/omg_performance/lib/omg_performance/http_rpc/watcher_client.ex
pdobacz: OK, since this got looped into this repo under `priv` it seems like the path is much more paved now. Will `priv` be visible and accessible from `omg_performance` level though?

Another question is whether we want to plunge into this now kinda "separately", or maybe when those tests get rephrased into the `specs` language (or whatever approach is used for the integration tests).

InoMurko:
> @InoMurko
> 
> > In my taste, performance "tools" or tests, need to be run against a deployed environment that looks like a deployment so that you can extrapolate the results. Mimicking childchain and watcher by spawning them in the background `:mix_based_child_chain` and `:mix_based_watcher` does not resemble a deployed environment, making the results of the tests... invalid.
> 
> Just to clarify this one quickly: it is **only** the test of the test that runs against the stuff spawned in the background. When you actually test to see how it performs, you point it at whatever you like (by specifying urls and contract addresses)

Aha!!! That's perfect! What we can do then is to make these perf tests run during cirleci jobs `test_barebone_release` and `test_docker_compose_release` release right?

pdobacz:
> What we can do then is to make these perf tests run during cirleci jobs `test_barebone_release` and `test_docker_compose_release` release right?

as in the nightly builds? yeah, we could. My usual doubt there is how do you gauge the assertions, since the jobs would (pardon my ignorance, if that's not the case) run in a uncontrolled environment and timings can vary wildly and => flaky tests.

The initial approach was to cater for being run from a local setup (possibly against non-local services ofc) on demand.

But if there's a way to make the perf tests results useful this way or another in the CI, I'm very supportive.

InoMurko:
> > What we can do then is to make these perf tests run during cirleci jobs `test_barebone_release` and `test_docker_compose_release` release right?
> 
> as in the nightly builds? yeah, we could. My usual doubt there is how do you gauge the assertions, since the jobs would (pardon my ignorance, if that's not the case) run in a uncontrolled environment and timings can vary wildly and => flaky tests.
> 
> The initial approach was to cater for being run from a local setup (possibly against non-local services ofc) on demand.
> 
> But if there's a way to make the perf tests results useful this way or another in the CI, I'm very supportive.

I'm missing something...How does the code in this PR gets tested if not in CI? And why are they not "flaky"?

pdobacz:
> I'm missing something...How does the code in this PR gets tested if not in CI? And why are they not "flaky"?

OK, so let me reiterate, and pls correct my terminology, if it's confusing. There's three kinds of things in play here:

1. the **perf test machinery**, that this PR expands. Basically that's a set of convenience functions wrapping various API calls allowing stuff to be called many, many times
2. the **test of the perf test machinery**, which executes a very small-sized perf test. It runs quickly, we don't care about the "results" (timing etc), only whether it passes. This tells us whether the machinery works and is ran by CI. That part would probably end up moving to wherever the current `:integration` tests are moved. 
3. the actual **act of perf testing**. That _currently_ is not ran by the CI. The intent is that it is fired manually by someone interested and against a local or non-local deployment of things (see :point_up: ). 

So to answer your question: code in this PR is tested in CI (2.) and is not flaky because the test cases are smallish and we're not exposed to flakiness coming from uncontrollable environment - we're _not_ measuring performance in those tests (2.)

Now, going back, how I understood your suggestion:

> What we can do then is to make these perf tests run during cirleci jobs `test_barebone_release` and `test_docker_compose_release` release right?

is to start doing (3.) in CI and my reply in the previous commit concerned this specifically.

InoMurko:
My final thoughts: we need to remove these "non flaky" unit tests from the codebase, make perf tests run against `test_barebone_release` and `test_docker_compose_release`. 
That should allow us to delete `OMG.Performance.Fixtures`
cc @achiurizo

pdobacz:
Ah, an old thread. The most recent comment must've landed during my sick-leave in Dec and the discussion stalled, sorry about this.

Let's reread the takeaways so far :point_up: and pick this up.

On my end, shelving this stalls #855 (actually doing the perf testing) further. I am conscious of the concerns with approach, but I'd again opt for delivering an imperfect tool now and clean-up/engineer (change approach to testing the tool, refactor) at a later stage.

The testing approach, which I suppose you're not OK with in saying:

>  remove these "non flaky" unit tests from the codebase

is the same as with "old" integration tests. When those get migrated to `whitebread`/`cabbage`, migrating `omg_perfomance` tests will follow the same path.

Having tests like this _early_ allows us to pick up issues like https://github.com/omisego/security-issues/issues/7. From some preliminary experiments, I've seen there are more take-aways like these to gather from these tests and, in my opinion, we (or the perf/loadtest team) could leverage this now.

pdobacz:
Closing as the work here done has become stale, considering the directions taken with the perf testing effort.

PR #1139, opened 2019-11-15
Bump plasma-contracts SHA to 1c7cd867

unnawut:
Closes #1138

## Overview

Upgrade plasma-contracts to SHA `1c7cd867` and related changes to align with the updated contract interfaces.

## Changes

- Upgrade plasma-contracts to `1c7cd867b1353ae72c825a6d15b2d382689f30ad`
- Add `OMG.Eth.Encoding.encode_constructor_params/1`
- Add `OMG.Eth.Encoding.ContractConstructor.extract_params/1`

Align with latest plasma-contracts' interfaces
- Start standard exits uses `uint256` for `utxo_pos`
- `EthDepositVerifier` deployment requires `transaction_type` and `output_type`
- `Erc20DepositVerifier` deployment requires `transaction_type` and `output_type`
- `EthVault` deployment requires `safe_gas_stipend`
- `Erc20Vault` deployment requires `safe_gas_stipend`
- `PaymentExitGame` deployment requires `safe_gas_stipend`

## Testing

- `mix test` should pass
- `mix test --only common` should pass
- `mix test --integration` should pass
- `docker-compose up` should be successful

Issue #1041, opened 2019-10-11
Leverage exw3 or similar for convenient interface to contracts


pdobacz:
There have been some efforts to leverage https://github.com/hswick/exw3, instead of maintaining the `Eth.RootChain`. Since now the wealth of the contract's APIs is significantly increasing, this maintenance becomes more and more of a hassle.

Let's reconsider this issue. First let's re-assess exw3 and alternatives (if there are any).

Old notes and intermediate results from @pik694 (pasted from pivotal tracker, note this is ~year old):

>
> @pdobacz
> https://github.com/hswick/exw3/ might solve this well enough, most likely we shouldn't do more than use this (in case it works well)
>
> @pik694
> The decision was made to split this task into two steps.
>
> 1. Rewrite OMG.Eth.RootChain in order to use exw3. This step is a test, whether exw3 offers all the required functionality (ex. indexed fields in events).
> 1. Prepare macros, that would use exw3 and would produce all the code during compilation time, basing on contract's abi.
>
> For now, we managed to refactor the OMG.Eth and OMG.Eth.RootChain to a point where we are able to deploy (needed for testing) the contract via exw3 and the contract registers itself to exw3 - having given the address and abi.
>
> Now we are refactoring functions which call the contract - exw3 seems to be doing the work. Refactor of a one call is just a simple exchange of our previous call into a call of exw3 method - pretty straightforward. We managed to refactor 3-4 functions.
>
> Next step is to rewrite all the remaining calls, then adapt exw3 event's handling.
> Some clean up is also needed if it is decided to merge it the the main development.
>
> @pdobacz
> WIP in branch https://github.com/omisego/elixir-omg/tree/feature/omg-66_macroify_creation_of_contract_proxy_modules

unnawut:
I believe this is taken care of by https://github.com/omgnetwork/ex_plasma

PR #1152, opened 2019-11-20
fix: standard exit struct specs

InoMurko:
I was working on: https://github.com/omisego/security-issues/issues/7 when I noticed this 🗡 

## Overview

https://github.com/omisego/plasma-contracts/blob/a360d5f68929f1f727104a7b4ab2e63bbc8a4327/plasma_framework/contracts/src/exits/payment/PaymentExitDataModel.sol#L21

## Changes

- uint size incorrect on elixir side

## Testing

/


unnawut:
I should've caught this in #1139 😅

pdobacz:
#1041 would help here, waiting for that courageous person to pick up and continue the effort...

PR #1175, opened 2019-12-01
Eth client snapshot

InoMurko:
## Overview

This PR is the beginning of decoupling of elixir-omg and plasma contracts. 
A snapshot of geth DB gets created immediately after the deploy of contracts, here: https://github.com/omisego-images/docker-elixir-omg/pull/9/files
with a genesis.json defined as https://github.com/omisego-images/docker-elixir-omg/blob/45ea9df3926f7da1ae6f3b985c6537556653da90/contracts/data/geth/genesis.json

When you bump contracts, you would bump them in first in docker-elixir-omg, which will spit out a snapshot location URL, which you place in the `Makefile:init_test`, if addresses of contract changed, you need to swap them in `docker-compose.yml` and in `apps/omg_eth/test/fixtures.exs`.

## Changes

- `--dev` on geth with `--datadir` does not permit snapshots
- an environment (local) gets initilised for testing with `make init_test` that pulls in the snapshot and unpacks it. all unittests always create a copy of the DB with briefly, so it remains unaffected.
- `DependencyConformance.Signature` tests were exvcr recorded because they relied on `PaymentEip712LibMock` contract. 
- Fixture `apps/omg_eth/test/fixtures.exs` does not deploy contracts anymore. they're hardcoded. 
- parity code was removed because it's unused
- whitebread tests are hardened 

## Testing

Same tests apply!

# Next step, remove the remaining deployments - like ERC20 token contracts!


apps/omg/test/omg/dependency_conformance/signature_test.exs
```diff
@@ -18,93 +18,117 @@ defmodule OMG.DependencyConformance.SignatureTest do
...
+    ExVCR.Config.cassette_library_dir(vcr_path)
 
-    {:ok, _, signtest_addr} = Deployer.create_new("PaymentEip712LibMock", root_path, Eth.Encoding.from_hex(addr), [])
+    :ok
+  end
 
-    # impose our testing signature contract wrapper (mock) as the validating contract, which normally would be
-    # plasma framework
+  setup_all do
+    %{plasma_framework: plasma_framework} = Application.get_env(:omg_eth, :contract_addr)
```
pnowosie: There could be more key-values in this map. The way you restore it after test, only `plasma_framework` gets restored.
```suggestion
    contract_addresses = Application.get_env(:omg_eth, :contract_addr)
```
and restore :point_down:

apps/omg/test/omg/dependency_conformance/signature_test.exs
```diff
@@ -18,93 +18,117 @@ defmodule OMG.DependencyConformance.SignatureTest do
...
     end)
 
     [contract: signtest_addr]
   end
 
-  test "signature test empty transaction", context do
-    contract = context[:contract]
-    tx = TestHelper.create_signed([], [])
-    verify(contract, tx)
+  test "signature with empty transaction", context do
```
pnowosie: If you'd return a map from setup_all you could patter-match on it safely
```suggestion
  test "signature with empty transaction", %{contract: contract} do
```

apps/omg_eth/test/support/root_chain_helper.ex
```diff
@@ -24,7 +24,7 @@ defmodule Support.RootChainHelper do
 
   import OMG.Eth.Encoding, only: [to_hex: 1, from_hex: 1]
 
-  @tx_defaults Eth.Defaults.tx_defaults()
+  @tx_defaults OMG.Eth.Defaults.tx_defaults()
```
pnowosie: :question:

priv/apps/itest/lib/account.ex
```diff
@@ -2,15 +2,19 @@ defmodule Itest.Account do
...
+
   alias Itest.Transactions.Currency
-  @accounts_path "../data/ganache/data/account_keys.txt"
-  @accounts_state '/tmp/used_accounts'
+  alias Itest.Transactions.Encoding
+  import Itest.Poller, only: [wait_on_receipt_confirmed: 2]
+
+  @retry_count 60
 
+  # checksummed!!!
```
pnowosie: I don't follow what you mean by `checksummed!!!`

apps/omg_eth/test/support/root_chain_helper.ex
InoMurko: I don't like these partial aliases, because it's difficult to find where a module is used!

priv/apps/itest/lib/account.ex
InoMurko: ugh, yeah, sorted in the next commit!

priv/features/standard_exits.feature
```diff
@@ -4,4 +4,4 @@ Feature: Standard Exits
     Then Alice should have "1" ETH on the network
     When Alice starts a standard exit on the network
     Then Alice should have "0" ETH on the network after finality margin
-    And Alice should have "100000" ETH on the blockchain
```
InoMurko: because the used accounts are not prefunded with 100 eth!

pdobacz:
My general take on this is that it is a step taken in the right direction (and huge :clap: for putting this together), however I think it might be going either too far or too fast.

Not going into details yet, here are my general thoughts:
1. this decreases the ease to develop things related to contracts changing. This is still a rapid process and the "you would bump them in first in docker-elixir-omg..." part seems like it would really make the development cycle slow and error prone. It is very useful to be free to choose the `plasma-contracts` revision (or even edit contract code if exploring) and have those changes reflect in `elixir-omg` behavior and tests, without going through all this. Imagine a situation, when you're about to bump contracts and since your `mix.lock` they got 5 big commits. You should be free to choose how you tackle that (all 5 at once or 1-by-1 etc.) end then follow this chosen path without this hassle
2. this hinders the `propcheck` based testing effort, since you can't exvcr those. We just got off a call with the https://github.com/orgs/omisego/projects/23 team and sentiment is that such tests are useful, considering recent findings about encoding etc. If this PR got merged, `propcheck` tests would need to find a new home in a different repo, that would be capable of deploying the proxy contracts (like `PaymentEip712LibMock`; it's not a mock really). While "finding new home" might be 100% justified, it will delay things.
3. it will impact timelines. We're preparing for "you-know-what" now and there's tons of things to sort out. So assuming this change has a positive net impact in the long run, it might take us a couple of iterations to polish it out and get used to it, and that will clash with other efforts underway.

Now to reach out, can you enumerate what are you actually trying to achieve here (more specific than "decoupling", which we agree on), and in what timeline?

What I would personally love to be able to do, and what this PR accomplishes is:
 - get rid of `deployer(s).ex` in favor of anything `plasma-contracts` inherently uses, like `truffle` today - no questioning of this, `deployers` were a temporary solution from day 1
 - run integration tests quicker using tools like exvcr, but only with the option to revert to running without exvcr, easily, against a mold-able `plasma-contracts` code

Hence my question about your specific goals - what beyond this list just above you're trying solve and when?

InoMurko:
1. I don't think there's anything like "rapid" contract development. I think you personally know more about this, since you've integrated ALD. But nothing like that (in size) is on the roadmap. And if it were, _the rules of the game_ are now different. We integrate towards a very precise specification document, and after that, we provide end2end tests. There's unit tests that will support your **implementation**, but not your **integration**! For (integration) end2end tests, there's really minimal things you need to do, to point to a new contract set.
2. Propcheck tests should not be used (in my book) to figure out whether two systems are fully compatible. If systems A uses RLP library and system B uses another RLP library, than compatibility of the systems needs to work on that level, no need to check if systems work together. RLP compatibility has nothing to do with plasma, per se, it's a serialization library.
3. Indeed we are preparing, but thats exactly why we're doing this. To cleanup and have good solid practices of testing that will lift our confidence in testing.

What I was addressing and what I'm solving with this:
- step 1 in reducing the test code (not tests, but code that supports testing but isn't tested)
- reduce complexity of the test code 
- reduce scope of the code (it was reaching into other systems, rather than work with them). limiting responsibility limits complexity and increases simplicty

boolafish:
> I don't think there's anything like rapid contract development

Not sure what "rapid" is defined here, but at least currently there is always still some refactoring, bug fixing and would potentially lead to interface or interaction changes. (eg. recently we add a new function to patch the protocol hole on IFE bond returning)

Also, the goal of ALD is to make adding new feature possible so I would imagine new feature being added in the future.

> We integrate towards a very precise specification document, and after that, we provide end2end tests.

Let's have a discussion on how to make this happen (probably off GH and in a call?) as a team. I understand the integration process is painful during this iteration, but at the same time, we (or at least I) would need to learn what is the better development process to make this happen. 


> Propcheck tests should not be used (in my book) to figure out whether two systems are fully compatible

I don't have much experience on prop check so I would not really comment on this. I guess the alternative is to do prop check testing against a spec that what the encode/decode/signature of transaction should look like.

However, I have to admit recent effort of Piotr's prop tests do make us aware some details of the spec. eg. https://github.com/omisego/plasma-contracts/issues/500. This looks trivial, however, test of invalid cases (not decoding an array) would easily to be missed.

Though it is honest not a great way to discover by checking the result divergent of the two implementations. And quite likely there are still things not covered. Just wondering what are better alternative to discover those.

pnowosie:
> If systems A uses RLP library and system B uses another RLP library, than compatibility of the systems needs to work on that level, no need to check if systems work together.

Unfortunately it isn't so easy. Prop check as a part of conformance test checks the compatibility of Eip-712 hashing/signing implementations. RLP is unavoidable to call code contract code.
It might be that they together causing issue - that cannot be detected in separation (or be harder to find).

Also preparing separate test environment is an effort we might avoid. 
And Eip-712 structure will change to match new transaction types.

InoMurko:
@boolafish the recent effort of plasma-contracts integration doc is already been proven as _the way to go_. Incredibly helpful. And as we go, we refine the process of documentation even further.  The omisego/plasma-contracts#500 issue comes more from lack of extensive testing on plasma-contracts. Especially the *negative* cases like in the issue.

@pnowosie @pthomalla now we're talking about RLP. The way propecheck works is that it's trying to create as many variations of your model as it can, and when it breaks, it tries to reduce the problem to the minimum. To help you debug the issue.  I think what @pthomalla you're saying is, that the best way to find our problems, is to use our transaction as a model to generate new RLPs. Correct? What me and @achiurizo are saying is that to RLP and to RLP decoding on solidity, "transaction models" are irrelevant, because the issue is greated than that. It's https://github.com/mana-ethereum/ex_rlp/pull/24 `choose optimal length encoding`. Ayrat pointed out: https://github.com/mana-ethereum/ex_rlp/pull/24#issuecomment-557862347

boolafish:
> comes more from lack of extensive testing on plasma-contracts. Especially the negative cases like in the issue.

probably a chicken egg problem. Definitely that is the cause, but then the problem would be how do we figure out we are missing invalid test cases from the beginning...

Intuitively I don't think conformance is a good nor reliable approach *but* at least it does prove that it find some issue. But this might be off topic, and we can do a one time conformance test to try to figure our problem and then add invalid test cases for what we have found.

I am personally not a huge fan of conformance test as well, and I like your proposal and approach in general. But just want to make sure we have something that can cover the work that the conformance tests as it has proven that it does have value.


>  RLP and to RLP decoding on solidity, "transaction models" are irrelevant

Yes. But a huge gap of the RLP meltdown we were solving on contract is actually from the step of RLP -> transaction struct. Most (new) issues are from RLP, by spec, only defines "string" (well...and array) but our struct needs to decode to "uint", "bytes32", "address"..... So how do we limit 1:1 from string -> our custom field is most of the problem we've found. (eg. we cannot map both strings "00001" and "1" to uint 1)

So though there is some test improvement needed on RLP level directly, quite a lot is on a higher level.

InoMurko:
This is a great comment @boolafish. It really explains what the issue is. I think we can all figure out how to do extensive testing on contracts side (property based testing exists for javascript as well!)

The other thing @pdobacz you mentioned:
> 2 .this hinders the propcheck based testing effort, since you can't exvcr those. We just got off a call with the https://github.com/orgs/omisego/projects/23 team and sentiment is that such tests are useful, considering recent findings about encoding etc. If this PR got merged, propcheck tests would need to find a new home in a different repo, that would be capable of deploying the proxy contracts (like PaymentEip712LibMock; it's not a mock really). While "finding new home" might be 100% justified, it will delay things.

I think we could provide a development flag to truffle that would deploy Mock contracts as well. Because this PR, does not really enforce exvcr, it just removes contract deployment responsibility by snapshoting them.

achiurizo:
> this hinders the propcheck based testing effort, since you can't exvcr those. We just got off a call with the https://github.com/orgs/omisego/projects/23 team and sentiment is that such tests are useful, considering recent findings about encoding etc. If this PR got merged, propcheck tests would need to find a new home in a different repo, that would be capable of deploying the proxy contracts (like PaymentEip712LibMock; it's not a mock really). While "finding new home" might be 100% justified, it will delay things.

propcheck testing can still continue on in its branch, it just doesn't need to re-base against master. This will give us time to find propcheck tests a new home *and* still build it out here to not delay things and move forward with simplifying and speeding up our tests.

> There are situations that one wants to interact with other contract version...

We should follow up with updates on the `readme`  on how to do the work flow with these changes.

boolafish:
> I think we could provide a development flag to truffle that would deploy Mock contracts as well. 

yeah, this should be doable

pdobacz:
Great discussion! First some replies from my PoV:

> I don't think there's anything like rapid contract development

I think, it is what is happening right now, basically what Andy wrote. If you want it to stop and go slower, timelines will be shifted, hence my pt (3.). Example: we need a new audit deployment after `plasma-contracts` progresses. Integration docs help, but they're not enough. If things break, you want to do little steps and hunt down things missed (also possibly missed in the int.-doc). Bear in mind that the stack is inherently unfriendly to us (e.g. solc, evm and all).

> /general critique of conformance test approach/

Two things here: https://github.com/omisego/plasma-contracts/issues/500 was discovered a bit by accident, as it is in fact a problem solely on the contracts end. This is just a bonus. 

Conformance testing really gives you something more: you compare whether 2 implementations behave the same. Now the problem is that their behavior is unstructured data (hash) where only bitwise comparison makes sense. So there is no test limited to just one implementation you can make, that will tell you whether a hash is correct. We spoke about this when talking about merkling and I thought we reached some consensus there.

It is true, that the particular tests in #1164 conflate conformance with correctness, but this is a result of limited resources. It's just trying do 2 (3) things with same code (what @pnowosie brought up), maybe imperfect in terms of design, but quite useful. Ability to pick up https://github.com/omisego/plasma-contracts/issues/500 proves this, it is a bonus gain.

> one time conformance test to try to figure our problem and then add invalid test cases for what we have found

that's the plan. Propcheck needs time, so it fits well with nightly builds or (to start off with) manual runs. Proper CI would only carry on the test cases discovered laid out explicitly.

> I think we could provide a development flag to truffle that would deploy Mock contracts as well. Because this PR does not really enforce exvcr...

I think that would be good, with the remark on:

> ...it just removes contract deployment responsibility by snapshoting them

To be clear, it is the snapshoting that requires one to route via a docker image push is problematic to me (my point 1., scoped broader than conformance tests). If the snapshot is done by default on every test run and _optionally_ once-until-manually-cleaned (when one doesn't touch contracts), I'm ok with it.

> propcheck testing can still continue on in its branch, it just doesn't need to re-base against master

I wish it could be an option and we find _new home_ ( :) ), but go back to my 1st point - I need to rebase against master because `plasma-contracts` is a moving target and `elixir-omg` follows along.

Property tests as of today get stuck on the shortcommings (#500) of current impl, so I actually already need to rebase today.

----------------------

Now, aware of all the arguments, may I come forward with a suggestion:
 - delay this PR till propcheck/conformance gets merged to `master`
 - then, limit the removal of stuffs in this PR, so that those still can operate as they do in #1164; _Lots_ of stuff you don't like will still be removed (`bundle_deployer.ex`)
 - do the snapshots as I suggested above, allowing one a more tinker-able setup
 - work doing more of the changes in this PR, e.g. allowing the conf. tests to run off of `truffle --devel`
 - continue from there.

? WDYT

InoMurko:
Sadly, I don't agree on the timeline you suggested. I think we need to increase friction on plasma contract usage ASAP on elixir-omg because it'll be beneficial in terms of code structure and design, approach to integration, etc. 

> I think, it is what is happening right now, basically what Andy wrote. If you want it to stop and go slower, timelines will be shifted, hence my pt (3.). Example: we need a new audit deployment after plasma-contracts progresses. Integration docs help, but they're not enough. If things break, you want to do little steps and hunt down things missed (also possibly missed in the int.-doc). Bear in mind that the stack is inherently unfriendly to us (e.g. solc, evm and all).

I think you're overreacting on the implications of having contract deployment in a separate repository. Every commit change you will make you can track with branches (and PRs) in `docker-elixir-omg` - so you can reference every commit during the "manual" contract bump changes you need to do in elixir-omg. If you think this would shift deliverables ... sorry, I think you're wrong. At least end2end testing that we ALL need to shift towards is much faster. It used to take 4-5 minutes to get `docker-compose up` running (and pruning), now it literally takes seconds. Same goes for all unit (integration tests) since deployment of contracts is completely omitted. 

>  So there is no test limited to just one implementation you can make, that will tell you whether a hash is correct. 

Can you elaborate this? I'm not sure I understand.

TLDR: This PR does not remove your ability to do property tests inside elixir-omg in ANY way with the --devel flag.
```
docker/migrations/1_deploy_712_lib_mock.js
var Contract = artifacts.require("PaymentEip712LibMock");

module.exports = function(deployer) {
  deployer.deploy(Contract);
};
```

-----------------------------------------------------------------------------------------
At this point, I don't think we will come to an agreement we all like. I suggest an executive decision so that we don't prolong this for too long. @kasima @achiurizo WDYT?

pdobacz:
OK! Yay indeed! I think the baby-stepping is going to pay off and we'll reach sth we all like. Now I need to tidy #1164 (plasma-contracts#500 blocking and some hiccups with `geth` :/) quick. I'll punt all the nice-to-haves there and :rocket: 

I'll also will have more breathing space to get the hang out of the "via-image" contract routine and generally we'll see how that works. I'm still inclined to doubt, if it's going to be as workable as `modify_contract_sha_or_code; mix test integration` I'm doing now (this runs in OoM of seconds for me BTW), but let's check this out!

achiurizo:
It's been over a week now I'd like to get us moving with this. While I understand the importance of the conformance test, this to me is critical for us to move forward quickly with the other pieces of the project. 

We'll need to spend time to update the conformance test to work wit this. Also I believe Piotr is out on sick leave for sometime so that delays that even further. :pray:

PR #1177, opened 2019-12-02
feature: decrese polling

InoMurko:
## Overview
We've been pulling events every 500 ms, which is obviously too much. The pull value should be less than frequency of pushed ethereum blocks (mainnet and rinkeby) + processing time 

## Changes

- Increase
- remove stale configuration

## Testing
/

### DIFF
--- a/apps/omg_child_chain/config/dev.exs
+++ b/apps/omg_child_chain/config/dev.exs
@@ -2,5 +2,4 @@
 use Mix.Config
 
 config :omg_child_chain,
-  ethereum_events_check_interval_ms: 500,
   block_queue_eth_height_check_interval_ms: 1_000
--- a/apps/omg/config/config.exs
+++ b/apps/omg/config/config.exs
@@ -4,7 +4,7 @@ use Mix.Config
 
 config :omg,
   deposit_finality_margin: 10,
-  ethereum_events_check_interval_ms: 500,
+  ethereum_events_check_interval_ms: 8_000,
   coordinator_eth_height_check_interval_ms: 6_000,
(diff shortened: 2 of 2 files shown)


jbunce:
Is this global for each module?

InoMurko:
I'm not sure what you mean @jbunce. It affects everything and everyone in the VM that uses this configuration.

unnawut:
1. Can we make this into a config provider? It seems to me a value that we would want to adjust per environment, and changes over time depending on the network's behaviour.

2. RE: config per module, that would be nice to have but I would say we don't need that level of granularity yet?

InoMurko:
we can definitely make this into a config provider, but I would wait until we get a sufficient requirement

jbunce:
@InoMurko thanks, that would be useful to write in the PR for reviewers

apps/omg_child_chain/config/dev.exs
```diff
@@ -2,5 +2,4 @@
 use Mix.Config
 
 config :omg_child_chain,
-  ethereum_events_check_interval_ms: 500,
```
pdobacz: :( missed this going through. In `dev` one often does 1-second blocks, so this was indeed useful

InoMurko: where are you reading ethereum_events_check_interval_ms from the childchain application @pdobacz ?

pdobacz: ahaaaaa, nice, :man_facepalming:

unnawut:
In resume-speak: Single-handedly reduced Ethereum calls by 1600%

PR #1179, opened 2019-12-02
Introduce adapter logic for fees loading

T-Dnzt:
## Overview

This PR takes the "fees loading from file" strategy from the `FeeServer` and wraps it into an adapter module (`Fees.FileAdapter`). This change allows the introduction of new adapters in the future (one will be made for the `FeeGate` service incoming soon).

## Changes

- Add a configurable adapter (hardcoded for now, will prob be moved to an ENV var later)
- Move the file loading to a different module

## Testing

Everything should be working the same way it is now. You can try updating the fee spec JSON file and see the changes reflected in `/fees.all`.


apps/omg_child_chain/lib/omg_child_chain/fees/file_adapter.ex
```diff
@@ -0,0 +1,75 @@
...
+      {:ok, %File.Stat{mtime: mtime}} ->
+        mtime
+
+      # possibly wrong path - returns current timestamp to force file reload where file errors are handled
+      _ ->
+        :os.system_time(:second)
+    end
+  end
+
+  defp get_path do
```
InoMurko: Do you prefer brackets or no brackets?

T-Dnzt: I prefer without if there are no arguments :) Can change if we follow a different style.

InoMurko: Pls double check what it says.

T-Dnzt: Can you be more explicit? I don't get it. Thought you meant parenthesis at first, now I'm not sure anymore. What's `it` also? Would be great to have less cryptic comments to avoid this kind of back and forth - if I did something wrong, just say so.

InoMurko: hey! I had no bad intentions. My comment was only saying that you should look at the Elixir style guide we're using... I didn't know what the answer is as well! But now I checked: https://github.com/lexmag/elixir-style-guide#fun-parens

T-Dnzt: @InoMurko Ah, I see, thanks! And, no worries, I was just super confused which led to some frustration. You're totally right about the style, I'll add the parenthesis to follow the guide :)

unnawut: I assumed for 2 full years that the style guide is to not have zero-arity fn parentheses :mindblown:

PR #1181, opened 2019-12-02
Inomurko/events check

InoMurko:
@unnawut was right.

We need to be able to modify ethereum event listeners, because local env. produce blocks every second.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -104,7 +104,7 @@ services:
       - DD_HOSTNAME=datadog
       - DD_DISABLED=true
       - DB_PATH=/app/.omg/data
-      - ETHEREUM_EVENTS_CHECK_INTERVAL_MS=1000
+      - ETHEREUM_EVENTS_CHECK_INTERVAL_MS=800
     restart: always
     ports:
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -104,6 +104,7 @@ services:
       - DD_HOSTNAME=datadog
       - DD_DISABLED=true
       - DB_PATH=/app/.omg/data
+      - ETHEREUM_EVENTS_CHECK_INTERVAL_MS=800
```
achiurizo:
```suggestion
      - ETHEREUM_EVENTS_CHECK_INTERVAL_MS=8000
```

docker-compose.yml
```diff
@@ -137,6 +138,7 @@ services:
...
```
achiurizo:
```suggestion
      - ETHEREUM_EVENTS_CHECK_INTERVAL_MS=8000
```

docker-compose.yml
InoMurko: why `8000`? it needs to be less then 1 second because thats the block production rate of the local geth

PR #1186, opened 2019-12-05
fix: harden end2end tests

InoMurko:
/

## Overview

Hardening the specs tests!

## Changes

- formatters
- account creation that works towards geth as well
- geth/ganache calls can fail, also API calls, so these are hardened to retry

## Testing


InoMurko:
I've extracted this from other PR's because they're on hold.

Issue #1176, opened 2019-12-02
Strict RLP decoding / Enforce PaymentTransaction according to spec


thec00n:
### Description
Recent efforts to achieve a 1:1 mapping from an RLP tx to a decoded PaymentTransaction have recently been merged into the root chain contracts ( PR https://github.com/omisego/plasma-contracts/pull/476). This issue proposes to implement the same mechanisms to achieve a 1:1 mapping and to decode/encode only well-formed PaymentTransactions. 

### Enforce non overlapping RLP decoding
Achieving a 1:1 relationship between decoded and encoded RLP message is not trivial and there are some edge cases that can easily be overlooked. Check out the test inputs in https://github.com/omisego/plasma-contracts/blob/182908d4c7ef669e426356c0bce528a5ebe75593/plasma_framework/test/src/utils/fixture/ and the RLP related test runners in https://github.com/omisego/plasma-contracts/blob/182908d4c7ef669e426356c0bce528a5ebe75593/plasma_framework/test/src/utils/. The implementation in Elixir should cover all of the tests that have been specific on the plasma-contracts.

#### Define Data types 

The current RLPReader implements RLP conform decoding as specified here https://github.com/ethereum/wiki/wiki/RLP.  It also introduces 3 data types that are application specific. 

- [x] RLPReader.toUint():  treats leading zeros as invalid. Must be between 1 and 32 bytes. 
- [x] RLPReader.toAddress() requires the item data to be exactly 20 bytes.
- [x] RLPReader.toBytes32() requires the item data to be exactly 32 bytes.

#### Decode/Encode only valid PaymentTransaction 

During PaymentTransaction decoding it is also checked that the tx is well formed and according to spec: 

- [x]  Transaction inputs must be decoded as bytes32
- [x]  Transaction inputs cannot be empty (all zeros)
- [x]  Transaction outputs cannot be empty (a list that has zero items)
- [x]  Transaction output.amount must be greater than zero
- [x]  Transaction output.outputType must be greater than zero
- [x]  Transaction metadata must be bytes32

CC @kevsul feel free to add details.

thec00n:
Additionally update https://github.com/omisego/elixir-omg/blob/master/docs/transaction_validation.md if changes are required.

pdobacz:
Thanks for the detailed write up, this is very helpful.

I'm currently on top of this, as part of the https://github.com/omisego/security-issues/issues/2 effort and #1164.

[...]

PR #1193, opened 2019-12-07
Avoid code is loaded check for version/service name

pdobacz:
blocks bumping contracts for #1176 a bit. 

## Overview

Adding version and service name to the API responses was using `Code.is_loaded?` to check what service is being run. This was marked to be changed and was bugging me, also it broke some configurations of `mix test`, for some reasons I wouldn't dare investigate.

Instead, we add a Plug that discovers this and is parametrized by the application being run. This plug assigns some data, which is later picked up by views

## Changes

 - as an additional change, I revamped the `health_test` a bit, so that it doesn't get fooled by alarms that are sometimes thrown as background noise (e.g. `disk_almost_full` alarms)
 - I don't like the way the `app_infos` must be verbosely injected into the response in every single view. Phoenix-gurus out there - is there a way to have sth like a custom JSON view template, so that some fields are always present?

## Testing

`mix test`

### DIFF
--- a/apps/omg_child_chain_rpc/lib/omg_child_chain_rpc/web/views/error.ex
+++ b/apps/omg_child_chain_rpc/lib/omg_child_chain_rpc/web/views/error.ex
@@ -23,19 +23,19 @@ defmodule OMG.ChildChainRPC.Web.Views.Error do
   Handles client errors, e.g. malformed json in request body
   """
   def render("400.json", _) do
-    Error.serialize("operation:bad_request", "Server has failed to parse the request.")
+    Error.serialize("operation:bad_request", "Server has failed to parse the request.", %{})
   end
 
   @doc """
   Supports internal server error thrown by Phoenix.
   """
   def render("500.json", %{reason: %{message: message}}) do
-    Error.serialize("server:internal_server_error", message)
+    Error.serialize("server:internal_server_error", message, %{})
   end
 
   # In case no render clause matches or no
   # template is found, let's render it as 500
   def template_not_found(_template, _assigns) do
-    Error.serialize("server:internal_server_error", nil)
+    Error.serialize("server:internal_server_error", nil, %{})
   end
 end
--- a/apps/omg_child_chain_rpc/test/omg_child_chain_rpc/plugs/application_info_test.exs
+++ b/apps/omg_child_chain_rpc/test/omg_child_chain_rpc/plugs/application_info_test.exs
@@ -1,0 +1,33 @@
+# Copyright 2019 OmiseGO Pte Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.ChildChainRPC.Plugs.ApplicationInfoTest do
+  @moduledoc false
+  use ExUnit.Case, async: false
+  use Phoenix.ConnTest
+
+  alias OMG.Utils.HttpRPC.Plugs.ApplicationInfo
+
+  describe "call/2" do
+    test "service name is appended" do
+      conn = build_conn() |> ApplicationInfo.call(application: :omg_child_chain_rpc)
+      assert "child_chain" == conn.assigns.app_infos.service_name
+    end
+
+    test "version is appended and follows semver" do
+      conn = build_conn() |> ApplicationInfo.call(application: :omg_child_chain_rpc)
+      assert {:ok, _} = Version.parse(conn.assigns.app_infos.version)
+    end
+  end
+end
--- a/apps/omg_utils/lib/omg_utils/http_rpc/plugs/application_info_plug.ex
+++ b/apps/omg_utils/lib/omg_utils/http_rpc/plugs/application_info_plug.ex
@@ -1,0 +1,44 @@
+# Copyright 2019 OmiseGO Pte Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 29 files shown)


unnawut:
I'm on this now

pdobacz:
Actually, I have a late thought, whether it wouldn't be good to make the version getting more lax. That is, if it still gets an `:undefined` when getting `:vsn`, it could just return `:undefined+<sha>` instead of throwing up. WDYT?

InoMurko:
I don’t think defaults like that are beneficial.

InoMurko:
@unnawut should this take watcher_info jnto account?

unnawut:
> if it still gets an :undefined when getting :vsn, it could just return :undefined+<sha> instead of throwing up. WDYT?

Enforcing `:vsn` sounds simpler & should be doable? The exception would then be a nice way to ensure enforcement.

---

> @unnawut should this take watcher_info jnto account?

I think so. Although I'm not sure how best to get that info...

InoMurko:
maybe ETS and `start_phases`?

pdobacz:
:+1: on the raise on `:vsn` missing, you're right, that was just a loose thought.

> @unnawut should this take watcher_info jnto account?

this works on a `_rpc` basis, so at this point, `_info` doesn't its own. As soon as it will, it would be trivial to have. In general this sounds like something pending more design and split work on a deeper level.

Are we good to go with this as the starting point to remove the `is_loaded?` thing or no good enough?

unnawut:
> this works on a _rpc basis, so at this point, _info doesn't its own. As soon as it will, it would be trivial to have. In general this sounds like something pending more design and split work on a deeper level.

Sounds good to me. Isn't a blocker here.

apps/omg_child_chain_rpc/lib/omg_child_chain_rpc/web/views/error.ex
```diff
@@ -23,19 +23,19 @@ defmodule OMG.ChildChainRPC.Web.Views.Error do
   Handles client errors, e.g. malformed json in request body
   """
   def render("400.json", _) do
-    Error.serialize("operation:bad_request", "Server has failed to parse the request.")
+    Error.serialize("operation:bad_request", "Server has failed to parse the request.", %{})
```
InoMurko: Why an empty map?

apps/omg_child_chain_rpc/test/omg_child_chain_rpc/plugs/application_info_test.exs
```diff
@@ -0,0 +1,33 @@
...
+  alias OMG.Utils.HttpRPC.Plugs.ApplicationInfo
+
+  describe "call/2" do
+    test "service name is appended" do
+      conn = build_conn() |> ApplicationInfo.call(application: :omg_child_chain_rpc)
+      assert "child_chain" == conn.assigns.app_infos.service_name
+    end
+
+    test "version is appended and follows semver" do
+      conn = build_conn() |> ApplicationInfo.call(application: :omg_child_chain_rpc)
```
InoMurko: Wrong use of pipe

apps/omg_utils/lib/omg_utils/http_rpc/plugs/application_info_plug.ex
```diff
@@ -0,0 +1,44 @@
...
+  Assigns per-application meta-information like version and service name to the connection, for later use
+  """
+  @behaviour Plug
+
+  @sha String.replace(elem(System.cmd("git", ["rev-parse", "--short=7", "HEAD"]), 0), "\n", "")
+
+  def init(options), do: options
+
+  def call(conn, options) do
+    conn
```
InoMurko: Wrong use of pipe

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/views/transaction_test.exs
```diff
@@ -58,7 +58,7 @@ defmodule OMG.WatcherRPC.Web.View.TransactionTest do
         }
       }
 
-      rendered = View.Transaction.render("transactions.json", %{response: paginator})
+      rendered = View.Transaction.render("transactions.json", %{response: paginator, app_infos: %{}})
```
InoMurko: Why are we permitting empty app info map?

apps/omg_watcher_rpc/test/omg_watcher_rpc/plugs/application_info_test.exs
```diff
@@ -0,0 +1,33 @@
...
+  alias OMG.Utils.HttpRPC.Plugs.ApplicationInfo
+
+  describe "call/2" do
+    test "service name is appended" do
+      conn = build_conn() |> ApplicationInfo.call(application: :omg_watcher_rpc)
+      assert "watcher" == conn.assigns.app_infos.service_name
+    end
+
+    test "version is appended and follows semver" do
+      conn = build_conn() |> ApplicationInfo.call(application: :omg_watcher_rpc)
```
InoMurko: Wrong use of pipe

apps/omg_watcher_rpc/test/omg_watcher_rpc/plugs/application_info_test.exs
```diff
@@ -0,0 +1,33 @@
...
+defmodule OMG.WatcherRPC.Plugs.ApplicationInfoTest do
+  @moduledoc false
+  use ExUnit.Case, async: false
+  use Phoenix.ConnTest
+
+  alias OMG.Utils.HttpRPC.Plugs.ApplicationInfo
+
+  describe "call/2" do
+    test "service name is appended" do
+      conn = build_conn() |> ApplicationInfo.call(application: :omg_watcher_rpc)
```
InoMurko: Wrong use of pipe

apps/omg_watcher_rpc/lib/web/views/error.ex
```diff
@@ -23,19 +23,19 @@ defmodule OMG.WatcherRPC.Web.Views.Error do
...
   def render("500.json", %{reason: %{message: message}} = _conn) do
-    Error.serialize("server:internal_server_error", message)
+    Error.serialize("server:internal_server_error", message, %{})
   end
 
   # In case no render clause matches or no
   # template is found, let's render it as 500
   def template_not_found(_template, _assigns) do
-    Error.serialize("server:internal_server_error", nil)
+    Error.serialize("server:internal_server_error", nil, %{})
```
InoMurko: Why is there no app info here?

unnawut:
`"service_name": "watcher_info"` tracked as https://github.com/omisego/elixir-omg/issues/1201

InoMurko:
@unnawut if your intention is to work on this pas part of #1201 I suggest you keep this PR open and polish it. At the moment this PR doesn’t solve any immediate issues.

unnawut:
I'm stealing this PR so it could also address #1198 and #1201 as Ino suggested.

unnawut:
Closing in favor of #1266 so it can be reviewed properly. Thanks @pdobacz !

Issue #1129, opened 2019-11-12
Is transaction.submit security critical API?


InoMurko:
The security critical watcher  allows you to use `/transaction.submit Sends transaction to Child chain.` but it does not allow you to retrieve any of the UTXOs or balances.

```/account.get_balance Returns the balance of each currency for the given account address.  
/account.get_utxos Gets all utxos belonging to the given address. 
```
The question is... 

- [ ] should we remove `transaction.submit`from the security critical watcher?
- [ ] what do we do with tests (for example `OMG.Watcher.Integration.BlockGetterTest`) that are part of the "secure" implementation of Watcher but cross the border to informational api specs (this goes well with another issue https://github.com/omisego/elixir-omg/issues/1122)

unnawut:
Or the other way around, `/transaction.submit` is a security critical API, plus bring `/account.get_utxos` into security critical API. Since both are required to make a transaction. And misinformation/unavailability of these two endpoints could inability to interact with the funds?

This would leave the rest of the Informational APIs purely for data retrieval, except for maybe `/transaction.submit_typed`.

pnowosie:
@InoMurko - this is correct. Security critical API is not for transacting on the network, but to provide network's security data. So if one wants utxo selection features for transacting it's InfoApi piece of cake. (Security Api provides however all addresses utxos for exit)

`/transaction.submit` is considered security critical because it **checks validity of the chain** before it forward the tx to the child chain. (Note: **checking validity isn't yet implemented** there - so it might confuse you).

InoMurko:
as we found out with @unnawut 
```
/account.get_utxos
```
actually use PG to retrieve utxos per account. So that would be rather non trivial to use on the "security" side of watcher.

Hm... so @pnowosie so one should use informational watcher to retrieve UTXOs and than if he wants to securely transact, he should use  `/transaction.submit` on security watcher? But then your statement is contradicting itself 
```
Security critical API is not for transacting on the network, but to provide network's security data. 
```
why is `/transaction.submit` there if API's purpose is to provide network's security data?

unnawut:
> So if one wants utxo selection features for transacting it's InfoApi piece of cake

Note that that's `/transaction.create` though which I agree to have it in informational. 

But `/transaction.submit` can't be done unless it at least has info from `/account.get_utxos.` Which makes a security API dependent on info API, hence my suggestion to move /account.get_utxos into security.

jrhite:
[...]

PR #1194, opened 2019-12-09
fix: ecto create

InoMurko:
Failing master nightly build

## Overview

Adding ecto setup where needed and cleanup.

## Changes

- OMG watcher breaches scope of testing and it needs ecto because of that
- Removed cookie setting
- Docs

## Testing

/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,7 +87,7 @@ jobs:
       - run:
           name: Integration Tests
           command: |
-            # Slow, serial integration test, run nightly. Here to make sure the standard `mix test --only integration` works
+            # Slow, serial integration test, run nightly. Here to make sure the standard `mix test --trace --only integration` works
             export SHELL=/bin/bash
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -626,15 +626,17 @@ workflows:
                 - master
     jobs:
       - build
-      - integration_tests:
-          requires: [build]
+      #- integration_tests:
```
unnawut: Can just remove the two lines I think

InoMurko: I just wanted to see it pass on the PR branch and put it back into nightly ;)

unnawut:
> would it be feasible to put into the TODO (or maybe some other place visible, like PR desc) the list of tests that do the breach - I suppose those are the ones that failed in one of the nighlies?

Added to https://github.com/omisego/elixir-omg/issues/1129#issuecomment-564891289

PR #1197, opened 2019-12-10
feat: use .tool-versions to pin elixir/erlang for dev env to the builder's versions

unnawut:
## Overview

RE: @okalouti 

> Hello everyone. You might already be aware of this, but FYI just in case: I tried to run tests (mix test) on elixir-omg yesterday using Elixir v. 1.9.2 and they didn't work because the fixtures were not being loaded (e.g. module OMG.Fixtures is not loaded and could not be found ) . Debugged it with @unnawut for some time before figuring out that setting Elixir version to 1.8.0 fixed this problem.

Had a few occurrences where this happens so I'd like to add a `.tool-versions` so those of us with [asdf](https://github.com/asdf-vm/asdf) can work on the same versions. For non-asdf users, looking up the versions in `.tool-versions` is convenient too.

## Changes

- Add `.tool-versions`

## Testing

- If you use [asdf](https://github.com/asdf-vm/asdf), running `asdf current` on elixir-omg directory should give:

    ```
    elixir         1.8.1    (set by <your_path_to_elixir_omg>/.tool-versions)
    erlang         21.2.4   (set by <your_path_to_elixir_omg>/.tool-versions)
    ```

- If you don't use asdf: no impact.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,10 +112,10 @@ jobs:
           source ~/.bash_profile
           asdf plugin-add elixir https://github.com/asdf-vm/asdf-elixir.git
           asdf plugin-add erlang https://github.com/asdf-vm/asdf-erlang.git
-          asdf install erlang 21.0
+          asdf install erlang 21.2.4
           asdf install elixir 1.8.2
           asdf global elixir 1.8.2
-          asdf global erlang 21.0
+          asdf global erlang 21.2.4
       - run:
           command: ./bin/setup
           no_output_timeout: 2400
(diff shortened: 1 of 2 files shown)


jrhite:
Yeah, totally agree with this. When I setup with Ino using asdf, we set my versions to: 
```
erlang 21.3.8.5
elixir 1.8.1-otp-21
```
Not sure if the OTP matters...never had problems with these versions....

InoMurko:
Normally these specific installation files (IDE, asdf...) don't belong in the common repository.

unnawut:
> Normally these specific installation files (IDE, asdf...) don't belong in the common repository.

Hmm good point. I'll propose a new generic solution on top of `make diagnostics`.

PR #1205, opened 2019-12-11
feat: add `mix xomg.watcher_info.start`

unnawut:
## Overview

Adds `mix xomg.watcher_info.start` in addition to the existing `mix xomg.watcher.start` and `mix xomg.child_chain.start`. I forgot about this in the main code split PR.

## Changes

- Removed convenience api option from watcher.start and replace with a new task for watcher_info so it has closer UX to API and releases.

## Testing

Run `mix xomg.watcher_info.start`. Couldn't think of a way to automate test without significant overhead. And this is a dev tool so 🤷‍♂

### DIFF
--- a/apps/xomg_tasks/lib/mix/tasks/child_chain.ex
+++ b/apps/xomg_tasks/lib/mix/tasks/child_chain.ex
@@ -14,14 +14,14 @@
 
 defmodule Mix.Tasks.Xomg.ChildChain.Start do
   @moduledoc """
-    Contains mix.task to run the child chain server
+  Contains mix.task to run the child chain server.
   """
 
   use Mix.Task
 
   import XomgTasks.Utils
 
-  @shortdoc "Start the child chain server. See Mix.Tasks.ChildChain"
+  @shortdoc "Start the child chain server. See Mix.Tasks.Xomg.ChildChain.Start."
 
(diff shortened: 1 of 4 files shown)


InoMurko:
I actually think this is an antipattern we shouldn't support anymore. Any reason we're adding this?

unnawut:
> I actually think this is an antipattern we shouldn't support anymore. Any reason we're adding this?

Hmm ok I thought this is still a supported way, or at least @mederic-p has been using it. And so his workflow broke when the code split PR landed.

I believe the main reason it's still being used is, again back to, the auto-hot reload on code changes, for quick debugging.

unnawut:
Side note: Since this auto hot reload thing popped up a few times, I'll document it down in `make help` once it's concluded.

InoMurko:
If there's a way to extract these tasks out ->
- so that there's no way they could affect tests (so that someone doesn't get an idea to use them in fixtures). We kinda agreed with @achiurizo that we need to impose friction on certain areas so that test scope is enforced.
- or that we would somehow be able to "break" them and would that force us to support these tasks with much hassle

I would definitely be up for helping developers develop quicker.

unnawut:
Sounds good. I'm all in for removing these tasks from fixtures.

PR #1207, opened 2019-12-13
Fix Watcher Info deployments

jbunce:
## Overview

Fix CD for Watcher Info


jbunce:
This is due to naming constraints with Kubernetes: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

PR #1212, opened 2019-12-17
Refactor: consolidate InputPointer modules

achiurizo:
## Overview

This PR does a first pass at pushing towards integrating [ex_plasma]() with `elixir-omg`'s encoding/decoding mechanics. In this PR, it consolidates the `InputPointer` modules and leverages `ExPlasma.Utxo` underneath to handle the conversion of input Utxos.

## Changes

* Consolidate `InputPointer.UtxoPosition` - It was not being called much and effectively was there to `reconstruct` a binary encoded input back into a utxo position tuple, e.g `{;utxo_position, blknum, txindex, oindex}`
* Remove the `InputPointerProtocol`. Currently we only support 1 type so this is un-necessary at the moment. Also following some reviews [in other PRs](https://github.com/omisego/elixir-omg/pull/1182#discussion_r357472928), we want to move away from having the type behaviour being specified in the application config.
* Remove `Utxo.position` macro. It's un-necssary to have a macro for this tuple. In a follow up PR, will revisit how we should pass this structure around. I think a map or struct here makes more sense and makes it much clearer at what kind of data it contains. This also removes the need for us to have to move between an atom lead- tuple form `{:utxo_position, x, y, z}` to a "db tuple" form `{x, y, z}`.
* Increases `InputPointer` responsibility and add doctests and docs.
* cleans up unused alias/require calls to `OMG.Utxo`

## Testing

```sh
mix test
# and CI/CD
```


apps/omg/test/omg/state/core_test.exs
```diff
@@ -680,23 +680,29 @@ defmodule OMG.State.CoreTest do
...
     # this test checks whether all ways of calling `get_exiting_utxo_positions/2` translates
     # to given exiting utxo positions
-    utxo_pos_exits = [Utxo.position(@blknum1, 0, 0), Utxo.position(@blknum1, 0, 1)]
+    expected_utxo_pos_exits = [{:utxo_position, @blknum1, 0, 0}, {:utxo_position, @blknum1, 0, 1}]
 
-    assert utxo_pos_exits ==
+    utxo_pos_exits = [
+      %{blknum: @blknum1, txindex: 0, oindex: 0},
+      %{blknum: @blknum1, txindex: 0, oindex: 1}
+    ]
```
achiurizo: Will come back to re-visit this in a follow up PR, but I think we should move the data structure into a map or struct. I believe this will provide more clarity and mileage then sticking with the various atom-led tuple formats e.g:
*  `{:utxo_position, x, y, z}`
* `{:input_pointer, @input_pointer_type_marker, {x, y, z}`
* `{x, y, z}`

where `x,y,z` respectively is `blknum, txindex, oindex`

apps/omg/test/omg/state/core_test.exs
```diff
@@ -680,23 +680,29 @@ defmodule OMG.State.CoreTest do
...
+
+    assert expected_utxo_pos_exits ==
              utxo_pos_exits
-             |> Enum.map(&%{call_data: %{utxo_pos: Utxo.Position.encode(&1)}})
+             |> Enum.map(&%{call_data: %{utxo_pos: ExPlasma.Utxo.pos(&1)}})
```
achiurizo: For this usage, It is seamless to use the `ExPlasma.Utxo` library directly since we want to generate values such as the integer input utxo position instead of having a wrapper around it.

apps/omg/lib/omg/input_pointer.ex
```diff
@@ -14,33 +14,132 @@
...
+  def decode(encoded) when is_binary(encoded) and byte_size(encoded) == 32,
+    do: encoded |> :binary.decode_unsigned(:big) |> decode()
+
+  def decode(encoded) when is_integer(encoded) and encoded <= 0,
+    do: @position_too_low_error_tuple
+
+  def decode(encoded) when is_integer(encoded) do
+    %ExPlasma.Utxo{blknum: blknum, txindex: txindex, oindex: oindex} = ExPlasma.Utxo.new(encoded)
+    {:ok, {:utxo_position, blknum, txindex, oindex}}
+  end
```
achiurizo: Eventually, I'd like to remove this call as well, but because I'm unsure where we care about this error tuple, I'm leaving it in another follow up PR(or if comments suggest otherwise) to clean this portion up.

apps/omg/lib/omg/input_pointer.ex
```diff
@@ -14,33 +14,135 @@
...
+          :utxo_position,
+          # blknum
+          non_neg_integer,
+          # txindex
+          non_neg_integer,
+          # oindex
+          non_neg_integer
+        }
+
+  # This is the 'legacy' db format that the code is using to maintain backwards
```
InoMurko: no reason to keep this comment I presume. or rephrase it so it makes more sense :-)

apps/omg/lib/omg/input_pointer.ex
pdobacz: oh I think you changed the code and the comment is dangling without proper context now

apps/omg/lib/omg/input_pointer.ex
```diff
@@ -14,34 +14,142 @@
...
+  @type db_key_tuple() :: {
+          :input_pointer,
+          <<_::8>>,
+          utxo_pos_db_tuple()
+        }
+
+  @position_too_low_error_tuple {:error, :encoded_utxo_position_too_low}
+  @type position_too_low_error_tuple() :: {:error, :encoded_utxo_position_too_low}
+
+  defstruct [:blknum, :txindex, :oindex]
```
pdobacz: I think a macro-wrapped `Utxo.position(x, y, z)` implementation was more readable and to-the-point. A Utxo position is naturally a triplet. We would consider maps, if we were considering expanding to new fields or any other modifications, but that is not the case. 

Also, I find `encode(Utxo.position(...))` more readable than `.pos(some_map_with_fields)`. The latter would also happily encode the position for anything _similar_ to a UTXO position, which is scary. 

I don't think this ever was unclear what `Utxo.position(...)` represents (as soon as someone knows what a UTXO position is) and it avoided the repetitive and noisy chatter `blknum: blknum, txindex: txindex` etc.

achiurizo: >  think a macro-wrapped Utxo.position(x, y, z) implementation was more readable and to-the-point. A Utxo position is naturally a triplet. We would consider maps, if we were considering expanding to new fields or any other modifications, but that is not the case.

I think the underlying issue is we _should_ use maps instead. The tuple format while simple, doesn't really buy us much except cost us alot of code transition between a `:utxo_position` lead atom vs one with out (see the db key 
tuple).

> Also, I find encode(Utxo.position(...)) more readable than .pos(some_map_with_fields). The latter would also happily encode the position for anything similar to a UTXO position, which is scary.

I'm not sure the former really does much to prevent that either. You could [happily encode any value into the tuple as well](https://github.com/omisego/elixir-omg/blob/master/apps/omg/lib/omg/utxo.ex#L33). With the map, while more keys could be placed in there, the function pulls out the value it requires to build the utxo position, the remainder has no effect. the `ExPlasma.Utxo.pos/1` could also guard against valid values if needs be(it's already pattern matching the values out as a function, just need to include a guard).

> I don't think this ever was unclear what Utxo.position(...) represents (as soon as someone knows what a UTXO position is) and it avoided the repetitive and noisy chatter blknum: blknum, txindex: txindex etc.

The purpose of this removal was to move us away from using a tuple to represent this data. I think a map here provides much more clarity(and flexibility) when it comes to representing this data. Additionally, this PR wants to move towars restructuring how we organize this data. We go through a bit of effort to map a position just to point it back into the output utxo repreentation (see input utxo pos -> {output data, creating_tx_hash mapping).  I think we can simplify this and merge the input utxo information with that of its corresponding output utxo information. It only needs to be pulled out [individually when it comes to encoding](https://github.com/omisego/ex_plasma/blob/master/lib/ex_plasma/transaction.ex#L123).

I don't think it encourages that much noise, and even if the consensus deems it so, we could also do the same macro approach to produce the underlying map(instead of the current tuple).

apps/omg/lib/omg/output.ex
```diff
@@ -14,53 +14,108 @@
...
-  # default clause for backwards compatibility
-  def from_db_value(%{} = db_value), do: OMG.Output.FungibleMoreVPToken.from_db_value(db_value)
-end
+  # TODO(achiurizo)
+  # Need to fix that this method is able to re-build from it's own generated rlp data.
+  # ex: [<<1>>, <<1::160>>, <<1::160>>, 1] (last number is an integer instead of binary)
+  @doc """
+  Returns a OMG.Output struct from a map
 
-defprotocol OMG.Output.Protocol do
```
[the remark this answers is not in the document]
achiurizo: This is behaviour I want to move into ex_plasma. Secondly I disagree that protocols is the right approach here(why not just use a [behaviour?](https://github.com/omisego/ex_plasma/blob/master/lib/ex_plasma/transaction.ex#L12) You can extend [additional types like payment transactions](https://github.com/omisego/ex_plasma/blob/master/lib/ex_plasma/transactions/payment.ex#L11). The dispatching happens [all within the module](https://github.com/omisego/ex_plasma/blob/master/lib/ex_plasma/transaction.ex#L129).

While the current ex_plasma has implemented such dispatching for output_type, to implment it with the current setup won't be difficult.

apps/omg/lib/omg/state/core.ex
```diff
@@ -72,7 +68,7 @@ defmodule OMG.State.Core do
           utxo_db_updates: list(db_update()),
           # NOTE: because UTXO set is not loaded from DB entirely, we need to remember the UTXOs spent in already
           # processed transaction before they get removed from DB on form_block.
-          recently_spent: MapSet.t(InputPointer.Protocol.t())
+          recently_spent: MapSet.t(tuple())
```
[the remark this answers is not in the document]
achiurizo: Perhaps I'm not understanding you correctly, but this is coupling us to another module which is the protocol. I think also there's a larger de-coupling that needs to happen as the current code base cross calls too many modules/contexts.

apps/omg/lib/omg/input_pointer.ex
```diff
@@ -14,33 +14,142 @@
...
+    iex> OMG.InputPointer.from_db_key({:input_pointer, <<1>>, {1, 2,3}})
+    %OMG.InputPointer{blknum: 1, txindex: 2, oindex: 3}
+
+    # Convert a legacy db key tuple for compatiblity(the _plain_ tuple)
+    iex> OMG.InputPointer.from_db_key({1, 2, 3})
+    %OMG.InputPointer{blknum: 1, txindex: 2, oindex: 3}
   """
+  @spec from_db_key(db_key_tuple() | utxo_pos_db_tuple()) :: utxo_pos_tuple()
+  def from_db_key({:input_pointer, @input_pointer_type_marker, db_value}), do: from_db_key(db_value)
+  def from_db_key({blknum, txindex, oindex}), do: %OMG.InputPointer{blknum: blknum, txindex: txindex, oindex: oindex}
```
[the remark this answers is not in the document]
achiurizo: Okay, can you explain what the `@input_pointer_type_marker` is in reference to? We have an `output_type` on an output utxo. Is this what this is?

achiurizo:
@pdobacz the end goal which this PR is beginning to introduce is the over-all decoupling of plasma framework encoding/decoding protocol into another library(re: ex_plasma). The purpose is to start breaking out those behaviors and logic into that library to handle for us so the main repository can focus on everything but encoding/decoding(think persistence, scaling, API).

apps/omg/lib/omg/state/core.ex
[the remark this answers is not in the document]
achiurizo: Also with the new `OMG.InputerPointer` and it's associated `OMG.InputPointer.Protocol` and `OMG.InputPointer.UtxoPosition`, I don't see how we de-coupled anything. I'd argue now we are coupled to even more modules.

See [OMG.InputPointer.UtxoPosition.from_db_key/1](https://github.com/omisego/elixir-omg/blob/master/apps/omg/lib/omg/input_pointer/utxo_position.ex#L24) which now introduces additional modules, but still calls out functionality back into `OMG.Utxo.Position` for backwards compatibility.

apps/omg/lib/omg/output.ex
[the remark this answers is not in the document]
InoMurko: "future code", "readyiness for future code", always a bad sign. don't ever do it if you can avoid it.

apps/omg/lib/omg/input_pointer.ex
InoMurko: I will just add that, a macro-based-tripplet + macro guard validator is a very uncommon approach. From what I've seen, things were made more confusing than actually helping. Definitely for removal and usage of a common data structure like a struct.

achiurizo:
Closing this PR in favor of #1218, which will introduce the integration in smaller scopes to make the review easier, and integration more palatable.

apps/omg/lib/omg/state/core.ex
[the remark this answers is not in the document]
pdobacz: By "coupling" I mean

```
          recently_spent: MapSet.t(tuple())
```

this thing this `MapSet` holds is (or was, until being postponed, see other thread) about to be extended with a 32-byte hash-based binary (`InputPointer.OutputID`), that's not a tuple anymore. 

And more generally speaking, this module here, should not care about how the inputs are pointed to ("input pointer"), it's (or should be, and that was the direction we were heading) agnostic of that.

apps/omg/lib/omg/input_pointer.ex
achiurizo:
```
alias OMG.Utxo
require Utxo

Utxo.position(blknum, txindex, oindex)

# vs
%OMG.InputPointer{blknum: blknum, txindex: txindex, oindex: oindex}
```

Okay, let's be more fair. You need the `require` to invoke the macro.

> I'm surprised about this being confusing and the (small) uncommonness being a problem. I think what is built on top (MoreVP/State logic/Watcher logic/challenges/whatnot) is uncommon and confusing, and we could use this syntax sugar to make the higher-level constructs less confusing.

I am all for syntactic sugar to make code nicer. I think having a struct here can increase clarity and flexibility in our code then having a macro that returns us a atom led tuple that we convert back and forth into, a non atom-led tuple:

```
{:utxo_position, 1, 2, 3}
{1, 2, 3}
```

apps/omg/lib/omg/state/core.ex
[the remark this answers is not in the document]
achiurizo: > And more generally speaking, this module here, should not care about how the inputs are pointed to ("input pointer"), it's (or should be, and that was the direction we were heading) agnostic of that.

I think this sounds exactly like why we should move the encoding/decoding and representation of the structures into a library like `ex_plasma`. When it comes to the childchain or watcher, the responsiblity of these apps should be how we store and organize the transactions(and also the utxos). Anything specifically related to encoding and how we send it down to the contracts belongs into a library like `ex_plasma`.

apps/omg/lib/omg/output.ex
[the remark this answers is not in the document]
achiurizo: > So, please, let's try to keep the critique constructive and considerate.

If it comes off that our discussion are not constructive or considerate, then I apologize. Discussions like these can be contentious so that we can understand the motivations behind it and how to move forward.

>  It is hard to see the plan, if it lives in a separate repo.)

I agree, but I that's the separation of concerns I think we need to move towards. The responsibility of how to interact with the contract is making it much more difficult for us to focus on the other responsibilities of the child chain and watcher.

> BTW, why wouldn't we rather fix in elixir-omg and split ex_plasma out later or split out "as-is" and fix after split. It would be much easier to see the plan through.

This I accept responsibility. I wanted to move quickly to understand the inner workings of the encoding/decoding process and build a fuller test suite that I can run quickly with. I also didn't want to make such large changes that would affect the push towards ALD directly. By coming in with full unit tests et al, my goal was to be able to introduce these changes effectively without much disruption (beyond the removal of the code-paths that are doing in repo).

PR #1214, opened 2019-12-17
bump contracts to c05b445

pdobacz:
N/A

## Overview

Provides contract compatibility as title

## Changes

- conform to some updates of the contracts: WireTransaction refactor, txData being added, output data being wrapped in an array, fixes
- I needed to bring back the `plasma_contracts` dependency just for the `Eip712Mock` - it still relied on that being there for `vcr_recording`. At some point this will be resolved by `--devel` truffle migration in some separate issue/PR
- conformed to the updates to format & contracts in `itest`
- did some minor tweaks in `itest` infrasturcture (`Poller` will fail if it gets an unexpected status in the receipt)
- refactored the contract addresses used in `docker-compose.yml` into an `env` file to ease the process of bumping

## Testing

the usual


pdobacz:
while doing the bump I updated @unnawut 's https://github.com/omisego/devops/wiki/Upgrading-elixir-omg-to-a-new-plasma-contracts-SHA. The instructions now should be good to go through the bumping process

Dockerfile.child_chain
```diff
@@ -1,7 +1,7 @@
 FROM alpine:3.8
 
 LABEL maintainer="OmiseGO Team <omg@omise.co>"
-LABEL description="Official image for OmiseGO (Watcher) Plasma Network"
+LABEL description="Official image for OmiseGO (Child Chain) Plasma Network"
```
InoMurko: lol, good one :D

apps/omg/lib/omg/state/transaction/payment.ex
```diff
@@ -80,11 +80,11 @@ defmodule OMG.State.Transaction.Payment do
   @doc """
   Transaform the structure of RLP items after a successful RLP decode of a raw transaction, into a structure instance
   """
-  def reconstruct([inputs_rlp, outputs_rlp | rest_rlp])
-      when rest_rlp == [] or length(rest_rlp) == 1 do
+  def reconstruct([inputs_rlp, outputs_rlp, tx_data_rlp, metadata_rlp]) do
     with {:ok, inputs} <- reconstruct_inputs(inputs_rlp),
          {:ok, outputs} <- reconstruct_outputs(outputs_rlp),
-         {:ok, metadata} <- reconstruct_metadata(rest_rlp),
+         :ok <- reconstruct_tx_data(tx_data_rlp),
```
achiurizo: nitpick: since it's written this way, would it be better to just have this follow the same call style?

```elixir
{:ok, _tx_data} = reconstruct_tx_data(metadata_rlp)
```

apps/omg/test/omg/state/transaction/recovered_test.exs
```diff
@@ -108,52 +109,81 @@ defmodule OMG.State.Transaction.RecoveredTest do
...
       assert {:error, :malformed_witnesses} ==
-               Transaction.Recovered.recover_from(ExRLP.encode([[<<1>>, <<1>>], @payment_marker, inputs, outputs]))
+               Transaction.Recovered.recover_from(
+                 ExRLP.encode([[<<1>>, <<1>>], @payment_marker, inputs, outputs, 0, <<0::256>>])
+               )
 
       assert {:error, :malformed_witnesses} ==
-               Transaction.Recovered.recover_from(ExRLP.encode([<<1>>, @payment_marker, inputs, outputs]))
+               Transaction.Recovered.recover_from(
+                 ExRLP.encode([<<1>>, @payment_marker, inputs, outputs, 0, <<0::256>>])
```
achiurizo:
```suggestion
                 ExRLP.encode([1, @payment_marker, inputs, outputs, 0, <<0::256>>])
```

Just want to highlight we should start moving towards the transaction format's specification, which I think transaction type should be 1? https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/docs/integration-docs/integration-doc.md#payment-transaction-rlp-encoding

priv/apps/itest/lib/transactions/encoding.ex
```diff
@@ -1,13 +1,13 @@
...
+  Provides helper functions for encoding and decoding data.
   """
-  @output_type_marker <<1>>
+  @payment_tx_type_marker <<1>>
 
   alias Itest.Transactions.Deposit
 
   def get_data_for_rlp(%Deposit{inputs: inputs, outputs: outputs, metadata: metadata}),
-    do: [@output_type_marker, inputs, outputs, metadata]
+    do: [@payment_tx_type_marker, inputs, outputs, "", metadata]
```
achiurizo: would it be more clearer to have this as a `0` instead according to the new format? https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/docs/integration-docs/integration-doc.md#payment-transaction-rlp-encoding

apps/omg/test/omg/state/transaction/recovered_test.exs
pdobacz: ah yes, good catch

priv/apps/itest/lib/transactions/encoding.ex
pdobacz: argh, was going too fast here, 0 absolutely is better. I'll see if I haven't done a similar slip-up elsewhere

apps/omg/lib/omg/state/transaction/payment.ex
pdobacz: I was wondering what was less confusing, since we're disregarding the value as long as it's `0`. Maybe I'll just rename the `reconstruct` to `verify` in this case?

apps/omg/test/omg/state/transaction/recovered_test.exs
```diff
@@ -96,10 +96,11 @@ defmodule OMG.State.Transaction.RecoveredTest do
         Transaction.Payment.new([{1, 0, 0}, {2, 0, 0}], [{alice.addr, @eth, 12}])
         |> DevCrypto.sign([alice.priv, alice.priv])
 
-      [_payment_marker, inputs, outputs, _metadata] = Transaction.raw_txbytes(tx) |> ExRLP.decode()
+      [_payment_marker, inputs, outputs, _txdata, _metadata] = Transaction.raw_txbytes(tx) |> ExRLP.decode()
```
InoMurko:
```suggestion
      [_payment_marker, inputs, outputs, _txdata, _metadata] = tx |> Transaction.raw_txbytes() |> ExRLP.decode()
```

apps/omg/test/omg/state/transaction/recovered_test.exs
pdobacz: ah no, that's not the type, those are malformed witnesses, payment marker comes next

apps/omg/lib/omg/state/transaction/payment.ex
achiurizo: @pdobacz comment was mostly for consistency with the call stack here. The outcome of what the code is clear, but in terms of aesthetics of the code, it would conform nicely if it followed the rest of the stack which is also doing:

```elixir
    with {:ok, inputs} <- reconstruct_inputs(inputs_rlp),
         {:ok, outputs} <- reconstruct_outputs(outputs_rlp),
         {:ok, metadata} <- reconstruct_metadata(rest_rlp),
         {:ok, _tx_data} <- reconstruct_tx_data(tx_data_rlp),
```

apps/omg/lib/omg/state/transaction/payment.ex
```diff
@@ -80,11 +80,11 @@ defmodule OMG.State.Transaction.Payment do
   @doc """
   Transaform the structure of RLP items after a successful RLP decode of a raw transaction, into a structure instance
```
T-Dnzt: typo: `transaform` -> `transform`

apps/omg/lib/omg/typed_data_hash/tools.ex
```diff
@@ -95,12 +95,16 @@ defmodule OMG.TypedDataHash.Tools do
       |> Stream.concat(Stream.cycle([empty_output_hash]))
       |> Enum.take(Transaction.Payment.max_outputs())
 
+    tx_data = ABI.TypeEncoder.encode_raw([0], [{:uint, 256}])
+    metadata = metadata || <<0::256>>
```
T-Dnzt: I'm nitpicking, but the second item in the list below doesn't get its own variable, not sure these two deserve it 😛 They should either all get intermediate variables, or just all not get any. 

```
ABI.TypeEncoder.encode_raw([:binary.decode_unsigned(plasma_framework_tx_type)], [{:uint, 256}])
```

priv/apps/itest/lib/transactions/encoding.ex
```diff
@@ -1,13 +1,13 @@
 defmodule Itest.Transactions.Encoding do
   @moduledoc """
-    Provides helper functions for encoding and decoding data.
+  Provides helper functions for encoding and decoding data.
   """
-  @output_type_marker <<1>>
+  @payment_tx_type_marker 1
```
T-Dnzt: ❤️

apps/omg/lib/omg/state/transaction/payment.ex
pdobacz: if we're never caring about the output, I think it's better to not return it and have a representative function name.

apps/omg/lib/omg/output/fungible_morevp_token.ex
```diff
@@ -30,14 +30,22 @@ defmodule OMG.Output.FungibleMoreVPToken do
...
   end
 
-  def reconstruct([owner, currency, bin_amount]) do
-    with {:ok, cur12} <- parse_address(currency),
-         {:ok, owner} <- parse_address(owner),
-         {:ok, int_amount} <- parse_int(bin_amount),
+  @doc """
+  Reconstructs the structure from a list of RLP items
+
+  NOTE: the of items should be a single-item list holding a list of three items. The output type has been parsed earlier
```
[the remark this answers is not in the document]
pdobacz: ha, but that's what it is - a list inside a list within a list :). I'll fix this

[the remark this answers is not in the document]
pdobacz: I refactored so there's no nesting there and `NOTE` is unnecessary

apps/omg/test/omg/state/core_test.exs
```diff
@@ -615,9 +615,7 @@ defmodule OMG.State.CoreTest do
              }, _, _}, _} = form_block_check(state)
 
     # precomputed fixed hash to check compliance with hashing algo
-    assert block_hash ==
-             <<168, 134, 99, 72, 67, 39, 246, 127, 88, 163, 200, 169, 167, 105, 15, 124, 143, 240, 174, 27, 84, 112, 48,
-               53, 41, 53, 236, 243, 192, 188, 116, 226>>
+    assert <<220, 51, 45, 150, 11, 157, 177, 120, 76, 168>> <> _ = block_hash
```
pnowosie: Does it makes any difference checking the first 10 bytes instead of full value? It does not. 
I'm against checking any kind of such values so when we have a test where value of the hash is important for compliance I'm opting to compare on full value.

apps/omg/lib/omg/state/transaction/payment.ex
```diff
@@ -118,9 +118,12 @@ defmodule OMG.State.Transaction.Payment do
          do: {:ok, outputs}
   end
 
-  defp reconstruct_metadata([]), do: {:ok, @zero_metadata}
-  defp reconstruct_metadata([metadata]) when Transaction.is_metadata(metadata), do: {:ok, metadata}
-  defp reconstruct_metadata([_]), do: {:error, :malformed_metadata}
+  # txData is required to be zero in the contract, zero comes out as `""` from `ExRLP.decode/1`
+  defp check_tx_data(""), do: :ok
```
pnowosie: Since it's `uint256` we can check the value of `RawData.parse_uint256(tx_data) == 0` and hide the weirdness of RLP from here

apps/omg/test/omg/state/transaction_test.exs
```diff
@@ -46,9 +46,7 @@ defmodule OMG.State.TransactionTest do
   end
 
   test "raw transaction hash is invariant" do
-    assert Transaction.raw_txhash(@transaction) ==
-             <<117, 35, 30, 107, 221, 12, 246, 144, 217, 144, 147, 52, 82, 0, 223, 240, 60, 26, 79, 183, 28, 57, 22,
-               132, 169, 12, 58, 233, 132, 52, 131, 219>>
+    assert <<21, 94, 181, 22, 125, 2, 47, 124, 113>> <> _ = Transaction.raw_txhash(@transaction)
```
pnowosie: As above if the value is important for the test please leave it as whole

apps/omg/lib/omg/state/transaction/payment.ex
pdobacz: ah ok, I realized this earlier too and was thinking that your PR would help here :sweat_smile:. I'll try to do that if not too much hassle

apps/omg/test/omg/state/core_test.exs
pdobacz: so this is not compliance (conformance) really, just "invariance", so a test that goes red whenever you touch hashing, so that you know that you touched hashing. This convention is followed in other tests like this (albeit some consistency more would help - separate cleanup PRs pending).

I just thought when changing this line anyway, I'll adapt to the (I think) cleanest convention out there. WDYT?

kasima:
:raised_hands:

pdobacz:
for our future selves, this bump in fact bumped to `835d10cbe8d7ef56feb745f71de638f85f572f7d`, fixing the title. Sorry about the confusion

Issue #1211, opened 2019-12-16
Update ENVs to allow deployment on multiple Ethereum networks


jbunce:
```
grep -r RINKEBY_AUTHORITY_ADDRESS .                            
                                                            
./docker-compose-watcher.yml:      - RINKEBY_AUTHORITY_ADDRESS=0x3272b97b7f1b74b338cb0fdda167cf76bc4da3b6
./docker-compose-watcher.yml:      - RINKEBY_AUTHORITY_ADDRESS=0x3272b97b7f1b74b338cb0fdda167cf76bc4da3b6
./docs/deployment_configuration.md:- "RINKEBY_AUTHORITY_ADDRESS"
./apps/omg_eth/test/omg_eth/release_tasks/set_contract_test.exs:      :ok = System.put_env("RINKEBY_AUTHORITY_ADDRESS", "authority_address_value")
./apps/omg_eth/test/omg_eth/release_tasks/set_contract_test.exs:      :ok = System.delete_env("RINKEBY_AUTHORITY_ADDRESS")
./apps/omg_eth/test/omg_eth/release_tasks/set_contract_test.exs:      :ok = System.put_env("RINKEBY_AUTHORITY_ADDRESS", "Authority_address_value")
./apps/omg_eth/test/omg_eth/release_tasks/set_contract_test.exs:      :ok = System.delete_env("RINKEBY_AUTHORITY_ADDRESS")
./apps/omg_eth/test/omg_eth/release_tasks/set_contract_test.exs:      :ok = System.put_env("RINKEBY_AUTHORITY_ADDRESS", "authority_address_value")
./apps/omg_eth/test/omg_eth/release_tasks/set_contract_test.exs:      :ok = System.delete_env("RINKEBY_AUTHORITY_ADDRESS")
./apps/omg_eth/test/omg_eth/release_tasks/set_contract_test.exs:      :ok = System.put_env("RINKEBY_AUTHORITY_ADDRESS", "authority_address_value")
./apps/omg_eth/test/omg_eth/release_tasks/set_contract_test.exs:      :ok = System.delete_env("RINKEBY_AUTHORITY_ADDRESS")
./apps/omg_eth/test/omg_eth/release_tasks/set_contract_test.exs:    :ok = System.put_env("RINKEBY_AUTHORITY_ADDRESS", "authority_address_value")
[...]

PR #1216, opened 2019-12-18
feat: Allow services to be deployed on multiple networks

jbunce:
## Overview

Updates the ENVs taken by the service to account for deployment on multiple networks. Added support for Ropsten and Mainnet.

## Testing

CI


jbunce:
Heads up @kendricktan this is going to force an update of our Helm charts :)

jbunce:
https://github.com/omisego/elixir-omg/issues/1211

jbunce:
Land after https://github.com/omisego/elixir-omg/pull/1214

docs/deployment_configuration.md
```diff
@@ -41,21 +41,21 @@ We allow a static configuration or a dynamic one, served as a http endpoint (one
 Static configuration
 
 - "ETHEREUM_NETWORK" - RINKEBY or LOCALCHAIN
```
InoMurko: update this doc pls!

InoMurko:
> Codewise, LGTM. But before merging, was there a specific reason for the `LOCALCHAIN_`, etc. prefix? @InoMurko

No 🙃

jbunce:
>was there a specific reason for the LOCALCHAIN_, etc. prefix?

This was legacy configuration from when we used `launcher.py` to start the services

InoMurko:
> > Codewise, LGTM. But before merging, was there a specific reason for the `LOCALCHAIN_`, etc. prefix? @InoMurko
> 
> No 🙃

docker-compose.yml
```diff
@@ -65,12 +65,12 @@ services:
     command: "full_local"
     environment:
       - ETHEREUM_NETWORK=LOCALCHAIN
-      - LOCALCHAIN_AUTHORITY_ADDRESS=0xc0f780dfc35075979b0def588d999225b7ecc56f
-      - LOCALCHAIN_TXHASH_CONTRACT=0xc47317b0de4c6ccc9a9fb858ad3ee858c2ee8ade3aed0fa9bcab7f7dad27fe12
-      - LOCALCHAIN_CONTRACT_ADDRESS_PLASMA_FRAMEWORK=0xc673e4ffcb8464faff908a6804fe0e635af0ea2f
-      - LOCALCHAIN_CONTRACT_ADDRESS_ETH_VAULT=0x0433420dee34412b5bf1e29fbf988ad037cc5db7
-      - LOCALCHAIN_CONTRACT_ADDRESS_ERC20_VAULT=0x04badc20426bc146453c5b879417b25029fa6c73
-      - LOCALCHAIN_CONTRACT_ADDRESS_PAYMENT_EXIT_GAME=0x1d92a9bef49bd340c7a11da536fb3ac9efbb1806
+      - AUTHORITY_ADDRESS=0xc0f780dfc35075979b0def588d999225b7ecc56f
```
pdobacz: heads up: when you rebase on `master`, this is gone and you'll only make the change once in a `.env` file added next to this `.yml` file.

apps/omg_eth/lib/omg_eth/release_tasks/set_contract.ex
```diff
@@ -84,24 +84,30 @@ defmodule OMG.Eth.ReleaseTasks.SetContract do
   end
 
   defp apply_static_settings(network) do
-    network =
+    _network =
```
pdobacz: since now you don't need the network var, you can put this whole bit (`case`) in a function like `verify_network_name` (if we want such safeguard to be in place ofc). It can return `:ok` on success and do this `exit(@error)` on failure, to not complicate/change things too much

docker-compose.yml
jbunce: Thanks

pdobacz: oh, wait, I think I maybe miswrote. `localchain_contract_addresses.env` should be enough to be modified, no need to readd those values here, see https://github.com/omisego/elixir-omg/pull/1216/files#diff-4e5e90c6228fd48698d074241c2ba760R67

PR #1220, opened 2019-12-19
streamline contracts updating and build speedup

InoMurko:
/

## Overview

@pdobacz gave me an idea with his introduction of `localchain_contract_addresses.env` file and I took that a step further by consolidating all our outside dependencies to contracts into that file.

So if you now need to change anything related to our contract dependencies, you only need to do it in `localchain_contract_addresses.env` and it'll get picked up by everyone. 
I have introduced `SNAPSHOT` variable in our Makefile, that allows you to override your snapshot with
`SNAPSHOT="your url, maybe the one with 120 seconds min exit period" make init_test`
## Changes

- Change how we install erlang and elixir for barebone builds because asdf is recompiling everything and is slow
- Make tests and others pull contract data from the env. variable file ^^
- update Makefile for pulling contract snapshots

## Testing

Same tests!


apps/omg_eth/test/fixtures.exs
```diff
@@ -38,16 +38,36 @@ defmodule OMG.Eth.Fixtures do
 
   deffixture contract(eth_node) do
     :ok = eth_node
+    contract_addreses = Path.join([File.cwd!(), "../../", "localchain_contract_addresses.env"])
+
+    addresses_to_parse = [
+      "LOCALCHAIN_AUTHORITY_ADDRESS",
```
jbunce: `LOCALCHAIN_AUTHORITY_ADDRESS ` -> `AUTHORITY_ADDRESS`

InoMurko: good catch. tnx!

apps/omg/test/omg/conformance/signature_test.exs
```diff
@@ -109,4 +109,27 @@ defmodule OMG.Conformance.SignatureTest do
 
     assert solidity_hash == OMG.TypedDataHash.hash_struct(tx)
   end
+
+  defp parse_contracts() do
```
pdobacz: I'd leave a comment to a new reader of this. There used to be `taken from the `plasma-contracts` deployment snapshot` there. A single sentence justifying this operation.

Oh, and I'd leave a hint that there might be several implementations of `parse_contracts`, so that someone who goes about modifying this bit of code doesn't create a fork by accident. (or actually just put it somewhere under `OMG.Eth.SomeModule`, I think this is justified)

apps/omg/test/omg/conformance/signature_test.exs
```diff
@@ -45,8 +45,8 @@ defmodule OMG.Conformance.SignatureTest do
   setup_all do
     {:ok, exit_fn} = Support.DevNode.start()
 
-    # taken from the `plasma-contracts` deployment snapshot
-    signtest_addr_hex = "0x19925cc645720fbb61f76304ee15501e3197f3a9"
+    parsable_data = parse_contracts()
```
pdobacz: a nitpick: `parsable_data` as a result from `parse` looks like a leftover name

bin/variables
```diff
@@ -9,12 +9,12 @@ export ETHEREUM_RPC_URL=http://127.0.0.1:8545
...
-export AUTHORITY_ADDRESS=0xc0f780dfc35075979b0def588d999225b7ecc56f
-export TXHASH_CONTRACT=0xc47317b0de4c6ccc9a9fb858ad3ee858c2ee8ade3aed0fa9bcab7f7dad27fe12
-export CONTRACT_ADDRESS_PLASMA_FRAMEWORK=0xc673e4ffcb8464faff908a6804fe0e635af0ea2f
-export CONTRACT_ADDRESS_ETH_VAULT=0x0433420dee34412b5bf1e29fbf988ad037cc5db7
-export CONTRACT_ADDRESS_ERC20_VAULT=0x04badc20426bc146453c5b879417b25029fa6c73
-export CONTRACT_ADDRESS_PAYMENT_EXIT_GAME=0x1d92a9bef49bd340c7a11da536fb3ac9efbb1806
 export DATABASE_URL=postgresql://omisego_dev:omisego_dev@127.0.0.1:5432/omisego_dev
 export CHILD_CHAIN_URL=http://127.0.0.1:9656
 export ETHEREUM_EVENTS_CHECK_INTERVAL_MS=800
+# expects it's executed from the root of the project
```
pdobacz: `sh` magic :tophat:

localchain_contract_addresses.env
```diff
@@ -4,3 +4,8 @@ CONTRACT_ADDRESS_PLASMA_FRAMEWORK=0xc673e4ffcb8464faff908a6804fe0e635af0ea2f
 CONTRACT_ADDRESS_ETH_VAULT=0x0433420dee34412b5bf1e29fbf988ad037cc5db7
 CONTRACT_ADDRESS_ERC20_VAULT=0x04badc20426bc146453c5b879417b25029fa6c73
 CONTRACT_ADDRESS_PAYMENT_EXIT_GAME=0x92ce4d7773c57d96210c46a07b89acf725057f21
+CONTRACT_SHA=835d10cbe8d7ef56feb745f71de638f85f572f7d
```
jbunce: Have you considered moving these into ENVs in Circle? I know it's slightly less deterministic because once an ENV is added you cannot view it, but it would enable changes to be made without a PR/rebasing etc.

Makefile
```diff
@@ -164,7 +165,8 @@ build-test: deps-elixir-omg
 init_test:
 	mkdir data/ || true && \
 	rm -rf data/* || true && \
-	wget https://storage.googleapis.com/circleci-docker-artifacts/data-elixir-omg-tester-plasma-deployer-dev-551a2b5-MIN_EXIT_PERIOD-20-PLASMA_CONTRACTS_SHA-835d10cbe8d7ef56feb745f71de638f85f572f7d.tar.gz -O data/snapshot.tar.gz && \
+	URL=$$(grep "^$(SNAPSHOT)" localchain_contract_addresses.env | cut -d'=' -f2-) && \
```
unnawut: :mindblown:

localchain_contract_addresses.env
unnawut: I think having these info here helps pinning/tracing elixir-omg code base to the contracts it's intended to support though. Very useful for local developments. 🎉

InoMurko: yeah I think @unnawut is right, it would also mean that not anyone could bump contracts but only those that have admin rights on circleci or they'd have to ask for someone else to do it :|

InoMurko:
> Awesome idea, though can we consolidate this with a single .env? I imagine a committed .env here is for a localchain development(therefore the default development env) and other envs we need could be called like [.test.env](http://www.virtuouscode.com/2014/01/17/dotenv-for-multiple-environments/)

Elixir 1.9 brings some interesting changes on how to deal with configuration. It's def. an interesting idea, but for the simple reason of not doing it again when we introduce elixir 1.9 ... lets bump it down the road.

localchain_contract_addresses.env
unnawut: I just saw @jbunce's PR on another repo which seems similar. Looks like we already have it in elixir-omg: https://github.com/omisego/elixir-omg/blob/master/.circleci/ci_deploy.sh#L45

PR #1223, opened 2019-12-20
fix: remove obsolete envs that got re-introduced from parallel branches

unnawut:
## Overview

Remove obsolete docker-compose env vars which now should reside in `localchain_contract_addresses.env`.

## Changes

The env vars were moved to `localchain_contract_addresses.env` in [#1214](https://github.com/omisego/elixir-omg/pull/1214/files#diff-4e5e90c6228fd48698d074241c2ba760).

But there was also changes in parallel in https://github.com/omisego/elixir-omg/commit/40f899bc32797bc1ff3f286aef2b4c11f586af5c#diff-4e5e90c6228fd48698d074241c2ba760 and git took this as line additions, so the removed lines were added back.

This PR re-removed those obsolete envs.

## Testing

`docker-compose up` continues to work without errors.

I also checked that the removed values are the same as in `localchain_contract_addresses.env` except for `CONTRACT_ADDRESS_PAYMENT_EXIT_GAME`, but the one in `localchain_contract_addresses.env` aligns with https://github.com/omisego-images/docker-elixir-omg/releases so it should be okay.


unnawut:
@nicholasmueller had the problem firing up docker-compose and this change solved it

PR #1224, opened 2019-12-20
feat: add creating_txhash and spending_txhash to /transaction.get and /transaction.all txoutputs

### LINKED ISSUE #1162: Add creating_tx_hash to /transaction.get's inputs and outputs
Part of the user story for the block explorer is to be able to visualise the details of a given transaction, which includes a breakdown of participating inputs and outputs. 

For a given input or output, we would like to navigate to the transaction that created it. In order to do this, we would like `/transaction.get` to include `creating_transaction` in the **input** and **output** objects. 

Something along the below lines. Only change to current behaviour is `creating_tx`

``` 
"inputs": [
      {
        "amount": 499999999999992300,
        "blknum": 91000,
        "currency": "0x0000000000000000000000000000000000000000",
        "oindex": 1,
        "owner": "0xe4c8c1f49fe031f879278853a01634004022f2eb",
        "txindex": 0,
         "creating_tx": "0x5861ae7fdb18c9cc4d5c26e68c0aae5902108d70f456b34202009e42037e0dbc"
        "utxo_pos": 91000000000001
      }
    ],

```

[Issue #72](https://github.com/omisego/blockexplorer/issues/72) in Block Explorer.

unnawut:
Closes #1162

## Overview

Add `inputs.*.creating_txhash` and `outputs.*.spending_txhash` to `/transaction.get`, `/transaction.all` and `/transaction.create`

Originally intended for `/transaction.get` but since the serializers are shared between endpoints, why not let them be consistent.

## Changes

- Add `transaction_factory/0` and `txoutput_factory/0` test data factories
- Add `:creating_txhash` and `:spending_txhash` to `Serializer.Base.to_utxo/1`
- Refactored `/transaction.get` away from complex fixtures setup and towards test data factories

### Performance impact

Data is already there. This PR simply puts the data on to the API response so there should be minimal perf impact. Response size gets slightly larger.

## Testing

These too shall pass:

```elixir
mix test test/omg_watcher_rpc/web/controllers/transaction_test.exs
```

```shell
curl -X POST http://localhost:7534/transaction.all \
-H "Content-Type: application/json" \
-d '{
  "limit": 10,
  "page": 1
}' \
-v -w "\n"
```

```shell
curl -X POST http://localhost:7534/transaction.get \
-H "Content-Type: application/json" \
-d '{
  "id": "0xYour_transaction_hash"
}' \
-v -w "\n"
```

Example `/transaction.all` response:

```json
{
  "block": {
    "blknum": 2000,
    "eth_height": 2,
    "hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "timestamp": 200
  },
  "metadata": "0x014ec6769ad805c1",
  "txbytes": "0x01cdf26a0afa1b20",
  "txhash": "0x0000000000000000000000000000000000000000000000000000000000000009",
  "txindex": 0,
  "inputs": [
    {
      "amount": 100,
      "blknum": 1000,
      "creating_txhash": "0x0000000000000000000000000000000000000000000000000000000000000004",
      "currency": "0x0000000000000000000000000000000000000000",
      "oindex": 0,
      "owner": "0xbecd4b58796ecc9f81acdbb6b0d1d915f75c4c3a",
      "spending_txhash": "0x0000000000000000000000000000000000000000000000000000000000000009",
      "txindex": 0,
      "utxo_pos": 2000000000000
    },
    {
      "amount": 100,
      "blknum": 1000,
      "creating_txhash": "0x0000000000000000000000000000000000000000000000000000000000000002",
      "currency": "0x0000000000000000000000000000000000000000",
      "oindex": 0,
      "owner": "0xa0eaaec1048ca9b65410b4ef215665812465b899",
      "spending_txhash": "0x0000000000000000000000000000000000000000000000000000000000000009",
      "txindex": 0,
      "utxo_pos": 1000000000000
    }
  ],
  "outputs": [
    {
      "amount": 100,
      "blknum": 2000,
      "creating_txhash": "0x0000000000000000000000000000000000000000000000000000000000000009",
      "currency": "0x0000000000000000000000000000000000000000",
      "oindex": 0,
      "owner": "0x96a0d682d19324735022a91224c866aa58081e7e",
      "spending_txhash": null,
      "txindex": 0,
      "utxo_pos": 4000000000000
    },
    {
      "amount": 100,
      "blknum": 2000,
      "creating_txhash": "0x0000000000000000000000000000000000000000000000000000000000000009",
      "currency": "0x0000000000000000000000000000000000000000",
      "oindex": 0,
      "owner": "0xb4e1e6b3b30f5a5cc77ccb6ec00347854d68718b",
      "spending_txhash": null,
      "txindex": 0,
      "utxo_pos": 3000000000000
    }
  ]
}
```

Notice that inputs have both `creating_txhash` and `spending_txhash` values while outputs only have `creating_txhash` and `spending_txhash: nil`.


apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -29,11 +31,11 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
...
   @eth OMG.Eth.RootChain.eth_pseudo_address()
   @other_token <<127::160>>
-  @eth_hex @eth |> Encoding.to_hex()
-  @other_token_hex @other_token |> Encoding.to_hex()
+  @eth_hex Encoding.to_hex(@eth)
+  @other_token_hex Encoding.to_hex(@other_token)
   @default_data_paging %{"limit" => 200, "page" => 1}
 
-  describe "getting transaction by id" do
+  describe "/transaction.get" do
```
pdobacz: don't want to throw this at you for this PR, but we could standardize on this `describe` convention in the entire file, when you're at it. I looked and it seems straightforward, maybe 2 `describe` sections into one, since they share an endpoint.

unnawut: Would this work? https://github.com/omisego/elixir-omg/pull/1224/commits/75207a787f3c6866d3089a66d7a0caea5940136f

apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
```diff
@@ -108,8 +114,14 @@ defmodule OMG.Watcher.Integration.BlockGetterTest do
...
-    assert WatcherHelper.get_utxos(alice.addr) == WatcherHelper.get_exitable_utxos(alice.addr)
-    assert [] == WatcherHelper.get_utxos(alice.addr)
+    # Drop watcher_info's creating_txhash and spending_txhash before comparing to security's exitable utxos
+    utxos =
+      alice.addr
+      |> WatcherHelper.get_utxos()
+      |> Enum.map(fn utxo -> Map.drop(utxo, ["creating_txhash", "spending_txhash"]) end)
+
+    assert WatcherHelper.get_exitable_utxos(alice.addr) == utxos
+    assert WatcherHelper.get_utxos(alice.addr) == []
```
unnawut: This line is to align with https://github.com/lexmag/elixir-style-guide#exunit

> When asserting (or refuting) something with comparison operators (such as ==, <, >=, and similar), put the expression being tested on the left-hand side of the operator and the value you're testing against on the right-hand side.
>
> ```elixir
> # Bad
> assert "héllo" == Atom.to_string(:"héllo")
> 
> # Good
> assert Atom.to_string(:"héllo") == "héllo"
> ```

apps/omg_watcher_info/test/support/factory.ex
```diff
@@ -66,4 +73,49 @@ defmodule OMG.WatcherInfo.Factory do
...
+  Generates a transaction without any transaction output and no associated block.
+
+  To generate a transaction with closest data to production, consider associating the transaction
+  to a block and generating transaction outputs associated with this transaction.
+  """
+  def transaction_factory() do
+    %DB.Transaction{
+      txhash: sequence(:block_hash, fn seq -> <<seq::256>> end),
+      txindex: 0,
+      txbytes: :crypto.strong_rand_bytes(8),
```
pnowosie: I'm against draining entropy for no reason, if you share my point alternative could be 
`0..255 |> Enum.shuffle() |> Enum.take(32) |> :erlang.list_to_binary()`

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/account_test.exs
```diff
@@ -128,7 +128,13 @@ defmodule OMG.WatcherRPC.Web.Controller.AccountTest do
          }}
       ])
 
-      assert WatcherHelper.get_exitable_utxos(alice.addr) == WatcherHelper.get_utxos(alice.addr)
+      # Drop watcher_info's creating_txhash and spending_txhash before comparing to security's exitable utxos
```
pnowosie: Also same as before - fix it for good

unnawut: Addressed in https://github.com/omisego/elixir-omg/pull/1224/commits/0943dbe1fd8a9874f716e0be44f8c9c958ed9d0d

apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
```diff
@@ -68,7 +68,13 @@ defmodule OMG.Watcher.Integration.BlockGetterTest do
              %{"blknum" => ^block_nr}
            ] = WatcherHelper.get_utxos(alice.addr)
 
-    assert WatcherHelper.get_utxos(alice.addr) == WatcherHelper.get_exitable_utxos(alice.addr)
+    # Drop watcher_info's creating_txhash and spending_txhash before comparing to security's exitable utxos
```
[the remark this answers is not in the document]
unnawut: Addressed in https://github.com/omisego/elixir-omg/pull/1224/commits/0943dbe1fd8a9874f716e0be44f8c9c958ed9d0d

apps/omg_watcher_info/test/support/factory.ex
unnawut: Addressed in https://github.com/omisego/elixir-omg/pull/1224/commits/3649aa7c17b835428005deba8ad33e3ef85ab9c1

unnawut:
> Have you missed update the swagger specs?

Addressed in https://github.com/omisego/elixir-omg/pull/1224/commits/c2c9ae36513b495d787860035ef01ae8dcdf3b09. I need to get better at this 🤦‍♂

apps/omg_watcher_info/test/support/factory.ex
```diff
@@ -58,12 +58,69 @@ defmodule OMG.WatcherInfo.Factory do
...
+  @doc """
+  Block factory.
+
+  Generates a block in an incremental blknum of 1000, 2000, 3000, etc.
+  """
   def block_factory() do
     %DB.Block{
       blknum: sequence(:block_blknum, fn seq -> seq * 1000 end),
-      hash: sequence(:block_hash, fn seq -> <<seq::256>> end),
+      hash: insecure_random_bytes(32),
```
jrhite: I recommend changing back to using `sequence()` here.
```suggestion
      hash: sequence(:block_hash, fn seq -> <<seq::256>> end),
```

apps/omg_watcher_info/test/support/factory.ex
```diff
@@ -58,12 +58,69 @@ defmodule OMG.WatcherInfo.Factory do
...
+      txhash: sequence(:transaction_hash, fn seq -> <<seq::256>> end),
+      txindex: 0,
+      txbytes: insecure_random_bytes(32),
+      sent_at: DateTime.utc_now(),
+      metadata: insecure_random_bytes(32),
+      block: nil,
+      inputs: [],
+      outputs: []
+    }
+  end
```
jrhite: same as above. please avoid random test data and remove use of `insecure_random_bytes(32)` and `DateTime.utc_now()`

```suggestion
      def transaction_factory(attrs \\ nil) do
        # a default 'sent_at' DateTime to use if none is passed in via args
        {:ok, default_test_datetime, 0} = DateTime.from_iso8601("2019-12-12T11:11:11Z")

        sent_at = attrs[:sent_at] || default_test_datetime

        transaction = %OMG.WatcherInfo.DB.Transaction{
          txhash: sequence(:transaction_hash, fn seq -> <<seq::256>> end),
          txindex: sequence(:transaction_txindex, fn seq -> seq end),
          txbytes: sequence(:transaction_txbytes, fn seq -> <<seq::256>> end),
          sent_at: sequence(:transaction_sent_at, fn seq -> DateTime.add(sent_at, seq) end),
          metadata: sequence(:transaction_metadata, fn seq -> <<seq::256>> end),
          block: attrs[:block] || build(:block)
        }

        merge_attributes(transaction, attrs)
```

apps/omg_watcher_info/test/support/factory.ex
```diff
@@ -58,12 +58,69 @@ defmodule OMG.WatcherInfo.Factory do
...
+      owner: owner.addr,
+      amount: 100,
+      currency: @eth,
+      proof: insecure_random_bytes(32),
+      spending_tx_oindex: 0,
+      child_chain_utxohash: insecure_random_bytes(32),
+      creating_transaction: insert(:transaction),
+      spending_transaction: nil
+    }
+  end
```
jrhite: Generally, do we want the default test objects that the factories create to be:
1. consistent and valid with our business logic (meaning parent and/or children objects block <-> transaction <-> utxo may need to be created together via `build()` calls in the factories or using `params_with_assocs()`
2. or at a minimum the object can be inserted into the DB and not violate DB constraints although it might be business rules may be violated.

I think 1) seems like the most reasonable thing to grow the factories to produce as defaults. But only over time as more tests are moved over.

2) database insertable default object seems like a good baseline to start with though.

Just some things that caught my eye, if you're interested in putting them this PR...

1. The UTXO being created in this factory gets its `blknum` from `blknum: insert(:block).blknum`. But will that be the same blknum as the transaction that created it? `creating_transaction: insert(:transaction)` is called, but the default transaction factory set its `block` to nil.
2. Does `insert(:transaction)` work ok in this PR given the DB key constraints, but the nil transaction block?
3. Correct sequencing of `blknum, txindex, oindex`. Eg: if we want to create a deposit then is_deposit(blknum, txindex, oindex)` should return true. see [code](https://github.com/omisego/elixir-omg/blob/master/apps/omg/lib/omg/utxo/position.ex#L84-L86)

jrhite: ☝️ I'm thinking about this stuff in my PR too.

jrhite: One last thing I forgot to ask... why the use of `insert()` in the factory instead of `build()`.  ExMachina docs discourage that a bit.

To have the `block` and `transaction` objects built and inserted in the txoutput factory can you use `params_with_assocs(:txoutput)`. That worked well for me.

apps/omg_watcher_info/test/support/factory.ex
unnawut: Can you explain your rationale a bit?

apps/omg_watcher_info/test/support/factory.ex
unnawut: Disagreeing here. I think random test data and DateTime is fine.

unnawut:
> Main thing though is, I hope we can avoid introducing random test data into tests as that often leads to flaky tests. Due to the randomness of tests using random data, an exact test failure or success is not repeatable. I don't see the random values in this PR directly causing problems, but who knows....and the random values don't seem to provide any specific benefits here, so I'd rather the test data be predictable and the test runs reproducible.

I have to disagree here. Random test data in the cases I use for this PR is to mimick closer to production value. E.g. the hashes. It should be beneficial if we see flaky tests arising from these random data.

unnawut:
> (also got me wondering if ExMachina provided utilities for generating sequences, but omitted random value utilities on purpose)

Because that random function can be used/built easily from existing Elixir core library.

apps/omg_watcher_info/test/support/factory.ex
unnawut: On your first comment, I was thinking the same too but chose the 2nd approach because different use cases demand data to be constructed in different ways. E.g. a spent utxo will have both `creating_txhash` and `spending_txhash`, then you'll need to construct two transactions. While an unspent utxo has only `creating_txhash` and only one transaction is needed, etc. etc.

So in this case I chose to let each tests determine its own needs while the factories take care of inserted defaults that the tests can chose not to care about.

If these primitive factories are built with in "consistent, valid with our business logic" then we end up with highly-coupling fixtures again.

I'm supportive of building more derived factories on top to fit certain use cases though.

unnawut: > ExMachina docs discourage that a bit.

Can you point me to that part? I'm having trouble finding it.

apps/omg_watcher_info/test/support/factory.ex
unnawut: Nevermind I saw your comment after!

unnawut:
> Does insert(:transaction) work ok in this PR given the DB key constraints, but the nil transaction block?

Interesting this does work per current code.

@pnowosie Seeing you're the original committer for [20180813131706_create_transaction_table.exs](https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher_info/priv/repo/migrations/20180813131706_create_transaction_table.exs), do you have some context on the TODO there?

unnawut:
> The UTXO being created in this factory gets its blknum from blknum: insert(:block).blknum. But will that be the same blknum as the transaction that created it? creating_transaction: insert(:transaction) is called, but the default transaction factory set its block to nil.

> Correct sequencing of blknum, txindex, oindex. Eg: if we want to create a deposit then is_deposit(blknum, txindex, oindex)` should return true.

I'm supportive of finding a better middleground between business-valid vs primitive factories. Agreeing they're a little too primitive. But there's not enough test code using these factories to experiment I think. Happy to revisit this say after /deposit.all is merged though. 🎉

apps/omg_watcher_info/test/support/factory.ex
jrhite: https://hexdocs.pm/ex_machina/readme.html#ecto-associations
> Using insert/2 in factory definitions may lead to performance issues and bugs, as records will be saved unnecessarily.

In terms of your idea:
> So in this case I chose to let each tests determine its own needs while the factories take care of inserted defaults that the tests can chose not to care about.

I was thinking about this as well and was able to get along just fine using only `build()` and `params_with_assocs(:txoutput)` which lets us build a txoutput and not care about the parent `transaction` about and the `transaction`'s parent object `block`. I does require taking full control of the factory though to make sure the blknum of the txoutput and the blknum of the `creating_transaction` are the same.

If I want to specifically override the `txoutput`'s `transaction` and/or `block`, using attr overrides can be used. Eg:

`build(:txoutput,: creating_transaction: my_creating_txn)` 
or using pipes `build(:txoutput) |> with_creating_txn(...) |> with_spending_txn(...)`

There is also the question with the code, specifically:
```
    %DB.TxOutput{
      blknum: insert(:block).blknum,
      <snip>
      creating_transaction: insert(:transaction),
      <snip>
    }
```

From a business logic perspective the `creating_transaction` above should have a `blknum` that is equal to the txoutput's `blknum`. But building it this way, it's not clear that that is true?

jrhite:
> > Main thing though is, I hope we can avoid introducing random test data into tests as that often leads to flaky tests. Due to the randomness of tests using random data, an exact test failure or success is not repeatable. I don't see the random values in this PR directly causing problems, but who knows....and the random values don't seem to provide any specific benefits here, so I'd rather the test data be predictable and the test runs reproducible.
> 
> I have to disagree here. Random test data in the cases I use for this PR is to mimick closer to production value. E.g. the hashes. It should be beneficial if we see flaky tests arising from these random data.

`txbytes` and `metadata` are not random data and have a specific encoded format, so I not sure how well random data mimics it.

Also if we want to make use random data are we drawing a clear line of when and where to use it? `owner: owner.addr` could also be random.

jrhite:
Summary here from my perspective...about using random data in these types of tests or not, I feel ok to agree to disagree for now. Definitely not a blocking.

The only blocker for me that I'd still like to see addressed are the comments about the factories... https://github.com/omisego/elixir-omg/pull/1224#discussion_r360647704

If get can align on that, then this PR is 👍 .

unnawut:
> `txbytes` and `metadata` are not random data and have a specific encoded format, so I not sure how well random data mimics it.

In the context of `WatcherInfo.DB` and downstream usage, I think they're pretty random. E.g. we don't & shouldn't be deriving data from those fields. If needed, those derivations need to be done before it enters `WatcherInfo.DB`. So for these tests I think it's closer to random binary.

However, if we're intending these factories to be used out of `WatcherInfo.DB`, you have a good point which also defeats the use of both sequential and random data 😅 I don't think that's the plan though. If we do so then we'll need new factories independent of `OMG.WatcherInfo.DB.*` schemas so I think we can keep it the way it is.

unnawut:
> The only blocker for me that I'd still like to see addressed are the comments about the factories... #1224 (comment)

Agreed on using `build` and avoid `insert` within factory implementations. 👍 Updating now.

apps/omg_watcher_info/test/support/factory.ex
unnawut: How about https://github.com/omisego/elixir-omg/pull/1224/commits/3a450bf94f08c3a4e8985fe37e52d12346733624 now? The txoutput_factory should now have consistent block and transaction.

apps/omg_watcher_info/test/support/factory.ex
```diff
@@ -98,21 +98,28 @@ defmodule OMG.WatcherInfo.Factory do
...
+  Generates a txoutput. A new block and transaction are generated per each txoutput built.
+  Therefore, if you are overriding some values, also consider its relation to other values. E.g:
+
+    - To override `blknum`, also consider overriding `txindex`.
+    - To override `creating_transaction`, also consider overriding `txindex` and `oindex`.
+    - To override `spending_transaction`, also consider overriding `spending_tx_oindex`
   """
   def txoutput_factory() do
+    block = build(:block)
+    transaction = build(:transaction, blknum: block.blknum)
```
jrhite: Because of this mapping in `db/transaction.ex`
`belongs_to(:block, DB.Block, foreign_key: :blknum, references: :blknum, type: :integer)`

I think `transaction = build(:transaction, block: block)` is a more natural and consistent use of the factories rather than `blknum: block.blknum`

You decide what to do for you now, it's not blocking.

unnawut: Yup this code did fail.

PR #1228, opened 2019-12-20
fix: CompileError ExMachina.Ecto is not loaded on dev

### LINKED ISSUE #1229: CompileError module ExMachina.Ecto is not loaded and could not be found on dev
@mederic-p and @pnowosie reported:

```
==> omg_watcher_info
Compiling 26 files (.ex)
== Compilation error in file test/support/factory.ex ==
** (CompileError) test/support/factory.ex:57: module ExMachina.Ecto is not loaded and could not be found
    (elixir) expanding macro: Kernel.use/2
    test/support/factory.ex:57: OMG.WatcherInfo.Factory (module)
    (elixir) lib/kernel/parallel_compiler.ex:208: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/6
```

@mederic-p's investigation and proposal:

> It introduces the use of ExMachina (https://github.com/omisego/elixir-omg/pull/1192/files#diff-fe97b7849956d06fb1edb6d35b79d23fR57), the problem is that with the current config in `mix.exs` of the `watcher_info` app, the `test/support` folder is compiled for all environment except `:prod` which conflicts with the deps config that is only imported for `:test` env.
> So the issue now is that we can’t run a normal `iex -S mix run --no-start` (`:dev` env by default), it can’t compile `OMG.WatcherInfo.Factory` because ExMachina is used but not imported in the deps.
>
> I can see that for all `mix.exs` of this project we use
> ```
> defp elixirc_paths(:prod), do: ["lib"]
> defp elixirc_paths(_), do: ["lib", "test/support"]
>```
> Any special reason for this? and why we don’t use
> ```
> defp elixirc_paths(:test), do: ["lib", "test/support"]
> defp elixirc_paths(_), do: ["lib"]
> ```
> instead, which would fix the issue.
>
> I believe that “test/support" is only required when running tests.

unnawut:
Closes #1229

## Overview

Be more specific on the compilation paths for `:prod`, `:dev`, `:test` per each subapp.

### Workaround

Meanwhile waiting for this PR to merge, a quick workaround is to find the 2 occurrences of:

```elixir
{:ex_machina, "~> 2.3", only: [:test], runtime: false}
```

and replace with:

```elixir
{:ex_machina, "~> 2.3", only: [:test, :dev], runtime: false}
```

## Changes

The subapp approach threw lots of cyclic dependency because `Support.*` depend a lot on `OMG.Eth.*` functionalities. So I go with allowing `"test/support"` compilation on `:dev` only for `OMG` and `OMG.Eth` where most `Support.*` are residing:

```elixir
defp elixirc_paths(:dev), do: ["lib", "test/support"]
defp elixirc_paths(:test), do: ["lib", "test/support"]
```

While the rest are more or less:

```elixir
defp elixirc_paths(:dev), do: ["lib"]
defp elixirc_paths(:test), do: ["lib", "test/support"]
```

This way we're scoping down `"test/support"` in `:dev` to a minimum and `:ex_machina` can still be limited to `only: [:test]`.

Work is still needed to remove `"test/support"` from `:dev` but is not covered in this PR.

## Testing

These commands should pass:

```shell
mix compile
mix test

# Support.*
MIX_ENV=dev iex -S mix run --no-start -e "true = Code.ensure_loaded?(Support.DevHelper); IO.puts('ok'); System.halt()"
MIX_ENV=test iex -S mix run --no-start -e "true = Code.ensure_loaded?(Support.DevHelper); IO.puts('ok'); System.halt()"
MIX_ENV=prod iex -S mix run --no-start -e "false = Code.ensure_loaded?(Support.DevHelper); IO.puts('ok'); System.halt()"

# ExMachina.*
MIX_ENV=dev iex -S mix run --no-start -e "false = Code.ensure_loaded?(ExMachina); IO.puts('ok'); System.halt()"
MIX_ENV=test iex -S mix run --no-start -e "true = Code.ensure_loaded?(ExMachina); IO.puts('ok'); System.halt()"
MIX_ENV=prod iex -S mix run --no-start -e "false = Code.ensure_loaded?(ExMachina); IO.puts('ok'); System.halt()"
```


pdobacz:
yeah, I am not sure how to resolve this. The other side of the mirror is that there is code in `test/support` that is useful when doing various stuffs in `iex -S mix`. From slack:

----------

**piotr  1 hour ago**
While `elixirc_paths` excluding `test/support` in `dev` sounds clean and reasonable, it would limit the usefulness of running various scripts from `iex`. 
(ex_machina naturally doesn't belong in `dev`, so I see the point of not liking the workaround)

scripts are: ad hoc hacking and experimenting scripts . They're using things like root_chain_helper (also used in tests) and so forth. Probably also scripts to run `omg_performance` tests will be affected in `:dev`

**unnawut  24 minutes ago**
Can we always run `omg_performance` in `:test`? If so then we don’t need to compile `"test/support"` in `:dev`.

**piotr  < 1 minute ago**
well `test` is gauged (in some aspects like config) towards testing, `dev` is more `prod`-resembling but with all the non-prod, handy tools present. `omg_performance` likes `dev` much better, so do the ad-hoc scripts I sometimes use.

---------------

Having thought about this, if we don't care that an awful lot about cleanliness of the `dev` env, we could invite the possibility of having `test` specific code (like `ex_machina`) there - it will be useless, but do we care that much? Dunno.

unnawut:
1. Or we add a new compile path for dev & test...

    **Common elixir configuration:**
    - `lib` for application code -> compiled in `[:dev, :test, :prod]`
    - `test/support` for test support modules -> compiled in `[:test]`

    **So we add:**
    - `dev/support` for dev support modules -> compiled in `[:dev, :test]`

2. Or move those dev & test support modules into a new `:omg_dev` subapp. Add as a dependency with `only: [:dev, :test]`, and not including it in distillery configs.

pdobacz:
Both 1 and 2 look good and would work I think.

1/ seems slightly easier and "smaller"
2/ seems slightly cleaner and more standard

There's also
0/ just `ex_machina, only: [:dev, :test]` and yolo, dirtiest and simplest.

unnawut:
I'm trying the `:omg_dev` subapp approach and let you know how it turns out.

I'm avoiding 0/ because it'll be harder to fix down the road, and 1/ is just not conventional. But I might change my mind. 😅

unnawut:
It turns out the subapp approach throws lots of cyclic dependency because `Support.*` depend a lot on `OMG.Eth.*` functionalities.

So I stop here instead, with only `OMG` and `OMG.Eth`, where most `Support.*` reside in, having:

```elixir
defp elixirc_paths(:dev), do: ["lib", "test/support"]
defp elixirc_paths(:test), do: ["lib", "test/support"]
```

While the rest are more or less:

```elixir
defp elixirc_paths(:dev), do: ["lib"]
defp elixirc_paths(:test), do: ["lib", "test/support"]
```

This way we're scoping down `"test/support"` in `:dev` to a minimum and `:ex_machina` can still be limited to `only: [:test]`.

unnawut:
@pdobacz @pnowosie Can you help trying this PR out? I don't have visibility to what scripting is used so I can't test it thoroughly. Or you can let me know what you need as well.

unnawut:
Also if we're doing some scripting, I'd like to wrap them under some kind of tests so we can make sure it gets covered by new changes.

PR #1218, opened 2019-12-19
Consolidate output and input utxo behaviour

achiurizo:
:clipboard:  https://github.com/omisego/elixir-omg/pull/1212  larger scoped PR

## Overview

This is a first pass at integrating our [ex_plasma]() library to consolidate our encoding/decoding protocol with the plasma framework contracts. Currently, the codebase has begun to future code the ALD behavior. This reverts that and encourages the protocol related changes to be moved over to the library.

This is multi-stage integration. This first pass starts offloading the input utxo calculation and cleans up the structure to use more of `ex_plasma` data modeling. This also consolidates the output utxo modules a bit too. Following the discussion on #1212, this is intentionally a _smaller scope_ as we move towards the integration of said library.

Follow up tasks:

- [ ] refactor output utxo behavior/calculations and leverage `ExPlasma.Utxo`
- [ ] refactor transactions to use `ExPlasma.Transaction` , and it's other types like `Payment` and `Deposit`.
- [ ] port Fees transaction to `ExPlasma` and leverage that.
- [ ] Remove dispatching module code, like `WireTransactionTypes`. This behavior should exist in the library.

## Changes

* Consolidates:

  * `OMG.InputPointer`  - this module "*represents the data that's used to mention outputs that are intended to be inputs to a transaction*"
  * `OMG.InputPointer.UtxoPosition` this "*represents the utxo position interpreted as an implementation of an input pointer type*"
  * `OMG.InputPointer.Protocol` - this "*captures the varying behavior of how the outputs can be pointed to in child chain transactions*"

  into `OMG.Utxo.Position`, a pre-existing module that the modules above invoke in the call stack. This is a "*representation of a UTXO position in the child chain, providing encoding/decoding to/from formats digestible in 'Eth' and in the `OMG.DB`*".

  Consolidating it into `Utxo.Position` seems like the best first pass as all the above modules end up [calling functions](https://github.com/omisego/elixir-omg/compare/achiurizo/ex_utxo_2?expand=1#diff-64563a0f252706476c8cc34790e20564L30) in `Utxo.Position` [in their call stack](https://github.com/omisego/elixir-omg/compare/achiurizo/ex_utxo_2?expand=1#diff-417dc52abc6b8f820e190f8726f725d7L27). Once we've merged the module behaviors, we can start delegating the functionality to `ex_plasma` more. Currently there are some behaviors that need to be aligned and refactored(`OMG.Output` error tuple behavior is one such example)

[...]

PR #1240, opened 2019-12-28
Refactor OMG.Output guards and usage

achiurizo:
:clipboard: #1218 the first refactor series

## Overview

This refactors `OMG.Output` to de-tangle some of its coupling with the rest of the code. This is a small refactor to pave wave for a larger refactor of the `OMG.Utxo`, `OMG.Utxo.Position`, and `OMG.OUtput` tree.

## Changes

* Move guards into modules that use them. The guards that exist often are only used once. We should move them to where they are used.
* Leverage ExPlasma to do the RLP data -> struct conversion.
* Break out the error tuple generation into it's own method. Eventually we'll port this to `ex_plasma` to handle returning error messages around decoding.
* Remove `dispatching_reconstruct`. We only have 1 output to generate. For the future output models, we should be building these encoding structures into `ex_plasma`.
* Add test files for output and utxo. Currently not testing much but next PR will fill out the cases for `OMG.Utxo`.

## Testing

```elixir
mix test
mix test --only integration
# have CI run
```

### DIFF
--- a/apps/omg/lib/omg/output.ex
+++ b/apps/omg/lib/omg/output.ex
@@ -58,6 +58,7 @@ defmodule OMG.Output do
 
   defp validate_data([raw_type, [owner, currency, amount]]) do
     with {:ok, _} <- RawData.parse_uint256(raw_type),
+         {:ok, _} <- valid_output_type?(raw_type),
          {:ok, _} <- RawData.parse_address(owner),
          {:ok, _} <- non_zero_owner?(owner),
          {:ok, _} <- RawData.parse_address(currency),
@@ -67,4 +68,7 @@ defmodule OMG.Output do
 
   defp non_zero_owner?(<<0::160>>), do: {:error, :output_guard_cant_be_zero}
   defp non_zero_owner?(_), do: {:ok, :valid}
+
+  defp valid_output_type?(<<1>>), do: {:ok, :valid}
+  defp valid_output_type?(_), do: {:error, :unrecognized_output_type}
 end
--- a/apps/omg/lib/omg/output.ex
+++ b/apps/omg/lib/omg/output.ex
@@ -41,7 +41,7 @@ defmodule OMG.Output do
     end
   end
 
-  def reconstruct(rlp_data), do: {:error, :malformed_outputs}
+  def reconstruct(_), do: {:error, :malformed_outputs}
 
   def from_db_value(%{owner: owner, currency: currency, amount: amount, output_type: output_type})
       when is_binary(owner) and is_binary(currency) and is_integer(amount) and is_integer(output_type) do
--- a/apps/omg/lib/omg/output.ex
+++ b/apps/omg/lib/omg/output.ex
@@ -38,33 +38,17 @@ defmodule OMG.Output do
   @output_types_modules OMG.WireFormatTypes.output_type_modules()
   @output_types Map.keys(@output_types_modules)
 
-  def reconstruct([raw_type, rlp_decoded_chunks]) when is_binary(raw_type) do
-    case RawData.parse_uint256(raw_type) do
-      {:ok, output_type} when output_type in @output_types ->
-        reconstruct([output_type, rlp_decoded_chunks])
-
-      {:ok, _unrecognized_type} ->
-        {:error, :unrecognized_output_type}
-    end
-  end
-
   @doc """
   Reconstructs the structure from a list of RLP items
   """
-  def reconstruct([output_type, [owner_rlp, currency_rlp, amount_rlp]]) do
-    with {:ok, cur12} <- RawData.parse_address(currency_rlp),
-         {:ok, owner} <- RawData.parse_address(owner_rlp),
-         :ok <- non_zero_owner(owner),
-         {:ok, amount} <- RawData.parse_amount(amount_rlp),
-         do: %__MODULE__{output_type: output_type, owner: owner, currency: cur12, amount: amount}
+  def reconstruct([raw_type, [_owner, _currency, _amount]] = rlp_data) when is_binary(raw_type) do
+    with :ok <- validate_data(rlp_data) do
+      utxo = ExPlasma.Utxo.new(rlp_data)
+      %__MODULE__{output_type: utxo.output_type, owner: utxo.owner, currency: utxo.currency, amount: utxo.amount}
+    end
   end
 
-  def reconstruct(_), do: {:error, :malformed_outputs}
-
-  defp non_zero_owner(<<0::160>>), do: {:error, :output_guard_cant_be_zero}
-  defp non_zero_owner(_), do: :ok
-
-  def dispatching_reconstruct(_), do: {:error, :malformed_outputs}
+  def reconstruct(rlp_data), do: {:error, :malformed_outputs}
 
   def from_db_value(%{type: output_type} = db_value), do: @output_types_modules[output_type].from_db_value(db_value)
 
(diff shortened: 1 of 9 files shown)


apps/omg/lib/omg/output.ex
```diff
@@ -30,64 +31,44 @@ defmodule OMG.Output do
...
-        reconstruct([output_type, rlp_decoded_chunks])
-
-      {:ok, _unrecognized_type} ->
-        {:error, :unrecognized_output_type}
-    end
-  end
-
   @doc """
   Reconstructs the structure from a list of RLP items
   """
```
InoMurko: are there no dialyzer specs available?

apps/omg/lib/omg/output.ex
```diff
@@ -30,64 +31,44 @@ defmodule OMG.Output do
...
-  defp non_zero_owner(<<0::160>>), do: {:error, :output_guard_cant_be_zero}
-  defp non_zero_owner(_), do: :ok
-
-  def dispatching_reconstruct(_), do: {:error, :malformed_outputs}
-
-  def from_db_value(%{type: output_type} = db_value), do: @output_types_modules[output_type].from_db_value(db_value)
+  def reconstruct(rlp_data), do: {:error, :malformed_outputs}
 
   def from_db_value(%{owner: owner, currency: currency, amount: amount, output_type: output_type})
       when is_binary(owner) and is_binary(currency) and is_integer(amount) and is_integer(output_type) do
```
InoMurko: what is the point of these is_binary/1 checks?

apps/omg/lib/omg/state/core.ex
```diff
@@ -425,7 +424,7 @@ defmodule OMG.State.Core do
     |> Transaction.get_outputs()
     |> Enum.with_index()
     |> Enum.map(fn {output, oindex} ->
-      {Output.input_pointer(output, blknum, tx_index, oindex, tx, hash), output}
+      {Utxo.position(blknum, tx_index, oindex), output}
```
achiurizo: This usage here was interesting. Since this only cares for the utxo position data, I'm not sure why we created a separate method just capture the other data and not use it. 🤔 If there's a reason we can revisit, but simplifying it to just do what it was doing underneath:

```elixir
def input_pointer(%__MODULE__{}, blknum, tx_index, oindex, _, _),
    do: Utxo.position(blknum, tx_index, oindex)
```

apps/omg/lib/omg/state/transaction/validator.ex
```diff
@@ -78,8 +78,10 @@ defmodule OMG.State.Transaction.Validator do
   defp authorized?(outputs_spent, witnesses, raw_tx) do
     outputs_spent
     |> Enum.with_index()
-    |> Enum.map(fn {output_spent, idx} -> OMG.Output.can_spend?(output_spent, witnesses[idx], raw_tx) end)
+    |> Enum.map(fn {output_spent, idx} -> can_spend?(output_spent, witnesses[idx]) end)
```
InoMurko: nice

apps/omg/lib/omg/utxo.ex
```diff
@@ -36,25 +36,7 @@ defmodule OMG.Utxo do
...
-           when is_tuple(position) and tuple_size(position) == 4 and
-                  is_position(elem(position, 1), elem(position, 2), elem(position, 3)) and
-                  rem(elem(position, 1), @interval) != 0
-
-  defmacrop is_nil_or_binary(binary) do
-    quote do
-      is_binary(unquote(binary)) or is_nil(unquote(binary))
-    end
-  end
+  defguardp is_nil_or_binary(creating_tx_hash) when is_nil(creating_tx_hash) or is_binary(creating_tx_hash)
```
InoMurko: looking at where this guard is used I don't understand what it's guarding us from and why? do you know @achiurizo ? what's the purpose of this guard?

apps/omg/lib/omg/output.ex
achiurizo: @InoMurko no there weren't dialyzer's availlable. In the follow up PR to this, I'll add them to `OMG.Utxo`. The end goal is to consolidate this back into that module.

apps/omg/lib/omg/utxo.ex
achiurizo: I think the guard is being overly defensive here. But to minimize the amount of changes, I left the guard in until a follow up PR.

apps/omg/lib/omg/output.ex
achiurizo: I think it's overly defensive here. What we can do in a later PR is to ensure that we validate user input and db persistence of data to handle all these types of checks.

apps/omg/lib/omg/output.ex
InoMurko: ok

apps/omg/lib/omg/utxo.ex
InoMurko: ok

apps/omg/lib/omg/output.ex
```diff
@@ -30,64 +31,44 @@ defmodule OMG.Output do
...
+         {:ok, _} <- non_zero_owner?(owner),
+         {:ok, _} <- RawData.parse_address(currency),
+         {:ok, _} <- RawData.parse_amount(amount),
+         do: :ok
+  end
+
+  defp non_zero_owner?(<<0::160>>), do: {:error, :output_guard_cant_be_zero}
+  defp non_zero_owner?(_), do: {:ok, :valid}
+
+  defp valid_output_type?(<<1>>), do: {:ok, :valid}
```
pdobacz: I think `output_fee_token_claim: 2` was a valid output type catered for earlier (only partially though, since fee support is WIP still). If this lands in current form anyway, cc @pnowosie please make sure to re-add the support.

apps/omg/lib/omg/state/core.ex
pdobacz: To reiterate: this caters(-ed) for the output-type to input-pointer-type binding as part of https://github.com/omisego/elixir-omg/issues/844. The implementation you're putting in instead will work, it's just specific and not compatible e.g. with `output_id` input pointer, support of which used to be required (or at least impending). It also couples the `State.Core` to a specific way outputs can be referenced, so similar comments as with #1212 apply here.

apps/omg/lib/omg/output.ex
```diff
@@ -30,64 +31,44 @@ defmodule OMG.Output do
...
-
   def to_db_value(%__MODULE__{owner: owner, currency: currency, amount: amount, output_type: output_type})
       when is_binary(owner) and is_binary(currency) and is_integer(amount) and is_integer(output_type) do
     %{owner: owner, currency: currency, amount: amount, output_type: output_type}
   end
 
   def get_data_for_rlp(%__MODULE__{owner: owner, currency: currency, amount: amount, output_type: output_type}),
     do: [output_type, [owner, currency, amount]]
+
+  defp validate_data([raw_type, [owner, currency, amount]]) do
```
pdobacz: I think this function could use a comment on why are we abandoning the results here, in favor of the results of the call to `ex_plasma` (if this is temporary and will be fixed, a `TODO` should do).

apps/omg_watcher/lib/omg_watcher/api/utxo.ex
```diff
@@ -41,7 +45,7 @@ defmodule OMG.Watcher.API.Utxo do
   end
 
   @spec compose_utxo_exit(Utxo.Position.t()) :: {:ok, exit_t()} | {:error, :utxo_not_found}
-  def compose_utxo_exit(utxo_pos) when is_deposit(utxo_pos) do
+  def compose_utxo_exit({:utxo_position, blknum, _, _} = utxo_pos) when is_deposit(blknum) do
```
pdobacz: verbose use of a `{:utxo_position, ...}` should be avoided in favor of `Utxo.position(...)`, even if this is an intermediate result.

achiurizo:
> aren't we moving away from macro usage in function headers?

@InoMurko yep, but going to pull those changes into a follow up PR. cc @pdobacz

PR #1244, opened 2020-01-07
Automatically generate fee txs in each block

pnowosie:
Fixes omisego/feat-fee#28
Closes omisego/feat-fee/issues/55

## Overview

* Calculates fee amounts available to claim from pending transactions
* Generate FeeTokenClaim transaction at `form_block` and execute them on state which checks the fee amounts again and adds fee-transactions at the end of the block
* Make sure fee-tx will not be accepted when simply submitted via API

## Changes

Most significant changes were made in `State.Core` and `State.Transaction.Validator` modules to collect surpluses and allow fee tx go through `exec` function.

Other changes allows fee tx to be recovered from encoded/binary form.
Prevents fee tx to be submitted directly to ch-ch 

## Testing

Extended `State.Core` module tests. 
Other tests or test fixes.


apps/omg/test/omg/state/core_test.exs
```diff
@@ -897,6 +898,148 @@ defmodule OMG.State.CoreTest do
...
+    setup do
+      fee_claimer = OMG.TestHelper.generate_entity()
+
+      {:ok, child_block_interval} = OMG.Eth.RootChain.get_child_block_interval()
+      {:ok, state} = Core.extract_initial_state(0, child_block_interval, fee_claimer.addr)
+
+      alice = OMG.TestHelper.generate_entity()
+      fees = %{@eth => %{amount: 2}}
+
+      # Transaction requires fee of 2wei ETH but Alice actually is paying 3wei
```
mederic-p: Note that this should fail once @T-Dnzt prevent over-payment

apps/omg/config/config.exs
```diff
@@ -6,7 +6,8 @@ config :omg,
   deposit_finality_margin: 10,
   ethereum_events_check_interval_ms: 8_000,
   coordinator_eth_height_check_interval_ms: 6_000,
-  metrics_collection_interval: 60_000
+  metrics_collection_interval: 60_000,
+  fee_claimer_address: "NO FEE CLAIMER ADDR!"
```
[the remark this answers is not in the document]
InoMurko: if this is a configuration value that needs to be set before the application is deployed I suggest using `ReleaseTasks` (don't forget to let devops know of this requirement and updating the docs)

apps/omg/lib/omg/state.ex
```diff
@@ -92,9 +96,11 @@ defmodule OMG.State do
   def init(:ok) do
     {:ok, height_query_result} = DB.get_single_value(:child_top_block_number)
     {:ok, child_block_interval} = Eth.RootChain.get_child_block_interval()
+    fee_claimer_address = Application.get_env(:omg, :fee_claimer_address, @no_fee_claimer_address)
```
InoMurko: if it's not to big of a burden, the proper pattern is to pass the configuration values from the supervisor down, not read it in the configuration.

apps/omg/lib/omg/state/transaction/fee_token_claim.ex
```diff
@@ -114,14 +136,28 @@ defimpl OMG.State.Transaction.Protocol, for: OMG.State.Transaction.FeeTokenClaim
   def get_inputs(%Transaction.FeeTokenClaim{}), do: []
 
   @doc """
-  Fee claiming transaction is not used to transfer funds
+  Tells whether Fee claiming transaction is valid
   """
   @spec valid?(Transaction.FeeTokenClaim.t(), Transaction.Signed.t()) :: {:error, atom()}
```
InoMurko: the error tuple is not good enough. needs to be more restrictive.

apps/omg/lib/omg/state/transaction/fee_token_claim.ex
```diff
@@ -114,14 +136,28 @@ defimpl OMG.State.Transaction.Protocol, for: OMG.State.Transaction.FeeTokenClaim
...
+         true <- output.amount > 0 || {:error, :fee_output_amount_has_to_be_positive},
+         do: true
+  end
 
   @doc """
   Fee claiming transaction is not used to transfer funds
   """
-  @spec can_apply?(Transaction.FeeTokenClaim.t(), list(Output.t())) :: {:error, atom()}
-  def can_apply?(%Transaction.FeeTokenClaim{}, _outputs_spent), do: {:error, :not_implemented}
+  @spec can_apply?(Transaction.FeeTokenClaim.t(), list(Output.t())) :: {:ok, map()} | {:error, atom()}
```
InoMurko: error tuple could be more restrictive

apps/omg_child_chain/lib/child_chain.ex
```diff
@@ -66,6 +67,11 @@ defmodule OMG.ChildChain do
     |> result_with_logging()
   end
 
+  defp is_supported(%Transaction.Recovered{signed_tx: %Transaction.Signed{raw_tx: raw_tx}})
+       when is_map(raw_tx) do
+    Map.get(raw_tx, :__struct__) in [Transaction.Payment]
```
InoMurko: is there a better approach to accomplish the same result?

apps/omg/config/config.exs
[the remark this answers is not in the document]
pnowosie: Good point @T-Dnzt @InoMurko - especially Release tasks!
We also have demo/play-with mode where one can set `:fees_not_required` instead of `fee json`. Because surpluses are collected any way regardless this setting there should be some non-zero `fee claimer` address (can't be zero because of validation rules when tx gets recovered from binary)... For now I can make this config value obligatory.

apps/omg/lib/omg/state/transaction/validator.ex
```diff
@@ -17,7 +17,8 @@ defmodule OMG.State.Transaction.Validator do
   Provides functions for stateful transaction validation for transaction processing in OMG.State.Core.
   """
 
-  @maximum_block_size 65_536
+  @maximum_fee_transactions 128
```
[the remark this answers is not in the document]
pnowosie: Ouch, it slipped through - isn't used anywhere prpbly

apps/omg/lib/omg/state/transaction/validator.ex
```diff
@@ -36,27 +37,71 @@ defmodule OMG.State.Transaction.Validator do
...
+  @spec can_process(state :: Core.t(), tx :: Transaction.Recovered.t(), fees :: Fees.optional_fee_t()) ::
+          {:ok, :apply_spend, map()} | {:ok, :claim_fees, map()} | {{:error, process_error()}, Core.t()}
+  def can_process(
+        %Core{} = state,
+        %Transaction.Recovered{signed_tx: %{raw_tx: raw_tx}} = tx,
         fees
       ) do
+    case raw_tx do
+      %Transaction.Payment{} ->
+        can_apply_spend(state, tx, fees)
```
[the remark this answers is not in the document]
pnowosie: Yeah it also makes dialyzer unhappy - I'll try to get it working

apps/omg/lib/omg/state/transaction/fee_token_claim.ex
```diff
@@ -43,18 +43,40 @@ defmodule OMG.State.Transaction.FeeTokenClaim do
...
+  @spec claim_collected(
+          blknum :: non_neg_integer(),
+          owner :: Crypto.address_t(),
+          fees_paid :: %{Crypto.address_t() => pos_integer()}
+        ) :: list(t())
+  def claim_collected(blknum, owner, fees_paid) do
+    fees_paid
+    |> Enum.reject(fn {_token, amount} -> amount == 0 end)
+    |> Enum.map(fn {token, amount} -> new(blknum, {owner, token, amount}) end)
+    |> Enum.sort_by(fn %__MODULE__{outputs: [output]} -> output.currency end)
```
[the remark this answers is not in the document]
pnowosie: Please note that these transactions are included in block hash calculation. If I'm correct `Map` keys order are unspecified (above `Enum.map`) so I don't want to leave it to a chance.

apps/omg/test/omg/state/core_test.exs
pnowosie: I guess there would be still some margin :) We'll see :D

apps/omg_child_chain/lib/child_chain.ex
pnowosie: Yeah, I thing this may be part of the `Transaction.Protocol` so we don't miss it implementing new tx types.

apps/omg/lib/omg/state/core.ex
```diff
@@ -161,16 +183,25 @@ defmodule OMG.State.Core do
   """
   @spec exec(state :: t(), tx :: Transaction.Recovered.t(), fees :: Fees.optional_fee_t()) ::
           {:ok, {Transaction.tx_hash(), pos_integer, non_neg_integer}, t()}
-          | {{:error, Validator.exec_error()}, t()}
+          | {{:error, Validator.process_error()}, t()}
```
pdobacz: naming nitpick for this and `can_process` function. I fear of overloading the process word - is this an erlang process, exit process, process process...

The renamings I'd propose:
`Validator.can_apply_spend` -> `Validator.can_apply_tx`
`Validator.exec_error` -> `Validator.can_apply_error`
`State.Core.apply_spend` -> `State.Core.apply_tx`

apps/omg/lib/omg/state/core.ex
```diff
@@ -161,16 +183,25 @@ defmodule OMG.State.Core do
...
   def exec(%Core{} = state, %Transaction.Recovered{} = tx, fees) do
     tx_hash = Transaction.raw_txhash(tx)
 
-    case Validator.can_apply_spend(state, tx, fees) do
-      true ->
+    case Validator.can_process(state, tx, fees) do
+      {:ok, :apply_spend, fees_paid} ->
```
pdobacz: I'm not a fan of the Validator telling State.Core how to apply the tx like this. But I see where this is coming from, hmmm...

apps/omg/lib/omg/state/core.ex
```diff
@@ -228,7 +263,9 @@ defmodule OMG.State.Core do
         height: height + child_block_interval,
         pending_txs: [],
         utxo_db_updates: [],
-        recently_spent: MapSet.new()
+        recently_spent: MapSet.new(),
+        fees_paid: %{},
```
pdobacz: wouldn't these two final updates fit better under `claim_fees/1`?

apps/omg/lib/omg/state/core.ex
```diff
@@ -388,6 +425,44 @@ defmodule OMG.State.Core do
...
+  defp collect_fees(%Core{fees_paid: fees_paid} = state, tx_diffs) do
+    tx_fees =
+      tx_diffs
+      |> Enum.reject(fn {_token, amount} -> amount == 0 end)
+      |> Map.new()
+
+    %Core{
+      state
+      | fees_paid:
+          Map.merge(fees_paid, tx_fees, fn
```
pdobacz: split a variable out `updated_fees_paid` or `fees_paid_with_new` or sth.

apps/omg/lib/omg/state/core.ex
```diff
@@ -388,6 +425,44 @@ defmodule OMG.State.Core do
     }
   end
 
+  defp collect_fees(%Core{fees_paid: fees_paid} = state, tx_diffs) do
+    tx_fees =
+      tx_diffs
```
pdobacz: rename `tx_diffs` to sth more specific... `token_surplussed` or `collectable_fee_per_token`.

apps/omg/lib/omg/state/core.ex
```diff
@@ -388,6 +425,44 @@ defmodule OMG.State.Core do
...
+      | fees_paid:
+          Map.merge(fees_paid, tx_fees, fn
+            _token, collected, tx_paid -> collected + tx_paid
+          end)
+    }
+  end
+
+  defp disallow_payments(state), do: %Core{state | fee_claiming_started: true}
+
+  defp claim_token(state, token), do: %Core{state | fees_paid: Map.delete(state.fees_paid, token)}
```
pdobacz: naming doesn't click for me here. Do `claim_token` and `claim_fees` have the same meaning to the word "claim"? Also more specifically maybe `claim_token` -> `flush_collected_fees_for_token`. A bit longish, but this is what it is

apps/omg/lib/omg/state/core.ex
```diff
@@ -161,16 +183,25 @@ defmodule OMG.State.Core do
...
+         |> add_pending_tx(tx)
+         |> collect_fees(fees_paid)}
+
+      {:ok, :claim_fees, claimed_token} ->
         {:ok, {tx_hash, state.height, state.tx_index},
          state
          |> apply_spend(tx)
-         |> add_pending_tx(tx)}
+         |> add_pending_tx(tx)
+         |> claim_token(claimed_token |> Map.keys() |> hd())
```
pdobacz: so this `keys |> hd` looks like aftermath from the overloading of `Transaction.Protocol.can_apply?`.

Having thought about this a bit, I think you should abide by what `Transaction.Protocol.can_apply?` originally required the implementors to return, being `{:ok, fees paid per token}`. This is an abstract thing and let's not touch it, i.e. return 0 for `FeeClaimTransaction`s

Now to be able to do what you want, consider putting the fee-claiming pipeline outside of this `case Validator.can_process` pipeline.

So:
1. the `Validator` told you to `apply_spend` and `add_pending_tx` (which now happens uniformly for all tx types, so no need to branch _here_
2. if this happened to be a Fee transaction then do these two new steps (`claim_token` & `disallow_payments`)

In other words, handle the fee transactions as a post-processing step instead of bloating the `Transaction.Protocol` abstraction

apps/omg/lib/omg/state/transaction/fee_token_claim.ex
```diff
@@ -114,14 +136,28 @@ defimpl OMG.State.Transaction.Protocol, for: OMG.State.Transaction.FeeTokenClaim
...
   @doc """
   Fee claiming transaction is not used to transfer funds
   """
-  @spec can_apply?(Transaction.FeeTokenClaim.t(), list(Output.t())) :: {:error, atom()}
-  def can_apply?(%Transaction.FeeTokenClaim{}, _outputs_spent), do: {:error, :not_implemented}
+  @spec can_apply?(Transaction.FeeTokenClaim.t(), list(Output.t())) :: {:ok, map()} | {:error, atom()}
+  def can_apply?(%Transaction.FeeTokenClaim{outputs: [claimed]}, outputs) do
+    with %Output{} = collected <-
+           Enum.find(outputs, {:error, :surplus_in_token_not_collected}, fn o -> o.currency == claimed.currency end),
+         true <- collected.amount == claimed.amount || {:error, :claimed_and_collected_amounts_mismatch},
```
pdobacz: question: is this the `error` that will pop up in the byzantine event of operator attempting to drain too much of fees in a block?

apps/omg/lib/omg/state/transaction/payment.ex
```diff
@@ -227,13 +227,17 @@ defimpl OMG.State.Transaction.Protocol, for: OMG.State.Transaction.Payment do
...
   end
 
-  defp fees_paid(input_amounts_by_currency, output_amounts_by_currency) do
-    input_amounts_by_currency
-    |> Enum.into(%{}, fn {input_currency, input_amount} ->
-      # fee is implicit - it's the difference between funds owned and spend
-      implicit_paid_fee = input_amount - Map.get(output_amounts_by_currency, input_currency, 0)
-      {input_currency, implicit_paid_fee}
-    end)
+  defp input_output_diffs(input_amounts_by_currency, output_amounts_by_currency) do
```
pdobacz: similar here, `diffs` is too generic here. If the old name was not ok, then consider just `surplus_by_token`. Also observe that the naming _within_ the function remain's in the "fee world" so the end effect is a bit strange. Maybe easiest is to revert to `fees_paid`, this isn't that bad :thinking:

apps/omg/lib/omg/state/transaction/signed.ex
```diff
@@ -77,6 +80,8 @@ defmodule OMG.State.Transaction.Signed do
             |> Enum.into(%{}, fn {witness, idx} -> {idx, witness} end)}
   end
 
+  defp get_witnesses_from_raw_tx(%Transaction.FeeTokenClaim{}, _raw_witnesses), do: {:ok, %{}}
```
pdobacz: hm, is this branching on the tx type necessary here? if the FeeTokenClaim doesn't provide any witnesses, it should result in `{:ok, %{}}`, doesn't it? This module should at all cost remain raw_tx_type agnostic

apps/omg/lib/omg/state/transaction/validator.ex
```diff
@@ -36,27 +37,71 @@ defmodule OMG.State.Transaction.Validator do
...
-          true | {{:error, exec_error()}, Core.t()}
-  def can_apply_spend(
-        %Core{utxos: utxos} = state,
-        %Transaction.Recovered{signed_tx: %{raw_tx: raw_tx}, witnesses: witnesses} = tx,
+  @type fee_claim_error :: :claiming_unsupported_token | :claiming_more_than_collected
+
+  @type process_error :: exec_error | fee_claim_error
+
+  @doc """
+  can process transaction?
```
pdobacz: this doc doesn't tell us much more than the function name, let's either be more helpful here or drop altogether.

A more useful doc would say sth like "Checks, whether at a given state of the ledger, a particular transaction can be applied (!) to it, subject to particular fee requirements"

apps/omg/lib/omg/state/transaction/validator.ex
```diff
@@ -36,27 +37,71 @@ defmodule OMG.State.Transaction.Validator do
...
+  end
+
+  @spec can_claim_fees(Core.t(), Transaction.Recovered.t()) ::
+          {:ok, :claim_fees} | {{:error, fee_claim_error()}, Core.t()}
+  defp can_claim_fees(
+         %Core{fee_claimer_address: owner, fees_paid: fees_paid} = state,
+         %Transaction.Recovered{signed_tx: %{raw_tx: fee_tx}}
+       ) do
+    with outputs = make_outputs(owner, fees_paid),
+         {:ok, claimed_fee_token} <- Transaction.Protocol.can_apply?(fee_tx, outputs) do
```
pdobacz: oh, this looks like overloading of the return of `Transaction.Protocol.can_apply?` too much.

apps/omg/lib/omg/state/core.ex
pnowosie: Regarging `can_process` full fn signature use just once in `State.Core` looks quite OK
```Validator.can_process(state, tx, fees)```

I will rename to `can_process_tx`

File: apps/omg/lib/omg/state/transaction/signed.ex (hunk 19 above)
Comment (pnowosie) [reply apps/omg/lib/omg/state/transaction/signed.ex]: I witnessed issues pushing fee tx through the old pipeline. Will try to resolve differently.

File: apps/omg/lib/omg/state/transaction/validator.ex (hunk 8 above)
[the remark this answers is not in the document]
Comment (pnowosie) [reply apps/omg/lib/omg/state/transaction/validator.ex]: I guess this branching is ever needed for special fee claiming case, payments will go with the above payment pipeline. Indeed each transaction type provides its validator in `Transaction.Protocol.can_apply()`. Is it answers your concern @T-Dnzt :question:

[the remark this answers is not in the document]
Comment (T-Dnzt) [reply apps/omg/lib/omg/state/transaction/validator.ex]: Hmm, not sure I understand your comment. What I meant it to have `can_apply_tx` and `can_claim_fees` moved into two other modules and to keep `Validator` as more of a dispatcher calling the appropriate validation flow depending on the tx type.

File: apps/omg/lib/omg/state/transaction/fee_token_claim.ex (hunk 17 above)
Comment (pnowosie) [reply apps/omg/lib/omg/state/transaction/fee_token_claim.ex]: So there could be to cases:
* `{:error, :surplus_in_token_not_collected}`  - fee tx claiming the token that state knows nothing about,
* `{:error, :claimed_and_collected_amounts_mismatch}` - collected and claimed aren't equal (we have to claim exactly the amount of collected)

**Comment by pnowosie** [comment]
@InoMurko can you pls take a look particularly at the (last) commit changes `feat: pass fee claimer address from env var via release task`. Never done release task before.

(I known I need to fix the tests the change broke)

File: apps/omg/lib/omg/release_tasks/set_fee_claimer_address.ex [hunk 22]
```diff
@@ -0,0 +1,57 @@
...
+    _ = Application.ensure_all_started(:logger)
+
+    fee_claimer_address = get_fee_claimer_address()
+    Application.put_env(@app, :fee_claimer_address, fee_claimer_address, persistent: true)
+
+    :ok
+  end
+
+  defp get_fee_claimer_address do
+    address = validate_address(get_env("FEE_CLAIMER_ADDRESS"), @default_fee_claimer_address)
```
mederic-p: Can we pipe this?
```
"FEE_CLAIMER_ADDRESS"
|> get_env()
|> validate_address(@default_fee_claimer_address)
```

apps/omg/lib/omg/state/core.ex
```diff
@@ -388,6 +432,49 @@ defmodule OMG.State.Core do
...
+    }
+  end
+
+  defp disallow_payments(state), do: %Core{state | fee_claiming_started: true}
+
+  defp flush_collected_fees_for_token(state, %Output{currency: token}),
+    do: %Core{state | fees_paid: Map.delete(state.fees_paid, token)}
+
+  @spec claim_fees(state :: t()) :: t()
+  defp claim_fees(
```
pdobacz: um, wait a second, now after the cleanups I'm seeing sth strange here. Why aren't we calling `exec` from `OMG.State.form_block`? It looks like if the Watcher calls into here it will `exec` the fee transactions twice (or do sth comparably strange).

apps/omg/lib/omg/state/core.ex
```diff
@@ -388,6 +432,49 @@ defmodule OMG.State.Core do
...
+      |> Enum.map(fn fee_tx ->
+        Transaction.Signed.encode(%Transaction.Signed{raw_tx: fee_tx, sigs: []})
+      end)
+      |> Enum.reduce(state, fn rlp_tx, curr_state ->
+        {:ok, tx} = Transaction.Recovered.recover_from(rlp_tx)
+        {:ok, _, new_state} = exec(curr_state, tx, :no_fees_required)
+        new_state
+      end)
+
+    %Core{new_state | fees_paid: %{}, fee_claiming_started: false}
```
pdobacz: also `fees_paid: %{}` seems unnecessary at 1st glance - doesn't the `flush` function do that already?

apps/omg/lib/omg/state/transaction/fee_token_claim.ex
```diff
@@ -114,14 +136,30 @@ defimpl OMG.State.Transaction.Protocol, for: OMG.State.Transaction.FeeTokenClaim
...
+    # we're able to check structure validity => single output with amount > 0
+    with outputs = Transaction.get_outputs(fee_tx),
+         true <- length(outputs) == 1 || {:error, :wrong_number_of_fee_outputs},
+         [output] = outputs,
+         true <- output.amount > 0 || {:error, :fee_output_amount_has_to_be_positive},
+         do: true
+  end
 
   @doc """
   Fee claiming transaction is not used to transfer funds
```
pdobacz: this requires a comment on the `outputs` input - that it is expected to be prepared in line with the protocol interface. So we're having the `FeeTokenClaim` spend pseudo-outputs based on the `fees_paid`.

These explanations should be split between this `@doc` here and the spot where it's called

apps/omg/lib/omg/state/transaction/validator.ex
```diff
@@ -14,74 +14,49 @@
...
+  def can_process_tx(%Core{} = state, %Transaction.Recovered{} = tx, fees) do
+    with :ok <- validate_block_size(state), do: dispatch_validation(state, tx, fees)
   end
 
-  defp validate_block_size(%Core{tx_index: number_of_transactions_in_block}) do
-    case number_of_transactions_in_block == @maximum_block_size do
+  defp validate_block_size(%Core{tx_index: number_of_transactions_in_block, fees_paid: fees_paid}) do
+    fee_transactions_count = Enum.count(fees_paid)
+
+    case number_of_transactions_in_block + fee_transactions_count >= @maximum_block_size do
```
pdobacz: this might sound like a nitpick, but if `Enum.count(fees_paid) == 1` and there is a single spot "left" in the block, and the last transaction to hit this pays in a distinct token (making `count(fees_paid) == 2`) it looks like it will bork.

So you should prepare 2 slots and possibly "waste" one if `count` remains at 1

Would be nice to have a unit test for this :sweat_smile:

apps/omg/test/omg/state/core_test.exs
```diff
@@ -897,6 +898,133 @@ defmodule OMG.State.CoreTest do
            )
   end
 
+  describe "Automatic fees claiming" do
+    setup do
```
pdobacz: missing behaviors to test:
 - surpluses adding up for same-token-fees paid in a block
 - surpluses for > 1 token collected, can do 2 fee transactions after that
 - big block tests (see other comment)?
 - surpluses from multi-token outputs are all collected (alternative phrasing: all fees paid from multi-input/output transactions are claimable, so have 4in/4out and each use a different token? A bit of hassle, but I think this is required)
 - if surplus is zero (we have 2ins/2outs tx, both in fee token possibly, only one in/out pay the fee), it won't be collected _at all_ (checks the `Enum.filter(amount > 0)` out there somewhere)

apps/omg_child_chain/test/omg_child_chain/integration/happy_path_test.exs
```diff
@@ -244,6 +248,13 @@ defmodule OMG.ChildChain.Integration.HappyPathTest do
     assert {:error, %{"code" => "submit:utxo_not_found"}} = tx |> Transaction.Signed.encode() |> submit_transaction()
   end
 
+  @tag fixtures: [:alice, :in_beam_child_chain]
+  test "check submitted fee transaction is rejected", %{alice: alice} do
+    fee_tx = OMG.TestHelper.create_encoded_fee_tx(1000, alice.addr, @eth, 1000)
+
+    assert {:error, %{"code" => "submit:transaction_not_supported"}} = submit_transaction(fee_tx)
+  end
+
```
pdobacz: can I assume Watcher integration tests with fee-tx-rich blocks are pending a separate issue?

pnowosie: Yes, Watcher integration is the next task

apps/omg/lib/omg/state/core.ex
```diff
@@ -161,22 +185,38 @@ defmodule OMG.State.Core do
...
         error
     end
   end
 
+  # Post-processing step of transaction execution. It either collect fees from `Payment` transaction
+  # or claim fees with `FeeTokenClaim` transaction
+  @spec handle_fees(state :: t(), oper :: exec_mode(), Transaction.Recovered.t(), map()) :: t()
+  defp handle_fees(state, :apply_spend, _tx, fees_paid) do
+    state
+    |> collect_fees(fees_paid)
```
InoMurko:
```suggestion
    collect_fees(state, fees_paid)
```

apps/omg/lib/omg/state/core.ex
```diff
@@ -209,10 +249,14 @@ defmodule OMG.State.Core do
...
-        child_block_interval,
-        %Core{pending_txs: reversed_txs, height: height, utxo_db_updates: reversed_utxo_db_updates} = state
-      ) do
+  def form_block(child_block_interval, %Core{} = state) do
+    # important: `claim_fees` changes state significantly, overriding the parameter
+    %Core{
+      height: height,
+      pending_txs: reversed_txs,
+      utxo_db_updates: reversed_utxo_db_updates
+    } = state = claim_fees(state)
```
InoMurko: is this safe? maybe rather `new_state`?

apps/omg/lib/omg/state/transaction/fee_token_claim.ex
```diff
@@ -114,14 +136,30 @@ defimpl OMG.State.Transaction.Protocol, for: OMG.State.Transaction.FeeTokenClaim
...
   """
-  @spec can_apply?(Transaction.FeeTokenClaim.t(), list(Output.t())) :: {:error, atom()}
-  def can_apply?(%Transaction.FeeTokenClaim{}, _outputs_spent), do: {:error, :not_implemented}
+  @spec can_apply?(Transaction.FeeTokenClaim.t(), list(Output.t())) ::
+          {:ok, map()}
+          | {:error, :surplus_in_token_not_collected | :claimed_and_collected_amounts_mismatch}
+  def can_apply?(%Transaction.FeeTokenClaim{outputs: [claimed]}, outputs) do
```
InoMurko: maybe rewrite this into a more readable form.

apps/omg/test/support/test_helper.ex
```diff
@@ -99,6 +99,17 @@ defmodule OMG.TestHelper do
     create_signed(inputs, outputs) |> Transaction.Signed.encode()
   end
 
+  def create_encoded_fee_tx(blknum, owner, currency, amount) do
```
InoMurko: this is used only in `happy_path_test.exs `, why is this in a test_helper?

apps/omg/lib/omg/state.ex
```diff
@@ -33,12 +33,14 @@ defmodule OMG.State do
 
   require Utxo
 
-  @type exec_error :: Validator.exec_error()
+  @type exec_error :: Validator.can_process_tx_error()
+
+  @fallback_fee_claimer_address Application.fetch_env!(:omg, :fee_claimer_address)
```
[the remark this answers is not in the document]
pnowosie: Yes, it is. The other place setting this config is release task run only for ch-ch. So for other parties this value is irrelevant (Watcher, perftest, tests) it fallbacks to this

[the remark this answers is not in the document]
InoMurko: I don't think this is as safe as it should be. If someone omits the parameter from the supervisor, nothing will prevent this from working with the default parameter. 
Omg state should prevent booting if the parameter in child chain isn't there.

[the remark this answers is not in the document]
InoMurko: Would the child chain tests protect us from starting OMG.State without a correct fee claimer address?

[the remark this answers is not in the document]
pnowosie: I think I see your concern here. I can do this config value mandatory for the ch-ch in the release task. What I cannot see is how it affect running ch-ch in other scenarios like integration tests or running bare-metal / docker (in case the fallback option isn't possible).

I'm opting for creating an issue to provide this config better way. WDYT @InoMurko :question:

apps/omg/lib/omg/state/core.ex
pnowosie: This is very good catch sir :exclamation:

apps/omg/test/support/test_helper.ex
pnowosie: Good question :thinking:

InoMurko:
A test I would like to see and I think @achiurizo as well is a cabbage test that send a bunch of transactions between alice and bob and at the end you assert the balance of the `fee_claimer_address`.

achiurizo:
> A test I would like to see and I think @achiurizo as well is a cabbage test that send a bunch of transactions between alice and bob and at the end you assert the balance of the fee_claimer_address.

Second this. Cabbage has been great tool for integration. If you want some :eyes: or pairing, happy to spend time to get those additions in and help us move away from fixtures usage. :smile:  :pray:

pnowosie:
> Second this. Cabbage has been great tool for integration. If you want some eyes or pairing

Can't wait to put my hand on this... but review comments first. Let's complete it.

apps/omg/lib/omg/state.ex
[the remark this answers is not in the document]
InoMurko: Maybe implementing fallbacks is a trap we shouldn't be putting them in place.
I would rather see us updating the tests so that the correct value is passed.
I look at this as a VERY important setting. If this isn't set correctly - our software would continue working fine (there's no specific monitoring for the claimer fee address) but you'd be in production without generating any revenue...

apps/omg/lib/omg/state.ex
```diff
@@ -204,7 +208,12 @@ defmodule OMG.State do
   """
```
pdobacz: include the new responsibility (fees) in the description above.

apps/omg/test/omg/state/core_test.exs
```diff
@@ -924,6 +1174,7 @@ defmodule OMG.State.CoreTest do
   # used to check the invariants in form_block
   # use this throughout this test module instead of Core.form_block
   defp form_block_check(state) do
+    state = Core.claim_fees(state)
```
pdobacz: I think it's better to split it out and explicitly call `claim_fees` where applicable / expected.

apps/omg/test/omg/state/core_test.exs
```diff
@@ -897,6 +898,255 @@ defmodule OMG.State.CoreTest do
...
+        |> success?()
+
+      can_claim_fees = [
+        create_recovered_fee_tx(1000, state.fee_claimer_address, @eth, 2),
+        create_recovered_fee_tx(1000, state.fee_claimer_address, @not_eth, 3),
+        create_recovered_fee_tx(1000, state.fee_claimer_address, not_eth_1, 4),
+        create_recovered_fee_tx(1000, state.fee_claimer_address, not_eth_2, 5)
+      ]
+
+      %Core{pending_txs: pending_txs} = Core.claim_fees(state)
```
pdobacz: ekhm, internals

apps/omg/test/omg/state/core_test.exs
```diff
@@ -897,6 +898,255 @@ defmodule OMG.State.CoreTest do
...
+
+      # First tx is applied just to make below transactions generation easier
+      state =
+        state
+        |> do_deposit(alice, %{amount: amount_for_fees, currency: @eth, blknum: 1})
+        |> Core.exec(create_recovered([{1, 0, 0, alice}], @eth, [{alice, available_after_1st_tx}]), fees)
+        |> success?()
+
+      # sanity check
+      assert 1 == Enum.count(state.fees_paid)
```
pdobacz: internals again

apps/omg/test/omg/state/core_test.exs
```diff
@@ -897,6 +898,255 @@ defmodule OMG.State.CoreTest do
...
+      # but we have to claim the exact amount
+      state
+      |> Core.exec(create_recovered_fee_tx(1000, state.fee_claimer_address, @eth, collected + 1), fees)
+      |> fail?(:claimed_and_collected_amounts_mismatch)
+      |> Core.exec(create_recovered_fee_tx(1000, state.fee_claimer_address, @eth, collected - 1), fees)
+      |> fail?(:claimed_and_collected_amounts_mismatch)
+    end
+
+    # this test takes ~26 seconds on my machine
+    @tag skip: true
```
pdobacz: on 2nd thought, this is not unreasonable speed. We were doing our perf tests on multiple cores, this is just one here, and thats 65K ecrecovers there, 26 seconds makes sense :(.

I don't know what would be a quick fix here. I don't like a skipped test, 26 seconds of `mix test` is bad too. Proper fix would be to parametrize block size in `Core.extract...` saving 2**16 as a default for convenience, probably better put outside of the scope of this PR.

So my propo is this:
 - unskip
 - tag `:slow` so that `mix test --exclude slow` is possible for development, for the time being
 - (optionally, unless you just do it in a follow-up PR) put an issue to parametrize block size
 - describe in a NOTE here, or even better `Logger.warn("slow test is running, use --exclude slow to skip")`
 - don't forget about it :)

apps/omg/test/omg/state/core_test.exs
```diff
@@ -911,6 +913,227 @@ defmodule OMG.State.CoreTest do
...
+        state_empty
+        |> do_deposit(alice, %{amount: 10, currency: @eth, blknum: 1})
+        |> Core.exec(create_recovered([{1, 0, 0, alice}], @eth, [{alice, 8}]), fees)
+        |> success?()
+
+      {:ok, [state: state, alice: alice, fees: fees, fee_claimer: fee_claimer, state_empty: state_empty]}
+    end
+
+    test "should append fee txs in block", %{state: state} do
+      state = Core.claim_fees(state)
```
InoMurko: where does `form_block_check` come from? 
maybe we could start describing test steps and the intentions. For example:

The test starts with claiming and clearing any current fees.
_Not sure about form_block_check_
It continues to analyze all transactions in a block and asserts that it found a payment transaction and a fee transaction.

apps/omg/test/omg/state/core_test.exs
```diff
@@ -911,6 +913,227 @@ defmodule OMG.State.CoreTest do
...
+
+      %Transaction.Recovered{signed_tx: %Transaction.Signed{raw_tx: fee_tx}} =
+        Transaction.Recovered.recover_from!(fee_txbytes)
+
+      assert %Transaction.Payment{} = payment_tx
+      assert %Transaction.Fee{} = fee_tx
+    end
+
+    test "fee txs are appended even when fees aren't required", %{state: state} do
+      state = Core.claim_fees(state)
```
InoMurko: why do we append fee txs when fees aren't required? didn't we say we don't support `no-fees` anymore?

apps/omg/test/omg/state/core_test.exs
```diff
@@ -911,6 +913,227 @@ defmodule OMG.State.CoreTest do
...
+      {:ok, {block, _dbupdates}, _state} = form_block_check(state)
+
+      assert 2 = length(block.transactions)
+    end
+
+    test "should create utxos from claimed fee", %{alice: alice, state: state, fee_claimer: fee_claimer} do
+      state = Core.claim_fees(state)
```
InoMurko: bare with me here... why do we need to create UTXOs from claimed fees?

apps/omg/test/omg/state/core_test.exs
```diff
@@ -911,6 +913,227 @@ defmodule OMG.State.CoreTest do
...
+      {:ok, _, state} = form_block_check(state)
+
+      state
+      |> Core.exec(create_recovered([{1000, 1, 0, fee_claimer}], @eth, [{alice, 2}]), :no_fees_required)
+      |> success?()
+    end
+
+    test "fee txs cannot be intermixed with payments", %{alice: alice, state: state, fees: fees} do
+      fee_tx = create_recovered_fee_tx(1000, state.fee_claimer_address, @eth, 2)
```
InoMurko: what do you mean by intermixed? that they need to be the last transactions in a block but not in the middle?

apps/omg/test/omg/state/core_test.exs
```diff
@@ -911,6 +913,227 @@ defmodule OMG.State.CoreTest do
...
+      new_state
+      |> Core.exec(fee_tx, fees)
+      |> fail?(:surplus_in_token_not_collected)
+    end
+
+    test "fee is paid in one token only, many surpluses prohibited", %{alice: alice, state: state, fees: fees} do
+      state
+      |> do_deposit(alice, %{amount: 100, currency: @not_eth, blknum: 2})
+      |> Core.exec(
+        create_recovered([{1000, 0, 0, alice}, {2, 0, 0, alice}], [{alice, @eth, 5}, {alice, @not_eth, 90}]),
```
InoMurko: `create_recovered` is a very bad name for what it really does. looking at this no one could know it's really creating transactions with inputs and ouputs

apps/omg/test/omg/state/core_test.exs
```diff
@@ -911,6 +913,227 @@ defmodule OMG.State.CoreTest do
...
+
+      # we just send 1 payment and this reserves 1 spot for fee
+      already_reserved = 2
+
+      # available space is block_size
+      ntx_to_apply = maximum_block_size - (1 + maximum_inputs_size + already_reserved)
+
+      {state, _} =
+        0..ntx_to_apply
+        |> Enum.reduce({state, available_after_1st_tx}, fn idx, {curr_state, amount} ->
```
InoMurko:
```suggestion
        Enum.reduce(0..ntx_to_apply, {state, available_after_1st_tx}, fn idx, {curr_state, amount} ->
```

apps/omg/test/omg/state/core_test.exs
```diff
@@ -911,6 +913,227 @@ defmodule OMG.State.CoreTest do
...
+          new_state =
+            curr_state
+            |> Core.exec(create_recovered([{1000, idx, 0, alice}], @eth, [{alice, new_amount}]), fees)
+            |> success?()
+
+          {new_state, new_amount}
+        end)
+
+      state
+      # NOTE: I don't care about existing utxo actual position or available amount because block size is checked first
```
InoMurko: 👏

apps/omg/test/omg/state/persistence_test.exs
```diff
@@ -130,6 +132,38 @@ defmodule OMG.State.PersistenceTest do
              tx |> Transaction.get_inputs() |> hd() |> Utxo.Position.to_input_db_key() |> OMG.DB.spent_blknum()
   end
 
+  @tag fixtures: [:alice]
+  test "collected surpluses are persisted", %{alice: alice} do
+    token1 = <<1::160>>
+    token2 = <<2::160>>
+
+    tx =
+      create_recovered(
```
InoMurko: where does create_recovered come from? is it imported? pls declare imports.

apps/omg/test/omg/state/transaction/fee_test.exs
```diff
@@ -0,0 +1,94 @@
...
+  use ExUnit.Case, async: true
+
+  alias OMG.Output
+  alias OMG.State.Transaction
+
+  @eth OMG.Eth.RootChain.eth_pseudo_address()
+  @other_token <<127::160>>
+
+  describe "new/2" do
+    @tag fixtures: [:alice]
```
InoMurko: can we step back and stop using fixtures for new features and code? cc @achiurizo @unnawut

apps/omg/test/omg/state/transaction/fee_test.exs
```diff
@@ -0,0 +1,94 @@
...
+              %Transaction.Recovered{
+                signed_tx: %Transaction.Signed{raw_tx: ^fee_tx}
+              }} = Transaction.Recovered.recover_from(tx_rlp)
+    end
+  end
+
+  describe "claim_collected/3" do
+    @tag fixtures: [:alice]
+    test "no fees collected result in empty fee-txs list", %{alice: owner} do
+      assert [] == Transaction.Fee.claim_collected(1000, owner, %{})
```
InoMurko:
```suggestion
      assert Transaction.Fee.claim_collected(1000, owner, %{}) == []
```

apps/omg_child_chain/lib/omg_child_chain/release_tasks/set_fee_claimer_address.ex
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  defp validate_address("0x" <> value), do: validate_address(value)
+
+  defp validate_address(value) when is_binary(value) do
+    with {:ok, address} when byte_size(address) == 20 and address != @zero_address <-
+           value |> String.upcase() |> Base.decode16() do
+      address
+    else
+      :error -> exit("Fee claimer address has to be HEX-encoded string")
```
InoMurko: 👏

apps/omg_child_chain/lib/omg_child_chain/supervisor.ex
```diff
@@ -34,9 +34,10 @@ defmodule OMG.ChildChain.Supervisor do
     # prevent booting if contracts are not ready
     :ok = RootChain.contract_ready()
     {:ok, _contract_deployment_height} = RootChain.get_root_deployment_height()
+    fee_claimer_address = Application.fetch_env!(:omg, :fee_claimer_address)
```
InoMurko: put the configuration fetcing into `OMG.Configuration`

```
def fee_claimer_address() do
    Application.fetch_env!(:omg, :fee_claimer_address)
  end
```
`OMG.Configuration.fee_claimer_address()`

apps/omg_child_chain/test/omg_child_chain/integration/happy_path_test.exs
```diff
@@ -141,7 +142,10 @@ defmodule OMG.ChildChain.Integration.HappyPathTest do
       in_flight_tx =
       OMG.TestHelper.create_signed([{blknum, txindex, 0, alice}, {blknum, txindex, 1, alice}], @eth, [{alice, 8}])
 
-    proof = Block.inclusion_proof([Transaction.Signed.encode(tx)], txindex)
+    fee_claimer = Application.fetch_env!(:omg, :fee_claimer_address)
```
InoMurko: 🗜 
```
OMG.Configuration.fee_claimer_address()
```

docs/deployment_configuration.md
```diff
@@ -15,6 +15,7 @@
 - "BATCH_SIZE" - Datadog batch size for APM
 - "SYNC_THRESHOLD" - Datadog sync threshold for APM
 - "ETHEREUM_EVENTS_CHECK_INTERVAL_MS" - the frequency of HTTP requests towards the Ethereum clients and scanning for interested events. Should be less then average block time (10 to 20 seconds) on Ethereum mainnet.
+- "FEE_CLAIMER_ADDRESS" - [ch-ch only!] 20-bytes HEX-encoded string of Ethereum address of Fee Claimer. Set via ENV vars only for Child Chain. For the Watcher this values is not important and fallbacks to "DEAD000000000000000000000000000000000000".
```
InoMurko:
```suggestion
- "FEE_CLAIMER_ADDRESS" - [Child Chain only!] 20-bytes HEX-encoded string of Ethereum address of Fee Claimer. Set via ENV vars only for Child Chain. For the Watcher this values is not important and fallbacks to "DEAD000000000000000000000000000000000000".
```

priv/apps/itest/test/itest/fee_claiming_test.exs
```diff
@@ -0,0 +1,119 @@
...
+    {:ok, state}
+  end
+
+  defthen ~r/^Operator has claimed the fees$/, _, %{fees_initial_balance: initial_balance} do
+    actual_balance =
+      @fee_claimer_address
+      |> Client.get_balance()
+      |> fix_balance_response()
+
+    assert_equal(
```
InoMurko: wohooo!

priv/apps/itest/test/itest/fee_claiming_test.exs
```diff
@@ -0,0 +1,119 @@
...
+  defthen ~r/^"(?<entity>[^"]+)" should have "(?<amount>[^"]+)" ETH on the child chain$/,
+          %{entity: entity, amount: amount},
+          state do
+    entity_address = get_entity(state, entity).address
+    expecting_amount = Currency.to_wei(amount)
+
+    balance =
+      entity_address
+      |> Client.get_balance(expecting_amount)
+      |> fix_balance_response()
```
InoMurko: I would suggest, for the sake of consistency, that we handle this malformed response the same way as in other tests, even if not optimal
```
child_chain_balance =
      case amount do
        "0" ->
          assert Client.get_balance(address, Currency.to_wei(amount)) == []
          0

        _ ->
          %{"amount" => network_amount} = Client.get_balance(address, Currency.to_wei(amount))
          assert network_amount == Currency.to_wei(amount)
          network_amount
      end

```

or let's change it everywhere else the way you did it here.

priv/apps/itest/test/itest/fee_claiming_test.exs
```diff
@@ -0,0 +1,119 @@
...
+
+    _ = Client.submit_transaction(typed_data, sign_hash, [sender.pkey])
+
+    {:ok, state}
+  end
+
+  defthen ~r/^"(?<entity>[^"]+)" should have "(?<amount>[^"]+)" ETH on the child chain$/,
+          %{entity: entity, amount: amount},
+          state do
+    entity_address = get_entity(state, entity).address
```
InoMurko: I don't see how the helper below is making things more readable...
```suggestion
    entity_address = state[entity].address
```

apps/omg/lib/omg/state.ex
```diff
@@ -89,12 +90,15 @@ defmodule OMG.State do
   @doc """
   Initializes the state. UTXO set is not loaded now.
   """
-  def init(:ok) do
+  def init(opts) do
     {:ok, height_query_result} = DB.get_single_value(:child_top_block_number)
     {:ok, child_block_interval} = Eth.RootChain.get_child_block_interval()
 
+    {:ok, fee_claimer_address} = Keyword.fetch(opts, :fee_claimer_address)
```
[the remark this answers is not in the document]
pnowosie: Agree, more clear error msgs when value is missing

apps/omg/lib/omg/state/core.ex
```diff
@@ -122,16 +140,20 @@ defmodule OMG.State.Core do
...
+    state = %__MODULE__{
+      height: height_query_result + child_block_interval,
+      fee_claimer_address: fee_claimer_address
+    }
 
     {:ok, state}
   end
 
-  def extract_initial_state(:not_found, _child_block_interval) do
+  def extract_initial_state(:not_found, _child_block_interval, _fee_claimer_address) do
```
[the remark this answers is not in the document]
pnowosie: This clause serves as __fail fast__ in case of corrupted DB... This is at least what I believe :)

apps/omg/lib/omg/state/transaction/validator/fee_claim.ex
```diff
@@ -0,0 +1,46 @@
...
+
+  @spec can_claim_fees(Core.t(), Transaction.Recovered.t()) ::
+          {:ok, %{}} | {{:error, fee_claim_error()}, Core.t()}
+  def can_claim_fees(
+        %Core{fee_claimer_address: owner, fees_paid: fees_paid} = state,
+        %Transaction.Recovered{signed_tx: %{raw_tx: fee_tx}}
+      ) do
+    # NOTE: Fee claiming transaction does not transfer funds. It spends pseudo-output resultant of fees collection
+    with outputs = make_outputs(owner, fees_paid),
+         {:ok, _} <- Transaction.Protocol.can_apply?(fee_tx, outputs) do
```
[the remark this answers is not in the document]
pnowosie: WAT?! Did `dialyzer` take vacation recently?
Great catch! :clap:

apps/omg/test/omg/state/core_test.exs
pnowosie: `form_block_check` is a test helper (private fn declared at bottom) heavily used here.
I cleared the test intention - should be clear now.

apps/omg/test/omg/state/core_test.exs
pnowosie: This approach was taken in omisego/feat-fee/issues/55 to simplify the flow - the only difference is that surpluses are not wasted/burnt but collected by the operator. Also we considered `:fees_not_required` isn't longer an option #1299

apps/omg/test/omg/state/core_test.exs
pnowosie: Totally fair question - it's by [design](https://github.com/omisego/elixir-omg/blob/master/docs/fee_design.md) (HINT: fee tx output == UTXO):
> To exit fees, an operator puts a special fee transaction into a plasma chain block. This allows the operator to spend the transaction fee output as a Payment transaction. By leveraging MoreVP, we only need a contract to represent the fee transaction type and can use the pre-existing exit game contract.

apps/omg/test/omg/state/core_test.exs
pnowosie: Correct - fee tx by design should claim all surpluses from the current block, so no other payment txs can appear after.

apps/omg/test/omg/state/core_test.exs
pnowosie: Can't say - probably... it exists in the test-helpers since the beginning so I got familiar with it ;)

apps/omg/test/omg/state/persistence_test.exs
pnowosie: It's in test helpers from ages ;) I'm open to improve tests, but rather in separate PR :persevere:

apps/omg/test/omg/state/transaction/fee_test.exs
pnowosie: Absolutely :exclamation:  Thanks for spotting this :+1:

priv/apps/itest/test/itest/fee_claiming_test.exs
pnowosie: IMHO we can fix it for all #1293 - and what you posted isn't this exact case (there are 3 clauses of `fix_balance_response()`

apps/omg/lib/omg/state/core.ex
```diff
@@ -161,22 +183,36 @@ defmodule OMG.State.Core do
...
     tx_hash = Transaction.raw_txhash(tx)
 
-    case Validator.can_apply_spend(state, tx, fees) do
-      true ->
-        {:ok, {tx_hash, state.height, state.tx_index},
-         state
-         |> apply_spend(tx)
-         |> add_pending_tx(tx)}
-
+    with {:ok, fees_paid} <- Validator.can_process_tx(state, tx, fees) do
```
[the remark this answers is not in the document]
pnowosie: > That's why case's exist.

No, this  is not the case :joy: 
IMO, `case` vs `with` is often misunderstood and the article you mention is very basic in its conclusions. You can probably use `case` and `with` interchangeably, but they are different at the __intention__ layer.

`case` is general-purpose branching concept it shows several equally important cases where flow can branch.

`with` is a concept of railway oriented programming pattern so there is happy-path flow and the error flow. In the happy-path functions use the results of the previous executions when successful to produce the final result. The first error is returned otherwise.

There is a great work of Scott Wlaschin who describes many of FP concepts (however in F# lang)

apps/omg/lib/omg/state/transaction/validator/fee_claim.ex
```diff
@@ -0,0 +1,46 @@
...
+  @type fee_claim_error :: :claiming_unsupported_token | :claiming_more_than_collected
+
+  @spec can_claim_fees(Core.t(), Transaction.Recovered.t()) ::
+          {:ok, %{}} | {{:error, fee_claim_error()}, Core.t()}
+  def can_claim_fees(
+        %Core{fee_claimer_address: owner, fees_paid: fees_paid} = state,
+        %Transaction.Recovered{signed_tx: %{raw_tx: fee_tx}}
+      ) do
+    # NOTE: Fee claiming transaction does not transfer funds. It spends pseudo-output resultant of fees collection
+    with outputs = make_outputs(owner, fees_paid),
```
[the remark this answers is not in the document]
pnowosie: It is not pattern-matching. I could write this `outputs <- make_outputs(...)` but I need to be clear there is no fail-case here. However fn returns is used in the latter, so it more readable inside `with` (I've seen some of the assignment cases in the `with` in the codebase)

apps/omg/lib/omg/state/core.ex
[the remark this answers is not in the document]
InoMurko: I would disagree, but I'll let it pass since most of this code will get removed or rewritten.

apps/omg/lib/omg/state/transaction/validator/fee_claim.ex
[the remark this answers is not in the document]
InoMurko: Wait now... you're contradicting yourself.
Earlier you said:
```
with is a concept of railway oriented programming pattern so there is happy-path flow and the error flow. In the happy-path functions use the results of the previous executions when successful to produce the final result. The first error is returned otherwise.
```
Where is any of it now?
Why is there an else clause?

apps/omg/lib/omg/state/transaction/validator/payment.ex
```diff
@@ -0,0 +1,78 @@
...
+  @spec can_apply_tx(state :: Core.t(), tx :: Transaction.Recovered.t(), fees :: Fees.optional_fee_t()) ::
+          {:ok, map()} | {{:error, can_apply_error()}, Core.t()}
+  def can_apply_tx(
+        %Core{utxos: utxos} = state,
+        %Transaction.Recovered{signed_tx: %{raw_tx: raw_tx}, witnesses: witnesses} = tx,
+        fees
+      ) do
+    inputs = Transaction.get_inputs(tx)
+
+    with true <- not state.fee_claiming_started || {:error, :payments_rejected_during_fee_claiming},
```
InoMurko: why are you negating `fee_claiming_started`?

priv/apps/itest/test/itest/fee_claiming_test.exs
InoMurko: I guess the response depends on whether you provide the amount you're looking for Client.get_balance/2 or `Client.get_balance/1`

apps/omg/lib/omg/state/transaction/validator/fee_claim.ex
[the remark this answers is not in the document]
pnowosie: In the above code a happy-path is a chain of functions:  `make_outputs` & `can_apply?` and 1st will not fail which is explicitly show by using `=`. Final result is implicit result of prev functions successful execution.

I would like to avoid `with-else` as much as possible. Here is simple wrapping error so I think is justified.

PR #1246, opened 2020-01-07
fix: finalize dropping websocket support by cutting out producers

pdobacz:
continues #1125, as discussed on Slack

## Overview

Follow up to #1125 which got rid of consumers + some more things cut out too

## Changes

All the preparing of data to be pushed to the late websocket-feeding `Eventer` is dropped, as are the corresponding tests. APis are thinned down and lighter and dead code greatly reduced

## Testing

CI/CD


pdobacz:
@unnawut 

> Crashing the party because I'm seeing Bus changes 

:beers: :) 

> 
> I assume `OMG.Bus.broadcast(...)` are removed from `:deposits` and `:finalize_exits` because they're already covered by https://github.com/omisego/elixir-omg/blob/master/apps/omg/lib/omg/ethereum_event_listener.ex#L150-L155 ?

I think what this PR removes is a different kind of events (unless I over-removed sth by accident). Those would be the events that were pushed out of the Watcher to clients connecting via websockets (by `Phoenix.Channels`). So not "pure" events from the Ethereum node, but rather "acknowledgements of being consumed" by the watcher, that a client could listen on.

Does this make more sense now?

unnawut:
Yup clear! We don't need the events you removed so far.

Issue #1136, opened 2019-11-14
child_block_minimal_enqueue_gap config is misinterpreted by the BlockQueue


pdobacz:
https://github.com/omisego/elixir-omg/blob/master/apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex#L366

should be `>=` not `>`. `child_block_minimal_enqueue_gap` is the minimum that we want "now" and "last submitted" to be apart. So if it's `1`, "now" is `15`, "last submitted" is `14`, we should form and submit a new block.

This causes the current prevalent configuration entry (`1`) to cause the child chain to form & submit every second ethereum block.

Maybe, some reworking of the respective unit tests (that put a stamp on this wrong behavior) is advised, so that the correct behavior is more apparent.

### Bonus

Let's change the confusing config name. I vote for Sirn's idea: `block_submit_every_nth`. `block_submit_every_nth==1` means every ethereum block.

pdobacz:
I'll give it a stab. Also I'll go with @sirn suggestion to rename the setting to `block_submit_every_nth`, I think it prevailed in the brief naming discussion we had on Slack :).

PR #1258, opened 2020-01-14
Clean block queue code

pdobacz:
## Overview

BlockQueue.Core (especially tests) aged a bit, which I spotted when solving #1136. This should give those a significant refactor.

As far as I can tell, this shouldn't change `BlockQueue`'s behavior in any way.

## Changes

This is a set of interleaving implementation and test changes. It might be **most convenient** to review commit-by-commit or in small batches
- **removing some dead code**
  - refactor: remove BlockQueue.Core.new/0 which wasn't really used
  - refactor: get rid of child_start_parent_height handling
- **mechanical and simple tidies to `CoreTests`**
  - test: remove duplicate test and tighten the one that remained after
  - style: rearrange existing BlockQueue.CoreTests into function `describe` blocks
  - refactor: ditch the recover/3 helper in test
- **larger refactor cleaning `Core.new/1` that had a multitude of named args**
  - refactor: hideoptional config parameters in tests and make consistent …
- **tidies of tests now grouped in the new `describe` blocks, adding some new ones** 
  - test: tidy new/1 tests
  - style: replace the confusing x * @child_block_interval DRYing
  - test: tidy enqueue_block and set_ethereum_height tests
  - test: simple tidies of get_blocks_to_submit/1 tests
- **large refactor/revamp of gas price tests** (actually, this was my biggest itch in here. Those tests used a lot of internal stuffs to test uncleanly. Now they are tested based on public API calls and observable actions like ethereum progressing/new blocks etc. 
  - refactor: change gas price tests to not use internals

**NOTE** on the gas price test refactors. This actually uncovered a bunch of surprising behaviors, so careful review required on the last commit :point_up: . So far I haven't attempted to "fix" this, since there hasn't been an issue with the mechanism as a whole. For now I only solidified existing behavior in the tests.

Overall, this aims to get closer to the `ExUnit` style we have, reduce the use of test helpers and make tests more approachable (hopefully).

## Testing

`mix test test/omg_child_chain` and :eyes:

### DIFF
--- a/apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
@@ -83,10 +83,10 @@ defmodule OMG.ChildChain.BlockQueue do
       {:ok, state} =
         with {:ok, _state} = result <-
                Core.new(
-                 mined_num,
-                 Enum.zip(range, known_hashes),
-                 top_mined_hash,
-                 parent_height,
+                 mined_child_block_num: mined_num,
+                 known_hashes: Enum.zip(range, known_hashes),
+                 top_mined_hash: top_mined_hash,
+                 parent_height: parent_height,
                  child_block_interval: child_block_interval,
                  block_submit_every_nth: Application.fetch_env!(:omg_child_chain, :block_submit_every_nth),
                  finality_threshold: finality_threshold
--- a/apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
@@ -55,10 +55,13 @@ defmodule OMG.ChildChain.BlockQueue do
     def handle_continue(:setup, %{}) do
       _ = Logger.info("Starting #{__MODULE__} service.")
       :ok = Eth.node_ready()
+
+      # prevent booting if contracts are not ready
       :ok = Eth.RootChain.contract_ready()
+      {:ok, parent_start} = Eth.RootChain.get_root_deployment_height()
+
       {:ok, parent_height} = EthereumHeight.get()
       {:ok, mined_num} = Eth.RootChain.get_mined_child_block()
-      {:ok, parent_start} = Eth.RootChain.get_root_deployment_height()
       {:ok, child_block_interval} = Eth.RootChain.get_child_block_interval()
       {:ok, stored_child_top_num} = OMG.DB.get_single_value(:child_top_block_number)
       {:ok, finality_threshold} = Application.fetch_env(:omg_child_chain, :submission_finality_margin)
@@ -85,7 +88,6 @@ defmodule OMG.ChildChain.BlockQueue do
                  top_mined_hash: top_mined_hash,
                  parent_height: parent_height,
                  child_block_interval: child_block_interval,
-                 chain_start_parent_height: parent_start,
                  block_submit_every_nth: Application.fetch_env!(:omg_child_chain, :block_submit_every_nth),
                  finality_threshold: finality_threshold,
                  last_enqueued_block_at_height: parent_height
--- a/apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
@@ -83,14 +83,13 @@ defmodule OMG.ChildChain.BlockQueue do
       {:ok, state} =
         with {:ok, _state} = result <-
                Core.new(
-                 mined_child_block_num: mined_num,
-                 known_hashes: Enum.zip(range, known_hashes),
-                 top_mined_hash: top_mined_hash,
-                 parent_height: parent_height,
+                 mined_num,
+                 Enum.zip(range, known_hashes),
+                 top_mined_hash,
+                 parent_height,
                  child_block_interval: child_block_interval,
                  block_submit_every_nth: Application.fetch_env!(:omg_child_chain, :block_submit_every_nth),
-                 finality_threshold: finality_threshold,
-                 last_enqueued_block_at_height: parent_height
+                 finality_threshold: finality_threshold
                ) do
           result
         else
--- a/apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
+++ b/apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
@@ -30,10 +30,21 @@ defmodule OMG.ChildChain.BlockQueue.CoreTest do
   @account_locked_response {:error, %{"code" => -32_000, "message" => "authentication needed: password or unlock"}}
 
   setup do
-    {:ok, empty} = Core.new(0, [], <<0::256>>, 0, child_block_interval: @child_block_interval)
+    {:ok, empty} =
+      Core.new(
+        mined_child_block_num: 0,
+        known_hashes: [],
+        top_mined_hash: <<0::256>>,
+        parent_height: 0,
+        child_block_interval: @child_block_interval
+      )
 
     {:ok, empty_high_max_gas_price} =
-      Core.new(0, [], <<0::size(256)>>, 0,
+      Core.new(
+        mined_child_block_num: 0,
+        known_hashes: [],
+        top_mined_hash: <<0::size(256)>>,
+        parent_height: 0,
         child_block_interval: @child_block_interval,
         gas_price_adj_params: %GasPriceAdjustment{max_gas_price: 40_000_000_000}
       )
@@ -73,10 +84,10 @@ defmodule OMG.ChildChain.BlockQueue.CoreTest do
     test "Recovers after restart to proper mined height" do
       assert [%{hash: "8", nonce: 8}, %{hash: "9", nonce: 9}] =
                Core.new(
-                 7000,
-                 [{5000, "5"}, {6000, "6"}, {7000, "7"}, {8000, "8"}, {9000, "9"}],
-                 "7",
-                 10,
+                 mined_child_block_num: 7000,
+                 known_hashes: [{5000, "5"}, {6000, "6"}, {7000, "7"}, {8000, "8"}, {9000, "9"}],
+                 top_mined_hash: "7",
+                 parent_height: 10,
                  child_block_interval: @child_block_interval
                )
                |> elem(1)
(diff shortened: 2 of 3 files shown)


apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
```diff
@@ -29,50 +30,27 @@ defmodule OMG.ChildChain.BlockQueue.CoreTest do
...
         child_block_interval: @child_block_interval,
-        chain_start_parent_height: 1,
-        block_submit_every_nth: 1,
-        finality_threshold: 12,
-        last_enqueued_block_at_height: 0
+        gas_price_adj_params: %GasPriceAdjustment{max_gas_price: 40_000_000_000}
       )
 
-    empty_with_gas_params = %{empty | formed_child_block_num: 5 * @child_block_interval, gas_price_to_use: 100}
+    empty_with_gas_params = %{empty | formed_child_block_num: 5000, gas_price_to_use: 100}
```
pnowosie: Why? IMHO the former was more descriptive.

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
```diff
@@ -29,50 +30,27 @@ defmodule OMG.ChildChain.BlockQueue.CoreTest do
...
 
-    {:do_form_block, empty_with_gas_params} =
-      Core.set_ethereum_status(empty_with_gas_params, 1, 3 * @child_block_interval, false)
+    {:do_form_block, empty_with_gas_params} = Core.set_ethereum_status(empty_with_gas_params, 1, 3000, false)
 
     # assertions - to be explicit how state looks like
-    child_block_mined = 3 * @child_block_interval
+    child_block_mined = 3000
```
pnowosie: Why? IMHO the former was more descriptive.
(and many more :point_down: in the file)

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
achiurizo: I second this. It's much clearer to understand what this value represents with the `@child_block_interval`.

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
```diff
@@ -29,50 +30,27 @@ defmodule OMG.ChildChain.BlockQueue.CoreTest do
...
-      block_submit_every_nth: 1,
-      finality_threshold: 12,
-      last_enqueued_block_at_height: 0
-    )
+    {:ok,
+     %{empty: empty, empty_with_gas_params: empty_with_gas_params, empty_high_max_gas_price: empty_high_max_gas_price}}
   end
 
-  describe "Block queue." do
+  describe "child_block_nums_to_init_with/4" do
```
achiurizo: ✨

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
```diff
@@ -89,173 +67,124 @@ defmodule OMG.ChildChain.BlockQueue.CoreTest do
...
-          parent_height: 6,
-          child_block_interval: @child_block_interval,
-          chain_start_parent_height: 1,
-          block_submit_every_nth: 1,
-          finality_threshold: finality_threshold,
-          last_enqueued_block_at_height: 0
+          6000,
+          [{5000, "5"}, {6000, "6"}, {7000, "7"}, {8000, "8"}, {9000, "9"}],
+          "6",
+          10,
```
achiurizo: Is there a reason we moved away from the named arguments? I think it helped make it clear what the arguments represented and how to invoke `Core.new` here.

apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
```diff
@@ -80,15 +83,13 @@ defmodule OMG.ChildChain.BlockQueue do
...
-                 known_hashes: Enum.zip(range, known_hashes),
-                 top_mined_hash: top_mined_hash,
-                 parent_height: parent_height,
+                 mined_num,
+                 Enum.zip(range, known_hashes),
+                 top_mined_hash,
+                 parent_height,
                  child_block_interval: child_block_interval,
-                 chain_start_parent_height: parent_start,
                  block_submit_every_nth: Application.fetch_env!(:omg_child_chain, :block_submit_every_nth),
```
InoMurko: A good pattern coming from OTP is that children never directly access configuration. Configuration values are always passed top to bottom. Application process to Supervisor to it's children.

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
InoMurko: now we need an explanation what the function name means :D

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
pdobacz: ah, so I liked those named args here too, but:
 - there were overall too many of those, and I wanted a clear and standard distinction between required and optional
 - I tried sth like `def new([required_1: ..., required_2: ... | optional])` to have best of both, but it seems not not be supported and the alternative: `def new([{:required_1, _}, {:required_2, _} | optional` either didn't work or it seemed weird and nonstandard. 
 - And then just `def new(opts \\ [])` and have the required args enforced in the body didn't click too well neither. But maybe this could be an alternative, it just would hide the required vs optional distinction.

I think 4 unnamed required params and then options is pretty standard. I can try the last `def new(opts \\ [])` if you like and see how it goes.

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
```diff
@@ -19,6 +19,7 @@ defmodule OMG.ChildChain.BlockQueue.CoreTest do
   import ExUnit.CaptureLog
```
InoMurko: I think a pattern that is quite good and I usually follow is to name the argument variables I'm using:
```
 Core.child_block_nums_to_init_with(0, 3000, @child_block_interval, 2)
```
I would call this:
```
mined_num = 0
until_child_block_num = 300, 
finality_threshold = 2
 Core.child_block_nums_to_init_with(mined_num, until_child_block_num, @child_block_interval, finality_threshold)
```
using this pattern, I do not need to store in memory what these arguments are or what they mean. I think we should follow this...

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
```diff
@@ -340,164 +348,218 @@ defmodule OMG.ChildChain.BlockQueue.CoreTest do
...
-      {:do_form_block, newstate} =
-        state
-        |> Core.enqueue_block(<<0>>, 6000, 1)
-        |> Core.set_ethereum_status(1 + eth_gap, 0, false)
-
-      assert current_price < newstate.gas_price_to_use
-
-      # assert the actual gas price based on parameters value - test could fail if params or calculation will change
-      assert 200 == newstate.gas_price_to_use
+    test "Progressing to enqueue height may raise price if threshold exceeded", %{empty_high_max_gas_price: empty} do
```
pnowosie: I had a hard time to understand test description, especially `may` & `if`
I'd like sth with more assurance ~ `exceeding the threshold height results with price raise`

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
```diff
@@ -340,164 +348,218 @@ defmodule OMG.ChildChain.BlockQueue.CoreTest do
...
-      {_, state} = Core.set_ethereum_status(state, 101, 0, false)
-      assert state.gas_price_to_use > gas_price
-    end
-
-    test "gas price changes only, when etherum advanses", %{empty_with_gas_params: state} do
-      gas_price = state.gas_price_to_use
-      eth_height = 0
+      assert [%{gas_price: 5}, %{gas_price: 5}] = Core.get_blocks_to_submit(state)
 
+      # no more successful submissions can change the price
```
pnowosie: I don't follow the test logic atm. If I'm correct you are doing the exact same thing 2 more times (after prev `1000-2` trials) with a comment that now it can change the price. Final asserts shows the price hasn't changed :question:

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
pdobacz: um, so the test and description are as un-assuring as the behavior. It means that despite the `set_ethereum_status` goes to exactly `enqueue_block`'s eth_height, it still raises the price.

I'll clarify to

```
    # NOTE: This behavior is strange - revisit along with the other ones (see `NOTEs` here)
    #       Why would detecting the exact height of enqueue cause the price to raise - we don't know if there's need to
    test "Despite progressing to enqueue height price may raise", %{empty_high_max_gas_price: empty} do
```

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
pdobacz: yeah, so the idea is to "reach" the bottom price (5) and do one more (actually I don't know why I did 2, a leftover, I'll fix this) and see it remain at 5

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
pdobacz: ok, did that, I think it looks better ff2a127946a4c8de40a752a1d2827674056c4788, thx!

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
pdobacz: I tried this too, but it really is unreadable, with cases like: 

```
      assert [2 * @child_block_interval, 3 * @child_block_interval] ==
               Core.child_block_nums_to_init_with(
                 2 * @child_block_interval,
                 3 * @child_block_interval,
                 @child_block_interval,
                 0
               )
```

and sometimes one can't pattern-match on the `n * ...` expression in the result. I opt towards leaving it consistently like:
 - when interval specifically is meant (as in a config param passed in) - `@child_block_interval`
 - when a blknum is meant `<n>000`.

(I have a full commit doing the change in a branch in case you're interested).

WDYT?

pdobacz:
OK, since there's :heavy_check_mark: 's and no further folluw-up comments in the style-related threads, merging this.

I'm going to open an issue to dig deeper into the strange behaviors spotted in the tests (everything seemed to work but maybe some edge cases could be better handled, when doing gas price selection).

PR #1260, opened 2020-01-15
Transaction.create automatic fee

mederic-p:
## Overview

This PR remove the `"amount"` parameter in the `fee` object when doing a `transaction.create` and fetch the correct amount from the childchain instead.

### DIFF
--- a/apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
+++ b/apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
@@ -1,4 +1,4 @@
-# Copyright 2019 OmiseGO Pte Ltd
+# Copyright 2019-2020 OmiseGO Pte Ltd
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,7 +14,7 @@
 
 defmodule OMG.WatcherInfo.OrderFeeFetcher do
   @moduledoc """
-  Handle fetching and foratting of fees for an order
+  Handle fetching and formatting of fees for an order
   """
 
   alias OMG.WatcherInfo.HttpRPC.Client
@@ -25,29 +25,27 @@ defmodule OMG.WatcherInfo.OrderFeeFetcher do
   @tx_type WireFormatTypes.tx_type_for(:tx_payment_v1)
   @str_tx_type Integer.to_string(@tx_type)
 
-  def add_fee_to_order(%{fee: %{currency: currency}} = order) do
-    child_chain_url = Application.get_env(:omg_watcher, :child_chain_url)
+  def add_fee_to_order(%{fee: %{currency: currency}} = order, url \\ nil) do
+    child_chain_url = url || Application.get_env(:omg_watcher_info, :child_chain_url)
     encoded_currency = Encoding.to_hex(currency)
+    params = %{"currencies" => [encoded_currency], "tx_types" => [@tx_type]}
 
-    %{"currencies" => [encoded_currency], "tx_types" => [@tx_type]}
-    |> Client.get_fees(child_chain_url)
-    |> parse_response(encoded_currency)
-    |> respond(order)
-  end
-
-  defp parse_response({:ok, fees}, currency) do
-    with %{@str_tx_type => [%{"amount" => amount, "currency" => ^currency} | _]} <- fees do
-      {:ok, amount}
+    with {:ok, fees} <- Client.get_fees(params, child_chain_url),
+         {:ok, amount} <- validate_child_chain_fees(fees, encoded_currency) do
+      {:ok, Kernel.put_in(order, [:fee, :amount], amount)}
     else
-      _ -> {:error, :unexpected_fee}
+      error ->
+        error
     end
   end
 
-  defp parse_response(error, _), do: error
+  defp validate_child_chain_fees(fees, currency) do
+    case fees do
+      %{@str_tx_type => [%{"amount" => amount, "currency" => ^currency} | _]} ->
+        {:ok, amount}
 
-  defp respond({:ok, amount}, %{fee: fee} = order) do
-    {:ok, Map.put(order, :fee, Map.put(fee, :amount, amount))}
+      _ ->
+        {:error, :unexpected_fee}
+    end
   end
-
-  defp respond(error, _), do: error
 end
--- a/apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
+++ b/apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
@@ -17,14 +17,29 @@ defmodule OMG.WatcherInfo.OrderFeeFetcher do
   Handle fetching and formatting of fees for an order
   """
 
-  alias OMG.WatcherInfo.HttpRPC.Client
-  alias OMG.WireFormatTypes
+  alias OMG.Crypto
+  alias OMG.State.Transaction
   alias OMG.Utils.HttpRPC.Encoding
+  alias OMG.WatcherInfo.HttpRPC.Client
+  alias OMG.WatcherInfo.UtxoSelection
+  alias OMG.WireFormatTypes
 
   # Note: Hardcoding the tx_type for now
   @tx_type WireFormatTypes.tx_type_for(:tx_payment_v1)
   @str_tx_type Integer.to_string(@tx_type)
 
+  @type order_without_fee_amount_t() :: %{
+          owner: Crypto.address_t(),
+          payments: nonempty_list(UtxoSelection.payment_t()),
+          fee: %{currency: Transaction.Payment.currency()},
+          metadata: binary() | nil
+        }
+
+  @doc """
+  Fetch the correct fee amount for the given fee currency from the childchain
+  and adds it to the order map.
+  """
+  @spec add_fee_to_order(order_without_fee_amount_t()) :: {:ok, UtxoSelection.order_t()} | {:error, atom()}
   def add_fee_to_order(%{fee: %{currency: currency}} = order, url \\ nil) do
     child_chain_url = url || Application.get_env(:omg_watcher_info, :child_chain_url)
     encoded_currency = Encoding.to_hex(currency)
--- a/apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
+++ b/apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
@@ -1,0 +1,53 @@
+# Copyright 2019 OmiseGO Pte Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.WatcherInfo.OrderFeeFetcher do
+  @moduledoc """
+  Handle fetching and foratting of fees for an order
+  """
+
+  alias OMG.WatcherInfo.HttpRPC.Client
+  alias OMG.WireFormatTypes
+  alias OMG.Utils.HttpRPC.Encoding
+
+  # Note: Hardcoding the tx_type for now
+  @tx_type WireFormatTypes.tx_type_for(:tx_payment_v1)
+  @str_tx_type Integer.to_string(@tx_type)
+
+  def add_fee_to_order(%{fee: %{currency: currency}} = order) do
+    child_chain_url = Application.get_env(:omg_watcher, :child_chain_url)
+    encoded_currency = Encoding.to_hex(currency)
+
+    %{"currencies" => [encoded_currency], "tx_types" => [@tx_type]}
+    |> Client.get_fees(child_chain_url)
+    |> parse_response(encoded_currency)
+    |> respond(order)
+  end
+
+  defp parse_response({:ok, fees}, currency) do
+    with %{@str_tx_type => [%{"amount" => amount, "currency" => ^currency} | _]} <- fees do
+      {:ok, amount}
+    else
+      _ -> {:error, :unexpected_fee}
+    end
+  end
+
+  defp parse_response(error, _), do: error
+
+  defp respond({:ok, amount}, %{fee: fee} = order) do
+    {:ok, Map.put(order, :fee, Map.put(fee, :amount, amount))}
+  end
+
+  defp respond(error, _), do: error
+end
--- a/apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
+++ b/apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
@@ -56,6 +56,9 @@ defmodule OMG.WatcherInfo.OrderFeeFetcher do
 
   defp validate_child_chain_fees(fees, currency) do
     case fees do
+      # Ensuring that the child chain response is correct, we should only have
+      # 1 currency for the given tx type as we filter it in the params of the request.
+      # We also take this opportunity to pattern match the amount.
       %{@str_tx_type => [%{"amount" => amount, "currency" => ^currency} | _]} ->
         {:ok, amount}
 
--- a/apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
+++ b/apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
@@ -24,7 +24,9 @@ defmodule OMG.WatcherInfo.OrderFeeFetcher do
   alias OMG.WatcherInfo.UtxoSelection
   alias OMG.WireFormatTypes
 
-  # Note: Hardcoding the tx_type for now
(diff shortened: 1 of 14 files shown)


apps/omg_watcher_info/test/omg_watcher_info/order_fee_fetcher_test.exs
```diff
@@ -0,0 +1,96 @@
...
+  alias OMG.WatcherInfo.TestServer
+  alias OMG.WireFormatTypes
+
+  @eth Eth.zero_address()
+  @not_eth <<1::160>>
+  @tx_type WireFormatTypes.tx_type_for(:tx_payment_v1)
+  @str_tx_type Integer.to_string(@tx_type)
+
+  describe "add_fee_to_order/2" do
+    @tag fixtures: [:test_server]
```
unnawut: Hoping to move away from fixtures. There aren't that many tests using `:test_server` but this PR adds quite a lot. Could we convert this into a helper function instead?

InoMurko:
If I understand this PR correctly ... every time someone sends a transaction to Watcher, this would fetch fees from ChildChain?

apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
```diff
@@ -0,0 +1,66 @@
...
+          metadata: binary() | nil
+        }
+
+  @doc """
+  Fetch the correct fee amount for the given fee currency from the childchain
+  and adds it to the order map.
+  """
+  @spec add_fee_to_order(order_without_fee_amount_t()) :: {:ok, UtxoSelection.order_t()} | {:error, atom()}
+  def add_fee_to_order(%{fee: %{currency: currency}} = order, url \\ nil) do
+    child_chain_url = url || Application.get_env(:omg_watcher_info, :child_chain_url)
```
InoMurko: I think a better pattern would be to inject the child chain URL as a plug on routes that need it. I would like us to start avoiding configuration fetches in this manner. A good pattern is to think about configuration in a hierarchical manner. For example: configuration must come from the top while propagated to the bottom (this pattern avoids shared state problems).

T-Dnzt: I disagree, I don't think accessing configuration values belong at the web/conn/plug level.

InoMurko: the problem comes down to shared state. Imagine having a life cycle of a process that is long lived (a websocket connection or a http request would do). 
Your process does several HTTP requests towards a pre-set-configurational value. While the process is still alive this particular configuration value changes. This could make the behavior of your process unpredictable. 
By injecting requirements as early as possible, you fully control the lifecycle of your process.

InoMurko: similarly to an OTP process, while attaching it to a supervision tree, you should always pass arguments to it and avoid reading configuration during the process lifecycle

apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
```diff
@@ -0,0 +1,66 @@
...
+    encoded_currency = Encoding.to_hex(currency)
+    params = %{"currencies" => [encoded_currency], "tx_types" => [@tx_type]}
+
+    with {:ok, fees} <- Client.get_fees(params, child_chain_url),
+         {:ok, amount} <- validate_child_chain_fees(fees, encoded_currency) do
+      {:ok, Kernel.put_in(order, [:fee, :amount], amount)}
+    else
```
pdobacz: an `error->error` `else` clause in `with` can be dropped

apps/omg_watcher_info/test/omg_watcher_info/order_fee_fetcher_test.exs
```diff
@@ -0,0 +1,96 @@
...
+      order = %{
+        fee: %{currency: @eth}
+      }
+
+      assert OrderFeeFetcher.add_fee_to_order(order, context.fake_addr) ==
+               {:ok, Kernel.put_in(order, [:fee, :amount], 2)}
+    end
+
+    @tag fixtures: [:test_server]
+    test "returns an `unexpected_fee` error when cc returns an unexpected fee value", %{test_server: context} do
```
pdobacz: cc -> child chain

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -844,12 +886,29 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
...
+    test "advice on merge does not merge single utxo", %{alice: alice, bob: bob, test_server: context} do
       max_spendable = max_amount_spendable_in_single_tx(alice.addr, @eth)
 
       payment = max_spendable + 1
 
+      fee_response = %{
+        @str_tx_type => [
+          %{
+            "currency" => @default_fee_currency,
+            "amount" => 0,
```
pdobacz: not sure I understand why amount is 0 here?

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -1069,15 +1157,30 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
...
+    @tag fixtures: [:alice, :more_utxos, :test_server]
     test "transaction without payments that burns funds in fees is created correctly and incorrect on decoding",
-         %{alice: alice} do
-      fee = 15
+         %{alice: alice, test_server: context} do
+      fee_response = %{
+        @str_tx_type => [
+          %{
+            "currency" => @other_token_hex,
+            "amount" => 15,
```
pdobacz: here, I don't know where the `15` used to come from, it seems strange - have you tried with the default? If it works with the default, you can simplify here

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -1136,18 +1239,35 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
...
-    test "empty transaction without payments list is not allowed", %{alice: alice} do
+    @tag fixtures: [:alice, :more_utxos, :test_server]
+    test "empty transaction without payments list is not allowed", %{alice: alice, test_server: context} do
       alice_addr = Encoding.to_hex(alice.addr)
 
+      fee_response = %{
+        @str_tx_type => [
+          %{
+            "currency" => @default_fee_currency,
+            "amount" => 0,
```
pdobacz: similar here, try with the default and see if it works. The fee amount should not be relevant here, at 1st glance. Maybe setting it to 0 was superfluous originally

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -699,18 +718,25 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
...
-    test "returns correctly formed transaction, identical with the verbose form", %{alice: alice, bob: bob} do
+    @tag fixtures: [:alice, :bob, :more_utxos, :test_server]
+    test "returns correctly formed transaction, identical with the verbose form", %{
+      alice: alice,
+      bob: bob,
+      test_server: context
+    } do
       alias OMG.State.Transaction
 
+      response = TestServer.make_response(@fee_response)
```
pdobacz: first see the comments about the `fee_response` being "customized" for tests. If the customizations can be undone, then you could move this to `setup`.

It seems that the fee response is auxiliary to `transaction.create` tests, so it could be uniformly done in `setup` for all tests.

Unless you wanted to test some interactions between those two behaviors :thinking:.

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
pdobacz: EDIT: ok it used to be 0 before. Like in other similar comments - try not customizing this to 0, it seems like it should work with `@fee_response`.

apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
InoMurko: thinking further... actually the problem here is our request process scope that's being dragged to deep in the application stack. if we wanted to limit the process scope and have clear separation of responsibilities we should introduce short lived processes that have one and only responsibility: transaction processing

apps/omg_watcher_info/test/omg_watcher_info/order_fee_fetcher_test.exs
```diff
@@ -0,0 +1,96 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.WatcherInfo.OrderFeeFetcherTest do
+  use ExUnitFixtures
+  use ExUnit.Case, async: false
```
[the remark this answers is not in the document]
InoMurko: good point. I think this is the TestServer - FakeServer dependency limitation because it changes the URL with each request - making Application.put_env/3 calls. We should step away from this dependency and write our own trivial webserver:
```
  defp start(port) do
    {:ok, sock} = :gen_tcp.listen(port, [{:active, false}])
    spawn(fn -> loop(sock) end)

    receive do
      :stop ->
        :gen_tcp.close(sock)
    end
  end

  defp loop(sock) do
    case :gen_tcp.accept(sock) do
      {:ok, conn} ->
        handler = spawn(fn -> handle(conn) end)
        :gen_tcp.controlling_process(conn, handler)
        loop(sock)

      _ ->
        :ok
    end
  end

  defp handle(conn) do
    body = Jason.encode!(@exchanger_body)
    :ok = :gen_tcp.send(conn, ["HTTP/1.0 ", Integer.to_charlist(200), "\r\n", [], "\r\n", body])
    :gen_tcp.close(conn)
  end
```
you can add ID to requests/responses and have it serve the whole test suite concurrently

apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
```diff
@@ -0,0 +1,66 @@
...
+  Fetch the correct fee amount for the given fee currency from the childchain
+  and adds it to the order map.
+  """
+  @spec add_fee_to_order(order_without_fee_amount_t()) :: {:ok, UtxoSelection.order_t()} | {:error, atom()}
+  def add_fee_to_order(%{fee: %{currency: currency}} = order, url \\ nil) do
+    child_chain_url = url || Application.get_env(:omg_watcher_info, :child_chain_url)
+    encoded_currency = Encoding.to_hex(currency)
+    params = %{"currencies" => [encoded_currency], "tx_types" => [@tx_type]}
+
+    with {:ok, fees} <- Client.get_fees(params, child_chain_url),
```
InoMurko: also... we're going through the troubles of maintaining swagger ... why are we not using it?

apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
```diff
@@ -0,0 +1,66 @@
...
+      {:ok, Kernel.put_in(order, [:fee, :amount], amount)}
+    else
+      error ->
+        error
+    end
+  end
+
+  defp validate_child_chain_fees(fees, currency) do
+    case fees do
+      %{@str_tx_type => [%{"amount" => amount, "currency" => ^currency} | _]} ->
```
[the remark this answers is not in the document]
mederic-p: yep, that's expected

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
mederic-p: I tried to keep the same test data that we had before to make it faster so I didn't have to understand each test individually. I could simplify some, but others still need a custom fee response.
Some of these tests will most likely be updated soon anyway with the work that @T-Dnzt is doing (preventing fee overpay).

apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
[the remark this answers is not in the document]
unnawut: Can you add a comment that this is intended? There were a few times this is used, but the author actually wanted to loop through all elements. 🙏

apps/omg_watcher_info/test/omg_watcher_info/order_fee_fetcher_test.exs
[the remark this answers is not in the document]
unnawut: Talked to @mederic-p a bit. The ID handling on TestServer side is quite straightforward, but injecting the ID into the feature code can add quite a bit of complexity. 

I'm thinking this test might not worth adding the complexity to make it async just yet?

apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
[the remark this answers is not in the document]
pnowosie: I think we put too much trust in ch-ch endpoint response which also made this code confusing to plp already.

I think for the clarity and future-proof it would be better to 
`case Enum.find(fees, <...fn_by_currency>) do`
so we just assume that fee in the currency exists in the response

apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
```diff
@@ -0,0 +1,69 @@
...
+  defp validate_child_chain_fees(fees, currency) do
+    case fees do
+      # Ensuring that the child chain response is correct, we should only have
+      # 1 currency for the given tx type as we filter it in the params of the request.
+      # We also take this opportunity to pattern match the amount.
+      %{@str_tx_type => [%{"amount" => amount, "currency" => ^currency} | _]} ->
+        {:ok, amount}
+
+      _ ->
+        {:error, :unexpected_fee}
```
pnowosie:
```suggestion
        {:error, :unexpected_fee_currency}
```

pnowosie:
> If I understand this PR correctly ... every time someone sends a transaction to Watcher, this would fetch fees from ChildChain?

@InoMurko Every time someone calls Watcher `transaction.create` which is let say inexperienced user friendly version. Watcher has also [`transaction.submit`](https://developer.omisego.co/elixir-omg/docs-ui/?url=master%2Foperator_api_specs.yaml&urls.primaryName=master%2Fsecurity_critical_api_specs#/Transaction/submit) so if one knows how to RLP can submit directly

apps/omg_watcher_info/lib/omg_watcher_info/order_fee_fetcher.ex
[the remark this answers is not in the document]
mederic-p: I could do that, but don't we want to just raise an error if the childchain is not returning what's expected? 🤔

apps/omg_watcher_rpc/lib/web/validators/order.ex
```diff
@@ -58,8 +58,8 @@ defmodule OMG.WatcherRPC.Web.Validator.Order do
   end
 
   defp parse_fee(map) when is_map(map) do
```
InoMurko: hey @mederic-p in line https://github.com/omisego/elixir-omg/pull/1260/files#diff-68ed50a7de191c4eb70067419fd0b717R44
there are outstanding module aliases and macros, could you move them at the top?

apps/omg_watcher_info/test/support/test_server.ex
```diff
@@ -0,0 +1,76 @@
...
+  **Please note: **
+  When the route is configured with a list of FakeServer.HTTP.Responses, the server will respond with the first element
+  in the list and then remove it. This will be repeated for each request made for this route.
+  Use `fn req -> response end` when you need to return always the same or modified response on every request
+
+  Also first use of `with_response` changes configuration variable to child chain api to fake server, so invoke this
+  function when fake response is needed.
+  """
+  def with_response(response_block, %{fake_addr: fake_addr, server_id: server_id} = _context, path) do
+    Application.put_env(:omg_watcher_info, :child_chain_url, fake_addr)
```
InoMurko: is this needed now?

apps/omg_watcher_rpc/lib/web/validators/order.ex
```diff
@@ -17,9 +17,13 @@ defmodule OMG.WatcherRPC.Web.Validator.Order do
   Validates `/transaction.create` request body.
   """
 
+  require OMG.State.Transaction.Payment
+
+  import OMG.Utils.HttpRPC.Validator.Base
```
InoMurko: pls specify what functions you're importing

apps/omg_watcher_info/test/support/test_server.ex
unnawut: Still needed for controller tests unfortunately. As discussed we should move towards https://github.com/parroty/exvcr but should be in a separate PR.

mederic-p: It is indeed still needed for testing from the controller level. (here for example: https://github.com/omisego/elixir-omg/pull/1260/files#diff-79e680b3736f828226c51ac2f9a1467fR756) where we don't have access to the extra `url` parameter and so it's pulling the config from the env var.

PR #1261, opened 2020-01-15
fix: bump ex_plasma for the fix omisego/ex_plasma#21

pdobacz:
omisego/ex_plasma#21

## Overview

This should fix the failing property tests

## Changes

See omisego/ex_plasma#23, bumping to a commit having this fix

Also it requires some updates how `elixir-omg` calls into `ex_plasma`.

## Testing

run property tests, see if CircleCI stops complaining about them failing

### DIFF
--- a/apps/omg_watcher/lib/omg_watcher/exit_processor/in_flight_exit_info.ex
+++ b/apps/omg_watcher/lib/omg_watcher/exit_processor/in_flight_exit_info.ex
@@ -83,7 +83,8 @@ defmodule OMG.Watcher.ExitProcessor.InFlightExitInfo do
           tx_seen_in_blocks_at: {Utxo.Position.t(), inclusion_proof :: binary()} | nil,
           timestamp: non_neg_integer(),
           contract_id: ife_contract_id(),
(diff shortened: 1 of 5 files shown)


apps/omg_watcher/lib/omg_watcher/exit_processor/in_flight_exit_info.ex
```diff
@@ -458,9 +462,24 @@ defmodule OMG.Watcher.ExitProcessor.InFlightExitInfo do
   defp do_is_viable_competitor?(seen_at_pos, oldest_pos, competitor_pos),
     do: is_older?(competitor_pos, seen_at_pos) and is_older?(competitor_pos, oldest_pos)
 
+  # no position is older than any real position
+  defp is_older?(Utxo.position(_, _, _), :no_position), do: true
```
InoMurko: is see that this module has private and public functions scattered around. public on top, private below?

pdobacz: true, this needs cleanup, but I suggest doing so in a separate PR

Issue #1241, opened 2020-01-03
mix test test/omg_status are slow and chatty


pdobacz:
Steps to reproduce and results below :point_down: 

My suggestion to fix this is a mixture of:
 - refactor the slow tests to be quicker or move them under `:integration` if this is impossible
 - `capture_log` the tests (selectively) that test warning or error yielding scenarios
 - lower the severity (to `info`) of those log messages that don't need to be warnings (e.g. `Sentry configuration provided.`)

```
user@user:~/sources/elixir-omg$ mix test test/omg_status --trace --slowest 2
==> omg_utils
Paths given to "mix test" did not match any directory/file: test/omg_status
==> xomg_tasks
Paths given to "mix test" did not match any directory/file: test/omg_status
==> omg_bus
Paths given to "mix test" did not match any directory/file: test/omg_status
==> omg_status
Excluding tags: [integration: true, property: true, wrappers: true]


OMG.Status.Metric.DatadogTest
  * test if exiting process/port sends an exit signal to the parent process (0.2ms)
2020-01-03 10:51:08.457 [error] module=gen_server function=error_info/7 ⋅GenServer #PID<0.412.0> terminating
** (stop) :portkill
Last message: {:EXIT, #PID<0.411.0>, :portkill}⋅

OMG.Status.Metric.StatsdMonitorTest
  * test if exiting process/port sends an exit signal to the parent process2020-01-03 10:51:08.617 [error] module=OMG.Status.Metric.StatsdMonitor function=handle_info/2 ⋅Monitored datadog connection process from statix died of reason :testkill ⋅
[os_mon] memory supervisor port (memsup): Erlang has closed
[os_mon] cpu supervisor port (cpu_sup): Erlang has closed
  * test if exiting process/port sends an exit signal to the parent process (94.2ms)
  * test if exiting process/port sends an exit signal to the parent process 22020-01-03 10:51:08.634 [error] module=OMG.Status.Metric.StatsdMonitor function=handle_info/2 ⋅Monitored datadog connection process from statix died of reason :testkill ⋅
[os_mon] memory supervisor port (memsup): Erlang has closed
[os_mon] cpu supervisor port (cpu_sup): Erlang has closed
  * test if exiting process/port sends an exit signal to the parent process 2 (15.0ms)

OMG.Status.Alert.AlarmPrinterTest
  * test if the process has a previous backoff set (0.08ms)
[...]

PR #1262, opened 2020-01-15
feature: set CAPTURE_LOG env var to unmute logs

InoMurko:
addresses the `chatty` part of https://github.com/omisego/elixir-omg/issues/1241
## Overview

If you
```
export CAPTURE_LOG=0
```
this would unmute your `mix test` invocation.

## Changes

- default mute logger

## Testing

 check the tests output


unnawut:
Sounds good! One thing though...

```shell
# happy
$ CAPTURE_LOG=0 iex -e 'System.get_env("CAPTURE_LOG") |> if(do: false, else: true) |> IO.inspect(); System.halt()'
false

# sad
$ CAPTURE_LOG=1 iex -e 'System.get_env("CAPTURE_LOG") |> if(do: false, else: true) |> IO.inspect(); System.halt()'
false
```

unnawut:
Following above, this solution is good enough imo. No need for fancy parsing.

```elixir
if(System.get_env("CAPTURE_LOG") == "1", do: true, else: false)
```

InoMurko:
> ```shell
> if(System.get_env("CAPTURE_LOG") == "1", do: true, else: false
> ```

in this case, the default would be false... do we want the default to be false or true?

unnawut:
The original intention is to mute log by default, no? Sounds good to me.

```elixir
$ CAPTURE_LOG=0 iex -e 'if(System.get_env("CAPTURE_LOG") == "1", do: true, else: false) |> IO.inspect(); System.halt()'
false 

$ CAPTURE_LOG=1 iex -e 'if(System.get_env("CAPTURE_LOG") == "1", do: true, else: false) |> IO.inspect(); System.halt()'
true

# disabled by default
$ iex -e 'if(System.get_env("CAPTURE_LOG") == "1", do: true, else: false) |> IO.inspect(); System.halt()'
false

# disabled by default
$ CAPTURE_LOG=inomurko iex -e 'if(System.get_env("CAPTURE_LOG") == "1", do: true, else: false) |> IO.inspect(); System.halt()'
false
```

unnawut:
My original suggestion is to address the issue that `CAPTURE_LOG=0` and `CAPTURE_LOG=1` both result in muted logs.

unnawut:
Okay I misunderstood that `CAPTURE_LOG=1` meant showing logs. It's actually the [opposite](https://hexdocs.pm/ex_unit/ExUnit.html#configure/1-options). Correction:

```elixir
if(System.get_env("CAPTURE_LOG") == "0", do: false, else: true
```

InoMurko:
Sure! It would be good we have a solution we can all agree with. I personally don't mind logs printed during testing so this PR came more as a friendly response to @achiurizo request.

If this was an easy quick PR to get it over with I would gladly keep this open and maintain... I'll close this and @achiurizo if this is still a requirement for you open an issue where we can discuss it further. 
At the moment I think we all have priorities more urgent than this!

InoMurko:
Oh yeah, the docker fixes ive extracted into https://github.com/omisego/elixir-omg/pull/1264/files pls review asap :D

Issue #1248, opened 2020-01-08
RocksDB dependency issue when running tests and using Make


okalouti:
As I work on the Informational Watcher API, two key elements of the development process are the following: 

* **Testing**: I write tests and run them with `mix test`

* **Visualising API changes:** If I write code to modify the API response, I would like to visualise this with `Insomnia` or `Postman`. To reflect changes on containers that were started with `docker-compose up`, I run `make docker-update-watcher_info`. 

This flow is hampered by `:rocksdb`. 

If I run tests, I will get the following error when I run `make docker-update-watcher_info`: 

```
** (Mix) Could not compile dependency :rocksdb, "/home/elixir-omg/.mix/rebar3 bare compile --paths "/app/_build/prod/lib/*/ebin"" command failed. You can recompile this dependency with "mix deps.compile rocksdb", update it with "mix deps.update rocksdb" or clean it with "mix deps.clean rocksdb"

```
Running `mix deps.clean --all` followed by `mix deps.get` will fix the issue, but then I will get the above error if I try to run tests. 

The result is that much time is lost cleaning and re-installing dependencies.

A fix would allow me to test and (re)build containers concurrently without this dependency issue as I work on `watcher_info`

N.B. @unnawut explains that this due to the dependency cache being shared - and that separate dependency caches may sort the issue.

unnawut:
Expanding on

> due to the dependency cache being shared - and that separate dependency caches may sort the issue.

It's this line https://github.com/omisego/elixir-omg/blob/master/Makefile#L244:

```makefile
docker run --rm -it \
  ...
  -v $(IMAGE_BUILD_DIR)/deps:/app/deps \
  ...
```

And because rocksdb is compiled from source, it's incompatible between the host's MacOS and docker container's linux, so the deps can't be shared.

InoMurko:
Yep exactly that @unnawut . Container mounts the local volume and it leaves compilations in the mounted volume. Containers are linux, host is osx and compilation outputs are not compatible.

Idk if there's a solution to this... if you go back and forth many times I suggest you use baremetal cluster.
```
make start-services
make start-child_chain
make start-watcher
```

InoMurko:
@okalouti would that work? ^^

InoMurko:
Thought of something:
There must be a way to redirect dependency build directories. So we could push rocksdb into
`/tmp/rocksdb/_docker_build`
`/tmp/rocksdb/_build`
and that should know where to get the build from based on OS.

[...]

PR #1267, opened 2020-01-17
refactor: build dir in docker

InoMurko:
https://github.com/omisego/elixir-omg/issues/1248
## Overview

This PR would identify a container build and use _build_docker and _deps_docker folders for compilation and dependencies.

## Changes

- builder and tester images now have an environment variables set called `DOCKER=true` https://github.com/omisego-images/docker-elixir-omg/pull/26
- if the ENV var is populated ^^, append `_docker` string in the directory name

## Testing

/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -382,8 +382,8 @@ jobs:
       - run:
           name: Unpack PLT cache
           command: |
-            mkdir -p _build/test
-            cp plts/dialyxir*.plt _build/test/ || true
+            mkdir -p _build_docker/test
+            cp plts/dialyxir*.plt _build_docker/test/ || true
             mkdir -p ~/.mix
             cp plts/dialyxir*.plt ~/.mix/ || true
       - run: mix dialyzer --plt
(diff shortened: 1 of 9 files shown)


InoMurko:
> despite throwing the `_build` thing in the air in #1248, I'm still a bit frightened by this. At least let's document it somewhere very visibly, as this is deviating from expected `mix` project structure considerably.
> 
> My bet is to put it here: https://github.com/omisego/elixir-omg/#testing--development.
> 
> Apart from this, LGTM!

I agree it's icky icky... but that's the cleanest approach from all of them (thanks again for remembering it!). 

I think we should just try it for a while... see if we like it. If not, we can revert, it's not ta big change after all.

unnawut:
I'm late. 😢  It might be possible to map one directory, then override its subdirectory to a different location: https://stackoverflow.com/questions/29181032/add-a-volume-to-docker-but-exclude-a-sub-folder

So we could map `/app`, then map `/app/_build` to the host's `_build_docker/` dir. This way we don't diverge from mix convention from the app's perspective.

I'll try this sometime later on this week.

PR #1271, opened 2020-01-20
fix: dev release startup

InoMurko:
my blocker 🗡 

## Overview

This moves the conformance property based tests into their own umbrella application.
Why?
```
== Compilation error in file test/support/conformance/property.ex ==
** (CompileError) test/support/conformance/property.ex:22: module PropCheck is not loaded and could not be found
    (elixir) expanding macro: Kernel.use/1
    test/support/conformance/property.ex:22: Support.Conformance.PropertyGenerators (module)
==> omg_child_chain
```
Propcheck was unable to work together with dev releases (those that don't pack the VM and runtime into a tar.gz, making local dev faster - if you're not on docker + linux).

It also enables `test_barebone_release` job. It's not a mandatory job you'd need to wait before you can merge. But since this gets regularly broken let's keep here. Nightly runs are easily ignored when they don't work as they should.
## Changes

- moved support and test files for procheck into their own umbrella app
- fixed test_barebone_release
- removed solc from bin/setup
- 

## Testing
Same tests apply! + test_barebone_release


InoMurko:
```
== Compilation error in file test/support/conformance/property.ex ==
** (CompileError) test/support/conformance/property.ex:22: module PropCheck is not loaded and could not be found
    (elixir) expanding macro: Kernel.use/1
    test/support/conformance/property.ex:22: Support.Conformance.PropertyGenerators (module)
==> omg_child_chain
```

.circleci/config.yml
```diff
@@ -493,15 +494,30 @@ jobs:
...
+          name: Run Child Chain, Watcher and Watcher Info
+          command: |
+            set -e
             make start-child_chain OVERRIDING_START=start &&
-            make start-watcher OVERRIDING_START=start
+            make start-watcher OVERRIDING_START=start &&
             make start-watcher_info OVERRIDING_START=start
+          background: true
           no_output_timeout: 2400
+      - run: sleep 2m
```
unnawut: 😮

Makefile
```diff
@@ -366,15 +366,15 @@ start-watcher:
...
+	echo "Building Watcher Info" && \
 	make build-watcher_info-${BAREBUILD_ENV} && \
 	echo "Potential cleanup" && \
 	rm -f ./_build/${BAREBUILD_ENV}/rel/watcher_info/var/sys.config || true && \
 	echo "Init Watcher DBs" && \
 	_build/${BAREBUILD_ENV}/rel/watcher_info/bin/watcher_info init_key_value_db && \
 	_build/${BAREBUILD_ENV}/rel/watcher_info/bin/watcher_info init_postgresql_db && \
 	echo "Init Watcher DBs DONE" && \
-	echo "Run Watcher" && \
+	echo "Run Watcher Info" && \
```
unnawut: Can you also update above to "Init Watcher Info DB DONE"?

Makefile
```diff
@@ -402,7 +402,7 @@ stop-watcher:
 	_build/dev/rel/watcher/bin/watcher stop
 
 stop-watcher_info:
-	_build/dev/rel/watcher/bin/watcher stop
+	_build/dev/rel/watcher_info/bin/watcher_info stop
```
unnawut: 🙇

apps/omg_conformance/mix.exs
```diff
@@ -0,0 +1,36 @@
...
+    ]
+  end
+
+  defp elixirc_paths(:prod), do: ["lib"]
+  defp elixirc_paths(:dev), do: ["lib"]
+  defp elixirc_paths(:test), do: ["lib", "test/support"]
+
+  defp deps do
+    [
+      {:propcheck, "~> 1.1", only: [:test, :dev]},
```
pdobacz: ok, so now since you don't compile it in dev anymore, you should be able to drop it from here too.

PR #1273, opened 2020-01-22
feat: add /stats.get endpoint

### LINKED ISSUE #1185: Watcher Info API: /stats.get for network stats

okalouti:
Closes #1185 

## Overview

Create a `/stats.get` endpoint to retrieve network statistics currently shown by the block explorer. 

Objectives: 
- Remove direct database calls from block explorer and move dependency to Watcher Info API completely. 
- Reduce likelihood of BE cluster errors due to database permissions. 

Example Response: 

```
{
  "data": {
    "average_block_interval": {
      "all_time": 2057.6373626373625,
      "last_24_hours": "N/A"
    },
    "block_count": {
      "all_time": 92,
      "last_24_hours": 0
    },
    "transaction_count": {
      "all_time": 93,
      "last_24_hours": 0
    }
  },
  "service_name": "watcher_info",
  "success": true,
  "version": "0.3.0+6a1a4cf"
}
```

## Changes

- Adds logic to `db/block.ex` to count blocks and retrieve timestamps (all time and 24 hours) 
- Adds logic to `db/transaction.ex` to count transactions
- Create `api/stats.ex` file (block interval averaging is done here)
- Create `controllers/stat.ex` file. 
- Create `views/stats.ex` file.
- Adds new route to `router.ex`
- Updates Swagger specs with new endpoints

## Testing

Tests written for all levels above. 

`mix test test/omg_watcher_rpc/web/controllers/stats_test.exs`
`mix test test/omg_watcher_info/api/stats_test.exs`
`mix test test/omg_watcher_info/db/block_test.exs`
`mix test test/omg_watcher_info/db/transaction_test.exs`

Note: 

I am aware the database count statements may pose problems in a scenario where we have a large number of transactions. I have not addressed this - I have simply moved equivalent calls that exist already in the BE NodeJS server to our Watcher Info API.


apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
```diff
@@ -0,0 +1,67 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.WatcherInfo.API.Stats do
+  @moduledoc """
+  Module provides operations related to network statistics.
+  """
+
+  alias OMG.WatcherInfo.DB.{Block, Transaction}
```
InoMurko: I haven't seen us using this pattern anywhere else. best to stick with consistency:
```
alias OMG.WatcherInfo.DB.Block
alias OMG.WatcherInfo.DB.Transaction
```

InoMurko:
A general observation:
Statistics can be greatly influenced by the consumer of OMG BUS.

For example:
The consumer's only promises that it will eventually process the events being produced. This means, that you can get a discrepancy between inserts and actual data production. (the moment they're produced and the moment they're inserted).
Without going to deep, are you accounting for that?

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
okalouti: Thanks.Done 98fbd15.

okalouti:
Hi @InoMurko. Thank you for bringing this up. To recap what you explained to me yesterday, the simplified flow is the following: 

1. Child chain pushes a block to Ethereum Plasma contract.
2. Security Watcher detects that a block has been pushed
3. Security Watcher pushes the block creation event onto the `OMG Bus` queue
4. Informational Watcher picks this up and inserts the block into the DB. The timestamp on the block represents the time of **insertion** and not the time of creation. 

If there is a lag time between steps `3` and `4`, for example we can have a situation where the time of block submission and the timestamp on the block in the `Watcher DB` show a significant discrepancy.

As discussed, although it is great to be aware of this, we will leave it aside for this PR. No major problem from the user perspective and the `BE` is not part of this cycle. 

For future reference, I have created this issue https://github.com/omisego/blockexplorer/issues/114. 

Thanks again Ino!

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
```diff
@@ -0,0 +1,68 @@
...
+        last_24_hours: get_average_block_interval(timestamps_last_24_hours)
+      }
+    }
+
+    {:ok, response}
+  end
+
+  @spec get_average_block_interval([%{timestamp: integer}]) :: float | String.t()
+  def get_average_block_interval(timestamps) do
+    case length(timestamps) do
```
unnawut: http://erlang.org/doc/efficiency_guide/commoncaveats.html#length-1

Would be better to do...

```elixir
case timestamps do
  [_, _, | _] ->
    # calculate

  _ ->
    # skip
end
```

The reason I'm suggesting this here is because this function could be calculating for all blocks, not limited by a specific period.

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
```diff
@@ -0,0 +1,68 @@
...
+      n when n < 2 ->
+        String.to_atom("N/A")
+
+      n when n >= 2 ->
+        timestamps
+        |> Enum.chunk_every(2, 1, :discard)
+        |> Enum.map(fn [a, b] -> b.timestamp - a.timestamp end)
+        |> average()
```
unnawut: I think this could be mathematically reduced to `latest_block.timestamp - earliest_block.timestamp / num_blocks`.

apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex
```diff
@@ -92,6 +92,56 @@ defmodule OMG.WatcherInfo.DB.Block do
     |> Paginator.set_data(paginator)
   end
 
+  @spec query_last_24_hour(Ecto.Query.t()) :: Ecto.Query.t()
+  def query_last_24_hour(query) do
+    now = DateTime.to_unix(DateTime.utc_now())
+    twenty_four_hours = 86_400
```
unnawut: This looks like a static truth. Can move to module annotation i.e.

```elixir
defmodule OMG.WatcherInfo.DB.Block do
  #...

  @seconds_in_twenty_four_hours 86_400

  #...
end

File: apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex [hunk 5]
```diff
@@ -92,6 +92,56 @@ defmodule OMG.WatcherInfo.DB.Block do
     |> Paginator.set_data(paginator)
   end
 
+  @spec query_last_24_hour(Ecto.Query.t()) :: Ecto.Query.t()
+  def query_last_24_hour(query) do
```
unnawut: This is `DB` level functions should be as generic as possible. E.g.

```elixir
defmodule OMG.WatcherInfo.DB.Block do
  def query_timestamp_between(start_datetime, end_datetime) do
    # ...
  end
end

```

Then we have `OMG.WatcherInfo.API.Stats` module doing:

```elixir
end_datetime = DateTime.to_unix(DateTime.utc_now())
start_datetime = end_datetime - @seconds_in_twenty_four_hours

DB.Block.all_by_timestamp_between(start_datetime, end_datetime)
```

apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex
```diff
@@ -92,6 +92,56 @@ defmodule OMG.WatcherInfo.DB.Block do
...
+    now = DateTime.to_unix(DateTime.utc_now())
+    twenty_four_hours = 86_400
+    from(block in query, where: block.timestamp >= ^(now - twenty_four_hours))
+  end
+
+  @doc """
+  Returns the total number of blocks in the last 24 hours
+  """
+  @spec get_count_last_24_hour() :: non_neg_integer()
+  def get_count_last_24_hour do
```
unnawut:
```suggestion
  def count_by_timestamp_between(start_datetime, end_datetime) do
```

Same reason as my suggestion to `query_last_24_hour`

apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex
```diff
@@ -92,6 +92,56 @@ defmodule OMG.WatcherInfo.DB.Block do
...
+    from(block in __MODULE__, select: count())
+  end
+
+  @spec query_timestamps :: Ecto.Query.t()
+  defp query_timestamps do
+    from(block in __MODULE__, select: %{timestamp: block.timestamp})
+  end
+
+  @spec get_timestamps :: []
+  def get_timestamps() do
```
unnawut:
```suggestion
  def all_timestamps() do
```

Avoid mental load by following [Ecto convention](https://hexdocs.pm/ecto/Ecto.Repo.html) where `all` returns multiple results and `get` returns single.

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
unnawut: However, before making changes to this. Please see https://github.com/omisego/elixir-omg/pull/1273/files#r369961022

apps/omg_watcher_info/test/omg_watcher_info/api/stats_test.exs
```diff
@@ -0,0 +1,126 @@
...
+      mined_block_2 = %{
+        transactions: [tx_3, tx_4],
+        blknum: 2000,
+        blkhash: "0x2000",
+        timestamp: within_today,
+        eth_height: 1
+      }
+
+      _ = DB.Block.insert_with_transactions(mined_block_1)
+      _ = DB.Block.insert_with_transactions(mined_block_2)
```
unnawut: To decouple from `DB.Block.insert_with_transactions/1`, which isn't directly related to `API.Stats.get/0`, could we populate the test data using these [test factories](https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher_info/test/support/factory.ex#L63-L96) instead.

Applicable to a few more places.

apps/omg_watcher_info/test/omg_watcher_info/api/stats_test.exs
```diff
@@ -0,0 +1,126 @@
...
+  use OMG.WatcherInfo.Fixtures
+
+  import OMG.WatcherInfo.Factory
+
+  alias OMG.WatcherInfo.API.Stats
+  alias OMG.WatcherInfo.DB
+
+  @eth OMG.Eth.RootChain.eth_pseudo_address()
+
+  describe "get_stats/0" do
```
unnawut:
```suggestion
  describe "get/0" do
```

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
```diff
@@ -0,0 +1,68 @@
...
+    end
+  end
+
+  @spec average([integer]) :: float
+  def average(diff_array) do
```
unnawut: Move this to `OMG.Utils`?

unnawut:
Love the extensive tests! 🎖 ❤️

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
unnawut: This comment serves no other purpose than showing off my rusty and potentially flawed middle school math skillz.

9BBD5C97-CD46-48B5-BEB6-291FB1EA90B3

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
okalouti: Removed this method in favour of your amazing mathematics.

apps/omg_watcher_info/test/omg_watcher_info/api/stats_test.exs
okalouti: Done 61afded

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
okalouti: Woah! Amazing! Done 61afded.

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
okalouti: Thanks for the syntactical (and time complexity) improvement! Done 61afded.

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    {:ok, response}
+  end
+
+  @spec get_average_block_interval([%{timestamp: integer}]) :: float | String.t()
+  def get_average_block_interval(timestamps) do
+    case length(timestamps) do
+      n when n < 2 ->
+        String.to_atom("N/A")
```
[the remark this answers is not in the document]
okalouti: Agreed. Done 61afded.

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
```diff
@@ -0,0 +1,68 @@
...
+    response = %{
+      transaction_count: %{
+        all_time: Transaction.get_count(),
+        last_24_hours: Transaction.get_count_last_24_hour()
+      },
+      block_count: %{
+        all_time: Block.get_count(),
+        last_24_hours: Block.get_count_last_24_hour()
+      },
+      average_block_interval: %{
```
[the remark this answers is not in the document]
okalouti: Done 61afded.

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
```diff
@@ -20,49 +20,58 @@ defmodule OMG.WatcherInfo.API.Stats do
...
+  @doc """
+  Calculates the average of the differences between timestamps.
+  """
+  @spec get_average_block_interval([%{timestamp: non_neg_integer()}]) :: float | String.t()
   def get_average_block_interval(timestamps) do
-    case length(timestamps) do
-      n when n < 2 ->
-        String.to_atom("N/A")
+    case timestamps do
+      [_, _ | _] ->
```
InoMurko: this case is a bit sub-optimal. 
for example:
```
[first, a | b]  = [1,2,3,4]
[1, 2, 3, 4]
iex(4)> a
2
iex(5)> b
[3, 4]
```

apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex
okalouti: Done 8946c62.

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
```diff
@@ -20,49 +20,58 @@ defmodule OMG.WatcherInfo.API.Stats do
...
+          |> List.last()
+          |> Map.get(:timestamp)
 
-  @spec average([integer]) :: float
-  def average(diff_array) do
-    diff_array
-    |> Enum.reduce(0, fn current, acc -> current + acc end)
-    |> Kernel./(length(diff_array))
+        # Formula for average of difference
+        (last - first) / (length(timestamps) - 1)
```
InoMurko: hm... so if, by coincidence, last item and first item have accidental huge difference (network latency, postgres latency)... this wouldn't be representable or am I wrong?

apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex
okalouti: Done in various files with 8946c62 and 6ea2c86.

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
InoMurko: https://github.com/omisego/elixir-omg/pull/1273/files/98fbd15582f2245d4b7c9095448ad87a66cae5aa..8946c62c598f52b5dbe7f2b718510119ded86b25#r370544428

apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex
okalouti: Done 8946c62 but opting for `count_all_between_timestamps`

apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex
okalouti: Done 8946c62.

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
InoMurko: but I didn't measure this so it might be just a slight difference (for the better or worse).

```
def get_average_block_interval([%{timestamp: first_timestamp} | timestamps] = all_timestamps) do
  %{timestamp: last_timestamp} = List.last(timestamps)
  (last_timestamp - first_timestamp) / (length(all_timestamps) - 1)
end
def get_average_block_interval(_), do: nil
```

unnawut: It's used to detect that the list has at least 2 items though. Otherwise we can't find a time difference.

apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex
```diff
@@ -92,6 +92,56 @@ defmodule OMG.WatcherInfo.DB.Block do
...
+  defp query_count do
+    from(block in __MODULE__, select: count())
+  end
+
+  @spec query_timestamps :: Ecto.Query.t()
+  defp query_timestamps do
+    from(block in __MODULE__, select: %{timestamp: block.timestamp})
+  end
+
+  @spec get_timestamps :: []
```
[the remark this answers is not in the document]
okalouti: Done as per the first suggestion `@spec get_timestamps :: [non_neg_integer()]`. 

I would have needed to define a type for a map `%{timestamp: non_neg_integer()}` and felt  like I did not want to define a type for returning a single field in `block`.

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
okalouti: Noted thanks @Ino! As @unnawut notes this will not pose a problem for this purpose specifically.

okalouti:
Added logic to order timestamps in ascending order: ade5186.

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
InoMurko: I don't understand the connection... 
`[_|_]` is pattern matching a non empty list. Head | rest. 
in your pattern you're matching first, second and the rest... why?

InoMurko: ahaaa, you need two. lol. brain fart!

okalouti: Also misunderstood. The above is what we are looking for! Thanks both.

unnawut:
Could you help double-check that the `block.timestamp` and `transaction.timestamp` columns are indexed since we're doing ordering and scoping on them?

apps/omg_watcher_info/test/omg_watcher_info/api/stats_test.exs
okalouti: Done cf39f9d in: 
- db/transaction_test.exs
- db/block_test.exs
- api/stats_test.exs
- controllers/stats_test.exs

apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex
```diff
@@ -125,6 +125,23 @@ defmodule OMG.WatcherInfo.DB.Block do
...
+  def get_timestamps() do
+    DB.Repo.all(query_timestamps())
+  end
+
+  @spec get_timestamps_last_24_hours :: []
+  def get_timestamps_last_24_hours() do
```
jrhite: Curious, could you please provide the SQL this code generates? Just asking because I'm wondering if it might make sense to change `block.timestamp` from a simple postgres `integer` to an actual `timestamp` object.

This makes our storage of datetime-like columns consistent across all tables/columns - a small thing, but nice to have. Another, possibly bigger benefit would be the ability to use some of postgres' built-in datetime functionality like `interval`, etc..

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
```diff
@@ -0,0 +1,77 @@
...
+          timestamps
+          |> List.last()
+          |> Map.get(:timestamp)
+
+        # Formula for average of difference
+        (last - first) / (length(timestamps) - 1)
+
+      _ ->
+        nil
+    end
```
jrhite: Why not perform this calculation in the database natively as part of the query? Will certainly be faster.

apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex
```diff
@@ -92,6 +92,67 @@ defmodule OMG.WatcherInfo.DB.Block do
...
+
+  @doc """
+  Returns the total number of blocks in between given timestamps
+  """
+  @spec count_all_between_timestamps(non_neg_integer(), non_neg_integer()) :: non_neg_integer()
+  def count_all_between_timestamps(start_datetime, end_datetime) do
+    query_count()
+    |> query_timestamp_between(start_datetime, end_datetime)
+    |> DB.Repo.all()
+    |> Enum.at(0)
```
jrhite: Instead of:
```
  |> DB.Repo.all()
  |> Enum.at(0)
```
how about 
```
DB.Repo.one!()
```
I think this represents what you want more.

apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex
```diff
@@ -92,6 +92,67 @@ defmodule OMG.WatcherInfo.DB.Block do
...
+  end
+
+  @doc """
+  Returns the total number of blocks
+  """
+  @spec count_all() :: non_neg_integer()
+  def count_all() do
+    query_count()
+    |> DB.Repo.all()
+    |> Enum.at(0)
```
jrhite: same: `DB.Repo.one!()`

jrhite:
> 4\. Informational Watcher picks this up and inserts the block into the DB. The timestamp on the block represents the time of **insertion** and not the time of creation.

Is this actually correct? I thought `timestamp` is an actual Ethereum block timestamp. Can we please confirm?

unnawut:
> 1. Currently `blocks.timestamp` is stored in the DB as a simple integer type.
> 
> ```
> select blknum, timestamp, to_timestamp(timestamp) from blocks;
>  blknum | timestamp |   to_timestamp    
> --------+------------+------------------------
>   1000 | 1578996566 | 2020-01-14 10:09:26+00
>   2000 | 1579061008 | 2020-01-15 04:03:28+00
>   3000 | 1579075913 | 2020-01-15 08:11:53+00
> ```
> 
> It might be worth considering changing and storing this value as a `timestamp without time zone(0)` (timestamp with seconds precision). Have a look. I think there may be some native postgres timestamp functionality we could leverage here which could be a big plus. The minus being is it ties us to postgres (don't see that changing though anywhere in the near term though).

FYI we had a previous discussion on this:

> Jacob Dec 27th, 2019 at 07:05
> Also on the topic of DateTime values in the DB, the blocks table has a timestamp column which is an integer. It's stored as seconds since epoch.
> Do we want to keep it this way? My thoughts and questions here are:
> 1. Seconds seems like not enough precision. Increase precision to at least millis.
> 2. Any objections/reasons not to store timestamp as utc_datetime_usec
> 
> 4 replies
> 
> unnawut  1 month ago
> I think that’s the rootchain block’s timestamp: https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher_info/lib/omg_watcher_info/block_application_consumer.ex
> If it’s just a direct copy I think we can keep it the same as the source (source is seconds precision)?
> 
> Jacob  1 month ago
> ah, wow...thought eth timestamp would be more precise....but i guess blocks aren't the quickest things to be created in the world.
> agreed though...just stick with eth units
> 
> unnawut  1 month ago
> I remember someone said it’s determined by the block miner so it’s not even used for anything sensitive :sweat_smile:
> 
> Jacob  1 month ago
> ah, ok....not even worth much of a discussion then...i'll just leave it as a integer timestamp that is epoch in seconds

> Seeing math calculations being done in the code vs the DB where they'd be quicker, you might consider moving the calculations to the DB.

I would argue the opposite, to keep the load off the database.

Sidenote: Although I see the point that it could be optimized a bit more to just query the first and last to do the calculation, instead of querying the whole range. But up to @okalouti to decide to include this in this PR or not.

> 3. The count queries in the PR are written as:
> ```
> |> DB.Repo.all()
> |> Enum.at(0)
> ```
> I think it's better written simply as
> ```
> |> DB.Repo.one!()
> ```
> which really captures your intention, including raising errors if there are programmatic problems.

👍

> 4. As we began to touch on in our chat, it's worth a discussion of creating an index for the `blocks.timestamp` column, let's talk more about this.

https://github.com/omisego/elixir-omg/pull/1273#issuecomment-578596180

jrhite:
> > 1. Currently `blocks.timestamp` is stored in the DB as a simple integer type.
> > 
> > ```
> > select blknum, timestamp, to_timestamp(timestamp) from blocks;
> >  blknum | timestamp |   to_timestamp    
> > --------+------------+------------------------
> >   1000 | 1578996566 | 2020-01-14 10:09:26+00
> >   2000 | 1579061008 | 2020-01-15 04:03:28+00
> >   3000 | 1579075913 | 2020-01-15 08:11:53+00
> > ```
> > 
> > 
> > It might be worth considering changing and storing this value as a `timestamp without time zone(0)` (timestamp with seconds precision). Have a look. I think there may be some native postgres timestamp functionality we could leverage here which could be a big plus. The minus being is it ties us to postgres (don't see that changing though anywhere in the near term though).
> 
> FYI we had a previous discussion on this:

Ah yes, I was searching for this chat, but was in a rush. Thank you for adding it here.

> 
> > Jacob Dec 27th, 2019 at 07:05
> > Also on the topic of DateTime values in the DB, the blocks table has a timestamp column which is an integer. It's stored as seconds since epoch.
> > Do we want to keep it this way? My thoughts and questions here are:
> > 
> > 1. Seconds seems like not enough precision. Increase precision to at least millis.
> > 2. Any objections/reasons not to store timestamp as utc_datetime_usec
> > 
> > 4 replies
> > unnawut  1 month ago
> > I think that’s the rootchain block’s timestamp: https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher_info/lib/omg_watcher_info/block_application_consumer.ex
> > If it’s just a direct copy I think we can keep it the same as the source (source is seconds precision)?
> > Jacob  1 month ago
> > ah, wow...thought eth timestamp would be more precise....but i guess blocks aren't the quickest things to be created in the world.
> > agreed though...just stick with eth units
> > unnawut  1 month ago
> > I remember someone said it’s determined by the block miner so it’s not even used for anything sensitive 😅
> > Jacob  1 month ago
> > ah, ok....not even worth much of a discussion then...i'll just leave it as a integer timestamp that is epoch in seconds

@unnawut @InoMurko - conclusion here: timestamp is ethereum block timestamp.

> 
> > Seeing math calculations being done in the code vs the DB where they'd be quicker, you might consider moving the calculations to the DB.
> 
> I would argue the opposite, to keep the load off the database.
> 
> Sidenote: Although I see the point that it could be optimized a bit more to just query the first and last to do the calculation, instead of querying the whole range. But up to @okalouti to decide to include this in this PR or not.

There is really no right answer to this. DBs do amazing optimizations on `avg` and other types of math operations that are 'map-reduce'-like in nature which is what avg is (ie: the input is a lot larger than the output). Moving this type of stuff to the DB can also be particularly helpful when joins start getting involved.

Regarding keeping load off the database...as far as I know we don't know where the load and bottlenecks will be yet (hope our feat-load team can help here!). In the Elixir code, in the DB queries, etc... Also it's not clear how our infrastructure will end up (ie: how will we scale the app code, how will we scale the DB servers, etc...). Infrastructure setup likely will be an analysis of where load/bottlenecks are, cost of spinning up app servers vs db servers, how many app servers a single db server can serve, etc...

Totally fair to leave code as is since we don't know where the bottlenecks will be yet. Premature optimization is not a good thing to spend time on.

> 
> > 1. The count queries in the PR are written as:
> > 
> > ```
> > |> DB.Repo.all()
> > |> Enum.at(0)
> > ```
> > 
> > 
> > I think it's better written simply as
> > ```
> > |> DB.Repo.one!()
> > ```
> > 
> > 
> > which really captures your intention, including raising errors if there are programmatic problems.
> 
> 👍
> 
> > 1. As we began to touch on in our chat, it's worth a discussion of creating an index for the `blocks.timestamp` column, let's talk more about this.
> 
> [#1273 (comment)](https://github.com/omisego/elixir-omg/pull/1273#issuecomment-578596180)

@okalouti one final thought regarding index creation on the `blocks.timestamp` column. I think this column, is for the most part, meaningless. In distributed systems, there is little chance of the ethereum miners being synced on time. (A block is what really becomes the unit of time.) It's possible (although unlikely) to get 2 ethereum blocks with identical timestamps...timestamps not being unique per block, might bust any hope of indexing on timestamp.

I have a PR adding `inserted_at` and `updated_at` to all tables which I'll ask for re-review shortly. I'd think `blocks.inserted_at` would be more `blocks.timestamp` for these calculations. Granted those columns don't exist, but will soon.

pnowosie:
> > 4. Informational Watcher picks this up and inserts the block into the DB. The timestamp on the block represents the time of **insertion** and not the time of creation.
> 
> Is this actually correct? I thought `timestamp` is an actual Ethereum block timestamp. Can we please confirm?

Regarding WatcherDB, you’re correct @jrhite  Timestamp of the Ethereum block in which plasma block merkle hash is published is persisted.

InoMurko:
> > > 1. Informational Watcher picks this up and inserts the block into the DB. The timestamp on the block represents the time of **insertion** and not the time of creation.
> > 
> > 
> > Is this actually correct? I thought `timestamp` is an actual Ethereum block timestamp. Can we please confirm?
> 
> Regarding WatcherDB, you’re correct @jrhite Timestamp of the Ethereum block in which plasma block merkle hash is published is persisted.

Indeed my statements were wrong. The block fetched from BlockGetter and passed down the rabbit hole comes with a timestamp.
```
  def get_child_chain(blknum, contract \\ %{}) do
    contract = Config.maybe_fetch_addr!(contract, :plasma_framework)
    Eth.call_contract(contract, "blocks(uint256)", [blknum], [{:bytes, 32}, {:uint, 256}])
  end
```
the second return argument (uint 256) is the timestamp. And this timestamp is persisted in PG
```
%OMG.WatcherInfo.DB.Block{
      __meta__: #Ecto.Schema.Metadata<:loaded, "blocks">,
      blknum: 4000,
      eth_height: 1725,
      hash: <<205, 240, 124, 4, 164, 142, 45, 71, 17, 153, 214, 221, 247, 168,
        247, 193, 44, 38, 36, 162, 10, 10, 254, 232, 225, 45, 178, 203, 193,
        101, 167, 246>>,
      timestamp: 1580163289,
      transactions: #Ecto.Association.NotLoaded<association :transactions is not loaded>,
      tx_count: 1
    }
```

apps/omg_watcher_rpc/priv/swagger/info_api_specs.yaml
```diff
@@ -2526,3 +2528,143 @@ paths:
...
+                      data:
+                        transaction_count:
+                          all_time: 4
+                          last_24_hours: 2
+                        block_count:
+                          all_time: 2
+                          last_24_hours: 1
+                        average_block_interval:
+                          all_time: 100
+                          last_24_hours: N/A
```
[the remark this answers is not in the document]
okalouti: Thanks for catching. c85a853.

apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex
okalouti: Thanks Jacob. Done 7fbf75b.

apps/omg_watcher_info/lib/omg_watcher_info/db/block.ex
okalouti: Done 7fbf75b.

apps/omg_watcher_info/lib/omg_watcher_info/db/transaction.ex
```diff
@@ -96,6 +96,39 @@ defmodule OMG.WatcherInfo.DB.Transaction do
...
 
+  @spec query_count :: Ecto.Query.t()
+  defp query_count do
+    from(transaction in __MODULE__, select: count())
+  end
+
+  @spec query_timestamp_between(Ecto.Query.t(), non_neg_integer(), non_neg_integer()) :: Ecto.Query.t()
+  def query_timestamp_between(query, start_datetime, end_datetime) do
+    from(transaction in query,
+      left_join: block in assoc(transaction, :block),
```
jrhite:
```suggestion
      join: block in assoc(transaction, :block),
```

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
InoMurko: you're wrong @InoMurko

apps/omg_watcher_info/lib/omg_watcher_info/db/transaction.ex
okalouti: Thanks Jacob for catching this and explaining. Done fdd87a1.

apps/omg_watcher_info/lib/omg_watcher_info/api/stats.ex
okalouti: Thanks Jacob. I refactored the code here. 
* The DB call now will directly fetch the `max` and `min` timestamps (either all time or within a time range) - instead of this being done as above in Elixir.
* Elixir will run the rest of the averaging formula `(max - min)/ count - 1`

No strong opinion as to whether it should all be in the database natively, but let's leave aside for now and optimise later if the need arises.

okalouti:
Hello everyone. Thanks a lot for your feedback and attention on this PR! 

In the last few commits, I have incorporated various points raised: 

**- 597c2fe: indexing blocks on `timestamp`**

**- e430145: refactor the logic to calculate average block interval.** 

The database call now fetches the `max` and the `min` for the averaging formula directly, as opposed to fetching all timestamps. The rest of the averaging logic is done on layer above.

**7fbf75b: switch from `Repo.all` to `Repo.one`** 

**fdd87a1**: switch from `left_join` to `join`

For clarity, things I am opting to set aside even if good points raised: 

- The question of whether the averaging should fully be done on the database level. Think it is fine as it is and we can optimise down the line should the necessity arise. 
- The type of `timestamp` and whether to use another column related to Jacob's upcoming PR.  
- Issues creating a migration file in `watcher_info`. The issue is probably with the missing `jason` dependency in `watcher_info` but getting stuck in loops of `mix deps.get` / `mix deps.clean` as I try to replicate this as an actual fix and think best left out of this PR. 

Thank you all again for the eye on this PR! Intend to merge tomorrow unless any final objections!

jrhite:
Well done @okalouti! This one was a long one. 😉 Thanks for your patience with all the comments.

I've approved this, but do have one suggestion you might want to do before merging. Since there is an actual DB migration in this PR, it would be desirable to actually test that the DB migration works. Unfortunately we don't have automated tests in place for DB migration.

If possible, I'd suggest running the migrations locally with:
1. a completely deleted omisego-dev DB (for devs spinning up new workspaces, etc...)
2. an omisego-dev DB with tables that are empty ((for devs spinning up new workspaces, etc...)
3. and last, but not least, grab a dump of the dev DB and try running the migration on the data there.

InoMurko:
> Well done @okalouti! This one was a long one. 😉 Thanks for your patience with all the comments.
> 
> I've approved this, but do have one suggestion you might want to do before merging. Since there is an actual DB migration in this PR, it would be desirable to actually test that the DB migration works. Unfortunately we don't have automated tests in place for DB migration.
> 
> If possible, I'd suggest running the migrations locally with:
> 
>     1. a completely deleted omisego-dev DB (for devs spinning up new workspaces, etc...)
> 
>     2. an omisego-dev DB with tables that are empty ((for devs spinning up new workspaces, etc...)
> 
>     3. and last, but not least, grab a dump of the dev DB and try running the migration on the data there.

or maybe do this:
- checkout master and let it sync with one of our networks
- create yourself a local release out of your branch
- run it towards the same database as in step 1

This would "migrate" your postgres to the new version. The difficult part is probably finding a network that has ANY traffic :)

PR #1281, opened 2020-01-28
Prevent overpaying fees

T-Dnzt:

## Overview

This PR prevents users of the network from overpaying fees.

## Changes

- Return the error `overpaying_fees` when a user tries to pay more than the defined amount of fees
- Update all tests & fee specs file to integrate the fee amount (mostly set to 1 ETH or 2 TOKENs)

## Testing

- Send a transaction overpaying fees -> get `overpaying_fees` errror
- Send a transaction with the correct amount of fees -> get tx included in a block

## To Do

- [x] Add integration tests for overpaying fees


apps/omg/test/omg/fees_test.exs
```diff
@@ -54,30 +54,35 @@ defmodule OMG.FeesTest do
...
-    test "returns false when the implicit fees currency does not match any of the supported fee currencies" do
-      other_currency = <<2::160>>
-      refute Fees.covered?(%{other_currency => 100}, @payment_fees)
+    test "returns fees_not_covered when fees do not cover the fee price" do
+      assert Fees.check_if_covered(%{@not_eth_1 => 1}, @payment_fees) == {:error, :fees_not_covered}
+    end
+
+    test "returns overpaying_fees when fees cover more than the fee price" do
+      assert Fees.check_if_covered(%{@not_eth_1 => 4}, @payment_fees) == {:error, :overpaying_fees}
     end
```
pdobacz: the new behavior added to the module is that `0` fees "don't count" and don't allow transactions through, I think this isn't tested.

BTW, are we testing for existence `: 0` entries in the fee specs early on, and rejecting such fee specs or we're ok with such entries?

priv/apps/itest/test/itest/in_flight_exits_test.exs
```diff
@@ -12,6 +12,7 @@ defmodule InFlightExitsTests do
   alias Itest.ApiModel.IfeOutputChallenge
   alias Itest.ApiModel.SubmitTransactionResponse
   alias Itest.Client
+  alias Itest.Fee
```
achiurizo: `cabbage` tests? ❤️ 😻 ❤️

apps/omg/test/omg/fees_test.exs
mederic-p: We were ok until now, will update the parser in this PR to reject `0` fee amounts

apps/omg/lib/omg/state/transaction/validator.ex
```diff
@@ -50,7 +50,7 @@ defmodule OMG.State.Transaction.Validator do
          {:ok, outputs_spent} <- UtxoSet.get_by_inputs(utxos, inputs),
          :ok <- authorized?(outputs_spent, witnesses),
          {:ok, implicit_paid_fee_by_currency} <- Transaction.Protocol.can_apply?(raw_tx, outputs_spent),
-         true <- Fees.covered?(implicit_paid_fee_by_currency, fees) || {:error, :fees_not_covered} do
+         :ok <- Fees.check_if_covered(implicit_paid_fee_by_currency, fees) do
```
unnawut: `@type exec_error` needs `:overpaying_fees` and `:multiple_potential_currency_fees`?

apps/omg/lib/omg/fees.ex
```diff
@@ -34,7 +34,7 @@ defmodule OMG.Fees do
   where fees is itself a map of token to fee spec
   """
   @type full_fee_t() :: %{non_neg_integer() => fee_t()}
-  @type optional_fee_t() :: fee_t() | :no_fees_required
+  @type optional_fee_t() :: fee_t() | :no_fees_required | :no_fees_allowed
```
pnowosie: IMHO `no_fees_required` fits better to `merge` txs in the sense that network does not require fees for such txs... <br/>:point_right:  `no_fees_allowed` sounds to me like warning message :)

How about `fees_waived` to diff from fees-free network?

PR #1285, opened 2020-01-30
feat: bump merkle_tree to 2.0.0 with second preimage attack prevention

unnawut:
## Overview

State of the art merkle tree implementation.

Or more like bumping package versions with a fancy PR name.

## Changes

- Bumped `:merkle_tree` to `2.0.0` from hex repository.
- Bumped `:ex_plasma` to latest master head, so it also uses merkle_tree 2.0.0
  - Because of this, a small bit of change to `OMG.Eth.RootChainTest` to align with the new `{:ok, deposit}` tuple signature returned by `ExPlasma.Transaction.Deposit.new/1`

## Testing

Whole CI workflow should pass.


unnawut:
Gonna merge right away since it's the same merkle_tree code from @pgebal's `prevent_second_preimage_attack` branch that's been used previously. 🎉

PR #1286, opened 2020-01-31
feature: get configuration api

InoMurko:
https://github.com/omisego/devops/issues/221

## Overview

Adds `configuration.get` endpoint that allows service to pull configuration values:

## Changes
- slight refactor of how we retrieve configuraton
- internal API layer for child chain
- Watchers and Child Chain now expose the deployed plasma contracts it's able to talk to
- Watchers will respond with
```
{
"contract_semver": "1.0.1+987saf2",
        "deposit_finality_margin": 10,
        "network": "LOCALCHAIN",
        "exit_processor_sla_margin": 5520
}
```

- Child Chain will respond with
```
{
"contract_semver": "1.0.1+987saf2",
        "deposit_finality_margin": 10,
        "network": "LOCALCHAIN"
}
```
We can add more configs if needed.
## Testing

Integrated into cabbage tests.


InoMurko:
For now, I did not add the ability for the elixir-omg services to periodically check if plasma contracts changed it's version (by calling ```
Eth.call_contract(Encoding.from_hex(plasma_framework_contract), "getVersion()", [], [{:tuple, [:string]}])
```)
so for now, this is read once at startup and that's it. if you ALDd contracts and bumped versions you'd need to restart elixir-omg services). I don't see this occurring anytime in the near future though.

**Comment by InoMurko** [comment]
@kevsul I think perf team might find this useful (for example, to wait deposit finality margin blocks before continuing the test)

File: apps/omg_watcher/lib/omg_watcher/exit_processor.ex [hunk 1]
```diff
@@ -209,18 +209,21 @@ defmodule OMG.Watcher.ExitProcessor do
...
+  def init(
+        exit_processor_sla_margin: exit_processor_sla_margin,
+        metrics_collection_interval: metrics_collection_interval
+      ) do
     {:ok, db_exits} = DB.exit_infos()
     {:ok, db_ifes} = DB.in_flight_exits_info()
     {:ok, db_competitors} = DB.competitors_info()
 
-    sla_margin = Application.fetch_env!(:omg_watcher, :exit_processor_sla_margin)
+    sla_margin = exit_processor_sla_margin
```
unnawut: re-assignment not needed anymore

apps/omg_watcher_rpc/priv/swagger/info_api_specs/configuration/schemas.yaml
```diff
@@ -0,0 +1,15 @@
+ConfigurationSchema:
+  type: object
+  properties:
+    deposit_finality_margin:
+      type: integer
+      format: int256
+    contract_semver:
+      type: string
+      format: binary
```
unnawut: I just realized that `format: binary` is [for describing files](https://swagger.io/docs/specification/data-models/data-types/#format). I think we can remove the `format` attribute.

Looks like you already removed it from some files but not all. Also applies to `network` below, and a few more other files.

InoMurko: I guess I can remove this from everywhere right?

PR #1287, opened 2020-02-02
fix: solidity install

InoMurko:
Fixing nightly builds and moving them to main build but they're not mandatory for merge.
This way they're easier to observe when they get broken by PRs.

## Overview

Mac build and barebone build fixes

## Changes

- Input and ouput sort as per DB query
- Install Solidity 5.11 via RB url 

## Testing
Tests were already failing. Hopefully they'll pass now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,10 +73,10 @@ jobs:
       - run:
           name: Integration Tests
           command: |
-            # Slow, serial integration test, run nightly. Here to make sure the standard `mix test --trace --only integration` works
+            # Slow, serial integration test, run nightly. Here to make sure the standard `mix test --only integration --trace` works
             export SHELL=/bin/bash
             make init_test
-            mix test --trace --only integration
+            mix test --only integration --trace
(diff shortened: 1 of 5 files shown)


InoMurko:
Need to pimp byzantine events tests 😫

apps/omg_watcher_info/test/support/factory.ex
```diff
@@ -105,19 +105,19 @@ defmodule OMG.WatcherInfo.Factory do
...
-  def txoutput_factory() do
+  def txoutput_factory(attrs) do
     block = build(:block)
     transaction = build(:transaction, block: block)
 
     %DB.TxOutput{
       blknum: block.blknum,
       txindex: 0,
-      oindex: 0,
+      oindex: Map.get(attrs, :oindex, 0),
```
unnawut: Actually ExMachina will auto override the values if the keys match so you don't even need Map.get stuff here.

Kind of obscure at first, but it's one of the big uses of ExMachina.

Issue #937, opened 2019-09-09
Raised and cleared alarms need to be sent to Datadog as events


InoMurko:
Each raised alarms (system and application) needs to be sent to Datadog and Slack with
- git sha
- environment
so that we're able to properly react.
https://docs.datadoghq.com/api/?lang=python#post-an-event

InoMurko:
We should expand this with events and service checks:
- byzantine events could be sent to datadog
- alarms, already mentioned (system - disk, ethereum client connectivity, invalid fee file)
- service checks for watcher -> child chain

Currently we use Statix (https://github.com/lexmag/statix) that lacks support for both of the above (https://github.com/lexmag/statix/pull/26).

InoMurko:
I suggest we fork `statix` and maintain a mix of 
https://github.com/bleacherreport/statix/commits/otp-21.3.8.4-support 
and https://github.com/lexmag/statix/commits/master
they diverge after https://github.com/lexmag/statix/commit/2245dfbe869d6e266401ba7f89d8097ff8f7180b.

InoMurko:
- [ ] expose watcher and child chain configurational values (contract sha and finality margin)
- [ ] change devops values for APP_ENV, propose the domain naming change to fix metrics filtering in datadog
- [ ] alarms and byzatnine events into datadog

jrhite:
See slack discussion for additional info.

Would be good to tighten up the scope and definition of this issue, or change the title of it, depending on how we decide to split up and/or coordinate the work and requirements.

cc: @achiurizo @kevsul @InoMurko

kevsul:
To follow up after various conversations:

We agreed that most of what performance testing needs are metrics and can use the existing telemetry system. 

If stress tests break the system by hitting its limits, then we may be able to identify some events that should have been fired. Depending on the progress of the this issue, we can add them here or create new issues.

PR #1294, opened 2020-02-04
fix: push events to DD

InoMurko:
https://github.com/omisego/elixir-omg/issues/937

## Overview

Watcher, Watcher Info and Childchain detect certain events that plasma contract emit. We want to detect them and be notified if the chain might have byzantine events.
Our system raises and clears alarms. We want to be notified of them as well.

## Changes

- Refactored RootChain (exabi function selector created from plasma contract json abis) and OMG.Eth (decoding and RPC calls moved elsewhere)
- Added worker processes for each event (worker process is subscribed to a OMG Bus topic, Ethereum Event Listeners publish events on these topics)
- Added worker alarm handler that forwards alarms as events to Datadog
- cleaned up some websocket dependencies 
- added Spandex tracers to Child Chain application
- some exvcr tests were stale - instead of re-recording them I fallback to geth snapshots
- for all encoding (events binaries are hex encoded so that they're readable on datadog) there are extensive tests
- for all plasma contract emitted events we now have coverage for all events

Whats left to do after this:
I want to polish Datadog emitted events based on the feedback I get from ya'll.
## Testing

All the tests.


apps/omg_eth/lib/omg_eth/root_chain/hack_naming.ex
```diff
@@ -0,0 +1,102 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.Eth.HackNaming do
+  @moduledoc """
+  Adapt to naming from contracts to elixir-omg.
+
+  I need to do this even though I'm bleeding out of my eyes.
```
jrhite: 👀 😆

apps/omg_eth/lib/omg_eth/root_chain/hack_naming.ex
```diff
@@ -0,0 +1,102 @@
...
+  """
+  def deposit(data) do
+    # key is naming coming from plasma contracts
+    # value is what we use
+    contracts_naming = [{"token", :currency}, {"depositor", :owner}, {"blknum", :blknum}, {"amount", :amount}]
+
+    reduce_naming(data, contracts_naming)
+  end
+
+  # we always call it output_index, which is kinda weird?
```
pdobacz: not that much. An `input` is also an `output` :)

pdobacz: ah OK, yeah, what adds to the confusion here is that it used to be always `output_index`, when there was a single event and inputs were 0->3 and outputs 4->7. So agreed on weird, this is a missed leftover after the hasty push to ALD compatibility :(.

When the overall way you want to handle this renaming crystalizes, ping me to help out on how to reduce the eyebleed in this case.

The goal at the end of the pipeline is to have `{:input | :output, index}` and there is `OMG.EthereumEventListener.Preprocessor` that already does part of this transformation. It could also translate the specific `"outputIndex" | "inputIndex"` into just `index`.

apps/omg_eth/lib/eth.ex
```diff
@@ -116,152 +116,4 @@ defmodule OMG.Eth do
...
-    # NOTE: we're using `put_new` here, because `merge` would allow us to overwrite data fields in case of conflict
-    result
-    |> Map.put_new(:eth_height, int_from_hex(eth_height))
-    |> Map.put_new(:root_chain_txhash, from_hex(root_chain_txhash))
-    |> Map.put_new(:log_index, int_from_hex(log_index))
-    # just copy `event_signature` over, if it's present (could use tidying up)
-    |> Map.put_new(:event_signature, event[:event_signature])
-  end
-
-  defp put_signature(events, signature), do: Enum.map(events, &Map.put(&1, :event_signature, signature))
```
achiurizo: 👋 🚒

apps/omg_eth/lib/omg_eth/root_chain/hack_naming.ex
achiurizo: 😆

apps/omg_eth/test/omg_eth/root_chain_test.exs
```diff
@@ -16,110 +16,112 @@ defmodule OMG.Eth.RootChainTest do
...
-                  blknum: 1,
-                  owner: ^authority_addr,
-                  currency: @eth,
-                  eth_height: height,
-                  log_index: 0,
-                  root_chain_txhash: ^root_chain_txhash
-                }
-              ]} = deposits
-    end
+    # not using OMG.ChildChain.Transaction to not depend on that in omg_eth tests
```
pdobacz: not a matter of your change, but these 2 lines of comment are stale. Maybe let's take the opportunity and remove them

apps/omg_watcher_rpc/lib/web/endpoint.ex
```diff
@@ -17,7 +17,7 @@ defmodule OMG.WatcherRPC.Web.Endpoint do
   use Sentry.Phoenix.Endpoint
 
   # NOTE: one connects to `ws://host:port/socket/websocket` here (the transport is appended)
-  socket("/socket", OMG.WatcherRPC.Web.Socket, websocket: false)
+  # socket("/socket", OMG.WatcherRPC.Web.Socket, websocket: false)
```
pdobacz: remove?

.circleci/config.yml
```diff
@@ -309,9 +309,9 @@ jobs:
             # Don't submit coverage report for forks, but let the build succeed
             export SHELL=/bin/bash
             if [[ -z "$COVERALLS_REPO_TOKEN" ]]; then
-              mix coveralls.html --parallel --umbrella --trace
+              mix coveralls.html --parallel --umbrella --trace --exclude common --exclude watcher --exclude watcher_info --exclude child_chain --exclude test
```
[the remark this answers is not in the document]
InoMurko: OMG!!! 

OK fixed

apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex
```diff
@@ -599,12 +599,12 @@ defmodule OMG.ChildChain.BlockQueue.Core do
   end
 
   defp log_known_tx(submission) do
-    _ = Logger.debug("Submission #{inspect(submission)} is known transaction - ignored")
+    _ = Logger.warn("Submission #{inspect(submission)} is known transaction - ignored")
```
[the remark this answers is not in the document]
InoMurko: didn't this warn us of a certain condition (the one where private keys got mismatched between environments?)

apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
```diff
@@ -0,0 +1,95 @@
...
+  Integrated with the help of: https://docs.datadoghq.com/api/?lang=bash#post-an-event
+  Most things from the doc doesn't work. Either because Statix doesn't work or Datadog.
+  Date_happened, aggregation_key, source_type_name doesn't seem to appear in Events list.
+  Hence we transform everything into a tag.
+  """
+
+  require Logger
+  alias OMG.ChildChain.DatadogEvent.Encode
+
+  @doc """
```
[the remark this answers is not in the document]
InoMurko: good point!

apps/omg_eth/lib/omg_eth/root_chain/hack_naming.ex
InoMurko: yeah awesome! I think we can give this another pass once things settles a little bit!

apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex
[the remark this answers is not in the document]
pdobacz: oh my, good point :thinking:. However, IIRC, that log was just something we investigated as strange, but I suspect you'll see it on any occasion where a submission isn't mined and gets re-submitted. We just never saw it because `:debug`.

So, yes, it did provide some info, but I'd argue this was more of a coincidence.

In the end, not sure what to do about this. We can try and see, and if it is too chatty we revert to `debug` and possibly think of other warnings/checks that could be more accurate. Let's just make a self-note to not forget to check the chatiness.

apps/omg_eth/lib/omg_eth/root_chain/decode.ex
```diff
@@ -0,0 +1,527 @@
...
+      types: [:address, {:bytes, 32}, {:uint, 16}]
+    }
+  end
+
+  defp in_flight_exit_output_piggybacked() do
+    %ABI.FunctionSelector{
+      function: "InFlightExitOutputPiggybacked",
+      input_names: ["exitTarget", "txHash", "outputIndex"],
+      inputs_indexed: [true, true, false],
+      method_id: "n͎y",
```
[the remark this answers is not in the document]
InoMurko: yeah this is just how binaries are represented sadly 
```
<<method_id::binary-size(4), _::binary>> = :keccakf1600.hash(:sha3_256, "InFlightExitChallengeResponded(address,bytes32,uint256)")
```

```
IO.inspect method_id, binaries: :as_binaries
<<110, 205, 142, 121>>
"n͎y"
```

apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex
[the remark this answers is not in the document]
InoMurko: put back to :debug as it was spamming tests :)

apps/omg_eth/lib/omg_eth/root_chain/decode.ex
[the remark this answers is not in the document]
pdobacz: well yeah, I am aware of this, the thing to consider here is to write it down as `<<110, ...>>` in this file, since the unicode string there just looks super weird.

PR #1296, opened 2020-02-05
Configurable IGNORE_FEES env var

unnawut:
## Overview

Allows configurable :ignore_fees config via `IGNORE_FEES` env var.

## Changes

Added the `OMG.ChildChain.ReleaseTasks.SetIgnoreFees` release task that retrieves the env var and set to the config.

Allows either `IGNORE_FEES="true"` or `IGNORE_FEES="false"` or the env var must be unset, otherwise an error is raised.

## Testing

To ignore the fees, set the environment variable to `true` on application startup:

```
export IGNORE_FEES=true
```

Then proceed to submit a transaction without a fee.

### DIFF
--- a/apps/omg_child_chain/lib/omg_child_chain/release_tasks/set_ignore_fees.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/release_tasks/set_ignore_fees.ex
@@ -26,7 +26,7 @@ defmodule OMG.ChildChain.ReleaseTasks.SetIgnoreFees do
     _ = Application.ensure_all_started(:logger)
     ignore_fees = ignore_fees()
 
-    :ok = Application.put_env(@app, @config_key, interval_ms, persistent: true)
+    :ok = Application.put_env(@app, @config_key, ignore_fees, persistent: true)
   end
 
   defp ignore_fees() do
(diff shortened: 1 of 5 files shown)


InoMurko:
I wonder if ‘ignore fees’ has a use case? I thought it doesn’t. If thats true maybe forcing us to adopt to fees is a better way forward. Is there a no fee usecase @T-Dnzt ?

jrhite:
cc @kevinsul

This came up in our load testing team meeting yesterday.

The team would very much appreciate keeping for now if it's not too
much trouble. Fees makes testing a bit trickier.

We do intend to test with fees, but we're just not there yet.

Kevin can follow up if more info is needed.

T-Dnzt:
@InoMurko I'd prefer if everyone switched to using fees, but I think there are still some use cases for no fees (development env maybe?). Let's keep the `ignore_fees` option for a little while and remove it once everyone got used to them.

@jrhite @kevsul The problem with load testing without fees is that you're not actually testing a system mirroring the Mainnet, so what's the point? I would very much recommend starting with fees right away.

jrhite:
@T-Dnzt absolutely agreed, we're getting there and the end goal of course is to test a system as close to production as possible, including fees of course.

jrhite:
Howdy 👋 

From load-test team, feel free to remove IGNORE_FEES support. We will test with fees.

unnawut:
Closing as this is no longer needed and `:ignore_fees` will be removed.

mederic-p:
https://github.com/omisego/elixir-omg/pull/1299

PR #1299, opened 2020-02-07
Remove ignore_fees from config

mederic-p:
## Overview

This PR removes `ignore_fees` from the config, enforcing fees to be paid.
Note that we only remove the possibility to configure the environment to ignore fees, however some internal calls (e.g. Watcher) can still pass `:ignore_fees` when needed.

## Changes

- Remove `:ignore_fees` config
- Update tests that were skipping fees to use a base fee of `1`

## Testing

Run tests!


apps/omg_child_chain/test/omg_child_chain/integration/fee_server_test.exs
```diff
@@ -163,31 +161,7 @@ defmodule OMG.ChildChain.Integration.FeeServerTest do
...
-
-      assert refresh_fees() =~ "Updates have no effect"
-
-      assert {:ok, :ignore_fees} == FeeServer.transaction_fees()
-      assert server_alive?()
-    end
-  end
-
-  defp start_fee_server() do
+  defp start_fee_server do
```
InoMurko:
```suggestion
  defp start_fee_server() do
```

unnawut: What... how possible....

mederic-p: My bad rebase 🙇

unnawut: No no, I mean credo should've complained. Looks like it's complaining but not exiting with bad status. Fixing in a separate PR.

apps/omg/test/omg/state/core_test.exs
```diff
@@ -56,9 +65,9 @@ defmodule OMG.State.CoreTest do
       # make some utxos
       state =
         state
-        |> Core.exec(create_recovered([{1, 0, 0, alice}], @eth, [{alice, 6}, {bob, 2}, {alice, 2}]), :ignore_fees)
```
pdobacz: hm, but `:ignore_fees` as an option to `Core.exec` still must be supported, since there are clients of `State.Core` that don't check the fees being paid (the Watcher). I see you're leaving `core.ex` without modification, so I suppose the tests of it should be OK without change.

In other words - we're only removing `ignore_fees: true/false` setting, which removes the possibility to ignore _for the child chain server_ only, correct?

mederic-p: Correct, `Core.exec` still supports `ignore_fees` only the config has been removed. Let me check and add a test if needed with `ignore_fees` to the Core.

pnowosie:
I've added `(e.g. Watcher)` in the [above description](https://github.com/omisego/elixir-omg/pull/1299#issue-372234152)

apps/omg_child_chain/test/omg_child_chain/integration/fee_server_test.exs
mederic-p: https://github.com/omisego/elixir-omg/pull/1303 (for reference)

Issue #1215, opened 2019-12-17
As an API designer I would like to have flexibility in how things can be filtered, paginated (and sorted) when working on new endpoints to meet the needs of the different User Interfaces we are building.


jrhite:
**For API requests returning single entities**
See below: the same filtering info and thoughts applies, excepting pagination and sorting.

**For API requests returning lists**

For reference and context have a look at the api [`/transaction.all` issue](https://github.com/omisego/elixir-omg/issues/954) and its [PR](https://github.com/omisego/elixir-omg/pull/1178)

For more context also see the api:
 [`/block.all` issue](https://github.com/omisego/elixir-omg/issues/1168) and its [PR](https://github.com/omisego/elixir-omg/issues/1168)
 [`/deposit.all` issue](https://github.com/omisego/elixir-omg/issues/1133) and its [PR](https://github.com/omisego/elixir-omg/issues/1199)

The most basic constraints for an api return a list of results will be limit and offset:
```
      {"limit", [:pos_integer, :optional]},
      {"page", [:pos_integer, :optional]}
```

[apps/omg_watcher_rpc/lib/web/validators/block_constraints.ex](https://github.com/omisego/elixir-omg/blob/f01a5e8a1e7644bcf97927e4a2e355ea7b096767/apps/omg_watcher_rpc/lib/web/validators/block_constraints.ex) is a new constraints file specific to blocks broken off from: [apps/omg_watcher_rpc/lib/web/validators/constraints.ex](https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher_rpc/lib/web/validators/constraints.ex). `constraints.ex` was kind of a free-for-all allowing filtering on pretty much any property even if the entity being fetched had no such a property. `block_constraints.ex` breaks this out into constraints that make sense for blocks.

Different entities will have different constraints depending on the entity. Eg: `/transaction.all` has its own constraints file with constraints like `address` and `blknum`. The `/deposit.all` endpoint is routed to `OMG.WatcherRPC.Web.Controller.Utxo.get_deposits()` and has its own constraints as well including `address` accepted by the request.

[...]

PR #1302, opened 2020-02-07
fix: mandatory limit

InoMurko:
## Overview

Swagger cannot generate optional parameters request body

## Changes
- Mandatory limit on get all transactions (it's dangerous to leave it without, we should also limit the amount!)

## Testing
@boolafish


pnowosie:
Can you help me understand this :pr: a little better.
If I got it correctly API client generators we use don't handle optional parameters well?
Therefore to fix it, we're faking our documentation (API specs) - is it what this PR delivers?

InoMurko:
Not really. This PR requires you to send a linit param in get all transactions because it’s a security issue to nit limit the response.

jrhite:
☝️ absolutely. the backend should always enforce such things. client can ask for what it wants, but server validates/adjusts the constraints.

A minor thing, but IMHO, I hope we can improve our pagination handling, there is another for ticket for that: https://github.com/omisego/elixir-omg/issues/1215)

But specific to this ticket, I think it's not a good API user experience. The server validates and adjusts the limit. But there is no indication of this to the user. This can make them think there is a bug in our code or they did something wrong etc...

Returning a response with error message explaining that their pagination limit request exceeds the max limit of #### would be more useful and informative and less confusing to the API user.

InoMurko:
I agree with @jrhite. 
@unnawut I think the server should set a cap on the limit param

Anyhow, this is mostly getting aroung swagger bugs that can’t handle all optional params in requesys

PR #1305, opened 2020-02-10
feat: hex and rebar version pinning

unnawut:
## Overview

Hex the package manager had been unstable lately with [v0.20.2 - v0.20.5 released within a little over a day](https://github.com/hexpm/hex/releases). The command `mix do local.hex --force, local.rebar --force` always fetch the latest, potentially buggy version, which caused many of our builds to fail.

Version & hash pinning also helps tightening the security in the event of package managers getting compromised.

## Changes

- Moved version pinning from `circleci/config.yml` into the main Makefile, so devs can also run `make install-hex-rebar` to get the same versions as on CI environments.

## Testing

- `make install-hex-rebar` should pass
- After the install command above, these commands should produce:
    ```
    $ ls ~/.mix/archives/hex* && \
      ~/.mix/rebar --version && \
      ~/.mix/rebar3 --version

    hex-0.20.5
    rebar 2.6.2 R15B03 20160628_123811 git 2.6.2
    rebar 3.6.1 on Erlang/OTP 21 Erts 10.0
    ```
  If you are using asdf, replace `~/.mix/...` above with the path to your .asdf mix folder e.g.
    ```
    $ ls ~/.asdf/installs/elixir/1.8.2/.mix/archives/hex* && \
      ~/.asdf/installs/elixir/1.8.2/.mix/rebar --version && \
      ~/.asdf/installs/elixir/1.8.2/.mix/rebar3 --version

    hex-0.20.5
    rebar 2.6.2 R15B03 20160628_123811 git 2.6.2
    rebar 3.6.1 on Erlang/OTP 21 Erts 10.0
    ```

- CI build should pass: https://app.circleci.com/github/omisego/elixir-omg/pipelines/9abaef0a-fc3d-45c5-8936-f4437b0f3622/workflows/ea16b18e-bd13-4ce0-b301-fc6499f38fa6


unnawut:
Btw. this PR is also needed as you see newer hex versions may introduce new metadata to mix.lock file so we need a way to install the same hex version on local dev machines.

InoMurko:
Good job hunting this down. It was causing quite a pain.

pnowosie:
> so devs can also run mix install-hex-rebar to get...

Did you mean `make install-hex-rebar` :question:

unnawut:
> > so devs can also run mix install-hex-rebar to get...
> 
> Did you mean `make install-hex-rebar` ❓

Yes thank you! Too many similar names🤦‍♂

PR #1311, opened 2020-02-11
feat: Increase block withholding threshold

jbunce:
## Overview

Increases the block withholding threshold to 15 hours in the Watcher.


jbunce:
@pdobacz correct, current assumptions.

Issue #1293, opened 2020-02-04
Watcher_Info autogenerated client `get_balance` varrying response


pnowosie:
Experienced in Cabbage tests (fee_claiming) 
Client `get_balance` response is either `map` or `list(map)`.

```
Balance response for Alice: %{
  "amount" => 3000000000000000000,
  "currency" => "0x0000000000000000000000000000000000000000"
}

11:54:04.164 [info]  		Then "Alice" should have "3" ETH on the child chain

11:54:04.188 [info]  		When "Alice" sends "Bob" "2" ETH on the child chain
Balance response for Bob: %{
  "amount" => 2000000000000000000,
  "currency" => "0x0000000000000000000000000000000000000000"
}

11:54:12.229 [info]  		Then "Bob" should have "2" ETH on the child chain
Balance response for FeeClaimer: [
  %{
    "amount" => 10000000000000000000,
    "currency" => "0x0000000000000000000000000000000000000000"
  }
]
```

InoMurko:
hm... is the issue on the cabbage tests or watcher info response?

pnowosie:
This is cabbage test client - Watcher responds with array tirelessly.
To reproduce see the usage of `Client.get_balance(...)` in `<elixir-omg>/priv/apps/itest/test/itest/fee_claiming_test.exs`

PR #1312, opened 2020-02-11
Watcher discovers fraudulent claim and publishes byzantine event

pnowosie:
Closes omisego/feat-fee/issues/29
Resolves omisego/feat-fee/issues/21

## Overview

After #1244 - operator claim fees with each block. Watcher needs to protect against claim frauds.

## Changes

Because [fee txs are __executed__ on state](https://github.com/omisego/feat-fee/issues/28#issuecomment-567440923) Watcher should already be warned by incorrect state when too much / not collected fees are claimed. We just need a test to demonstrate this.

## Testing
Added `test operator claim more fees than collected (OMG.Watcher.Integration.BlockGetterTest)`

### DIFF
--- a/apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
+++ b/apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
@@ -217,7 +217,7 @@ defmodule OMG.Watcher.Integration.BlockGetterTest do
   end
 
   @tag fixtures: [:in_beam_watcher, :mix_based_child_chain, :test_server, :stable_alice, :stable_alice_deposits]
-  test "operator claim more fees than collected", %{
+  test "operator claimed fees incorrectly (too much | little amount, not collected token)", %{
     stable_alice: alice,
     test_server: context,
     stable_alice_deposits: {deposit_blknum, _}
--- a/apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
+++ b/apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
@@ -215,4 +215,34 @@ defmodule OMG.Watcher.Integration.BlockGetterTest do
     # we should still be able to challenge this "unchallenged exit" - just smoke testing the endpoint, details elsewhere
     WatcherHelper.get_exit_challenge(exit_blknum, 0, 0)
   end
+
+  @tag fixtures: [:in_beam_watcher, :mix_based_child_chain, :test_server, :stable_alice, :stable_alice_deposits]
+  test "operator claim more fees than collected", %{
+    stable_alice: alice,
+    test_server: context,
+    stable_alice_deposits: {deposit_blknum, _}
+  } do
+    fee_claimer = OMG.Configuration.fee_claimer_address()
+
+    # preparing transactions for a fake block that overclaim fees
+    txs = [
+      OMG.TestHelper.create_recovered([{deposit_blknum, 0, 0, alice}], @eth, [{alice, 9}]),
+      OMG.TestHelper.create_recovered([{1000, 0, 0, alice}], @eth, [{alice, 8}]),
+      OMG.TestHelper.create_recovered_fee_tx(1000, fee_claimer, @eth, 3)
+    ]
+
+    block_overclaiming_fees = OMG.Block.hashed_txs_at(txs, 1000)
+
+    # from now on the child chain server is broken until end of test
+    BadChildChainServer.prepare_route_to_inject_bad_block(
+      context,
+      block_overclaiming_fees
+    )
+
+    # checking if both machines and humans learn about the byzantine condition
+    assert WatcherHelper.capture_log(fn ->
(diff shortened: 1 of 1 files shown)


apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
```diff
@@ -215,4 +215,34 @@ defmodule OMG.Watcher.Integration.BlockGetterTest do
     # we should still be able to challenge this "unchallenged exit" - just smoke testing the endpoint, details elsewhere
     WatcherHelper.get_exit_challenge(exit_blknum, 0, 0)
   end
+
+  @tag fixtures: [:in_beam_watcher, :mix_based_child_chain, :test_server, :stable_alice, :stable_alice_deposits]
+  test "operator claim more fees than collected", %{
```
InoMurko: let's talk about this test - @achiurizo might want to see this in cabbage, instead of 
```
:in_beam_watcher, :mix_based_child_chain
```

pdobacz: my 2 cents on this:

Does cabbage support invalidly behaving child chain servers in current `master`? I assume it doesn't yet, since [this test](https://github.com/omisego/elixir-omg/pull/1312/files#diff-cd2d99842e0db23a5cf13c37248c98a3R148) is still here.

When it will, moving all those tests over to cabbage can be done in bulk. Since this PR isn't adding the first test like this, I think it's acceptable and pragmatic to add such a test now and refactor later.

apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
```diff
@@ -215,4 +215,34 @@ defmodule OMG.Watcher.Integration.BlockGetterTest do
     # we should still be able to challenge this "unchallenged exit" - just smoke testing the endpoint, details elsewhere
     WatcherHelper.get_exit_challenge(exit_blknum, 0, 0)
   end
+
+  @tag fixtures: [:in_beam_watcher, :mix_based_child_chain, :test_server, :stable_alice, :stable_alice_deposits]
```
pdobacz: comment about the next line actually, putting here to not pollute thread :point_down: 

> test "operator claim more fees than collected", %{

(also about the comment saying)

>  that overclaim fees

Name and comment are accurate, but they do not convey that this holds for any problem with the fee txs. I think we can say that this intends to test operator's fee claiming being incorrect in general.

But not a big deal anyway.

apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
InoMurko: Cabbage can support whatever is needed. If it’s not built, it needs to be built.

No one is specifically assigned to port all the tests (and, by the amount of tests, this is truly a horrible task to give someone).
So to me, the pragmatic approach is: don’t add anymore tests that need to be rewritten or refactored, start doing the changes now.

pnowosie: My 2 cents, from the prev experience with cabbage, it’s great, but very immature, lets just recall #1293 or it does not provide tests isolation. We will overcome all these issues at some point, however porting this case now it’s way bigger than the task itself.

Anyway, it’s about priorities. 
CC @T-Dnzt

achiurizo: I'll return with 6 cents:

> Does cabbage support invalidly behaving child chain servers in current master? I assume it doesn't yet, since this test is still here.
> Since this PR isn't adding the first test like this, I think it's acceptable and pragmatic to add such a test now and refactor later.
> We will overcome all these issues at some point, however porting this case now it’s way bigger than the task itself. Anyway, it’s about priorities.

The points raised here make sense and I can see that adopting a new framework or technology could make this transition difficult for some. The team and myself are happy to assist you all to help us all _collectively_ move to a better testing setup that's going to benefit all the engineers and contributors of this codebase.

In order for us to move forward to make it better for everyone, we all need to contribute. If we casually push aside these improvements because "it wasn't built" or there was "prior art so we can add on it" or that its "not priorities", these improvements will fall short.
> So to me, the pragmatic approach is: don’t add anymore tests that need to be rewritten or refactored, start doing the changes now.

I think this is quite a fair stance. I think the history of the code base can show that the "punt now, refactor later" approach has not done well(recall docker/mix issues and documents being out of date, difficulty getting distillery introduced, testing, etc ...) and to continue that behavior here is only going to add to that same pile.

---

As mentioned before, happy to offer time to pair to move this forward with cabbage. I think it's safe to say we all want to improve the code base and that it should be our collective priority to improve this now, and enjoy the benefits later.

pnowosie: Note: We agreed to postpone cabbage test due to tight timeframes

PR #1316, opened 2020-02-13
feat: add more config data to /configuration.get endpoint

jrhite:
## Overview

Add more config data to the `/configuration.get` API endpoints for CC, Watcher and Watcher Info, to make it easier to get a full picture of a given server's configuration from a single source.

Putting this PR out as a quick draft to get a feel for people's thoughts on this. The initial commit exposes the `min_exit_period_seconds` config variable.

If the idea seems useful, I'd also like to add the following config data:
```
child_chain_url: <child_chain_url>
watcher_url: <watcher_url>
watcher_info_url: <watcher_info_url>
blockexplorer_url: <blockexplorer_url>
contract_transactions_url: <contract_transactions_url>
ethereum_network: <ethereum_network>
authority_address: <authority_address>
eth_vault: <eth_vault>
erc20_vault: <erc20_vault>
payment_exit_game: <payment_exit_game>
plasma_framework_tx_hash: <plasma_framework_tx_hash>
plasma_framework: <plasma_framework>
```
## Changes

Only added and exposed new configuration variable `min_exit_period_seconds` to `OMG.Eth.Configuration`, for the first commit to test the waters.

For simplicity I would like to add all the above variables to all 3 CC, Watcher and Watcher Info, but it's an open question. Should the config data be more tightly scoped to the service and the config variables the service actually uses? Eg: The CC's `/configuration.get` doesn't need to have the `child_chain_url: <child_chain_url>` key/value returned in its config, but it doesn't hurt. CC/Watcher/Watcher Info also don't use `blockexplorer_url: <blockexplorer_url>`, but still might be useful to have.

Regarding contract addresses, tx hashes and authority addresses, is there any config data that we would not want included?

## Testing
`curl http:/localhost/configuration.get` and verify that `min_exit_period_seconds` is in the output with a value of 120.

### DIFF
--- a/apps/omg_child_chain/lib/omg_child_chain/api/configuration.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/api/configuration.ex
@@ -23,8 +23,10 @@ defmodule OMG.ChildChain.API.Configuration do
   def get_configuration() do
     configuration = %{
       deposit_finality_margin: Configuration.deposit_finality_margin(),
+
       contract_semver: OMG.Eth.Configuration.contract_semver(),
-      network: OMG.Eth.Configuration.network()
+      min_exit_period_seconds: OMG.Eth.Configuration.min_exit_period_seconds(),
+      network: OMG.Eth.Configuration.network(),
     }
 
     {:ok, configuration}
--- a/apps/omg_eth/lib/omg_eth/configuration.ex
+++ b/apps/omg_eth/lib/omg_eth/configuration.ex
@@ -14,14 +14,19 @@
 
 defmodule OMG.Eth.Configuration do
   @moduledoc """
-  Provides access to applications configuration
+  Provides access to plasma-contracts configuration
   """
   @app :omg_eth
+
   def contract_semver() do
     Application.get_env(@app, :contract_semver)
   end
 
   def network() do
     Application.get_env(@app, :network)
+  end
+
+  def min_exit_period_seconds() do
+    Application.get_env(@app, :min_exit_period_seconds)
   end
 end
--- a/apps/omg_watcher/lib/omg_watcher/api/configuration.ex
+++ b/apps/omg_watcher/lib/omg_watcher/api/configuration.ex
@@ -23,11 +23,15 @@ defmodule OMG.Watcher.API.Configuration do
   def get_configuration() do
     configuration = %{
       deposit_finality_margin: Configuration.deposit_finality_margin(),
+
       contract_semver: OMG.Eth.Configuration.contract_semver(),
+      min_exit_period_seconds: OMG.Eth.Configuration.min_exit_period_seconds(),
       network: OMG.Eth.Configuration.network(),
+
       exit_processor_sla_margin: OMG.Watcher.Configuration.exit_processor_sla_margin()
     }
 
     {:ok, configuration}
   end
 end
+


InoMurko:
maybe it would be cool to add this to childchain configuration.get as well.

jrhite:
> maybe it would be cool to add this to childchain configuration.get as well.

yup it's there. added to cc, watcher and watcher info. cc for example:
https://github.com/omisego/elixir-omg/pull/1316/files#diff-3b2645b7f40c1e05232c36a778073f6f

InoMurko:
ah overlooked that. ok cool! needs fixing tests tho!

unnawut:
> Should the config data be more tightly scoped to the service and the config variables the service actually uses?

Yes please! Each service's binary may not have all the config variables. It also helps with debugging to not see the configs that are not used.

---

> eth_vault: <eth_vault>
> erc20_vault: <erc20_vault>
> payment_exit_game: <payment_exit_game>
> plasma_framework_tx_hash: <plasma_framework_tx_hash>
> plasma_framework: <plasma_framework>

These ones are already in `/status.get`. I'd like to keep it in `/status.get`, because they're more of network information, and have `/configuration.get` for app internal behaviour.

---

> ethereum_network: <ethereum_network>
> authority_address: <authority_address>

I'm supportive of having them, but in `/status.get` where the rest of network information resides.

---

> contract_transactions_url: <contract_transactions_url>

Don't think so. It's easily derivable from `ethereum_network` and `plasma_framework` combo.

---

> Should the config data be more tightly scoped to the service and the config variables the service actually uses? Eg: The CC's /configuration.get doesn't need to have the child_chain_url: <child_chain_url> key/value returned in its config, but it doesn't hurt. CC/Watcher/Watcher Info also don't use blockexplorer_url: <blockexplorer_url>, but still might be useful to have.
>
> child_chain_url: <child_chain_url>
> watcher_url: <watcher_url>
> watcher_info_url: <watcher_info_url>
> blockexplorer_url: <blockexplorer_url>

The childchain doesn't really know its own `child_chain_url`. A service could also be serving multiple domains. So I'd prefer not having them for the same services, but exposing watcher's `child_chain_url` would be useful, etc. etc. `blockexplorer_url` is a no-go for me as it introduces unnecessary reverse dependency.

---

Other configs that might be useful to expose:

All services:
- `Application.get_env(:omg, :ethereum_events_check_interval_ms)`
- `Application.get_env(:omg, :deposit_finality_margin)`
- `Application.get_env(:omg_eth, :min_exit_period_seconds)`
- `Application.get_env(:omg_eth, :ethereum_block_time_seconds)`
- `Application.get_env(:omg_eth, :ethereum_events_check_interval_ms)`
- `Application.get_env(:omg_eth, :ethereum_stalled_sync_threshold_ms)`

ChildChain
- `Application.get_env(:omg_child_chain, :submission_finality_margin)`
- `Application.get_env(:omg_child_chain, :block_queue_eth_height_check_interval_ms)`
- `Application.get_env(:omg_child_chain, :fee_file_check_interval_ms)`
- `Application.get_env(:omg_child_chain, :block_submit_every_nth)`
- `Application.get_env(:omg_child_chain, :submission_finality_margin)`

Watcher & WatcherInfo
- `Application.get_env(:omg_watcher, :exit_processor_sla_margin)`
- `Application.get_env(:omg_watcher, :exit_processor_sla_margin_forced)`
- `Application.get_env(:omg_watcher, :maximum_block_withholding_time_ms)`
- `Application.get_env(:omg_watcher, :block_getter_loops_interval_ms)`
- `Application.get_env(:omg_watcher, :maximum_number_of_unapplied_blocks)`
- `Application.get_env(:omg_watcher, :exit_finality_margin)`
- `Application.get_env(:omg_watcher, :block_getter_reorg_margin)`

unnawut:
In any case, additional config exposure will be super helpful 🚀🚀🚀🚀🚀

jrhite:
Great list. Will do. And additionally fees config info would be useful.

> ChildChain
> 
> * `Application.get_env(:omg_child_chain, :submission_finality_margin)`
> * `Application.get_env(:omg_child_chain, :block_queue_eth_height_check_interval_ms)`
> * `Application.get_env(:omg_child_chain, :fee_file_check_interval_ms)`
> * `Application.get_env(:omg_child_chain, :block_submit_every_nth)`
> * `Application.get_env(:omg_child_chain, :submission_finality_margin)`
> 

I would like to show what the fees are per-operator (only one operator --us-- for now, but useful to show for our different envs as we can now config fees per environment).

Some slimmed down version of the `./apps/omg_child_chain/priv/fee_specs.json` file.
```
{
    "1": [{ 
        "token": "0x0000000000000000000000000000000000000000",
        "amount": 1,
        "subunit_to_unit": 1000000000000000000,
        "pegged_amount": null,
        "pegged_currency": null,
        "pegged_subunit_to_unit": null,
        "updated_at": "2019-01-01T10:10:00+00:00"
    }]
}
```

Ah, just noticed differences the spec is different in helm: `deployments/childchain/templates/configmap.yaml` 🤔 

No matter, the source of truth will come from the actual running childchain.

InoMurko:
Don’t overbuild this. Add code per requirements.

InoMurko:
That is: product or infra requirements, not feels good.

unnawut:
> Don’t overbuild this. Add code per requirements.

True that. In that sense I guess we could start with only the values that have been discussed a lot lately, or vary across deployments.

From my side that would be `:ethereum_events_check_interval_ms` and `:ethereum_stalled_sync_threshold_ms`

unnawut:
Or actually, maybe not even `:ethereum_stalled_sync_threshold_ms` since so far it's only concerned when the alarm is raised. It could easily go as alarm metadata.

jrhite:
I like @unnawut's list and the break down by service.

I also agree with you @InoMurko that this should not go too far. A sensible limit to me is that adding any more logic than what can be simple retrieved via `Application.get_env(app_name, var_name)` would be reaching too far to get config data.

This is not a product thing nor an infra thing (although, could very well be useful to infra). lt's an eng thing/feature. Given the right priority (this is low - non-working hours type of thing), we get to do helpful eng stuff too, right? 

I think it's useful -- or worded differently, I tried to find the info at this endpoint in one place, but it was not there, so it would have been useful.

jrhite:
@unnawut @InoMurko 

`ethereum_events_check_interval_ms` is defined in both `apps/omg/config/config.exs` and `apps/omg_eth/config/config.exs`. 

Any reason we would want them to have different values? If not, we could remove the duplicated config variable.

PR #1321, opened 2020-02-14
feat: make Watchers refuse to boot when sla_margin is unsafe

### LINKED ISSUE #1275: Default values for exit_processor_sla_margin can cause the watcher to not raise unchallenged exits
### Description 
The current audit environment `73f526d` has become byzantine. Min exit period is set to 30 minutes and `exit_processor_sla_margin` was set to its default value. The watcher flagged invalid exits but never stopped forwarding tx to the chch even after a period of more than 24h. 

### Suggested Fix 
- There should be runtime fail safe checks that prevents running a watcher with values for `exit_processor_sla_margin` and finalization period that are not within ranges that make sense. 
- Check all other config parameters for watcher and child chain that have interdependencies and  implement  fail safe checks.

pdobacz:
Closes #1275

## Overview

Since there are combinations of our configuration that may lead to a dangerous setup of the Watcher (in particular - muting the reports of `unchallenged_exits`), we're enforcing that the `exit_processor_sla_margin` be not larger than `min_exit_period`.

This check can be overridden when one know's what one's doing (tests, dev environments, where the `min_exit_period` is very short).

## Changes

- introduce the checking and blocking facility in `ExitProcessor` boot up
- introduce `exit_processor_sla_margin_force` to provide the forceful override
- introduce `ethereum_block_time_seconds` to correlate the the settings. Note on this: The ugly part is that `min_exit_period` is measured in seconds and `sla_margin` is in Ethereum blocks.
    - probably don't want to consider `min_exit_period` in blocks, seconds makes much more sense there
    - likewise don't want to consider `sla_margin` in seconds. This would require us to fetch timestamps of events, requiring many RPC calls, as they aren't included in `getLogs` response
    - so we'll be requiring the deployer to specify what the average Ethereum block time is :(, just for this check
- docs (how I didn't miss any)
- docker composes

The new configuration values are explained to most extent in [docs/details.md](https://github.com/omisego/elixir-omg/compare/pdobacz/1275_fail_safe_sla_margin?expand=1#diff-a3529659d2141cf6f12ae1b4487c5a8d), but other docs were also updated

TODO

- [ ] probably some helm charts to be able to deploy this

## Testing

The usual. Also you can try `docker-compose up watcher` and see it works unless you remove the `EXIT_PROCESSOR_SLA_MARGIN_FORCE=TRUE`, it will then stop with `error: :sla_margin_too_big`.


unnawut:
Havn't forgotten this one. Will take a look in my upcoming working hours.

apps/omg_watcher/config/config.exs
```diff
@@ -5,6 +5,8 @@ config :omg_watcher, child_chain_url: "http://localhost:9656"
 config :omg_watcher,
   # 23 hours worth of blocks - this is how long the child chain server has to block spends from exiting utxos
   exit_processor_sla_margin: 23 * 60 * 4,
+  # this means we don't want the `sla_margin` above be larger than the `min_exit_period`
+  exit_processor_sla_margin_force: false,
```
unnawut: I think `force` -> `forced` is more natural + stops me from chuckling the StarWars pun :trollface:

apps/omg_watcher/lib/omg_watcher/exit_processor/core.ex
```diff
@@ -121,6 +121,26 @@ defmodule OMG.Watcher.ExitProcessor.Core do
      }}
   end
 
+  @doc """
+  Use to check if the settings regarding the `:exit_processor_sla_margin` config of `:omg_watcher` are OK.
```
unnawut: I like your PR description and don't want it to fade away with the closed PR. Can we add it here?

```elixir
@doc """
Check if the settings regarding the `:exit_processor_sla_margin` config of `:omg_watcher` are OK.

Since there are combinations of our configuration that may lead to a dangerous setup of the Watcher
(in particular - muting the reports of unchallenged_exits), we're enforcing that the exit_processor_sla_margin
be not larger than min_exit_period.

This check can be overridden (that is, allowing sla_margin to be larger than min_exit_period) when one know's what one's doing 
(e.g. in tests, dev environments, where the min_exit_period is very short), by passing `true` to `sla_margin_force`.
"""

```

pdobacz: :+1: taking the liberty to distribute the description to this here (the "Since..." part) and `details.md` other docs (the "can be overriden" part)

PR #1323, opened 2020-02-17
separate cabbage and perf for different tests

boolafish:
:clipboard: Slack discussion: [here](https://omisego.slack.com/archives/CS8STN2R0/p1581671495038500)

## Overview

We want to get mono-repo benefits while keeping integration tests and load tests isolated. Moving `priv` to `priv/cabbage` and add new umbrella app `priv/perf`.

## Changes

- Move `priv` to `priv/cabbage` and fix paths in Makefile, tests, circle ci configs to make it run!
- Create new umbrella app `perf` and add an app under it `load_test`.
- Add a dummy chaperon tests under `load_test` app.

## Testing
- testing `priv/perf`
```
cd priv/perf
mix deps.get
mix test
```


.formatter.exs
```diff
@@ -5,11 +5,10 @@
     "mix.exs",
     "apps/*/mix.exs",
     "apps/*/{lib,test,config}/**/*.{ex,exs}",
-    "priv/config/config.exs",
-    "priv/mix.exs",
-    "priv/features/contexts/*.exs",
```
boolafish: I don't think this `features` folder is still valid, can you confirm I am not missing anything when removing this? @InoMurko

InoMurko: you're right!

InoMurko: you're correct!

PR #1329, opened 2020-02-19
bump contracts and geth

InoMurko:
Latest contracts, latest geth! Onwards and upwards!

### DIFF
--- a/apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex
@@ -259,7 +259,8 @@ defmodule OMG.ChildChain.BlockQueue.Core do
     :ok
   end
 
-  # maybe this will get deprecated soon once the network migrates to 1.9.11. Look at the previous function header for commit reference.
+  # maybe this will get deprecated soon once the network migrates to 1.9.11.
+  # Look at the previous function header for commit reference.
   # `fmt.Errorf("known transaction: %x", hash)`  has been removed
   def process_submit_result(
         submission,
--- a/apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex
@@ -259,6 +259,17 @@ defmodule OMG.ChildChain.BlockQueue.Core do
     :ok
   end
 
+  # maybe this will get deprecated soon once the network migrated to 1.9.11. look at the previos function header
+  # `fmt.Errorf("known transaction: %x", hash)`  has been removed
+  def process_submit_result(
+        submission,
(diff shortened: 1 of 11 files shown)


apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex
```diff
@@ -249,6 +249,18 @@ defmodule OMG.ChildChain.BlockQueue.Core do
...
+  def process_submit_result(
+        submission,
+        {:error, %{"code" => -32_000, "message" => "already known"}},
+        _newest_mined_blknum
+      ) do
+    log_known_tx(submission)
+    :ok
+  end
+
+  # maybe this will get deprecated soon once the network migrated to 1.9.11. look at the previos function header
```
InoMurko:
```suggestion
  # maybe this will get deprecated soon once the network migrates to 1.9.11. Llook at the previos function header for commit reference.
```

apps/omg_eth/test/fixtures.exs
```diff
@@ -56,12 +56,12 @@ defmodule OMG.Eth.Fixtures do
     {:ok, true} =
       Ethereumex.HttpClient.request("personal_unlockAccount", ["0x6de4b3b9c28e9c3e84c2b2d3a875c947a84de68d", "", 0], [])
 
-    add_exit_queue =
-      RootChainHelper.add_exit_queue(1, @eth, %{
-        plasma_framework: Encoding.from_hex("0xc673e4ffcb8464faff908a6804fe0e635af0ea2f")
-      })
+    if RootChainHelper.has_exit_queue(1, @eth, contract.contract_addr) do
```
unnawut: Not sure if I read this right... If RootChainHelper has an exit queue, add another one?

apps/omg_eth/test/fixtures.exs
```diff
@@ -56,12 +56,16 @@ defmodule OMG.Eth.Fixtures do
...
-      RootChainHelper.add_exit_queue(1, @eth, %{
-        plasma_framework: Encoding.from_hex("0xc673e4ffcb8464faff908a6804fe0e635af0ea2f")
-      })
+    case RootChainHelper.has_exit_queue(1, @eth, contract.contract_addr) do
+      {:ok, false} ->
+        add_exit_queue = RootChainHelper.add_exit_queue(1, @eth, contract.contract_addr)
+
+        {:ok, _} = Support.DevHelper.transact_sync!(add_exit_queue)
+
+      _ ->
```
pdobacz: 1. pattern match on `{:ok, true}` for an early fail if the request borks
2. it is a bit strange that we didn't require the `case` earlier and now we do. I'm guessing there's a reason why sometimes the queue is there and sometimes it isn't? Though it is strange, since we're working off of the same snapshot every time.

PR #1341, opened 2020-02-24
validate your blocks

InoMurko:
docker-compose -f docker-compose-watcher.yml up


docker-compose-watcher.yml
```diff
@@ -17,24 +17,29 @@ services:
 
   watcher:
     #last stable staging and production watcher with ALD, pre-lumphini
-    image: omisego/watcher:70e9a68
+    image: omisego/watcher:a43605e
```
jbunce: `78f1a7f`

PR #1342, opened 2020-02-24
Merge v0.4 to master

unnawut:
Keep the code between versions from diverging for as long as possible


InoMurko:
good job

kasima:
Thanks, I want to keep us in a single history for as long as we can. Will work on new release management procedures next.

PR #1343, opened 2020-02-24
feat: lint version and publish docker images on version tags

unnawut:
## Overview

Checks that the version tag matches `VERSION`, then publishes the corresponding stable docker images. For example, `v0.5.0` git tag becomes docker images `child_chain:0.5.0`, `watcher:0.5.0`, `watcher_info:0.5.0`.

On the other hand, these unstable docker images will still be published regardless of the version check: 

1. Tip of `master` -> tagged with `latest`, and also commit SHA e.g. `watcher_info:3fbd762`
2. Tip of version branches e.g. `v0.5` branch -> tagged with `-dev` suffix, e.g. `watcher_info:0.5-dev`

**This doesn't prevent the release tag from happening but makes the discrepancy visible as early as possible.**

## Changes

1. Add a `lint_version` CI job that checks for a matching git tag and `VERSION`. This check must pass before stable docker images are published. Applies to stable version tags, e.g. `watcher_info:0.5.0`.

    Unstable tags e.g. `watcher_info:latest`, `watcher_info:3fbd762`, `watcher_info:0.5-dev` are published without this check.

2. Update `.circleci/config.yml` to do `publish_child_chain`, `publish_watcher` and `publish_watcher_info` when a tag matches `/v[0-9]+\.[0-9]+\.[0-9]+/`

## Testing

When a tag e.g. `v0.5.0` is tagged on Github, docker images should be published to these DockerHub image repos:

- https://hub.docker.com/r/omisego/child_chain
- https://hub.docker.com/r/omisego/watcher
- https://hub.docker.com/r/omisego/watcher_info


jrhite:
Generally looks good. Just a question, if we're doing semvar versioning (a good thing!), this implies that at times CC, WS and WI versions may differ from time to time.

x.y.z -> means to me that

- CC: y.z can diverge from the y.z of WS and WI, but WS and WI's x must always be equal to the x of CC.
- WS: y.z can diverge from the y.z of WI, but WI's x must always be equal to the x of WS.
- WI: can diverge in anyway as long as x >= the x of WS and CC

These details might not be exactly correct, but that's the general idea. Can we also add this info to the (external) docs somewhere (and internal if necessary)?

unnawut:
> Generally looks good. Just a question, if we're doing semvar versioning (a good thing!), this implies that at times CC, WS and WI versions may differ from time to time.
> 
> x.y.z -> means to me that
> 
> * CC: y.z can diverge from the y.z of WS and WI, but WS and WI's x must always be equal to the x of CC.
> * WS: y.z can diverge from the y.z of WI, but WI's x must always be equal to the x of WS.
> * WI: can diverge in anyway as long as x >= the x of WS and CC
> 
> These details might not be exactly correct, but that's the general idea. Can we also add this info to the (external) docs somewhere (and internal if necessary)?

Looks like we're not really following semver so far and so my title with MAJOR.MINOR.PATCH is misleading. So I updated it for now. If you'd like feel free to start a new discussion on the ideal versioning approach!

jrhite:
> > Generally looks good. Just a question, if we're doing semvar versioning (a good thing!), this implies that at times CC, WS and WI versions may differ from time to time.
> > x.y.z -> means to me that
> > 
> > * CC: y.z can diverge from the y.z of WS and WI, but WS and WI's x must always be equal to the x of CC.
> > * WS: y.z can diverge from the y.z of WI, but WI's x must always be equal to the x of WS.
> > * WI: can diverge in anyway as long as x >= the x of WS and CC
> > 
> > These details might not be exactly correct, but that's the general idea. Can we also add this info to the (external) docs somewhere (and internal if necessary)?
> 
> Looks like we're not really following semver so far and so my title with MAJOR.MINOR.PATCH is misleading. So I updated it for now. If you'd like feel free to start a new discussion on the ideal versioning approach!

That's fine, but for both internal and external developers, if we're not doing true semvar yet, let's document that so that it's clear.

Then we can start a discussion on real semvar versioning approach, if we can do it (in theory we can, that's indeed what the semantics of semar are for), then i think we should do and follow real semvar. also makes me think about extending the semvar discussion to versioning of the contracts themselves. WS and CC should be able to follow semvar conventions with the contracts as well.

I don't have anything concrete yet, but the meaning and intention are clear. Let me open a new issue for discussion points to mature and not hijack this issue. 😉

PR #1345, opened 2020-02-24
feat: add release name, app env and hostname to datadog events

InoMurko:
Consolidate statix tags under omg_status.

## Overview

Set statix configuration in one place.
## Testing

`mix test`


apps/omg_status/lib/omg_status/release_tasks/set_tracer.ex
```diff
@@ -46,6 +49,13 @@ defmodule OMG.Status.ReleaseTasks.SetTracer do
     :ok = Application.put_env(:spandex_datadog, :sync_threshold, get_sync_threshold(), persistent: true)
   end
 
+  defp get_hostname() do
+    hostname = validate_hostname(get_env("HOSTNAME"))
+
+    _ = Logger.info("CONFIGURATION: App: #{@app} Key: HOSTNAME Value: #{inspect(hostname)}.")
```
achiurizo:
```suggestion
    Logger.info("CONFIGURATION: App: #{@app} Key: HOSTNAME Value: #{inspect(hostname)}.")
```

nitpick: isn't this also valid? Not sure of the convention but seems like the match is un-necessary.

apps/omg_status/test/omg_status/release_tasks/set_tracer_test.exs
```diff
@@ -31,6 +31,14 @@ defmodule OMG.Status.ReleaseTasks.SetTracerTest do
...
       :ok = Application.put_env(@app, Tracer, @configuration_old, persistent: true)
       :ok = Application.put_env(@app, :spandex_datadog, @configuration_old_spandex_datadog, persistent: true)
+      :ok = System.delete_env("HOSTNAME")
+      :ok = System.delete_env("DD_DISABLED")
+      :ok = System.delete_env("APP_ENV")
+      :ok = System.delete_env("DD_PORT")
+      :ok = System.delete_env("DD_HOSTNAME")
+      :ok = System.delete_env("DD_APM_PORT")
+      :ok = System.delete_env("BATCH_SIZE")
+      :ok = System.delete_env("SYNC_THRESHOLD")
```
jrhite: This seems off to me. Deleting several System env vars, unrelated to `omg_status` here. If we need to delete these env vars, it seems like each var would have a home where it would make sense to delete them.

apps/omg_utils/test/omg_utils/http_rpc/response_test.exs
```diff
@@ -23,7 +23,6 @@ defmodule OMG.Utils.HttpRPC.ResponseTest do
     txbytes: nil,
     txhash: nil,
     txindex: nil,
-    metadata: nil,
```
jrhite: this seems unrelated to this PR. is this intentional?

jrhite:
This should be helpful to a PR I'm working on! 🤞

apps/omg_status/lib/omg_status/release_tasks/set_tracer.ex
InoMurko: Depends. in Elixir and Erlang, every line returns a value. If you don't want to use the value you need to be explicit.

apps/omg_status/test/omg_status/release_tasks/set_tracer_test.exs
InoMurko: they're related to omg status because that's where datadog connection boots from.

apps/omg_utils/test/omg_utils/http_rpc/response_test.exs
InoMurko: yes, there are two metadata fields in the struct

PR #1349, opened 2020-02-25
remove timestamps from api layer

jrhite:
introducing `inserted_at` and `updated_at` fields into the api change breaks `omg-js`, so removing them for now.

yolo


jrhite:
> 3 more test fixes and we're good 🎉 🎉 🎉

Yeah, saw that, thanks. Was just letting circle do the work in this case to save time. 😒 That never works, does it.

Those tests should now pass. 🤞

jrhite:
no longer removing timestamps that was causing `omg-js` issues. `omg-js` has been relaxed to accept new keys in the PR https://github.com/omisego/omg-js/pull/252

PR #1351, opened 2020-02-25
publish BlockSubmitted event

InoMurko:
## Overview

When Watchers detect a block submitted event from Ethereum, publish it on omg bus so that consumers can do whatever they want with it.

## Changes

- find where it's detected and publish it

### DIFF
--- a/apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
@@ -14,11 +14,8 @@
 
 defmodule OMG.ChildChain.DatadogEvent.ContractEventConsumer do
   @moduledoc """
-  Subscribes to new events from EthereumEventListeners and pushes them to Datadog
+  Subscribes to new events from EthereumEventListeners and BlockGetter and pushes them to Datadog
   Integrated with the help of: https://docs.datadoghq.com/api/?lang=bash#post-an-event
-  Most things from the doc doesn't work. Either because Statix doesn't work or Datadog.
-  Date_happened, aggregation_key, source_type_name doesn't seem to appear in Events list.
-  Hence we transform everything into a tag.
   """
 
   require Logger
@@ -75,7 +72,10 @@ defmodule OMG.ChildChain.DatadogEvent.ContractEventConsumer do
     {:noreply, state}
   end
 
-  # Listens to events via OMG BUS and send them off
+  @doc """
+    Listens to events via OMG BUS and send them off
+    the assumption is all events are of the same type
+  """
   def handle_info({:internal_event_bus, :data, data}, state) do
     %{event_signature: event_signature} = hd(data)
     [event_name, _] = String.split(event_signature, "(")
--- a/apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
@@ -76,8 +76,8 @@ defmodule OMG.ChildChain.DatadogEvent.ContractEventConsumer do
   end
 
   # Listens to events via OMG BUS and send them off
-  def handle_info({:internal_event_bus, :data, [data]}, state) do
-    %{event_signature: event_signature} = data
+  def handle_info({:internal_event_bus, :data, data}, state) do
+    %{event_signature: event_signature} = hd(data)
     [event_name, _] = String.split(event_signature, "(")
     aggregation_key = :root_chain
     timestamp = DateTime.to_unix(DateTime.utc_now(), :millisecond)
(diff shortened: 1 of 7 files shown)


apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
```diff
@@ -76,8 +76,8 @@ defmodule OMG.ChildChain.DatadogEvent.ContractEventConsumer do
   end
 
   # Listens to events via OMG BUS and send them off
-  def handle_info({:internal_event_bus, :data, [data]}, state) do
-    %{event_signature: event_signature} = data
+  def handle_info({:internal_event_bus, :data, data}, state) do
+    %{event_signature: event_signature} = hd(data)
```
unnawut: Do we also need to publish 2nd, 3rd events and so on?

InoMurko: Not sure I understand what you mean... we could fetch events coming from a range of blocks all would be the same type. so one type of event per message on the omg bus.

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -350,4 +350,12 @@ defmodule OMG.Watcher.BlockGetter do
   defp check_in_to_coordinator(synced_height), do: RootChainCoordinator.check_in(synced_height, :block_getter)
 
   defp update_status(%Core{} = state), do: Status.update(Core.chain_ok(state))
+
+  defp publish_data([%{event_signature: event_signature} | _] = data) do
+    # String.split("BlockSubmitted(uint256)", "(")
```
achiurizo: Maybe worthwhile to move this out of the method and a more formal comment on this line? While I think I understood the meaning behind this it looks like commented code that could be deleted.

apps/omg_watcher/lib/omg_watcher/datadog_event/contract_event_consumer.ex
```diff
@@ -65,8 +65,9 @@ defmodule OMG.Watcher.DatadogEvent.ContractEventConsumer do
   end
 
   # Listens to events via OMG BUS and send them off
-  def handle_info({:internal_event_bus, :data, [data]}, state) do
-    %{event_signature: event_signature} = data
+  # the assumption is all events are of the same type
```
achiurizo: should we be using exdoc here?

apps/omg_watcher/test/omg_watcher/datadog_event/contract_event_consumer_test.exs
```diff
@@ -49,6 +49,14 @@ defmodule OMG.Watcher.DatadogEvent.ContractEventConsumerTest do
     assert_receive {:event, _, _}
   end
 
+  test "if a list of event messages are put on omg bus is consumed by the event consumer and published on the publisher interface" do
+    pid = :erlang.pid_to_list(self())
+    topic = "#{pid}"
+    data = [%{event_signature: "#{pid}(bytes32)"}, %{event_signature: "#{pid}(bytes32)"}]
+    OMG.Bus.direct_local_broadcast(topic, {:data, data})
```
achiurizo: nitpick: seems like `topic` and `pid` are the same values. Could we do:

```elixir
topic = self() |> :erlang.pid_to_list() |> to_string()
sig = "#{topic}(bytes32)"
data = [%{event_signature: sig}, %{event_signature: sig}]
OMG.Bus.direct_local_broadcast(topic, {:data, data})
```

apps/omg_child_chain/test/omg_child_chain/datadog_event/contract_event_consumer_test.exs
```diff
@@ -50,6 +50,15 @@ defmodule OMG.ChildChain.DatadogEvent.ContractEventConsumerTest do
     assert_receive {:event, _, _}
   end
 
+  test "if a list of event message are put on omg bus is consumed by the event consumer and published on the publisher interface" do
+    pid = :erlang.pid_to_list(self())
+    topic = "#{pid}"
+    data = [%{event_signature: "#{pid}(bytes32)"}, %{event_signature: "#{pid}(bytes32)"}]
+    OMG.Bus.direct_local_broadcast(topic, {:data, data})
```
achiurizo: copypasta comment from below regarding pid interpolation.

apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
pnowosie: I think this :point_up: comment should be placed in the code - otherwise is not clear why we just take the 1st of `data` and ignore the rest

pnowosie: Also I spotted the related comment in [similar handler in Watcher](https://github.com/omisego/elixir-omg/pull/1351/files#diff-89f1e052b66822627d1f4f29970083d0R68) - which I believe should apear here as well

apps/omg_watcher/lib/omg_watcher/datadog_event/contract_event_consumer.ex
pnowosie: Same on ch-ch's handler :point_up:

PR #1352, opened 2020-02-25
feat: accepts LOGGING_BACKEND=INK for json logging

InoMurko:
MULTILINE LOGS

## Overview

It's been a real pain to read Datadogs logs, especially stacktraces that are spread through multiple lines...

## Changes

- added Ink, seems to be the most used json backend logger
- this adds some pressure on logging. we should monitor metrics to see how much pressure it really adds.
- for local development, I've set `:console` as the only backend
- Ink logger backend is what'll be used after deploy
## Testing

`mix test`


apps/omg_status/test/omg_status/release_tasks/set_logger_test.exs
```diff
@@ -0,0 +1,61 @@
...
+  test "if environment variables (INK) get applied in the configuration" do
+    :ok = System.put_env("LOGGER_BACKEND", "INK")
+    :ok = SetLogger.init([])
+    configuration = Application.get_env(@app, :backends)
+    assert Enum.member?(configuration, Ink) == true
+  end
+
+  test "if environment variables (CONSOLE) get applied in the configuration" do
+    # env var to console and asserting that Ink gets removed
+    :ok = System.put_env("LOGGER_BACKEND", "conSole")
```
unnawut: :i_see_what_you_did_there:

apps/omg_status/lib/omg_status/release_tasks/set_logger.ex
```diff
@@ -0,0 +1,53 @@
...
+
+    backends = logger_backends |> Kernel.--([remove]) |> Enum.concat([logger_backend]) |> Enum.uniq()
+    :ok = Application.put_env(@app, :backends, backends, persistent: true)
+  end
+
+  defp get_logger_backend() do
+    logger =
+      "LOGGER_BACKEND"
+      |> get_env()
+      |> validate_string(Ink)
```
unnawut: Can you move `Ink` up into a module attr `@default_backend Ink`? Almost missed seeing it here.

apps/omg_status/lib/omg_status/release_tasks/set_logger.ex
```diff
@@ -0,0 +1,53 @@
...
+    _ = Logger.info("CONFIGURATION: App: #{@app} Key: LOGGER_BACKEND Value: #{inspect(logger)}.")
+    logger
+  end
+
+  defp get_env(key), do: System.get_env(key)
+
+  defp validate_string(nil, default), do: default
+  defp validate_string(value, default), do: do_validate_string(String.upcase(value), default)
+  defp do_validate_string("CONSOLE", _default), do: :console
+  defp do_validate_string("INK", default), do: default
```
unnawut:
```suggestion
  defp do_validate_string("INK", _default), do: Ink
```

So it doesn't rely on the default being `Ink` to be correct.

unnawut: If at some point we change to `@default_backend NotInk`, one needs to realize this line needs to change too, which I think is quite hard to spot. Otherwise `LOGGER_BACKEND=INK` would use `NotInk`.

And this same argument would also apply even if we didn't turn to module attr and still use `|> validate_string(Ink)`.

jbunce:
LGTM. Have you tested it with DD?

InoMurko:
no @jbunce . we don't read logging via an agent, so I would have to be within infrastructure or setup minicube I guess. Let's see how it turns out in development ;)

docs/deployment_configuration.md
```diff
@@ -18,7 +18,7 @@
 - "FEE_CLAIMER_ADDRESS" - [Child Chain only!] 20-bytes HEX-encoded string of Ethereum address of Fee Claimer. Set via ENV vars only for Child Chain. For the Watcher this values is not important and fallbacks to "DEAD000000000000000000000000000000000000".
 - "EXIT_PROCESSOR_SLA_MARGIN" - [Watcher only!] - number of Ethereum blocks since start of an invalid exit, before `unchallenged_exit` is reported to prompt to mass exit. Must be smaller than "MIN_EXIT_PERIOD_SECONDS", unless "EXIT_PROCESSOR_SLA_MARGIN_FORCED=TRUE"
 - "ETHEREUM_BLOCK_TIME_SECONDS" - should mirror Ethereum network's setting, defaults to 15 seconds
-
+- "LOGGER_BACKEND" - Ink or console. Ink will encode logs as json (useful for Datadog). Console will use the default elixir Logger backend. Default is Ink.
```
pdobacz: leave blank line, otherwise md misrenders.

InoMurko: uh uh! done!

Issue #1355, opened 2020-02-26
insert_all_chunked fails with high txn volumes


jrhite:
`OMG.WatcherInfo.DB.Repo.insert_all_chunked/3` fails when there is a high number of txns per block and the block is inserted into postgres. There error is:

`(Postgrex.QueryError) postgresql protocol can not handle 80010 parameters, the maximum is 65535`

Two new columns were added to each table, `inserted_at` and `updated_at`, and they are not taken into account when the chunk size is calculated. Relevant code is here:

https://github.com/omisego/elixir-omg/blob/v0.4.3/apps/omg_watcher_info/lib/omg_watcher_info/db/repo.ex#L35-L44

The inserted chunks then exceed postgres limits and insertion fails.

jrhite:
Just to link things up this regression was caused by https://github.com/omisego/elixir-omg/pull/1322

pnowosie:
Since PR #1356 is merged I'm closing the issue. Reopen if needed.

PR #1356, opened 2020-02-26
fix insert_all_chunked chunk_size calculation

jrhite:
## Overview
This PR introduces fixes the broken `OMG.WatcherInfo.DB.Repo.insert_all_chunked/3` functionality described in the issue https://github.com/omisego/elixir-omg/issues/1355.

`chunk_size` was being calculated before the entry to be inserted in the database had all of its fields added causing `chunk_size` to be calculated incorrectly. To fix this, the calculation of `chunk_size` is done after the new `inserted_at` and `updated_at` fields are added.

## Changes
A minor refactoring to the `OMG.WatcherInfo.DB.Repo.insert_all_chunked/3` function was done. The code 
```
chunk_size = @max_params_count |> div(entries |> hd |> fields_count)
```
has been extracted into a function of its own for testability and readability purposes:
```
def chunk_size(entry), do: div(@max_params_count, fields_count(entry))
```

__Note__: `chunk_size/1` should be a private function, but I think it is an important function to test. Suggestions welcome on how to make it a private function, yet be easily testable. 😬 

## Testing
2 tests were added
- test that `insert_all_chunked/3` chunking works when the postgres parameter maximum is exceeded
- test that `chunk_size/1` returns the expected chunk size


jrhite:
I thought about this PR a bit more after a some sleep.

This PR intends to fix the regression caused by https://github.com/omisego/elixir-omg/pull/1322 and bring us back to the functional parity of `OMG.WatcherInfo.DB.Repo.insert_all_chunked/3` before the new columns broke it. That is the single goal of this PR.

However even before the regression, there was and is still a bottleneck. If we receive transactions at a rate where we continuously need to be chunking, we will eventually run out of memory. This can be solved, but would take this PR to the next level, and far out of this PR's scope.

It's likely premature to fix this bottleneck at this early stage, but I was considering creating an issue for it so we remember this problem exists. In the code I'd add a comment linking to the issue.

Thoughts?

unnawut:
> This can be solved, but would take this PR to the next level, and far out of this PR's scope.
>
> It's likely premature to fix this bottleneck at this early stage, but I was considering creating an issue for it so we remember this problem exists. In the code I'd add a comment linking to the issue.

Sounds good. I would go for a separate PR.

apps/omg_watcher_info/lib/omg_watcher_info/db/repo.ex
```diff
@@ -18,6 +18,7 @@ defmodule OMG.WatcherInfo.DB.Repo do
     adapter: Ecto.Adapters.Postgres
 
   @max_params_count 0xFFFF
+  def max_params_count(), do: @max_params_count
```
unnawut: This function is not used. I think we can delete this.

apps/omg_watcher_info/test/omg_watcher_info/db/repo_test.exs
```diff
@@ -25,6 +25,15 @@ defmodule OMG.WatcherInfo.DB.RepoTest do
   require Utxo
 
   describe "DB.Repo.insert_all_chunked/3" do
+    # The current number of columns on the transaction table allow up to 8190
```
unnawut: Can you also describe how to derive the number 8190? So we know how to re-calculate.

apps/omg_watcher_info/test/omg_watcher_info/db/repo_test.exs
```diff
@@ -25,6 +25,15 @@ defmodule OMG.WatcherInfo.DB.RepoTest do
...
+    # transactions to be inserted using `DB.Repo.insert_all/3` before chunking must
+    # be done to avoid hitting postgres limits.
+    @max_txns_before_chunking 8191
+
+    # The current number of columns on the transaction table require a chunk size of
+    # 10922 to avoid hitting postgres limits.
```
unnawut: Can you also describe how to derive the number 10922? So we know how to re-calculate.

apps/omg_watcher_info/lib/omg_watcher_info/db/repo.ex
```diff
@@ -33,16 +34,25 @@ defmodule OMG.WatcherInfo.DB.Repo do
...
     |> Enum.each(&insert_all(schema_or_source, &1, opts))
   end
 
+  # Note: an entry with 0 fields will cause a divide-by-zero error.
+  #
+  # DB.Repo.chunk_size(%{}) ==> (ArithmeticError) bad argument in arithmetic expression
+  #
+  # Do we want/need to be that defensive?
+  def chunk_size(entry) do
+    @max_params_count |> div(entry |> fields_count)
```
unnawut:
```suggestion
    fields_count = fields_count(entry)
    div(@max_params_count, fields_count)
```

I see this is old code but I think it's an overuse of pipes. Maybe this?

apps/omg_watcher_info/test/omg_watcher_info/db/repo_test.exs
```diff
@@ -60,4 +89,32 @@ defmodule OMG.WatcherInfo.DB.RepoTest do
...
+      [new_transaction(blknum, index) | acc]
+    end)
+  end
+
+  # Prefer using `ExMachina.build/2 which uses `OMG.WatcherInfo.Factory.Transaction`
+  # over this function. This function is built to be fast and simple.
+  #
+  # `ExMachina.params_for/2` could be used here to make use of `OMG.WatcherInfo.Factory.Transaction`.
+  # But the transaction factory does a lot of extra stuff unnecessary for this test. This stripped
+  # down version is about 15x faster. Also using `ExMachina.params_for/2` here also requires some
```
unnawut: How to get the number 15x?

Some other thoughts that don't block this PR.
1. I agree with "The 2 functions differ in the inputs they expect" so we need to diverge from `ExMachina.build/2`
2. I tend to disagree with having another factory function for speed and simplicity, in exchange with having different ways of building data that we need to be aware of and maintain.

apps/omg_watcher_info/lib/omg_watcher_info/db/repo.ex
```diff
@@ -33,16 +34,25 @@ defmodule OMG.WatcherInfo.DB.Repo do
...
-    |> Enum.map(fn entry -> Map.merge(entry, %{inserted_at: utc_now, updated_at: utc_now}) end)
     |> Stream.chunk_every(chunk_size)
     |> Enum.each(&insert_all(schema_or_source, &1, opts))
   end
 
+  # Note: an entry with 0 fields will cause a divide-by-zero error.
+  #
+  # DB.Repo.chunk_size(%{}) ==> (ArithmeticError) bad argument in arithmetic expression
+  #
+  # Do we want/need to be that defensive?
```
unnawut: I would say it's too defensive. I couldn't think of a case where this would happen. And even if we handle it here `|> Stream.chunk_every(0)` is not possible anyway.

apps/omg_watcher_info/test/omg_watcher_info/db/repo_test.exs
jrhite: I added a test that both shows how to derive the number and that the number is correct (there was a typo in the comment, the actual max is 8191).

apps/omg_watcher_info/test/omg_watcher_info/db/repo_test.exs
jrhite: Ah, this number is wrong. I made a mistake. Actually `@max_txns_before_chunking  == @txn_chunk_size == 8191` which makes sense.

The number `10_922` came from a bad implementation of the `"chunk_size/1 calculates the correct chunk...` test. Notice `inserted_at` and `updated_at` were not being added to the transaction.
```
block = insert(:block)
transaction = new_transaction(block.blknum, 1)
assert DB.Repo.chunk_size(transaction) == @txn_chunk_size
```
I've fixed the test and removed the constant `@txn_chunk_size` completely. But still I think testing `OMG.WatcherInfo.DB.Repo.chunk_size/1` directly is important and want to keep it.

If you wish to derive the number just change and an `IO.puts()` or `IO.inspect() in the `chunk_size/1` function. Then run the test `test "DB.Repo.insert_all for transactions (via postgres INSERT)...` in `OMG.WatcherInfo.DB.RepoTest` and see what it prints out. To verify its correctness change the `chunk_size/1` to `div(@max_params_count, fields_count(entry)) + 1` and run `test "DB.Repo.insert_all for transactions (via postgres INSERT)...` again and it will fail.

apps/omg_watcher_info/test/omg_watcher_info/db/repo_test.exs
jrhite: Regarding the 15x I just ran it with `ExMachina.params_for/2`, dropping the breaking keys. Logged the test start and end time. Then I ran the test with the `new_transactions/3` function and compared times.

My 'optimization' doesn't change the Big-O of the test, but is perceivably faster...instant test result, vs a 15 second wait.

I agree your other points as well regarding test factory stuff. Let's get to that later.

apps/omg_watcher_info/lib/omg_watcher_info/db/repo.ex
```diff
@@ -32,17 +32,29 @@ defmodule OMG.WatcherInfo.DB.Repo do
...
+  # by breaking large single INSERT statements into smaller ones, it also creates a bottleneck
+  # where out-of-memory errors could become a problem. See this issue for more information:
+  # https://github.com/omisego/elixir-omg/issues/1357
   def insert_all_chunked(schema_or_source, entries, opts) do
-    chunk_size = @max_params_count |> div(entries |> hd |> fields_count)
-
     utc_now = DateTime.utc_now()
+    entries = Enum.map(entries, fn entry -> Map.merge(entry, %{inserted_at: utc_now, updated_at: utc_now}) end)
+
+    chunk_size = chunk_size(entries |> hd)
```
InoMurko:
```suggestion
    chunk_size = entries |> hd() |> chunk_size()
```

apps/omg_watcher_info/lib/omg_watcher_info/db/repo.ex
```diff
@@ -32,17 +32,29 @@ defmodule OMG.WatcherInfo.DB.Repo do
...
 
+  # Note: an entry with 0 fields will cause a divide-by-zero error.
+  #
+  # DB.Repo.chunk_size(%{}) ==> (ArithmeticError) bad argument in arithmetic expression
+  #
+  # But we could not think of a case where this code happen, so no defensive
+  # checks here.
+  def chunk_size(entry), do: div(@max_params_count, fields_count(entry))
+
   defp fields_count(map) when is_map(map), do: map |> Kernel.map_size()
```
InoMurko:
```suggestion
  defp fields_count(map) when is_map(map), do: Kernel.map_size(map)
```

apps/omg_watcher_info/test/omg_watcher_info/db/repo_test.exs
unnawut: Umm I was just expecting just this formula but thanks

```
Formula for chunk size = floor(postgres_param_limit / schema_column_count)
Transaction columns = 8
Therefore, testing chunk size = floor(65535 / 8) = 8191
```

apps/omg_watcher_info/lib/omg_watcher_info/db/repo.ex
InoMurko: question @jrhite why do we even need this? when are we inserting a `single map` and would worry it has to many elements?

jrhite: Not sure I understand your question? I didn't write the original code and to avoid another unexpected regression tried to limit the changes as much as possible.

Number of fields is needed to calculate `chunk_size`.

apps/omg_watcher_info/lib/omg_watcher_info/db/repo.ex
```diff
@@ -32,17 +32,29 @@ defmodule OMG.WatcherInfo.DB.Repo do
 
   def insert_all_chunked(_schema_or_source, [], _opts), do: :ok
 
+  # Note that while this function allows us to get around postgres' max parameters limit
```
InoMurko: delete this comment pls and close the issue

jrhite: done

jrhite:
> Great work!
> Just one general comment, I'd like to recommend [commit message convention](https://omisego.atlassian.net/wiki/spaces/ENG/pages/63111218/Commit+Message+Conventions+OMG+Network) some of the team members follow.

yeah sorry. i don't get that doc really. 😄  on the other hand in my subsequent commits, i tried to follow it.

apps/omg_watcher_info/lib/omg_watcher_info/db/repo.ex
```diff
@@ -33,16 +33,24 @@ defmodule OMG.WatcherInfo.DB.Repo do
...
+  # Note: an entry with 0 fields will cause a divide-by-zero error.
+  #
+  # DB.Repo.chunk_size(%{}) ==> (ArithmeticError) bad argument in arithmetic expression
+  #
+  # But we could not think of a case where this code happen, so no defensive
+  # checks here.
+  def chunk_size(entry), do: div(@max_params_count, fields_count(entry))
+
+  defp fields_count(map) when is_map(map), do: Kernel.map_size(map)
   defp fields_count(list) when is_list(list), do: length(list)
```
InoMurko: as said in slack, this is not hit by any test, I propose we remove this and just be vigilant if it's going to be hit anywhere else
```suggestion
defp fields_count(map), do: Kernel.map_size(map)
```

apps/omg_watcher_info/lib/omg_watcher_info/db/repo.ex
```diff
@@ -33,16 +33,23 @@ defmodule OMG.WatcherInfo.DB.Repo do
...
-  defp fields_count(list) when is_list(list), do: length(list)
+  # Note: an entry with 0 fields will cause a divide-by-zero error.
+  #
+  # DB.Repo.chunk_size(%{}) ==> (ArithmeticError) bad argument in arithmetic expression
+  #
+  # But we could not think of a case where this code happen, so no defensive
+  # checks here.
+  def chunk_size(entry), do: div(@max_params_count, fields_count(entry))
+
+  defp fields_count(map) when is_map(map), do: Kernel.map_size(map)
```
InoMurko:
```suggestion
  defp fields_count(map), do: Kernel.map_size(map)
```

PR #1358, opened 2020-02-27
Add CI to priv/perf

boolafish:

## Overview

> how are you gonna push clean code to elixir-omg … if you don’t have a CI to validate you have clean code?

Thus....add CI to `priv/perf` 🎉

## Changes

- add credo. (copy pasted from `priv/cabbage`)
- check credo and format for `priv/perf` in CI
- run load test in `test_barebone_release` and `test_docker_compose_release`. I think in the future we will make it run low iterations here and run higher iterations against real env? We need a place to make sure our load tests code is actually run-able before merge 🤔 should there be better approach?
- add `run_load_test_on_deployed_env` job in circle that hits the deployed env. Test example (without requiring previous steps done): [here](https://app.circleci.com/jobs/github/omisego/elixir-omg/44483), slack discussion: [here](https://omisego.slack.com/archives/CMX790Q5V/p1582780736078200). Seems like a bit controversial? If we dislike the approach we can add that when we have clearer view on how we want to run the tests.

## Testing

CI running and passing


.circleci/config.yml
```diff
@@ -594,6 +620,32 @@ jobs:
       - checkout
       - run: DEPLOY=watcher_info sh .circleci/ci_deploy.sh
 
+  run_load_test_on_deployed_env:
```
pgebal: Right now this step is not being run in CI (which is good IMO), but is it intended?

PR #1359, opened 2020-02-27
feat: prepare v0.4.4 and relax tag publishing rules

unnawut:
## Overview

Bump up `VERSION` in preparation for `v0.4.4` and relax circleci publishing rules for tags

## Changes

- Bump `VERSION` to 0.4.4
- Loosen the CircleCI rules for publishing on tags. This is an attempt to debug why CircleCI is not publishing them. We only push tags for releases so loosen this up shouldn't be an issue.

## Testing

N/A


InoMurko:
when is the release rolling out?

jbunce:
@InoMurko when I get the image :)

InoMurko:
Aha! OK. 
1. docker-compose-watcher.yml @unnawut please bump the SHA tags
2. Infra needs to set `LOGGER_BACKEND=console` because json logs are not ready on the infra level yet

.circleci/config.yml
```diff
@@ -748,8 +747,7 @@ workflows:
                 - /^v[0-9]+\.[0-9]+/
```
jrhite: is this one left behind intentionally?

unnawut:
We still need to do a load test on this bundle, get all downstream products tested with this version and the actual publishing, so I'm guesstimating we're looking at up coming Monday 2 March the earliest.

.circleci/config.yml
```diff
@@ -725,8 +725,7 @@ workflows:
...
```
jrhite: is this one left behind intentionally?

jbunce:
@InoMurko can we keep this release just for the fix? I know it's tempting to get the new feature in now but we're right in the middle of a refactor for versioning of our Helm charts.

.circleci/config.yml
unnawut: Yup. The branches are working fine.

.circleci/config.yml
unnawut: Yup. The branches are working fine.

unnawut:
> docker-compose-watcher.yml @unnawut please bump the SHA tags

I keep forgetting this, and probably will keep forgetting this 🤦‍♂ Any idea if we can automate/lint this?

PR #1366, opened 2020-03-02
fix: run CircleCI builds on tags

unnawut:
## Overview

Fix CircleCI not running CI on git tags

## Changes

By default, tags do not trigger CI jobs and the only way to enable it is through `only:` config. And so because we add `only:` config we need to add one for branches too.

Sources:
- https://discuss.circleci.com/t/builds-for-tags-not-triggering/17681
- https://discuss.circleci.com/t/build-workflow-not-triggered-when-tag-is-pushed/19577

## Testing

Merge this PR. Push a new tag. CircleCI should run on tags now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -665,7 +665,12 @@ workflows:
           requires: [build]
   build-test-deploy:
     jobs:
-      - build
+      - build:
+          filters:
+            branches:
+              only: /.+/
+            tags:
+              only: /.+/
       - test_barebone_release:
           requires: [build]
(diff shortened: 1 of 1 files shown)


jbunce:
Shouldn't the branch filter be master? Then this reads: only run on branch master and with a tag?

unnawut:
> Shouldn't the branch filter be master? Then this reads: only run on branch master and with a tag?

It's the `build` job so we need `/.*/` so all PR branch pushes get build 🎉

PR #1344, opened 2020-02-24
Cabbagize invalid exit tests

pdobacz:
Fixes #1333 

## Overview

This is a still quite tentative take on how to turn invalid (standard) exit tests into cabbage tests.

I'm having many doubts about the approach and I'll probably experiment with alternative approaches (see below).

I'm keeping the tidied up "old" approach tests, as I'd like to first gain more confidence in the cabbage tests added.

## Changes

- does some necessary intro fixes to the "old" integration tests which were not strict enough
- separates those tests into two "kinds"
    - tests requiring an invalid block (not moving to cabbage yet)
    - tests not requiring an invalid block (this is moving)
- turns ~one~ three of those tests into cabbage tests
- tried to stick by conventions, but did some refactors: added `Itest.PlasmaFramework` to group related tools dispersed across modules, made `StandardExitClient` more flexible to allow invalid exits
- took the liberty to bump log level to `info`; I think the `debug` logs weren't very useful
- customized the tested services, to enable sla-margin-related tests to execute in reasonable time.
- ADDED: revamped the `in_flight_exits_test`, so that it leaves the state of the chain intact (no byzantine events)

## Testing

run tests as usual / CI

## TODOs

 - [ ] - ~standardize with the `in_flight_exits_test.exs` and `standard_exit_test.exs` somehow, but that's pending review of `invalid_standard_exit_test.exs` conventions~ would love to, but not in this PR, this is huge already

## Alternative approaches

Current gherkin wording is [this](https://github.com/omisego/elixir-omg/compare/pdobacz/1333_fix_cabbagize_invalid_exit_tests?expand=1#diff-ba1c55a4d50e4a09cea75b4a5c4199be)

It comes from the "happy path" `standard_exits.feature`.

However there's a couple of modifications I'm considering:
  - a very general concern - distinguish the `watcher-info`/`happy-path`-related scenarios like the `standard_exit.feature` from the `watcher-security`/`unhappy-path` ones, like `invalid_standard_exit.feature`. The latter should probably be much more detailed in the wording (e.g. named utxos and transactions, not unlike the alicebobbing-scenarios used in some of our docs). But this has drawbacks too
  - wrap up more of the setup under a single `Given Alice has ETH on a secure chain` or something, collapsing the first 3-4 lines into one (**EDIT**: in fact, I think I'll do it right away)

PR #1372, opened 2020-03-03
Fix "piggyback_available" popping back after challenge

### LINKED ISSUE #1318: Split invalidly processed exit from unchallenged exit events
Coming from #1275, which in the end opted to take a different approach to solve.

> separate the conflated unchallenged_exit from invalidly_processed_exit and have things called by their name. The conflating of those events is too much of a stop-gap solution to continue with.
> 
> Then things get easier - when an exit processes invalidly (which was Gerhard's case) we'll get a dedicated event invalidly_processed_exit.
> 
> And as a consequence, we'll have more freedom in setting exit_processor_sla_margin, which drives unchallenged_exit. This is because we'll then set it to something larger than min_exit_period, which matters in cases where min_exit_period is really small.

This has its own merit, since conflating of the notions of an unchallenged exit and of an exit that processed invalidly under a single report (`unchallenged_exit`) and logic is messy and hard to understand by someone who's looking.

### LINKED ISSUE #1371: Challenged, invalid piggybacks may pop back to `piggyback_available` state
This happens in the cases where the IFE tx never got included.

The reason is [`is_piggybacked?`](https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher/lib/omg_watcher/exit_processor/in_flight_exit_info.ex#L346) is used when figuring `piggyback_available` out. If `true`, that report won't be shown.

However, challenging a piggyback flips `:is_piggybacked` to `false` [here](https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher/lib/omg_watcher/exit_processor/in_flight_exit_info.ex#L296).

Nice fix: #1318 and introduce a better data model (enums _in lieu_ of a bunch of boolean values).

Quick fix: introduce `is_challenged` field in the `exit_map`

pdobacz:
Closes #1371 

## Overview

It is the "Quick fix" mentioned in #1371. The larger fix #1318 would do proper justice to the handling of piggyback state (description of that in #1318 still pending), which should not be modeled with a bunch of booleans. There should also be a revamp done to the overall "lifecycle" of the piggyback, to include invalidly processing ones.

## Changes

 - adds a simple test picking up the case in question, in `piggyback_test.exs`
 - refactors some other tests and inner working of `InFlightExitInfo`. We're interrested in the notions of 1/ "activity" of a piggyback, where it means it may finalize and we perhaps want to check if it is valid and 2/ "being piggybacked", where it means just this.
 - it finally adds an explicit notion of `is_challenged` to the IFE's piggyback model. Whenever a piggyback challenge happens it now turns that `true` instead of turning `is_piggybacked` to `false`, which led to the error

## Testing

the usual tests apply. Also I'll use this in the revamped IFE tests prepared as part of #1344, where we will challange invalid piggybacks and require that they do not show up as "`piggyback_available`" anymore

### DIFF
--- a/apps/omg_watcher/test/support/exit_processor/test_helper.ex
+++ b/apps/omg_watcher/test/support/exit_processor/test_helper.ex
@@ -25,6 +25,7 @@ defmodule OMG.Watcher.ExitProcessor.TestHelper do
 
   require Utxo
 
+  # default exit_id used when starting exits using `start_se_from` and `start_ife_from`
   @exit_id 9876
 
   def start_se_from(%Core{} = processor, tx, exiting_pos, opts \\ []) do
--- a/apps/omg_watcher/test/support/exit_processor/test_helper.ex
+++ b/apps/omg_watcher/test/support/exit_processor/test_helper.ex
@@ -25,7 +25,7 @@ defmodule OMG.Watcher.ExitProcessor.TestHelper do
 
   require Utxo
 
-  @exit_id 1
+  @exit_id 9876
 
   def start_se_from(%Core{} = processor, tx, exiting_pos, opts \\ []) do
     {event, status} = se_event_status(tx, exiting_pos, opts)
--- a/apps/omg_watcher/lib/omg_watcher/exit_processor/core.ex
+++ b/apps/omg_watcher/lib/omg_watcher/exit_processor/core.ex
@@ -513,14 +513,14 @@ defmodule OMG.Watcher.ExitProcessor.Core do
 
     available_inputs =
       input_witnesses
-      |> Enum.filter(fn {index, _} -> not InFlightExitInfo.is_active?(ife, {:input, index}) end)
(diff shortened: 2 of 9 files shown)


pdobacz:
@paulperegud @pnowosie

It seems I introduced a regression along the way. ef7ee5afc5c4659d14aa6724625f72d60874cb34 exposes this in a test and 766e69c fixes.

Would you be so kind to take another glance at this?

pdobacz:
_sigh_ and one more regression picked up and fixed in 73fdb85 & f417502

apps/omg_watcher/test/support/exit_processor/test_helper.ex
```diff
@@ -25,7 +25,7 @@ defmodule OMG.Watcher.ExitProcessor.TestHelper do
 
   require Utxo
 
-  @exit_id 1
+  @exit_id 9876
```
pnowosie: It's unlikely that somebody needs to change the value, and it'll be discovered quickly but maybe to be nice can you add a comment saying the value is used by other tests?
Thanks!

PR #1381, opened 2020-03-06
chore: add more Tesla middleware overriding for a more stable client on load tests

boolafish:

## Overview

Override more on Tesla middleware for stable client.

## Changes

- use Jason instead of Poison
- change the header to "perf" on Tesla here.
- add retry and retry logic to the Tesla client to keep the test alive against transient error. It is by default using exponential backoff logic ✨
- setup connection pool (`:perf_pool`) in `Application.start/2` and use the pool in Tesla client
- update README

### References

- Jason vs Poison: https://github.com/phoenixframework/phoenix/issues/2693#issuecomment-388881650
- Tesla Middleware: https://hexdocs.pm/tesla/Tesla.Middleware.Retry.html#content
- Hackney connection pool: https://github.com/benoitc/hackney#use-the-default-pool
- Setting up `opts` for hackney: https://github.com/teamon/tesla/wiki/0.x-to-1.0-Migration-Guide#adapter-options-need-to-be-wrapped-in-adapter-key

## Testing

The setup is mainly tested/POC in `omg-load-test` repo with overriding the generated client.
`OMG.Perf.LoadTest.WatcherInfo` is run with 1000 concurrency and was able to continuously generate load for hours.

```
  plug Tesla.Middleware.Headers, [{"user-agent", "Elixir"}]
  plug Tesla.Middleware.EncodeJson, engine: Poison
  plug Tesla.Middleware.Opts, [adapter: [recv_timeout: 30_000, pool: :mypool]]
  plug Tesla.Middleware.Retry,
    delay: 500,
    max_retries: 10,
    max_delay: 45_000,
    should_retry: fn
      {:ok, %{status: status} = resp} when status in [400, 500, 502] -> 
        Logger.error("tesla middleware log - retry with status error: #{inspect({:ok, resp})}")
        true
      {:ok, resp} -> 
        Logger.debug("tesla middleware log: #{inspect({:ok, resp})}")
        false
      {:error, resp} -> 
        Logger.error("tesla middleware log - retry with error: #{inspect({:error, resp})}")
        true
      resp ->
        Logger.error("tesla middleware log - retry with unknown: #{inspect(resp)}")
        true
    end
```

See load from datadog: [here](https://app.datadoghq.com/apm/service/web/request?env=stress-e043a92-ropsten-01&paused=false&start=1583459095197&end=1583473495197)

With the retry from Tesla here, it does not even trigger `retry_on_error` from Chaperon session. (I put some `retry_on_error` on the higher level but it seems not triggered at all, for `retry_on_error` see: https://github.com/omisego/omg-load-testing/pull/37)

Also, the smoke test here to ensure it is not breaking.

### DIFF
--- a/priv/perf/README.md
+++ b/priv/perf/README.md
@@ -3,14 +3,13 @@
 Umbrella app for performance/load/stress tests
 
 ## How to run the test
-1. Override the urls for the services by environment variables if needed. Otherwise would be set to the default localhost with port defined in docker-compose file (No need to override if running against local services). for example:
-    ```
-    CHILD_CHAIN_URL=http://localhost:7534
-    WATCHER_SECURITY_URL=http://localhost:7434
-    WATCHER_INFO_URL=http://localhost:7534
-    ```
+1. Change the urls inside the [config files](config/) for the environment you are targeting.
+1. To generate the open-api client (only need once): `make init`
+1. To run the test: `MIX_ENV=<your_env> mix test`. Or `mix test` if you want to run against local services.
 
+### Spin up local services for development
 1. To test with local services (by docker-compose): `make start-services`
-1. To generate the open-api client: `make init`
-1. To run the test: `make test`
 1. To turn the local services down: `make stop-services`
+
+### Increase connection pool size and connection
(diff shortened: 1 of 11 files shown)


InoMurko:
👏

PR #1391, opened 2020-03-15
pool size and transactions in block

InoMurko:
increase introspection

## Overview

Plot the size of child chain mempool and number of transactions in a block
## Changes

- telemetry in the right places
- handlers


InoMurko:
Screen Shot 2020-03-15 at 22 31 20

priv/cabbage/apps/itest/test/itest/transactions.exs
```diff
@@ -0,0 +1,174 @@
...
+  use Cabbage.Feature, async: true, file: "transactions.feature"
+
+  require Logger
+
+  alias Itest.Account
+
+  alias Itest.Client
+  alias Itest.Transactions.Currency
+
+  @num_accounts 50
```
InoMurko: needs to be an even number

priv/cabbage/apps/itest/lib/account.ex
```diff
@@ -9,7 +9,11 @@ defmodule Itest.Account do
   @spec take_accounts(integer()) :: map()
   def take_accounts(number_of_accounts) do
     1..number_of_accounts
-    |> Task.async_stream(fn _ -> account() end, timeout: 60_000, on_timeout: :kill_task)
+    |> Task.async_stream(fn _ -> account() end,
+      timeout: 60_000,
+      on_timeout: :kill_task,
+      max_concurrency: System.schedulers_online() * 2
```
pgebal: Why not default `max_concurrency`?

priv/cabbage/apps/itest/test/features/transactions.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Transactions
+
+  Scenario: 50 entities exchange ETH
+    When they deposit "1" ETH to the root chain
+    And they deposit "1" ETH to the root chain
+    Then they should have "2" ETH on the child chain
+    When they send others "1" ETH on the child chain
+    Then others should have "1" ETH on the child chain
```
pgebal: No new line at the end of the file.

priv/cabbage/apps/itest/test/itest/transactions.exs
```diff
@@ -0,0 +1,174 @@
+defmodule TransactionsTests do
```
pgebal: Please add copyrights.

priv/cabbage/apps/itest/test/itest/transactions.exs
```diff
@@ -0,0 +1,174 @@
...
+        geth_block_every = 1
+
+        {:ok, response} =
+          WatcherSecurityCriticalAPI.Api.Configuration.configuration_get(WatcherSecurityCriticalAPI.Connection.new())
+
+        watcher_security_critical_config =
+          WatcherSecurityCriticalConfiguration.to_struct(Jason.decode!(response.body)["data"])
+
+        finality_margin_blocks = watcher_security_critical_config.deposit_finality_margin
+        to_miliseconds = 1000
```
pgebal: miliseconds -> milliseconds

priv/cabbage/apps/itest/test/itest/transactions.exs
```diff
@@ -0,0 +1,174 @@
...
+  end
+
+  defthen ~r/^they should have "(?<amount>[^"]+)" ETH on the child chain$/,
+          %{amount: amount},
+          %{alices: alices} = state do
+    alices
+    |> Enum.with_index()
+    |> Task.async_stream(
+      fn {{alice_account, alice_pkey}, index} ->
+        geth_block_every = 1
```
pgebal: that can be defined closer to the place where it's used for the first time,
please put it above `to_milliseconds`, or just name it `get_block_every_milliseconds` and put it instead of `to_milliseconds`

priv/cabbage/apps/itest/test/itest/transactions.exs
```diff
@@ -0,0 +1,174 @@
...
+        finality_margin_blocks = watcher_security_critical_config.deposit_finality_margin
+        to_miliseconds = 1000
+
+        finality_margin_blocks
+        |> Kernel.*(geth_block_every)
+        |> Kernel.*(to_miliseconds)
+        |> Kernel.round()
+        |> Process.sleep()
+
+        expecting_amount = Currency.to_wei(amount)
```
pgebal: expecting_amount -> expected_amount

priv/cabbage/apps/itest/test/itest/transactions.exs
```diff
@@ -0,0 +1,174 @@
...
+        {:ok, [sign_hash, typed_data, _txbytes]} =
+          Client.create_transaction(
+            Currency.to_wei(amount),
+            alice_account,
+            bob_account
+          )
+
+        # Alice needs to sign 2 inputs of 1 Eth, 1 for Bob and 1 for the fees
+        submitted = Client.submit_transaction(typed_data, sign_hash, [alice_pkey, alice_pkey])
+        Logger.warn("submitted #{inspect(submitted)}")
```
pgebal: That line probably should be deleted. If we want to log it, then probably debug or info would do.

priv/cabbage/apps/itest/lib/account.ex
InoMurko: because at least on macs thats 4 cores, which means 4 processes - we can easily do more, geth doesn't so i just double the amount. if you try to do more you can crash geth in a container.

priv/cabbage/apps/itest/test/itest/transactions.exs
pdobacz: put this comment into code pls

priv/cabbage/apps/itest/test/itest/transactions.exs
pdobacz: we have a `credo` pass which should pick up missing licenses - it works at the root-level credo config. If the credo configs are separate for `priv` maybe let's copy it over to here?

priv/cabbage/apps/itest/lib/account.ex
InoMurko: geths idea of concurrency is different then erlangs, where you can have 100k concurrent processes on the beam, geth can have only a few. so we can't bombard the geth nodes too much. but 8 concurrent processes creating accounts is still faster than 4 (on my machine :) )

priv/cabbage/apps/itest/test/itest/transactions.exs
InoMurko: fixed it

Issue #1173, opened 2019-11-29
Lock down the dependencies for chch / watcher


thec00n:
### Description 
The chch / watcher are critical components of the OMG network. As we are getting closer to running production grade services there is a need to evaluate the current dependencies from a security perspective and lock them down.

### Suggested Fix
- Review all current dependencies 
- Lock down required dependencies 
- Decide on a upgrade process

InoMurko:
1. Code owners for mix.lock (that would summon all devs - more eyes more security?)
2. `DependABot` (review for dependecy bumps)
3. Security review of current dependencies 
4. Review if we really need all dependencies for production builds (try and lower these!) -> Prioritize review of dependencies (smaller to bigger - for example, cowboy is a dependency used a lot by the community)

InoMurko:
@thec00n awaiting your input.

thec00n:
Consider using  https://github.com/mirego/mix_audit

okalouti:
Structuring the components of this issue a little bit: 
- [ ] Review and implement automated dependency checking: [Dependabot](https://dependabot.com/)
- [ ] Review and implement automated dependency checking: [mix_audit](https://hexdocs.pm/mix_audit/readme.html)
- [ ] Identify list of dependencies used in release (versus ones for testing/development) to prioritise for scrutiny
- [ ] Identify any production dependencies that can be removed.
- [ ] Discuss and - if agreed - implement [CODEOWNERS](https://help.github.com/en/enterprise/2.15/user/articles/about-code-owners) for `mix.lock` and `mix.exs`

okalouti:
Closing following thread with Piotr: https://omisego.slack.com/archives/C9KNA675J/p1585058472388200

PR #1393, opened 2020-03-16
Adding automated dependency audits

okalouti:
Addresses https://github.com/omisego/elixir-omg/issues/1173

## Overview

Adding [mix_audit]() to dependencies and to the CI workflow.

## Changes

- `mix.exs` (root directory) and `mix.lock` adding `mix_audit`
-  Additions to `.circleci/config.yml` to add to CI workflow. 

## Testing

/


okalouti:
Updated `mix_audit` to version `0.1.3` in order to remove warning about our earlier Elixir version. 
They fixed this issue after I requested: https://github.com/mirego/mix_audit/issues/3

thec00n:
Does this PR include `Review all current dependencies`?

okalouti:
Hi @thec00n. This PR does not remove unnecessary dependencies, but `mix_audit` reviews all dependencies in `mix.lock`

okalouti:
Wondering if `audit_deps` becomes once this branch is merged to master? It isn't currently, but nothing in the`.circle` file that I see missing.

mederic-p:
Approving, let's wait for Ino's review before merging tho.

InoMurko:
The output of the audit is: “No vulnerabilities found.“

Seems a bit... simplistic. Can you give us an overview what it does? Maybe provide an example where and how it would yield ‘a hacker’!

okalouti:
Hi @Ino. Sorry I realized I linked the wrong URL above. See [here](https://github.com/mirego/mix_audit) for details on the dependency.

```
How does it work?

MixAudit builds two lists when it’s executed in a project:

- A list of security advisories fetched from the community-maintained elixir-security-advisories repository
- A list of Mix dependencies from the various mix.lock files in the project

Then, it loops through each project dependency and tries to find security advisories that apply to it (through its package name) and that match its version specification (through the advisory patched and unaffected version policies).

If one is found, a vulnerability (the combination of a security advisory and a project dependency) is then added to the report.

The task will exit with a 0 status only if the report passes (ie. it reports no vulnerabilities). Otherwise, it will exit with a 1 status.
```

They show examples of what it looks like if a vulnerability is found. Not sure how we can "test" that though.

InoMurko:
Cool @okalouti nice work. The security repo is here: https://github.com/dependabot/elixir-security-advisories/tree/master/packages

it's rather thin, hopefully that doesn't give us a false assumption our dependencies are all safe. It's a start of a process!

PR #1397, opened 2020-03-17
Fix get IFE data for spent deposit

### LINKED ISSUE #1310: Internal server error caused by certain IFE tx
The following API query returns an internal server error. 

```
POST /in_flight_exit.get_data HTTP/1.1
Content-Type: application/json
host: xxx
Content-Length: 413
Connection: close

{"txbytes":"0xf8b9f843b8418b9e7b949b9b3971074666b9f4814ad3805785612aed4a8f945b71138bcbe16720e8118e52d4e9213a43a0f2079c8ec3296f18af6620adf0a8caff3e9b0d8e4c1b01e1a000000000000000000000000000000000000000000000000000000e8d85ebca00eeed01eb94d42b31665b93c128541c8b89a0e545afb08b7dd89400000000000000000000000000000000000000000180a00000000000000000000000000000000000000000000000000000000000000000","jsonrpc":"2.0","id":0}
```

mederic-p:
Closes #1310 #1128 

## Overview

This PR fixes a server error when a client was requesting IFE data for a transaction that contained a spent deposit as an input.

It has been decided to do this fix first even if the root cause of the problem is that we should be able to get the data and not get an error.

A more in depth description of the issue is available [here](https://github.com/omisego/elixir-omg/issues/1408).

## Changes

- Handle missing clause.

## Testing

Added a test to cover deposits.


apps/omg_watcher/lib/omg_watcher/api/in_flight_exit.ex
```diff
@@ -106,6 +106,9 @@ defmodule OMG.Watcher.API.InFlightExit do
 
       {:error, :utxo_not_found} ->
         {:halt, {:error, :tx_for_input_not_found}}
+
+      {:error, error} = e when is_atom(error) ->
```
pdobacz: instead of a catch-all-errors clause, please pattern match the expected atom (`:no_deposit_for_given_blknum` I presume).

I would also advocate translating this particular error into something more meaningful like (`deposit_input_spent_ife_unsupported`)

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/in_flight_exit_test.exs
```diff
@@ -103,6 +103,17 @@ defmodule OMG.WatcherRPC.Web.Controller.InFlightExitTest do
...
     end
 
+    @tag fixtures: [:web_endpoint, :db_initialized, :bob]
+    test "behaves well if input is a spent deposit", %{bob: bob} do
+      in_flight_txbytes =
+        [{1, 0, 0, bob}]
+        |> OMG.TestHelper.create_encoded(@eth, [{bob, 150}])
+        |> Encoding.to_hex()
+
+      assert %{"code" => "exit:invalid", "description" => "Utxo was spent or does not exist."} =
```
pdobacz: here we can then provide a specific and more meaningful message

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/in_flight_exit_test.exs
```diff
@@ -103,6 +103,17 @@ defmodule OMG.WatcherRPC.Web.Controller.InFlightExitTest do
              } = WatcherHelper.no_success?("/in_flight_exit.get_data", %{"txbytes" => in_flight_txbytes})
     end
 
+    @tag fixtures: [:web_endpoint, :db_initialized, :bob]
+    test "behaves well if input is a spent deposit", %{bob: bob} do
```
pdobacz: it still doesn't "behave well", at least not fully. How about "provides a report on unsupported start IFE case if input is a spent deposit"?

PR #1404, opened 2020-03-17
fix accessing to contracts, encoding/decoding in tests, hierarchical pattern, root chain access, fetch vault address from contracts

InoMurko:
## Overview

This is further reducing our exposure to failed deployments and possibly corrupting an environment.
## Changes

- Exit games and Vaults are now derived from plasma framework contracts (passing in currency or payment type)
- parametrising our workers in supervision trees
- more consistent data retrieval from contracts
- code reorganisation 
- removed complexity in blockgetter, removed `_ = :sys.get_status(ExitProcessor, 10 * 60_000)`, idk why it was there, @pdobacz pls take a look

## Testing

mix test


apps/omg/lib/omg/state.ex
```diff
@@ -92,14 +92,13 @@ defmodule OMG.State do
   Initializes the state. UTXO set is not loaded now.
   """
   def init(opts) do
-    {:ok, height_query_result} = DB.get_single_value(:child_top_block_number)
-    {:ok, child_block_interval} = Eth.RootChain.get_child_block_interval()
-
+    {:ok, child_top_block_number} = DB.get_single_value(:child_top_block_number)
```
InoMurko: this was just renamed from `height_query_result ` to `child_top_block_number `

InoMurko:
I have made some structural mistakes and opened the PR too soon.

apps/omg_watcher/lib/omg_watcher/supervisor.ex
```diff
@@ -32,8 +32,6 @@ defmodule OMG.Watcher.Supervisor do
   end
 
   def init(:ok) do
-    # prevent booting if contracts are not ready
-    :ok = RootChain.contract_ready()
```
InoMurko: contracts cannot not be ready at this point, because we fetched contracts addresses from plasma 
 contracts in SetContracts module.

InoMurko:
This might look like a big PR, but the reason the PR seems big is because it's touching lots of files. The reason it touches lots of files, is because of how highly coupled our test suite is.

- Since the introduction of snapshots, there's not reason 90% of our functions expose optional patterns (contracts,opts). So I removed those.
- We were passing binary and hex contracts here and there, because there was no consistent agreement on which layer an encoding/decoding should have happened. I tried to address this as much as I could.
- The recent deployment issues we were having is largely related to the exposing vector to false deployments. Contracts for vaults and exit games are now fetched from contracts itself (as per integration document), when the application starts.
- RootChain access has now been formalised and standardzied with the use of ExAbi.

apps/omg/lib/omg/state.ex
```diff
@@ -284,7 +277,7 @@ defmodule OMG.State do
   end
 
   defp do_form_block(state) do
-    {:ok, child_block_interval} = Eth.RootChain.get_child_block_interval()
```
InoMurko: this was a call to configuration:
```
def get_child_block_interval(), do: Application.fetch_env(:omg_eth, :child_block_interval)
```
so now instead of reading this from config, we read it once on process init and it stays in the process state

apps/omg/lib/omg/state/core.ex
```diff
@@ -51,6 +51,7 @@ defmodule OMG.State.Core do
   defstruct [
     :height,
     :fee_claimer_address,
+    :child_block_interval,
```
InoMurko: expanding the process state struct

apps/omg/test/fixtures.exs
```diff
@@ -14,12 +14,12 @@
 
 defmodule OMG.Fixtures do
   use ExUnitFixtures.FixtureModule
+  import OMG.TestHelper
```
InoMurko: following the pattern where imports and macros are at the top, aliases below

apps/omg/test/fixtures.exs
```diff
@@ -14,12 +14,12 @@
...
 
+  alias OMG.Eth.Configuration
   alias OMG.State.Core
 
-  import OMG.TestHelper
-
-  @eth OMG.Eth.RootChain.eth_pseudo_address()
+  @eth OMG.Eth.zero_address()
```
InoMurko: `zero_address` was defined twice, once in RootChain module and the second time in OMG.Eth

apps/omg/test/omg/state/persistence_test.exs
```diff
@@ -43,8 +42,20 @@ defmodule OMG.State.PersistenceTest do
...
+    metrics_collection_interval = 60_000
+
+    {:ok, _} =
+      Supervisor.start_link(
+        [
+          {OMG.State,
+           [
+             fee_claimer_address: @fee_claimer_address,
+             child_block_interval: @interval,
+             metrics_collection_interval: metrics_collection_interval
```
InoMurko: two more init arguments for state

apps/omg/test/omg/state/transaction/recovered_test.exs
```diff
@@ -133,12 +133,19 @@ defmodule OMG.State.Transaction.RecoveredTest do
...
+                   outputs,
+                   0,
+                   <<0::256>>
+                 ])
                )
 
       assert {:error, :malformed_witnesses} ==
                Transaction.Recovered.recover_from(
-                 ExRLP.encode([<<1>>, @payment_tx_type, inputs, outputs, 0, <<0::256>>])
+                 ExRLP.encode([ExPlasma.payment_v1(), @payment_tx_type, inputs, outputs, 0, <<0::256>>])
```
InoMurko: ex plasma is now the holder of payment types

InoMurko: because of that, we unify the access to this identifier

apps/omg/test/support/integration/deposit_helper.ex
```diff
@@ -39,7 +39,7 @@ defmodule Support.Integration.DepositHelper do
   end
 
   def deposit_to_child_chain(to, value, token_addr) when is_binary(token_addr) and byte_size(token_addr) == 20 do
-    contract_addr = Config.maybe_fetch_addr!(nil, :erc20_vault)
```
InoMurko: `maybe_fetch_addr ` is one of the things that I'm most happy we could remove at this point.
snapshots provide as a consistent world view of contract addresses, no need for parametrization anymore

apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
```diff
@@ -76,46 +77,48 @@ defmodule OMG.ChildChain.BlockQueue do
   """
   def handle_continue(:setup, args) do
     _ = Logger.info("Starting #{__MODULE__} service.")
-    :ok = Eth.node_ready()
-
-    {:ok, parent_start} = Eth.RootChain.get_root_deployment_height()
+    finality_threshold = Keyword.fetch!(args, :submission_finality_margin)
```
InoMurko: block queue init params got cleaned up, so that they're not fetched from the init, if they can be passed into the process

apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
```diff
@@ -76,46 +77,48 @@ defmodule OMG.ChildChain.BlockQueue do
...
-             parent_height: parent_height,
-             child_block_interval: child_block_interval,
-             block_submit_every_nth: block_submit_every_nth,
-             finality_threshold: finality_threshold
-           ) do
+      case core do
         {:ok, _state} = result ->
           result
 
-        {:error, reason} = error when reason in [:mined_hash_not_found_in_db, :contract_ahead_of_db] ->
```
InoMurko: fixed a bug here, `mined_hash_not_found_in_db ` this atom doesn't exist in core

apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
```diff
@@ -125,9 +128,6 @@ defmodule OMG.ChildChain.BlockQueue do
             )
 
           error
-
-        other ->
```
InoMurko: we don't want to catch this, we know what we can handle (or log) thats it.

apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
```diff
@@ -139,7 +139,7 @@ defmodule OMG.ChildChain.BlockQueue do
     {:ok, _} = :timer.send_interval(metrics_collection_interval, self(), :send_metrics)
 
     _ = Logger.info("Started #{inspect(__MODULE__)}")
-    {:noreply, %Core{} = state}
```
InoMurko: no need to prove ourselves what type the state is

apps/omg_child_chain/lib/omg_child_chain/supervisor.ex
```diff
@@ -33,13 +36,19 @@ defmodule OMG.ChildChain.Supervisor do
   end
 
   def init(:ok) do
-    # prevent booting if contracts are not ready
-    :ok = RootChain.contract_ready()
+    :ok = Client.node_ready()
```
InoMurko: we couldn't have started if contracts were not ready because of `SetContract` release task

apps/omg_conformance/test/support/conformance/merkle_proofs.ex
```diff
@@ -49,8 +50,25 @@ defmodule Support.Conformance.MerkleProofs do
     # only geth is supported for the merkle proof conformance tests for now
     :geth = Application.fetch_env!(:omg_eth, :eth_node)
 
-    # revert from `Eth.call_contract` it returns something resembling a reason
+    # revert from `call_contract` it returns something resembling a reason
     # binary (beginning with 4-byte function selector). We need to assume that this is in fact a revert
     assert <<0::size(28)-unit(8)>> = binary_part(chopped_reason_binary_result, 4, 28)
   end
+
+  defp call_contract(contract, signature, args, return_types) do
```
InoMurko: this is the only use case for `"checkMembership(bytes,uint256,bytes32,bytes)"` contract call.because of this, it does not belong in production code.

apps/omg_conformance/test/support/conformance/signatures_hashes.ex
```diff
@@ -106,8 +107,9 @@ defmodule Support.Conformance.SignaturesHashes do
   # NOTE: `solidity_hash/2` returns something like `{:ok, <<8, 195, 121, 160, 0, 0, 0, more zeroes...>>}`, on contract
   #       revert, when using `:geth` Ethereum node. If an assertion fails with such a result, it indicates the contract
   #       rejected some transaction to signhash unexpectedly.
-  defp solidity_hash(encoded_tx, contract) when is_binary(encoded_tx),
-    do: Eth.call_contract(contract, "hashTx(address,bytes)", [contract, encoded_tx], [{:bytes, 32}])
+  defp solidity_hash(encoded_tx, contract) when is_binary(encoded_tx) do
+    call_contract(contract, "hashTx(address,bytes)", [contract, encoded_tx], [{:bytes, 32}])
```
InoMurko: similarly as above, this is the only use case for calling hashTx, hence, we don't want to put this in the deliverable code

apps/omg_eth/lib/eth.ex
```diff
@@ -85,35 +55,12 @@ defmodule OMG.Eth do
...
-          RootChain.optional_address_t()
-        ) ::
-          {:error, binary() | atom() | map()} | {:ok, binary()}
-  def submit_block(hash, nonce, gas_price, from \\ nil, contract \\ %{}) do
-    contract = Config.maybe_fetch_addr!(contract, :plasma_framework)
-    from = from || from_hex(Application.fetch_env!(:omg_eth, :authority_addr))
-    backend = Application.fetch_env!(:omg_eth, :eth_node)
+  @spec submit_block(binary(), pos_integer(), pos_integer()) ::
+          {:error, binary() | atom() | map()} | {:ok, <<_::256>>}
+  def submit_block(hash, nonce, gas_price) do
```
InoMurko: removing all the optional paramters here, unifying access to contracts and configuration

apps/omg_eth/lib/omg_eth/client.ex
```diff
@@ -0,0 +1,41 @@
...
+        {:ok, Encoding.int_from_hex(height_hex)}
+
+      other ->
+        other
+    end
+  end
+
+  @spec node_ready() :: :ok | {:error, :geth_still_syncing | :geth_not_listening}
+  @spec node_ready(module()) :: :ok | {:error, :geth_still_syncing | :geth_not_listening}
+  def node_ready(mock \\ Ethereumex.HttpClient) do
```
InoMurko: providing a mock param, because I can test better

apps/omg_eth/lib/omg_eth/release_tasks/set_contract.ex
```diff
@@ -34,89 +36,71 @@ defmodule OMG.Eth.ReleaseTasks.SetContract do
...
       contract_semver,
-      network
+      network,
+      child_block_interval
     )
+
+    (logger_apps ++ ethereumex_apps) |> Enum.reverse() |> Enum.each(&Application.stop/1)
+  end
+
+  defp get_external_data(plasma_framework) do
```
InoMurko: SetContract release task needed a refactor + I've expanded it to retrieve all contract data from plasma contract

apps/omg_eth/lib/omg_eth/release_tasks/set_contract.ex
```diff
@@ -168,4 +175,8 @@ defmodule OMG.Eth.ReleaseTasks.SetContract do
         exit(@error)
     end
   end
+
+  defp rpc() do
+    Application.get_env(:omg_eth, :rpc_api, Rpc)
```
InoMurko: ability to mock, for better testing

apps/omg_eth/lib/omg_eth/root_chain/abi.ex
```diff
@@ -29,16 +35,17 @@ defmodule OMG.Eth.RootChain.Abi do
       end)
 
     {function_spec, data} = ABI.find_and_decode(function_specs, enriched_data)
+    decode_function_call_result(function_spec, data)
```
InoMurko: this needed a slightly different approach to the calls we're making from Ethereum Aggregator, hence the refactor.
the returned data from events or calls already contains the data that allows us to identify which function selector we should use to decode the data.
for function calls this isn't so, so we expanded decode_function to `decode_function/2`

apps/omg_eth/lib/omg_eth/root_chain/abi_function_selector.ex
```diff
@@ -82,4 +82,103 @@ defmodule OMG.Eth.RootChain.AbiFunctionSelector do
...
+
+  # min_exit_period/0, get_version/0, exit_games/0, vaults/0 are
+  # victims of unfortinate bug: https://github.com/poanetwork/ex_abi/issues/25
+  # All these selectors were intially pulled in with
+  # `ABI.parse_specification(contract_abi_json_decoded,include_events?: true)`
+  # and later modified so that `types` hold what `returns` should have because of
+  # issue 25.
+  # the commented properties of the struct is what it was generated,
+  # the new types were added to mitigate the bug.
+  def min_exit_period() do
```
InoMurko: these selectors are our API to contracts

apps/omg_eth/test/fixtures.exs
```diff
@@ -100,8 +102,25 @@ defmodule OMG.Eth.Fixtures do
     :ok
   end
 
-  # Hexifies the entire contract map, assuming `contract_map` is a map of `%{atom => raw_binary_address}`
+  defp has_exit_queue(vault_id, token) do
+    plasma_framework = Configuration.contracts().plasma_framework
+    token = Encoding.from_hex(token)
+    call_contract(plasma_framework, "hasExitQueue(uint256,address)", [vault_id, token], [:bool])
```
InoMurko: this code also has no use case in deliverable, hence it's in tests only

InoMurko: the code was originally in `RootChainHelper ` but used now where else.

apps/omg_performance/lib/omg_performance/byzantine_events.ex
```diff
@@ -217,7 +217,6 @@ defmodule OMG.Performance.ByzantineEvents do
   defp map_contract_transaction(enumberable, transaction_function) do
     enumberable
     |> Enum.map(transaction_function)
-    # NOTE: infinity doesn't work, hence the large number
```
InoMurko: ... well... I too would like an infinite CPU machine but all I got is this lousy T-SHIRT?

apps/omg_performance/lib/performance.ex
```diff
@@ -93,12 +91,6 @@ defmodule OMG.Performance do
     :ok = Application.put_env(:ethereumex, :request_timeout, :infinity)
     :ok = Application.put_env(:ethereumex, :http_options, recv_timeout: :infinity)
     :ok = Application.put_env(:ethereumex, :url, opts[:ethereum_rpc_url])
-
-    :ok =
-      if opts[:contract_addr],
-        do: Application.put_env(:omg_eth, :contract_addr, contract_map_to_hex(opts[:contract_addr])),
```
InoMurko: I have no idea what was going on here... SetContracts release task takes care of this now

apps/omg_watcher/config/test.exs
```diff
@@ -21,5 +21,3 @@ config :omg_watcher,
 config :omg_watcher, OMG.Watcher.Tracer,
   disabled?: true,
   env: "test"
-
-config :omg_watcher, environment: :test
```
InoMurko: not used

apps/omg_watcher/lib/omg_watcher/api/status.ex
```diff
@@ -51,17 +55,20 @@ defmodule OMG.Watcher.API.Status do
   def get_status() do
```
InoMurko: in a better state for a much larger performance refactor

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -56,50 +56,47 @@ defmodule OMG.Watcher.BlockGetter do
...
   alias OMG.Watcher.HttpRPC.Client
 
-  def start_link(_args) do
-    GenServer.start_link(__MODULE__, :ok, name: __MODULE__)
-  end
-
   @doc """
   Retrieves the freshest information about `OMG.Watcher.BlockGetter`'s status, as stored by the slave process `Status`.
   """
   @spec get_events() :: {:ok, Core.chain_ok_response_t()}
```
InoMurko: api above the callbacks pattern

apps/omg_watcher/test/omg_watcher/integration/block_getter_test.exs
```diff
@@ -133,11 +131,18 @@ defmodule OMG.Watcher.Integration.BlockGetterTest do
     block_with_incorrect_hash = %{OMG.Block.hashed_txs_at([], 1000) | hash: different_hash}
 
     # from now on the child chain server is broken until end of test
-    BadChildChainServer.prepare_route_to_inject_bad_block(
```
InoMurko: the setting was forced via shared state (application setting) which is overall discouraged approach

mix.lock
```diff
@@ -12,18 +12,18 @@
...
   "decimal": {:hex, :decimal, "1.8.1", "a4ef3f5f3428bdbc0d35374029ffcf4ede8533536fa79896dd450168d9acdf3c", [:mix], [], "hexpm", "3cb154b00225ac687f6cbd4acc4b7960027c757a5152b369923ead9ddbca7aec"},
   "decorator": {:hex, :decorator, "1.3.2", "63b8ac9e23b28053390abdda33bb9e1f3dd9e8f9a981f47a06fc2f2fe2e2f772", [:mix], [], "hexpm", "b80bd089e3c8579e6d9ea84eed307b1597a0d94af25331e424a209477ad1a7fc"},
-  "dialyxir": {:hex, :dialyxir, "1.0.0-rc.7", "6287f8f2cb45df8584317a4be1075b8c9b8a69de8eeb82b4d9e6c761cf2664cd", [:mix], [{:erlex, ">= 0.2.5", [hex: :erlex, repo: "hexpm", optional: false]}], "hexpm", "506294d6c543e4e5282d4852aead19ace8a35bedeb043f9256a06a6336827122"},
+  "dialyxir": {:hex, :dialyxir, "1.0.0", "6a1fa629f7881a9f5aaf3a78f094b2a51a0357c843871b8bc98824e7342d00a5", [:mix], [{:erlex, ">= 0.2.6", [hex: :erlex, repo: "hexpm", optional: false]}], "hexpm", "aeb06588145fac14ca08d8061a142d52753dbc2cf7f0d00fc1013f53f8654654"},
```
InoMurko: stable dialyzer

apps/omg_child_chain/lib/omg_child_chain/supervisor.ex
```diff
@@ -36,7 +36,6 @@ defmodule OMG.ChildChain.Supervisor do
   end
 
   def init(:ok) do
-    :ok = Client.node_ready()
```
InoMurko: we're already talking to plasma framework at this point (setcontract)

apps/omg_eth/test/omg_eth/client_test.exs
```diff
@@ -0,0 +1,59 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.Eth.ClientTest do
+  use ExUnit.Case, async: true
+  alias OMG.Eth.Client
+
+  test "get_ethereum_height/0 returns the block number", %{test: test_name} do
```
unnawut: Hmm what's this `%{test: test_name}` sorcery?

apps/omg_watcher/test/test_helper.exs
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ExUnit.configure(exclude: [common: true, integration: true, property: true, wrappers: true])
+ExUnit.configure(exclude: [watcher: true, common: true, integration: true, property: true, wrappers: true])
```
unnawut: watcher doesn't test watcher?

apps/omg_child_chain/lib/omg_child_chain/supervisor.ex
```diff
@@ -33,13 +35,18 @@ defmodule OMG.ChildChain.Supervisor do
   end
 
   def init(:ok) do
-    # prevent booting if contracts are not ready
-    :ok = RootChain.contract_ready()
     {:ok, contract_deployment_height} = RootChain.get_root_deployment_height()
+    metrics_collection_interval = OMG.ChildChain.Configuration.metrics_collection_interval()
     fee_claimer_address = OMG.Configuration.fee_claimer_address()
+    child_block_interval = Configuration.child_block_interval()
```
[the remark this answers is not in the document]
InoMurko: ah yes! OK, so configuration local to the application we're in is aliased, others are not! cool idea

apps/omg_eth/test/omg_eth/client_test.exs
InoMurko: this is basically ExUnit state

apps/omg_watcher/test/test_helper.exs
InoMurko: these are tags for tests. so for example, when you run tests:
`mix test` you don't run tests that are tagged: `[watcher: true, common: true, integration: true, property: true, wrappers: true]`

If we look at our circleci config, it looks similar to this:
```
test:
	mix test --include test --exclude common --exclude watcher --exclude watcher_info --exclude child_chain

test-watcher:
	mix test --include watcher --exclude watcher_info --exclude child_chain --exclude common --exclude test

test-watcher_info:
	mix test --include watcher_info --exclude watcher --exclude child_chain --exclude common --exclude test

test-common:
	mix test --include common --exclude child_chain --exclude watcher --exclude watcher_info --exclude test

test-child_chain:
	mix test --include child_chain --exclude common --exclude watcher  --exclude watcher_info --exclude test
```

so based on the parallel job, it's excluding some and including one specific tag. PLS validate my statements.

InoMurko:
> Also I think we should cut a new stable version before this PR gets merged. Then this PR could go in the next-next release so we have less pressure on time to test this on dev/staging.
> 
> I'll keep this "changes requested" status until then but lemme know if anyone has a differing opinion.

I completely agree. The diff is already big enough in master as is.

apps/omg_watcher/lib/omg_watcher/api/status.ex
```diff
@@ -51,17 +55,20 @@ defmodule OMG.Watcher.API.Status do
...
     {:ok, eth_block_number} = EthereumHeight.get()
     {:ok, eth_block_timestamp} = Eth.get_block_timestamp_by_number(eth_block_number)
-    eth_syncing = Eth.syncing?()
+    eth_syncing = syncing?()
 
     validated_child_block_number = get_validated_child_block_number()
+    # wtf is eth diagnostics?
+    contracts = Eth.Diagnostics.get_child_chain_config()[:contract_addr]
```
unnawut:
```suggestion
    contracts = OMG.Eth.Configuration.contracts()
```

Should work right? Then `Eth.Diagnostics.get_child_chain_config/0` can be dropped.

And `OMG.Eth.Diagnostics.get_node_diagnostics/0` moved into BlockQueue the only place where it's used.

apps/omg_status/test/omg_status/alert/alarm_printer_test.exs
```diff
@@ -25,68 +27,48 @@ defmodule OMG.Status.Alert.AlarmPrinterTest do
   end
 
   test "if the process has a previous backoff set", %{alarm_printer: alarm_printer} do
-    :erlang.trace(alarm_printer, true, [:receive])
-    %{previous_backoff: previous_backoff} = :sys.get_state(alarm_printer)
-    assert is_number(previous_backoff)
+    assert capture_log(fn ->
```
unnawut: My ears feel lightened up 🙉

apps/omg_status/test/omg_status/alert/alarm_printer_test.exs
```diff
@@ -25,68 +27,48 @@ defmodule OMG.Status.Alert.AlarmPrinterTest do
...
-    Enum.each(1..5, fn _ ->
-      %{previous_backoff: backoff} = :sys.get_state(alarm_printer)
-      :print_alarms = send(alarm_printer, :print_alarms)
-      min = backoff
-      max = backoff * 3
-      %{previous_backoff: new_backoff} = :sys.get_state(alarm_printer)
-      assert min < new_backoff and new_backoff < max
-    end)
-  end
+             assert_receive {:trace, _, :send, {:notify, {:warn, _, {Logger, "An alarm was raised 1", {_, _}, _}}},
```
unnawut: Epic refactor 🎖

apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
```diff
@@ -76,46 +77,48 @@ defmodule OMG.ChildChain.BlockQueue do
   """
   def handle_continue(:setup, args) do
     _ = Logger.info("Starting #{__MODULE__} service.")
-    :ok = Eth.node_ready()
```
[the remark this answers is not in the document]
InoMurko: I think this is actually a valid concern - I think. The freakish part is that not a single test caught this. I'll put it back in.

What I'm not sure is ... should we have this check in multiple places? What if the eth client starts falling behind?

apps/omg_eth/lib/omg_eth/client.ex
InoMurko: yeah, very bad naming on my part lol

apps/omg_eth/lib/omg_eth/root_chain.ex
```diff
@@ -21,65 +21,52 @@ defmodule OMG.Eth.RootChain do
...
+    %{"block_number" => mined_num} = get_external_data(contract_address, "nextChildBlock()", [])
+    mined_num
+  end
 
-  @doc """
-  This is what the contract understands as the address of native Ether token
-  """
-  @spec eth_pseudo_address() :: <<_::160>>
-  def eth_pseudo_address(), do: Eth.zero_address()
+  def blocks(mined_num) do
```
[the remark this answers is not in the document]
InoMurko: hm hm, the reason I kept this naming is because `blocks/1` and `next_child_block/0` are direct calls to plasma contracts, and return the result "unmodified".
where as `get_mined_child_block /0` is calculated as a diff between the next child block and the leap (interval).
so... I was like, these are not the same in principal maybe I shouldn't name them the same. IDK... should I change?

apps/omg_eth/lib/omg_eth/root_chain/abi_function_selector.ex
InoMurko: I think the long term plan is to do the following:
`ex_plasma - is a thin protocol specification for transaction construction and validation.`

`ex_plasma-contracts - is a thin API for plasma-contract exposed functions and events.` 
But to achieve that, we need to work on ExABI a little bit, hopefully that clears in the following months. 
At the moment, the amount of bugs ExABI has, it would be of little value to put it in a separate library. 
Also, ex_plasma_contracts is made up name. Not sure yet how we should call it.

apps/omg_eth/test/fixtures.exs
```diff
@@ -43,47 +44,37 @@ defmodule OMG.Eth.Fixtures do
...
     token_addr
   end
 
   deffixture root_chain_contract_config(contract) do
-    contract_addr = contract_map_to_hex(contract.contract_addr)
-    Application.put_env(:omg_eth, :contract_addr, contract_addr, persistent: true)
-    Application.put_env(:omg_eth, :authority_addr, Encoding.to_hex(contract.authority_addr), persistent: true)
-    Application.put_env(:omg_eth, :txhash_contract, Encoding.to_hex(contract.txhash_contract), persistent: true)
-
+    _ = contract
```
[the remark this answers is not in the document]
InoMurko: yeah, otherwise the contract variable is `unused`, but the way these fixtures work is that they're chained together (and the order) based on the arguments between them. it's funny I tell you.

apps/omg_performance/test/omg_performance/extended_perftest_test.exs
```diff
@@ -28,8 +28,8 @@ defmodule OMG.Performance.ExtendedPerftestTest do
 
   # NOTE: still bound to fixtures :(, because of the child chain setup, but this will go eventually, so leaving as is
   deffixture perf_test(contract) do
-    %{contract_addr: contract_addr} = contract
-    :ok = Performance.init(contract_addr: contract_addr)
+    _ = contract
```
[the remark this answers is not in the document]
InoMurko: yep, it's my fav line!

apps/omg_status/test/omg_status/alert/alarm_printer_test.exs
InoMurko: yeah, sometimes you need to step away and think again what you did 3 month ago

apps/omg_watcher/lib/omg_watcher/api/status.ex
InoMurko: yeah, removed that and a couple of others

apps/omg_eth/test/fixtures.exs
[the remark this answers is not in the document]
unnawut: Ah i see.......................................

apps/omg_eth/lib/omg_eth/root_chain.ex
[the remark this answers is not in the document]
unnawut: >  these are not the same in principal maybe I shouldn't name them the same

I'm kinda inclined towards that being the underlying implementation that the function user doesn't need to know. My first reaction to this was "is this function changing something?", the `get_*` just makes it clear it's a getter and aligns with other module APIs e.g. Map, Ecto, Keyword, etc.

Subjective though, might just need a 3rd opinion @achiurizo

apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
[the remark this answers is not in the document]
unnawut: > should we have this check in multiple places? What if the eth client starts falling behind?

I think the check for eth client starts falling behind will be useful.

Apart from that, not sure where else do we need this though, we're not submitting tx anywhere else.

apps/omg_eth/lib/omg_eth/root_chain.ex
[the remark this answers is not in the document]
achiurizo: I think I started with `get_*` names initially but I found it harder to keep track of the different namings we had between elixir and the contracts. I'm leaning towards we match elixir namings to the contract function names(same with the variable names), so I'm for `def blocks`.

That means we'll need to think about some sort of standard for other functions around it, like the `get_mined_*` being calculated.  I think `get_*` could work well.

PR #1406, opened 2020-03-18
Implementing Code Owners for mix.lock

okalouti:
Addressing issue #1173 

## Overview

Implementing code owners for `mix.lock`. If any changes to this file are made, the listed individuals will be automatically requested for review. 

Alongside this PR, plan is to configure the repo to require code owners' review, which means at least **one** of the listed code owners will have to approve. @T-Dnzt 

Read more [here](https://github.blog/2017-07-06-introducing-code-owners/#an-extra-layer-of-code-security)

## Changes

Create `CODEOWNERS` file with listed individuals. 

## Testing

Created dummy PRs to validate that review requests go automatically to listed individuals if `mix.lock` is changed, even if the committed changes in question are made _after_ the PR is opened.

Note: while `CODEOWNERS` can be implemented for the rest of the repository, intentionally limiting the scope to `mix.lock` for now after some discussion.


boolafish:
Curious, if a dependency is changed, how do we expect the reviewer to review the dependency?

InoMurko:
> Curious, if a dependency is changed, how do we expect the reviewer to review the dependency?
One way to go is:
https://hex.pm/blog/announcing-hex-diff

It's not so much review as it's vetoing being "safe".

PR #1416, opened 2020-03-23
Prevent split/merge creation in /transaction.create

T-Dnzt:
## Overview

Since the introduction of fees (and the overpaying error), merges cannot be created properly via `/transaction.create`. Since merges are free, but the `/transaction.create` generates transactions with fees included, it's currently returning paid merges which are rejected by the childchain.

`/transation.create` is currently returning invalid txs to the users.

To counter that, we thought about a few options, like trying to detect merges and not setting fees for them. Unfortunately, due to the way this endpoint works (only specifying outputs and selecting UTXOs for inputs), it's really hard to deduce if the user wants to do a merge or a split. 

Therefore, we've decided to prevent the creation of merges and splits via `/transaction.create` entirely, simply by checking that the `owner` address and all the output addresses are the same. 

## Changes

- Return a new error (`transaction.create:self_transaction_not_supported`) if a user attempts to generate a merge or a split via `/transaction.create`.

## Testing

/

### DIFF
--- a/apps/omg_watcher_rpc/lib/web/validators/order.ex
+++ b/apps/omg_watcher_rpc/lib/web/validators/order.ex
@@ -34,7 +34,8 @@ defmodule OMG.WatcherRPC.Web.Validator.Order do
          {:ok, metadata} <- expect(params, "metadata", [:hash, :optional]),
          {:ok, fee} <- expect(params, "fee", map: &parse_fee/1),
          {:ok, payments} <- expect(params, "payments", list: &parse_payment/1),
-         {:ok, payments} <- fills_in_outputs?(payments) do
+         {:ok, payments} <- fills_in_outputs?(payments),
+         :ok <- ensure_not_self_transaction(owner, payments) do
       {:ok,
        %{
          owner: owner,
@@ -44,6 +45,19 @@ defmodule OMG.WatcherRPC.Web.Validator.Order do
        }}
     end
   end
+
+  defp ensure_not_self_transaction(owner, payments) when length(payments) > 0 do
+    payments
+    |> Enum.any?(fn payment ->
+      owner != payment[:owner]
+    end)
+    |> handle_self_tx_result()
+  end
+
+  defp ensure_not_self_transaction(_, _), do: :ok
+
+  defp handle_self_tx_result(true), do: :ok
+  defp handle_self_tx_result(false), do: {:error, :self_transaction_not_supported}
 
   defp fills_in_outputs?(payments) do
     if length(payments) <= Transaction.Payment.max_outputs(),
--- a/apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
+++ b/apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
@@ -1295,6 +1295,24 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
                )
     end
 
+    @tag fixtures: [:alice, :more_utxos]
+    test "returns an error when requester is equal to all the ouputs owner", %{alice: alice} do
+      params = %{
+        "owner" => Encoding.to_hex(alice.addr),
+        "payments" => [
+          %{"amount" => 1, "currency" => @eth_hex, "owner" => Encoding.to_hex(alice.addr)},
+          %{"amount" => 1, "currency" => @eth_hex, "owner" => Encoding.to_hex(alice.addr)}
+        ],
+        "fee" => %{"currency" => @default_fee_currency}
+      }
+
+      assert %{
+               "object" => "error",
+               "code" => "transaction.create:self_transaction_not_supported",
+               "description" => "This endpoint cannot be used to build merge or split transactions."
+             } == WatcherHelper.no_success?("transaction.create", params)
+    end
+
     defp balance_in_token(address, token) do
       currency = Encoding.to_hex(token)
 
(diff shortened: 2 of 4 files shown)


apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -1230,6 +1230,24 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
                )
     end
 
+    @tag fixtures: [:alice, :more_utxos]
```
InoMurko: do you need these fixtures to test this endpoint? it seems like it stateless (as long as watcher is running)?

do you need to have the same validation on the childchain as well?

T-Dnzt: Valid point on the fixtures, I'll try to remove them. Nothing to do on the Childchain side, the `/transaction.create` is a helper endpoint only present in the Watcher. The childchain properly detects merges because it has the list of inputs / ouputs.

apps/omg_watcher_rpc/lib/web/validators/order.ex
```diff
@@ -45,6 +46,19 @@ defmodule OMG.WatcherRPC.Web.Validator.Order do
...
 
+  defp ensure_not_self_transaction(owner, payments) when length(payments) > 0 do
+    payments
+    |> Enum.any?(fn payment ->
+      owner != payment[:owner]
+    end)
+    |> handle_self_tx_result()
+  end
+
+  defp ensure_not_self_transaction(_, _), do: :ok
```
boolafish: curious, when/how would this get hit?

T-Dnzt: So it's actually possible to have an empty list of payments at that point, as demonstrated by the two tests below:

- https://github.com/omisego/elixir-omg/blob/534ed37f61b4ad4722284c01f7343e0d8a5e579b/apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs#L1171-L1201
- https://github.com/omisego/elixir-omg/blob/534ed37f61b4ad4722284c01f7343e0d8a5e579b/apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs#L1204-L1231

I'm not sure why we don't validate it earlier (and I'm kinda afraid of changing it), but it gets checked later in the UTXO selection, so I had to handle that with this clause. :/

InoMurko: hmmm `empty transaction without payments list is not allowed` seems like if you had a clause `ensure_not_empty`, this would cover it before it hits OMG.State, right?

boolafish: interesting. So the idea is to allow it to bypass the check here and let it fail later. I see.

T-Dnzt: @InoMurko yep, that's right. I don't see a reason why we couldn't validate it early in `order.ex`, but I'm afraid I'm lacking some context. Maybe we could create an issue about it and get it fixed in another PR?

InoMurko: Sure! Can you open it?

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
T-Dnzt: So it all breaks if I remove the fixtures, unfortunately. I'm gonna qualify the work to figure out why this test fails without fixtures as out of scope for this PR. 

However, since we're not going to merge this right away, I'll try to revisit it once I've been through some high priorities tasks.

apps/omg_watcher_rpc/lib/web/validators/order.ex
T-Dnzt: Issue created: https://github.com/omisego/elixir-omg/issues/1419

kevsul:
Hey, sorry I'm late to this.

Is it really so hard to detect a merge tx?
 - all inputs have the same owner
 - all inputs have the same currency
 - inputs.length > 1
 - outputs.length == 0

Or am I missing something?

In any case, the main reason that `transaction.create` exists is to provide a way for users to create transactions without having to use a client side library (like omg-js). So you can make a plasma app in whatever language you like by using just the http api.

I'm concerned that removing useful functionality like this means a change of direction towards requiring a client side lib for various languages, with all the maintenance overhead that entails.

Was this already discussed somewhere?

boolafish:
> Is it really so hard to detect a merge tx

No, it is easy. The 116 message reasoning of what is hard is [in slack](https://omisego.slack.com/archives/CPXLGCUQK/p1583811865135200) 😅

TL;DR the hard part is this API to be consistently generate a submittable transaction. A user could have utxos that can generate. eg. A user have utxos: (100, 25, 25, 25, 25)
1. non merging transactions, eg. [(100) -> (30, 70)]
1. merging transactions eg. [(25, 25, 25, 25) -> (30, 70)]

If we simply generate a tx and check is that tx a merge, we might fail a user that tries to create a transaction and the user indeed can create a transaction without being a merge tx. Meanwhile, the input amount needed with merge and non-merge tx is different since we need to add fee in a non-merging tx.

kevsul:
Ah ok, I missed that whole Slack thread 😅 

I'm fine with moving the merge complexity out of `transaction.create` an into a separate `transaction.merge` endpoint.  That way we can still support thin clients.

Sorry for the interruption, carry on 😄

PR #1424, opened 2020-03-25
Filtering Input Parameters to Childchain/Watcher API depending on HTTP Method

okalouti:
Addresses #1411 

## Overview

This PR implements filtering for input parameters depending on the HTTP method used. 
- For a POST: `query_params` will be ignored and `body_params` will be used. 
- For a GET: `body_params` will be ignored and `query_params` will be used. 

## Changes
Creates  plug in `method_param_filter.ex` files which are added to `endpoint.ex`  in in both `omg_childchain_rpc` and `omg_watcher_rpc` to implement above logic. 

## Testing

- Tests written for `method_param_filter.ex` (two files: one for childchain and one for watcher) 
- Tests added to selected endpoints on Watcher Info (`/account.get_balance` and `/transaction.get`) and Child Chain (`/block.get`) to test that passing passing query parameters will return a bad request error. 

_For the second set of tests: while these do the job, iI could not think of how to target tests to `endpoint.ex`. Any insights welcome. Note that we have no `GET` methods that take input parameters, so no tests written as above. 

```elixir
mix test test/omg_child_chain_rpc/web/controllers/block_test.exs
mix test test/omg_watcher_rpc/web/controllers/block_test.exs
mix test test/omg_watcher_rpc/web/controllers/transaction_test.exs
mix test test/omg_child_chain_rpc/web/plugs/method_param_filter_test.exs
mix test test/omg_watcher_rpc/web/plugs/method_param_filter_test.exs
```

Thanks to @mederic-p who implemented the logic right before his baby was born. I just added a few tests.


okalouti:
Hi all, I just had a chat with @thec00n and there are a few things where we could use more experienced eyes: 

In `endpoint.ex`: 
```
    parsers: [:urlencoded, :multipart, :json],
    pass: ["*/*"],
```
@mederic-p changed to (child chain): 
```
    parsers: [:urlencoded, :json],
    pass: [],
```
@mederic-p changed to (watcher): 
```
   parsers: [:json],
   pass: [],
``

It looks fine in the sense that we want the endpoints to only process valid JSON, but I'm not sure if there's something else to consider here, and why `:urlencoded` stays in the child chain `endpoint.ex` file.

File: apps/omg_watcher_rpc/lib/web/endpoint.ex [hunk 1]
```diff
@@ -19,18 +19,20 @@ defmodule OMG.WatcherRPC.Web.Endpoint do
   plug(Plug.RequestId)
   plug(Plug.Logger, log: :debug)
 
+  if code_reloading? do
+    plug(Phoenix.CodeReloader)
+  end
+
   plug(
```
thec00n: Could we add a `: query_string_length` with a reasonable value (see. https://hexdocs.pm/plug/Plug.Parsers.html#module-options)?

InoMurko: Since we're using a mish mush API (something between HTTP RPC and JSON-RPC) we can actually disable accepting `query_string_length `, or set it to 0. None of our APIs is accepting or expecting the params to come in the URL.

thec00n: I this case couldn't we just not copy `query_params` to `params` in `OMG.ChildChainRPC.Web.Plugs.MethodParamFilter` ?

InoMurko:
@unnawut pls review!

okalouti:
yep sorry mistake, intended to push only changes to endpoint.

okalouti:
Thanks to @InoMurko for that one!

Issue #1428, opened 2020-03-25
CI is failing on master branch


okalouti:
cc: @unnawut 

https://circleci.com/gh/omisego/elixir-omg/tree/master

pgebal:
I am on that @okalouti.

InoMurko:
@okalouti @pgebal pls elaborate what both of you are looking at/reporting. 
The link is showing failing nightly builds and failing `run_load_test_on_deployed_env`. I think @pgebal is taking `run_load_test_on_deployed_env `. 
@okalouti are you looking at nightly builds failing?

pgebal:
Sorry for not being precise. I am taking `run_load_test_on_deployed_env` @okalouti

okalouti:
Hey @InoMurko. I was pointing to the nightly builds failing, yes.

InoMurko:
cool @okalouti ! there were some attempts already: https://github.com/omisego/elixir-omg/pull/1287

pgebal:
The part with failed load tests is solved. We decided to switch them off until we come up with a good way on how to run them for `build_test_deploy` CI pipeline.

InoMurko:
macos barebuild sometimes fails, I;ll look into that

PR #1429, opened 2020-03-25
test: do not run load tests on deployed dev env

pgebal:
References: https://github.com/omisego/elixir-omg/issues/1428

## Overview

As currently tests of load tests do not measure performance drops and are just smoke tests of load test scenarios functionality I removed running them from CI job.

## Changes

Removes running smoke load tests on `build_test_deploy` CI job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -629,32 +629,6 @@ jobs:
       - checkout
       - run: DEPLOY=watcher_info sh .circleci/ci_deploy.sh
 
-  run_load_test_on_deployed_env:
-    machine:
-      image: ubuntu-1604:201903-01
-    steps:
-      - checkout
-      - run:
-          name: Install Erlang and Elixir
-          command: |
-            set -e
-            wget https://packages.erlang-solutions.com/erlang-solutions_2.0_all.deb
-            sudo killall dpkg || true &&
-            sudo rm /var/lib/dpkg/lock || true &&
-            sudo rm /var/cache/apt/archives/lock || true &&
-            sudo dpkg --configure -a || true &&
-            sudo dpkg -i erlang-solutions_2.0_all.deb  &&
-            sudo apt-get update &&
-            sudo apt-get install esl-erlang=1:21.3.8.10-1
-            sudo apt-get install elixir=1.8.2-1
-      - run: make install-hex-rebar
-      - run:
-          name: Run load test
-          command: |
-            cd priv/perf
-            make init
-            MIX_ENV=dev mix test
-
   coveralls_report:
     docker:
       - image: omisegoimages/elixir-omg-circleci:v1.8-20190129-02
@@ -853,12 +827,6 @@ workflows:
                 - master
       - deploy_watcher_info:
           requires: [publish_child_chain, publish_watcher, publish_watcher_info]
-          filters:
-            branches:
-              only:
-                - master
-      - run_load_test_on_deployed_env:
-          requires: [deploy_child_chain, deploy_watcher, deploy_watcher_info]
           filters:
             branches:
               only:


boolafish:
wait, why?

pgebal:
> wait, why?

They just do not test anything regarding deployed env at the moment. We do not collect anything, compare it with prev. runs, etc. No point in running it unless we measure performance on the dev env.

Smoke tests alone are still being run on PRs. But they just test if scenarios themselves are working.

boolafish:
we do have something. At least you can go to Datadog afterward and see what is there. It can also potentially trigger error to be catch as regression though not on performance but on load. (Though there is no hook back to flag test as failed at this point)

However, my biggest concern is that unless we are sure that we do not want to automate this as a future goal. Otherwise we need step by step code merge anyway. This is discouraging moving that path. Also, it has seen to be a SOP that we want some load in new env from what I see in recent releases/deployment.

boolafish:
btw, I am looking at the new configs today. So if we want to move forward with this and keep the style of lots of env var. Probably we can just put all configs into one to simplify the configs 🤔 

The only reason to have multiple configs was to 1:1 mapping with environments we have, which I think will get a lot more stable after the [release process](https://omisego.slack.com/archives/CTRND1351/p1585037548217700) get fixed. I personally still think that makes good sense and fit greatly into a CI/CD pipeline with multiple env. 

Meanwhile, the config file could be like `ENV_VAR || default_vale` so that we don't need to pass the env var in MakeFile. It looks weird to me 🤔. I don't really think that makes much sense actually. (the contract addresses, private key, urls in make file) I think config should be preferred unless you are looking for extra control or the data is confidential.

pgebal:
> Meanwhile, the config file could be like `ENV_VAR || default_vale` so that we don't need to pass the env var in MakeFile. It looks weird to me thinking. I don't really think that makes much sense actually. (the contract addresses, private key, urls in make file) I think config should be preferred unless you are looking for extra control or the data is confidential.

On one hand it makes a lot of sense and makes things easier for our devs. But on the other, those default values are crafted for us. If people from outside omsisego clone the repo, starts running it, those default values probably won't make sense for them.
I agree that putting those env vars in `Makefile` isn't cool, but I have no good ideas how to handle it differently.

pgebal:
> we do have something. At least you can go to Datadog afterward and see what is there. It can also potentially trigger error to be catch as regression though not on performance but on load. (Though there is no hook back to flag test as failed at this point)

I think I get your point. I agree, proper automation of running perfomance test should be done. But I don't think it's now. We have just one smoke test running that puts a couple thousands of utxos into the child chain. We don't know yet 
what that tells us about degradation in performance. You are right we can check DataDog, but is anyone doing it? To make it sustainable it should be automated.

pgebal:
> Also, it has seen to be a SOP that we want some load in new env from what I see in recent releases/deployment.

That makes sense.

boolafish:
> If people from outside omsisego clone the repo, starts running it, those default values probably won't make sense for them.

I think I messed up a bit on the paragraph so leads a bit confusion. If we are going your way and give up automation until we are perfect, my proposal is having single config with default value to local host (docker-compose). So even an open source contributor can run against the default local service. Whoever wants to run in different environments, pass in the env var.

But honestly I am not really buying "other people might run it" concept personally. I know it is controversial, but practically speaking nobody else is running beside us. And even there are, I don't see reason to not optimize for ourselves. This code is mainly for us anyway. We can make their life easier only after we have made our life easier.

> We have just one smoke test running that puts a couple thousands of utxos into the child chain

We do have other scenarios in omg-load-testing not ported yet though. 

> what that tells us about degradation in performance. You are right we can check DataDog, but is anyone doing it? To make it sustainable it should be automated.

So what is the next step? My feeling is that we would just stop here, write some more scenarios and that is it. It is fine if we agree to do that due to resource/priority reason. But I don't think that is the ideal path.

> To make it sustainable it should be automated

probably we could try to call data dog api at the end of the test for two things:
1. General Error rate. This can potentially be used as an indication to rollback the deployment.
1. Performance metric. 

Even we might not know what is a good threshold yet, it is worth to have report to learn just like the gas cost in contract repo. 🤔

But I guess this depends on what feat-launch-n-load team want to do. cc. @ unnawut @kevsul

PR #1446, opened 2020-04-03
Add load test merge_utxos()

kevsul:
## Overview

Add merge_utxos() functionality. Useful for utxo hygiene on the faucet

## Changes

- Add LoadTest.ChildChain.Utxos.merge()
- Add LoadTest.Service.Faucet.merge_utxos()

### DIFF
--- a/priv/perf/apps/load_test/lib/service/faucet.ex
+++ b/priv/perf/apps/load_test/lib/service/faucet.ex
@@ -21,6 +21,12 @@ defmodule LoadTest.Service.Faucet do
   wait for it to finalize and then use that deposit utxo for funding accounts.
 
   This means that faucet account must have sufficient funds on the root chain.
+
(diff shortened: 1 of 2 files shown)


priv/perf/apps/load_test/lib/service/faucet.ex
```diff
@@ -111,6 +129,13 @@ defmodule LoadTest.Service.Faucet do
     {:reply, {:ok, user_utxo}, updated_state}
   end
 
+  def handle_call({:merge_utxos, currency}, _from, %__MODULE__{faucet_account: faucet_account} = state) do
```
InoMurko, line 132:
```suggestion
  def handle_call({:merge_utxos, currency}, _from, state) do
```
and later use `state.faucet_account`

InoMurko: it's just makes the code less bloated!

PR #1455, opened 2020-04-07
fix: require origin for contract event consumer topic

pgebal:
Fixes a bug in setting contract event consumer id and topic.

### DIFF
--- a/apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
@@ -27,17 +27,21 @@ defmodule OMG.ChildChain.DatadogEvent.ContractEventConsumer do
   Mandatory params are in Keyword form:
   - :publisher is Module that implements a function `event/3` (title, message, option). It's purpose is to forward
   events to a collector (for example, Datadog)
-  - :event is a OMG.Bus topic from which this process recieves data and sends them to the publisher and onwards
+  - :topic is a OMG.Bus topic from which this process recieves data and sends them to the publisher and onwards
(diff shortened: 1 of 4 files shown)


apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
```diff
@@ -27,17 +27,21 @@ defmodule OMG.ChildChain.DatadogEvent.ContractEventConsumer do
...
   - :current_version is semver of the current code
   """
   # sobelow_skip ["DOS.StringToAtom"]
   @spec prepare_child(keyword()) :: %{id: atom(), start: tuple()}
   def prepare_child(opts) do
-    topic = Keyword.fetch!(opts, :topic)
+    id =
+      case Keyword.fetch!(opts, :topic) do
+        {origin, topic_name} -> "#{origin}:#{topic_name}_worker"
+        other -> "#{other}_worker"
```
InoMurko, line 40: Hmmm. are we just being defensive or is there a use case for this?

T-Dnzt: @InoMurko A string is passed for `topic` in the tests, only place I could find that would need this `case`: https://github.com/omisego/elixir-omg/blob/master/apps/omg_child_chain/test/omg_child_chain/datadog_event/contract_event_consumer_test.exs#L34

I'd say let's switch to the tuple format everywhere and remove this case if possible. If not, let us know what's the argument for keeping it @pgebal.

pgebal: @T-Dnzt @InoMurko you are right, for contract event consumer we always know the origin, hece there should be no need for the second case.

PR #1460, opened 2020-04-08
fix: fix logging in contract event consumer

pgebal:
Fixes logging in contract event consumer

### DIFF
--- a/apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
@@ -58,12 +58,12 @@ defmodule OMG.ChildChain.DatadogEvent.ContractEventConsumer do
   """
   def init(args) do
     publisher = Keyword.fetch!(args, :publisher)
-    topic = Keyword.fetch!(args, :topic)
+    {:root_chain, event_name} = topic = Keyword.fetch!(args, :topic)
     release = Keyword.fetch!(args, :release)
     current_version = Keyword.fetch!(args, :current_version)
     :ok = OMG.Bus.subscribe(topic, link: true)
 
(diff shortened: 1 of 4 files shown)


apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
```diff
@@ -58,12 +58,12 @@ defmodule OMG.ChildChain.DatadogEvent.ContractEventConsumer do
...
   def init(args) do
     publisher = Keyword.fetch!(args, :publisher)
-    topic = Keyword.fetch!(args, :topic)
+    {:root_chain, event_name} = topic = Keyword.fetch!(args, :topic)
     release = Keyword.fetch!(args, :release)
     current_version = Keyword.fetch!(args, :current_version)
     :ok = OMG.Bus.subscribe(topic, link: true)
 
-    _ = Logger.info("Started #{inspect(__MODULE__)} for event #{inspect(topic)}")
+    _ = Logger.info("Started #{inspect(__MODULE__)} for event #{event_name}")
```
pnowosie, line 66: I don't like the change - it removes the context. In the previous PR we extended events with `:root_chain | :child_chain` and here we just log event name - it could increase the confusion.

What's wrong with former `#{inspect(topic)}`

pgebal: This module is intended only for contract events, that's why it's always `:root_chain`. This log message suits better in this context. 
The main reason for this PR is that in one of the GenServers we didn't have inspect and it was failing when starting on dev.

pnowosie: What's obvious now won't necessary be in the month. 
- Also we specifically added `:root_chain | :child_chain` to defend against event names clashing. 
- Also, it's common to try to find logic behind just by searching the logged message.

Let's log `... root_chain:#{event_name}`

PR #1462, opened 2020-04-08
prevent adding more single pipes

InoMurko:
## Overview

Increasing the readability of code with this little feature of credo. 

## Changes

- circleci check if the number of single pipes has increased
- adding `SinglePipe` check in subprojects 
- bumping Credo


InoMurko:
> We at the Elixir core team consider a smell to use the pipe just once: value |> Some.fun. The pipe was designed to express pipelines and you don’t have a pipeline if you have only one call. :slightly_smiling_face: Therefore it would be nice to have a check that guarantees you never pipe only once.

.circleci/config.yml
```diff
@@ -177,7 +177,8 @@ jobs:
     steps:
       - setup_elixir-omg_workspace
       - run: make install-hex-rebar
-      - run: mix do compile --warnings-as-errors --force, credo, format --check-formatted --dry-run
+      - run: mix do compile --warnings-as-errors --force, credo --ignore-checks Credo.Check.Readability.SinglePipe, format --check-formatted --dry-run
+      - run: mix credo --only Credo.Check.Readability.SinglePipe | grep -c "Use a function call when a pipeline is only one function long" | awk '$1 > 369'
```
pgebal, line 181: How does this work? What this 369 mean?

unnawut: Also why this needs to be separate from the command above?

achiurizo: Pretty clever!

@pgebal it's a few things.

* breaks apart the general credo routine to ignore single pipe checks
* runs the single pipe check separately, if its greater than the existing `369` occurrences of the single pipe, output the number

@InoMurko does this need to return explicitly return a failure exit code?

```suggestion
      - run: mix credo --only Credo.Check.Readability.SinglePipe | grep -c "Use a function call when a pipeline is only one function long" | awk '$1 > 369{exit 1}' 
```

pgebal: Thanks Arthur. @InoMurko Why don't we want to fix single pipelines and stick to 1 run of credo? Is it because fixing all the single pipelines is too time consuming?

InoMurko: yeah, there's 369 occurrences at the moment. But we don't want to add new ones, and gradually fix those 369.

Issue #1426, opened 2020-03-25
ignore duplicate events, break down events


InoMurko:
Events consumed from Ethereum Event Listeners pushed ontu the OMG Bus, consumed by `OMG.Watcher.DatadogEvent.ContractEventConsumer` and it's sibling in Child Chain can come in bulk.
For example `BlockSubmitted` events can have a list of blocks, that were validated
```
[%{blknum: 150000, eth_height: 9735388, event_signature: "BlockSubmitted(uint256)", log_index: 8, root_chain_txhash: "0x060f9502ed7bad7565a1a668700ff1201f25cbd47c129ae4eba39365be1deaf7"}, %{blknum: 151000, eth_height: 9735415, event_signature: "BlockSubmitted(uint256)", log_index: 67, root_chain_txhash: "0x440f28117a14e55ae373108594c3bb224167f8023241696bf432d39e887fb192"}, %{blknum: 152000, eth_height: 9735447, event_signature: "BlockSubmitted(uint256)", log_index: 72, root_chain_txhash: "0xf1a910b6373f32a96c860c32ad227182d4a6b79a7b06e87cc3d2b3333d0af08a"}]
```

We should break these down into separate events being sent to Datadog, but also keep a local list of events that were also received.

pnowosie:
> For example BlockSubmitted events can have a list of blocks, that were validated

Just clarification. Usually `BlockSubmitted` is a single event. However due to gas price changes, previous block could not be spotted before next one is ready. Then Ch-ch sends both with a higher gas.

unnawut:
@InoMurko So Souradeep is about to start the event dedup part. Something like this?

- Use `root_chain_txhash` as the unique id
- Store the `root_chain_txhash` locally somewhere. This is not super critical so within a genserver state should be enough?
- On new event received, compare the new event with the local store before deciding to push

souradeep-das:
Also, Is there any history of duplicated events that we have received?

unnawut:
Interesting I'm not seeing any log with "*Timestamp:*". I'm looking into it a bit more.

souradeep-das:
@unnawut Do you mean duplicate logs?

unnawut:
Actually I'm not seeing any logs from this code at all

souradeep-das:
Summing up the discussions with O -
- A lot of duplicate events show up in datadog i.e. events about the same activity with only differing timestamp
[...]

PR #1472, opened 2020-04-16
Break down incoming events to publish separately

souradeep-das:
:clipboard: Partly https://github.com/omisego/elixir-omg/issues/1426

## Overview

This enables to break down and send separate events to Datadog, when events consumed from Ethereum event listeners come in bulk

## Changes

- Update `OMG.Watcher.DatadogEvent.ContractEventConsumer` and the Childchain sibling to iterate and send received events individually 
- Update tests to check breaking into separate events

## Testing
```
mix test test/omg_watcher/datadog_event/contract_event_consumer_test.exs
mix test test/omg_child_chain/datadog_event/contract_event_consumer_test.exs
```


apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
```diff
@@ -84,9 +84,11 @@ defmodule OMG.ChildChain.DatadogEvent.ContractEventConsumer do
 
     options = tags(aggregation_key, state.release, state.current_version, timestamp)
     title = "#{event_name}"
-    message = "#{inspect(Encode.make_it_readable!(data))} - Timestamp: #{timestamp}"
 
-    :ok = apply(state.publisher, :event, create_event_data(title, message, options))
+    Enum.each(data, fn ev ->
+      message = "[#{inspect(Encode.make_it_readable!(ev))}] - Timestamp: #{timestamp}"
+      :ok = apply(state.publisher, :event, create_event_data(title, message, options))
```
unnawut, line 90: The `title` is derived from `hd(data)`, so it's assuming the first event name is the same as the rest.

In this PR where we're already breaking down per event, should we derive it from each event instead?

apps/omg_child_chain/test/omg_child_chain/datadog_event/contract_event_consumer_test.exs
```diff
@@ -41,22 +41,23 @@ defmodule OMG.ChildChain.DatadogEvent.ContractEventConsumerTest do
...
-  test "if a list of event message are put on omg bus is consumed by the event consumer and published on the publisher interface" do
+  test "if a list of events put on omg bus is consumed by the event consumer, broken down and published individually on the publisher interface" do
     topic_name = self() |> :erlang.pid_to_list() |> to_string()
     sig = "#{topic_name}(bytes32)"
     data = [%{event_signature: sig}, %{event_signature: sig}]
     {:root_chain, topic_name} |> OMG.Bus.Event.new(:data, data) |> OMG.Bus.direct_local_broadcast()
 
-    assert_receive {:event, _, _}
+    Enum.each(data, fn _ ->
+      assert_receive {:event, _, _}
```
unnawut, line 59: Now that it expect 2 events, should we check that they're indeed the two events, and not one event reported twice?

apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
souradeep-das: True, that would make more sense. Is there any event other than the `BlockSubmitted` which could come in bulk?

apps/omg_child_chain/test/omg_child_chain/datadog_event/contract_event_consumer_test.exs
souradeep-das: Absolutely, i'm taking the message into equation then

apps/omg_child_chain/lib/omg_child_chain/datadog_event/contract_event_consumer.ex
unnawut: I didn't look into specifics but I think `DepositCreated` and exit started events. Need double checking though, I havn't touched on them lately.

kevsul: And `ExitFinalized`, some others as well

PR #1473, opened 2020-04-17
Move db storage out of docker containers

kevsul:
## Overview

Previously when running docker-compose, the rocksdb data dir was stored inside the `childchain`, `watcher` and `watcher_info` docker containers, which meant that they were lost every time the container was restarted, so the services were out of sync with local geth, which meant that any time you had to start a new chain and redeploy the contracts again on every restart.

Moving the rocksdb data dir out of the container makes it persistent, so that the services can be restarted and they will pick up where they left off. 

## Changes

- In docker-compose.yaml, mount `/data` as a volume and use it as the rocksdb data dir for `childchain`, `watcher` and `watcher_info`. Same pattern as the `geth` container.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -411,6 +411,10 @@ jobs:
       image: ubuntu-1604:201903-01
     steps:
       - checkout
+      - run:
+          name: Setup data dir
+          command: |
+            [ -d data ] || mkdir data && chmod 777 data
       - run: make docker-child_chain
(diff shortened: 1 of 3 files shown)


unnawut:
Could you also update `Makefile`'s docker-nuke?

```diff
- docker-compose down --remove-orphans
+ docker-compose down --remove-orphans --volumes
```

So the volume doesn't hang around when geth snapshot is wiped.

PR #1474, opened 2020-04-17
feat: system memory monitor that considers buffered and cached memory

### LINKED ISSUE #1431: :system_memory_high_watermark alarm should consider cached memory as free memory
## Overview

`:system_memory_high_watermark` was raised when used memory + cached memory is > 80%. The alarm shouldn't consider cached memory as used memory.

## Details
```elixir
An alarm was raised {:system_memory_high_watermark, []}
```

The BEAM VM's used memory looks nominal (hey I get to use SpaceX launch control term here):

image

However, `system.mem.used` is over 80%:

image (1)

Running `free -m`:

```shell
$ free -m
             total       used       free     shared    buffers     cached
Mem:         15037      13051       1986          1        643       9218
-/+ buffers/cache:       3189      11848
Swap:            0          0          0
```

Running `:memsup.get_system_memory_data()` (which should be the values the alarm was seeing):

```elixir
iex> :memsup.get_system_memory_data()
[
  system_total_memory: 15767875584,
  free_swap: 0,
  total_swap: 0,
  cached_memory: 9666367488,
  buffered_memory: 676225024,
  free_memory: 2060222464,
  total_memory: 15767875584
]
```

Turns out this is a legacy issue with erlang's memsup that considers cached memory as used memory: http://erlang.org/pipermail/erlang-questions/2006-September/023144.html

We need to update our sysmem alarm to consider cached memory as free memory.

## Actual behaviour

`:system_memory_high_watermark` alarm is raised when used_memory + cached_memory > 80%

## Expected behaviour

`:system_memory_high_watermark` alarm is raised when used_memory > 80% (excluding cached_memory)

unnawut:
Closed #1431

## Overview

Introduces a new system memory monitor that replaces [`:memsup`](http://erlang.org/doc/man/memsup.html#get_system_memory_data-0)'s `:system_memory_high_watermark`.

The new alarm is called `:system_memory_too_high` to avoid ambiguity whether it's being raised by `:memsup` or this monitor.

## Changes

- Added `OMG.Status.Monitor.MemoryMonitor`
- Configured `:os_mon`'s `system_memory_high_watermark: 1.00` to prevent it from reporting
- Moved `OMG.Status.Metric.StatsdMonitor` into Monitor namespace

## Testing

```
mix test apps/omg_status/test/omg_status/monitor/memory_monitor_test.exs
```

Since it's to do with memory this will be a little tricky to test. But the CI/CD environment should stop reporting `:system_memory_high_watermark` after this PR is deployed.


apps/omg_status/lib/omg_status/configuration.ex
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  @spec system_memory_high_threshold() :: float() | no_return()
+  def system_memory_high_threshold() do
+    Application.fetch_env!(@app, :system_memory_high_threshold)
+  end
+
+  @spec datadog_disabled?() :: boolean()
+  def datadog_disabled?() do
+    Application.get_env(@app, Tracer)[:disabled?]
```
pgebal, line 35: Why do we use `get_env` here and `fetch_env!` in others?

apps/omg_status/lib/omg_status/monitor/memory_monitor.ex
```diff
@@ -0,0 +1,167 @@
...
+  use GenServer
+  require Logger
+
+  @type t :: %__MODULE__{
+          alarm_module: module(),
+          memsup_module: module(),
+          interval_ms: pos_integer(),
+          threshold: float(),
+          raised: boolean(),
+          tref: reference() | nil
```
pgebal, line 34: More meaningful name here would be good.

apps/omg_status/lib/omg_status/monitor/memory_monitor.ex
```diff
@@ -0,0 +1,167 @@
...
+  end
+
+  def handle_cast(:clear_alarm, state) do
+    {:noreply, %{state | raised: false}}
+  end
+
+  #
+  # gen_event handlers
+  #
+  def handle_call(_request, state), do: {:ok, :ok, state}
```
pgebal, line 99: Why not `:reply` instead of `:ok`?

pgebal:
> Isn't there something built into docker or GCP that we could use for monitoring memory? Why did we go with a custom solution?

I remember we were using appsignal. What happened that we stopped?

unnawut:
> > Isn't there something built into docker or GCP that we could use for monitoring memory? Why did we go with a custom solution?
> 
> I remember we were using appsignal. What happened that we stopped?

Two main references I found: https://github.com/omisego/devops/issues/79 and https://github.com/omisego/elixir-omg/pull/812

apps/omg_status/lib/omg_status/monitor/memory_monitor.ex
unnawut: This is a :gen_event callback so the pattern is different from the GenServer 's `handle_call/3` callback. See https://erlang.org/doc/man/gen_event.html#Module:handle_call-2

apps/omg_status/lib/omg_status/configuration.ex
unnawut: Good question! This config is always available so I've changed it to `fetch_env!`: https://github.com/omisego/elixir-omg/pull/1474/commits/fc7ed1dba27c1fefe2b7a6474630f4408e303197

apps/omg_status/lib/omg_status/monitor/memory_monitor.ex
unnawut: Done! https://github.com/omisego/elixir-omg/pull/1474/commits/fc29514f8bb1fb907d527b56f7ca49af56dd7610

unnawut:
By the way this is more like a bug fix to address this flood of false alarms and not quite a new feature. But because this fix replaces :memsup system memory monitor with its own logic so I prefix it with `feat:`.

image

Issue #1272, opened 2020-01-21
docker dependency on localchain_contract_addresses.env


kevsul:
docker-compose.yml references the `localchain_contract_addresses.env` file which means that none of the docker commands or make targets that use docker will work if this file hasn't been generated yet.

The way to generate it is to call `make init_test` which is not very intuitive if you're not running tests.
For example, the README says that to bring the entire system up, run `docker-compose up` but this doesn't work if you haven't run `make init_test` first.

InoMurko:
I have used one of my open PR's to quickly edit the README on the front page: https://github.com/omisego/elixir-omg/blame/master/README.md#L41
This slipped me when I introduced snapshots. 

What would make this more process more intuitive? It seems to me we should rewrite our documentation that reference docker-compose commands and use just Makefile. At least in our docs we would be able to maintain a flow that works.

unnawut:
I havn't tried but I think it's possible to somehow include the file generation with `docker-compose up`. 

One unelegant idea is a short live container that checks & generates the file on `up`.

jrhite:
We're using a Makefile as script, a bit odd, but about the issue, since
we're using a Makefile can't we use it for its strength in this case?

Makefile

target: target_dependency_1 target_dependency_2
    do stuff after dependencies are done first

Whatever commands require the env file, just add init_test Makefile target
as a dependency.

Would that work?

PR #1476, opened 2020-04-17
refactor: add prerequisites for makefile targets involving docker-compose

pgebal:
Closes https://github.com/omisego/elixir-omg/issues/1272

It does not solve the issue in general. It solves it only for `make` commands involving `docker-compose` with default docker-compose file.


Makefile
```diff
@@ -324,38 +324,38 @@ docker-build-start-cluster:
 	SNAPSHOT=SNAPSHOT_MIX_EXIT_PERIOD_SECONDS_120 make init_test && \
 	docker-compose build --no-cache && docker-compose up
 
-docker-stop-cluster:
+docker-stop-cluster: localchain_contract_addresses.env
```
InoMurko, line 327: whats the reason these were just added everywhere?

pgebal: It's added wherever we use default  `docker-compose` as it needs to have `localchain_contract_addresses.env` in current dir.

InoMurko: why do we need to pull fresh contracts when we `docker-stop-cluster` and again when we start `docker-start-cluster-with-datadog`?

why do we need to pull fresh contracts when we `docker-nuke`?

pgebal: We need to do it only when there is no `localchain_contract_addresses.env`.
Otherwise `docker-compose down` won't work.
We have this section in our `docker-compose.yml`:
```
env_file:
      - ./localchain_contract_addresses.env
```

InoMurko: so this checks if the file exists and if it does, it doesn't run the whole `$(MAKE) init-contracts`?

pgebal: yes

Issue #1480, opened 2020-04-21
Add root chain transaction hash to invalid_exit and unchallenged_exit events


okalouti:
Related to epic https://github.com/omisego/elixir-omg/issues/1471

Refers to Ethereum transaction hash of the exit in question

PR #1485, opened 2020-04-23
[2] Add root chain transaction hash to InvalidExit and UnchallengedExit events

okalouti:
The original PR broke the Watcher because existing DB entries did not have newly added keys. 

This is the original PR with some additions: 

- defaults `root_chain_txhash` to `nil` for previously existing database entries without this key.
- writes a test (exit_info_test.exs) to test for this and effectively prevent adding a new key without similar handling. 

Test with `mix test test/omg_watcher/exit_processor/exit_info_test.exs`

## Original PR

#1480 

## Overview

This adds a `rootchain_txhash` property to the `invalid_exit` and `unchallenged_exit` events, referring to the Ethereum transaction hash of the exit in question.

As seen by running the cabbage tests locally: 

```json
{
        "details": {
          "amount": 12000000000000000000,
          "currency": "0x0000000000000000000000000000000000000000",
          "eth_height": 899,
          "name": "unchallenged_exit",
          "owner": "0x37d5c1fbfa32b6582b425d5ea8d724478dac0aee",
          "root_chain_txhash": "0x5ebfbd6caff8d53e74590a646ffa5018e8776445c78ed735e29bdb52e55edd9c",
          "utxo_pos": 2001000000000
        },
        "event": "unchallenged_exit"
      },
      {
        "details": {
          "amount": 12000000000000000000,
          "currency": "0x0000000000000000000000000000000000000000",
          "eth_height": 899,
          "name": "invalid_exit",
          "owner": "0x37d5c1fbfa32b6582b425d5ea8d724478dac0aee",
          "root_chain_txhash": "0x5ebfbd6caff8d53e74590a646ffa5018e8776445c78ed735e29bdb52e55edd9c",
          "utxo_pos": 2001000000000
        },
        "event": "invalid_exit"
      }

```

## Changes

- Changes in `exit_processor/exit_info.ex` to persist the root chain transaction hash in the database. 
- Changes in `omg_watcher/event.ex` to add root chain transaction hash to the event struct.

## Testing

- No tests added but exit test helper updated to include a transaction hash of `<<0::256`. Open question whether we'd like a more advanced logic there in this PR. 

Note to @dmitrydao to update documentation [here](https://docs.omg.network/byzantine-conditions) and [here](https://docs.omg.network/challenging-exits) once this is merged.


apps/omg_watcher/test/omg_watcher/exit_processor/exit_info_test.exs
```diff
@@ -0,0 +1,56 @@
...
+  Test of the logic of the exit_info module
+  """
+  use OMG.Watcher.ExitProcessor.Case, async: true
+
+  alias OMG.Watcher.ExitProcessor.ExitInfo
+
+  import OMG.Watcher.ExitProcessor.TestHelper
+
+  @recently_added_keys [:root_chain_txhash]
+  @utxo_pos_1 {0, 0, 0}
```
pnowosie, line 26: Nah, this is special value. I'd like to have real-world values in tests
```suggestion
  @utxo_pos_1 {1000, 0, 0}
```

apps/omg_watcher/test/omg_watcher/exit_processor/exit_info_test.exs
```diff
@@ -0,0 +1,56 @@
...
+  @exit_1 %{
+    exit_id: 1,
+    amount: 1,
+    currency: random_bytes(20),
+    eth_height: 1,
+    exiting_txbytes: random_bytes(32),
```
pnowosie, line 32: This is misleading as it contains transaction bytes but suggest ~hash value. Let's make it obviously invalid
```suggestion
    exiting_txbytes: "not important tx bytes",
```

apps/omg_watcher/test/support/exit_processor/test_helper.ex
```diff
@@ -159,4 +176,21 @@ defmodule OMG.Watcher.ExitProcessor.TestHelper do
...
+  @spec binary_from_integer(integer(), integer()) :: binary()
+  def binary_from_integer(integer, num_bytes) do
+    num_bits = num_bytes * 8
+    <<integer::size(num_bits)>>
+  end
+
+  @doc """
+  Generates a certain length of random bytes. Not recommended for identifiers as uniqueness not guaranteed.
+  """
+  @spec random_bytes(integer()) :: binary()
```
pnowosie, line 192: Oh, no please remove - IMHO it gains nothing

apps/omg_watcher/test/support/exit_processor/test_helper.ex
```diff
@@ -159,4 +176,21 @@ defmodule OMG.Watcher.ExitProcessor.TestHelper do
 
     long
   end
+
+  @doc """
+  Generates a given length of bytes for the given integer.
+  """
+  @spec binary_from_integer(integer(), integer()) :: binary()
+  def binary_from_integer(integer, num_bytes) do
```
pnowosie, line 184: IMHO there is no gain having such functions. We usually put some non-sensical values e.g. <<1>> for binary, or
<<1::160>> to show it's 20-bytes length. KISS-rule

Whenever particular value isn't important we should make it look obviously non-sensical. Otherwise we're suggesting the value has  some more context.

apps/omg_watcher/lib/omg_watcher/exit_processor/exit_info.ex
```diff
@@ -125,7 +125,8 @@ defmodule OMG.Watcher.ExitProcessor.ExitInfo do
       exiting_txbytes: exit_info.exiting_txbytes,
       is_active: exit_info.is_active,
       eth_height: exit_info.eth_height,
-      root_chain_txhash: exit_info.root_chain_txhash
+      # defaults value to nil if non-existent in the DB.
+      root_chain_txhash: Map.get(exit_info, :root_chain_txhash, nil)
```
InoMurko, line 129: map get default behaviour is returning nil

InoMurko:
```
If default is not provided, nil is used.

```

apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
```diff
@@ -38,6 +38,35 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
+        eth_height: 1,
+        exit_id: 1,
+        exiting_txbytes: random_bytes(32),
+        is_active: false,
+        owner: random_bytes(20)
+      }
+
+      {:ok, %{exits: exits}} =
+        Core.init(
+          [
```
InoMurko: pls fix this formatting

apps/omg_watcher/test/support/exit_processor/test_helper.ex
InoMurko: yes please

apps/omg_watcher/test/omg_watcher/exit_processor/exit_info_test.exs
```diff
@@ -0,0 +1,56 @@
...
+
+defmodule OMG.Watcher.ExitProcessor.ExitInfoTest do
+  @moduledoc """
+  Test of the logic of the exit_info module
+  """
+  use OMG.Watcher.ExitProcessor.Case, async: true
+
+  alias OMG.Watcher.ExitProcessor.ExitInfo
+
+  import OMG.Watcher.ExitProcessor.TestHelper
```
InoMurko, line 23: what are you importing?

InoMurko:
You've said in the comment:
```
As seen by running the cabbage tests locally:

```

Do cabbage tests still work if `root_chain_txhash is nil`. Meaning, for old entries without root_chain_txhash?

apps/omg_watcher/lib/omg_watcher/exit_processor/exit_info.ex
okalouti: Yes. I left it intentionally for clarity - but I suppose the comment is sufficient.

apps/omg_watcher/test/omg_watcher/exit_processor/exit_info_test.exs
okalouti: it was for the usage of `random_bytes/1` but now removed along with the import

apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
okalouti: Believe this is an old commit. No changes to `core_test.exs`.

apps/omg_watcher/test/support/exit_processor/test_helper.ex
okalouti: Thanks for the feedback @pnowosie. Incorporated it in 7f8da07.

apps/omg_watcher/test/omg_watcher/exit_processor/exit_info_test.exs
okalouti: Makes sense not to have a null value! `OMG.Utxo.positon` will also have the `:utxo_position` atom, so I just switched the `blknum` to 1000.

okalouti:
> You've said in the comment:
> 
> ```
> As seen by running the cabbage tests locally:
> ```
> 
> Do cabbage tests still work if `root_chain_txhash is nil`. Meaning, for old entries without root_chain_txhash?

The cabbage tests would be persisting `ExitInfo` structs as per the new specification in the DB so you would not encounter a scenario where the `root_chain_txhash` is `nil` in the cabbage tests.

InoMurko:
But does the API still work if they are nil?

okalouti:
> But does the API still work if they are nil?

Not sure if I'm addressing your question here but you did draw my attention to one thing. 

The `byzantine_events` array returned by `/status.get` contains events as defined in `event.ex`, created in `ExitInfo.make_event_data/3`. 

I just updated the relevant event modules to accept `nil` for the `root_chain_txhash` (thanks for prompting that) 

If that is what you meant I believe the API should be good. Not sure if it's worth writing a cabbage test for this but I can simulate locally and confirm.

okalouti:
Confirming that `/status.get` can return events with `root_chain_txhash` equal to nil if none is persisted in the DB.

apps/omg_watcher/test/support/exit_processor/test_helper.ex
```diff
@@ -102,7 +117,9 @@ defmodule OMG.Watcher.ExitProcessor.TestHelper do
     competitor_position = Keyword.get(opts, :competitor_position)
 
     competitor_position =
-      if competitor_position, do: Utxo.Position.encode(competitor_position), else: not_included_competitor_pos()
+      if competitor_position,
```
InoMurko, line 120: this should have been a `case` clause, not an `if`

PR #1488, opened 2020-04-24
who monitors the monitor

InoMurko:
If you want to receive exit signals from processes you need to monitor them. I don't see any code implying that RootChainCoordinator monitors Ethereum Event Listeners and these two chunks of code would ever get called. So I'm removing them.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -186,7 +186,7 @@ jobs:
             _counter=$(mix credo --only Credo.Check.Readability.SinglePipe | grep -c "Use a function call when a pipeline is only one function long")
             echo "Current Credo.Check.Readability.SinglePipe occurrences:"
             echo $_counter
-            if [ $_counter -gt 367 ]; then
+            if [ $_counter -gt 359 ]; then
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -186,7 +186,7 @@ jobs:
             _counter=$(mix credo --only Credo.Check.Readability.SinglePipe | grep -c "Use a function call when a pipeline is only one function long")
             echo "Current Credo.Check.Readability.SinglePipe occurrences:"
             echo $_counter
-            if [ $_counter -gt 367 ]; then
+            if [ $_counter -gt 359 ]; then
```
InoMurko, line 189: bravo team!!!

unnawut:
This means that after this PR we would see the listener crashes in the logs?

InoMurko:
No. This changes nothing. Coordinator does not monitor listeners at all. I can’t find any code that would invoke monitoring.

unnawut:
> No. This changes nothing. Coordinator does not monitor listeners at all. I can’t find any code that would invoke monitoring.

Aha ok got it!

PR #1496, opened 2020-04-30
Papa/sec 27 watcher info ife support

pnowosie:
Fixes https://github.com/omisego/security-issues/issues/27

## Overview

Watcher-Info subscribes to Watcher's event bus to catch up on network changes. As the issue describes IFE events wasn't handled at all. Raw contract events just report changes they are not sufficient for Watcher-Info to be able to work on them without storing its own context and Watcher already has this context any way.

## Changes

This :pr: introduces 3 new watcher events (new category as we have contract's marked with `:root_chain` and plasma's `:child_chain` events)
- `{:watcher, "InFlightExitStarted"}` - this events extends data from [ExitProcessor's `new_in_flight_exits`]() that can be used to spend (all) in-flight transaction's inputs (with [`Eth_Event`]())
- `{:watcher, "InFlightTxOutputPiggybacked"}` - guides Info API to spend output when in-flight transaction's output is piggybacked, is broadcasted from [`piggyback_exits`].
- `{:watcher, "InFlightExitIOputProcessed"}` - propagated after IFE is processed in [`finalize_in_flight_exits`] - which contains UTXO that security Watcher spends. It is a bit late for Info API but still can be useful is edge cases or just as sanity check. 

For the Child-chain the first 2 events are sufficient to eagerly spend IFE's I/Oputs, it should suffice in most cases for Info API as well. However here there is some latency involved in when events about new blocks and exits are processed so I think having 3rd event as a last-resort is a good idea.

Only the first event is persisted as a spending eth event in Postgres. Action on these event is idempotent.

## Testing

- [Cabbage's in_fligth_exits_test] demonstrates that issue is fixed.
- More unit tests


apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -437,6 +454,11 @@ defmodule OMG.Watcher.ExitProcessor do
 
     {:ok, state3, db_updates} = Core.finalize_in_flight_exits(state2, finalizations, invalidities)
 
+    :ok =
+      events_with_utxos
+      |> Tools.to_bus_events_data()
+      |> publish_internal_bus_events("InFlightExitIOputProcessed")
```
achiurizo, line 460: What is `IOput`?

pnowosie: Input or Output (credits to @jbunce) alternative is Xput 👐 

More context: we already publish `InFlightInputWithdrawn` & `InFlightOutputWithdrawn`  event from rootchain. The above event from Watcher is focused to let consumer know that particular *put can be marked as spent (if not done yet) than on the ife processing... I'm open to better names

achiurizo: Okay, why not just call it `InFlightExitProcessed` if it can be either? Introducing yet another term for this isn't helpful since, both inputs and outputs are viable means to do an IFE. Also, let's consider that an `input` and `output` are references to an output.

priv/cabbage/apps/itest/test/features/in_flight_exits.feature
```diff
@@ -14,3 +14,24 @@ Feature: In Flight Exits
...
+  Scenario: In-flight exited transaction piggybacked output is not spendable
+    Given "Alice" deposits "10" ETH to the root chain
+    Then "Alice" should have "10" ETH on the child chain after finality margin
+    Given Alice creates a transaction for "5" ETH
+    And Alice sends the most recently created transaction
+    And Alice starts an in flight exit from the most recently created transaction
+    Then Alice verifies its in flight exit from the most recently created transaction
+    Given Alice piggybacks output from her most recent in flight exit
+    And "Alice" in flight transaction most recently piggybacked output is not spendable any more
+    Then Alice can processes her own most recent in flight exit
```
pgebal, line 37: New line at the end please.

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pgebal: I can't come up with any scenario where we need this. Watcher info can do without it as all the utxos involved here are either:
* inputs of ife tx and we have an eth_event spending entry for that, as we put it into the db when exit starts
* outputs, which had to be piggybacked and we put eth events for piggybacks into the db

priv/cabbage/apps/itest/test/features/in_flight_exits.feature
pgebal, line 37: It would be nice to have a test where:
1. ife and piggyback come first
2. then comes the ch-ch block with a transaction that created the piggybacked output
3. piggybacked output is not available to spend
it might be hard to implement that in cabbage, but maybe a test involving just the db would do?

pnowosie: Yes - that's exactly the egde case I included handling the on-finalization event.
We don't have such control of the events in cabbage test - I'll include it elsewhere and point here. Thanks :+1:

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pnowosie: @achiurizo: I'd go with `InFlightExitOutputWithdrawn` then - exit processed is fine however in my opinion it could be confused as Info-Watcher (consumer) process no exits

@pgebal: https://github.com/omisego/elixir-omg/pull/1496#discussion_r421416898

apps/omg_watcher_info/lib/omg_watcher_info/db/eth_event.ex
```diff
@@ -105,95 +108,138 @@ defmodule OMG.WatcherInfo.DB.EthEvent do
...
-      from(ethevent in __MODULE__,
-        where: ethevent.root_chain_txhash_event == ^root_chain_txhash_event,
-        left_join: txoutputs in assoc(ethevent, :txoutputs),
-        preload: [txoutputs: txoutputs]
-      )
-    )
+  # Tells whether `TxOutput` was already spent
+  # NOTE: it looks a little too deep into DB.TxOutput module, but I don't want to extent its API
+  @spec output_spend?(%DB.TxOutput{}) :: boolean()
+  defp output_spend?(%DB.TxOutput{spending_txhash: nil} = tx_output) do
```
kevsul, line 240: I think `output_spent` is more correct

priv/cabbage/apps/itest/test/features/in_flight_exits.feature
pnowosie: I've added ExitConsumer tests, however testing on DB level above scenario with all 3 steps above will be unnecessary bloat because in pt 1. we know we are trying to spent non-existing output then in pt 2. we're adding it and finally spending in 3. 

I've check these points in the separate tests.

apps/omg_watcher_info/test/omg_watcher_info/exit_consumer_test.exs
```diff
@@ -0,0 +1,162 @@
...
+          root_chain_txhash: expected_root_hash,
+          call_data: %{txhash: txhash, oindex: oindex}
+        }
+      ])
+
+      assert %DB.TxOutput{
+               ethevents: [%DB.EthEvent{log_index: ^expected_log_index, root_chain_txhash: ^expected_root_hash}]
+             } = DB.TxOutput.get_by_position(txo_pos)
+
+      event_data_2 = [
```
pgebal, line 110: Where is `event_data_1`? Could we use different name, maybe `event` would do?

apps/omg_watcher_info/test/omg_watcher_info/exit_consumer_test.exs
```diff
@@ -0,0 +1,162 @@
...
+        %{log_index: 2, root_chain_txhash: <<12::256>>, call_data: %{utxo_pos: pos_1}},
+        %{log_index: 1, root_chain_txhash: <<11::256>>, call_data: %{txhash: txhash, oindex: 1}}
+      ]
+
+      send_events_and_wait_until_processed(event_data)
+
+      assert_test_consumer_alive()
+    end
+
+    @tag fixtures: [:alice, :initial_blocks]
```
pgebal, line 60: It's optional, but maybe we could cheaply get rid off fixtures by copying some code into setup. I know we are trying to move away from fixtures. If there is no "quick fix" for that, leave it as it.

apps/omg_watcher_info/test/omg_watcher_info/exit_consumer_test.exs
```diff
@@ -0,0 +1,162 @@
...
+    {:ok, _} =
+      GenServer.start_link(
+        ExitConsumer,
+        [topic: :watcher_test_topic, bus_module: __MODULE__.FakeBus],
+        name: TestExitConsumer
+      )
+
+    _ =
+      on_exit(fn ->
+        with pid when is_pid(pid) <- GenServer.whereis(TestExitConsumer) do
```
pgebal, line 38: Couldn't we use pid returned by `GenServer.start_link` in line 29 ?

apps/omg_watcher_info/test/omg_watcher_info/db/eth_event_test.exs
```diff
@@ -272,4 +272,114 @@ defmodule OMG.WatcherInfo.DB.EthEventTest do
...
+             %DB.EthEvent{
+               log_index: ^expected_log_index,
+               root_chain_txhash: ^expected_eth_txhash,
+               event_type: ^expected_event_type
+             }
+           ] = txo2.ethevents
+  end
+
+  @tag fixtures: [:alice, :initial_blocks]
+  test "Can spend ife piggybacked output event", %{alice: alice} do
```
pgebal, line 331: Something's off with this name. I think this `event` at the is not needed.

apps/omg_watcher_info/test/omg_watcher_info/db/eth_event_test.exs
```diff
@@ -272,4 +272,114 @@ defmodule OMG.WatcherInfo.DB.EthEventTest do
...
+      %{
+        root_chain_txhash: expected_eth_txhash2,
+        log_index: expected_log_index2,
+        call_data: %{txhash: txhash2, oindex: oindex2}
+      }
+    ]
+
+    assert :ok = DB.EthEvent.insert_exits!(exits, expected_event_type)
+
+    txo1 = DB.TxOutput.get_by_output_id(txhash1, oindex1)
```
pgebal, line 363: This looks like it could extracted to a private function. It's repeated for `txo1` and `txo2`.

apps/omg_watcher_info/test/omg_watcher_info/exit_consumer_test.exs
```diff
@@ -0,0 +1,162 @@
...
+      txhash = Crypto.hash(<<pos_1>>)
+
+      event_data = [
+        %{log_index: 2, root_chain_txhash: <<12::256>>, call_data: %{utxo_pos: pos_1}},
+        %{log_index: 1, root_chain_txhash: <<11::256>>, call_data: %{txhash: txhash, oindex: 1}}
+      ]
+
+      send_events_and_wait_until_processed(event_data)
+
+      assert_test_consumer_alive()
```
pgebal, line 57: It's just a matter of preferences. Do as you like, but I like name `assert_consumer_alive` better.

apps/omg_watcher_info/test/omg_watcher_info/exit_consumer_test.exs
```diff
@@ -0,0 +1,162 @@
...
+    end
+
+    @tag fixtures: [:alice, :initial_blocks]
+    test "when receive ife started or finalized events spends given outputs", %{alice: alice} do
```
pgebal, line 61: I think `when receive` -> `receiving` is "more correct".

apps/omg_watcher_info/test/omg_watcher_info/exit_consumer_test.exs
```diff
@@ -0,0 +1,162 @@
...
+      spent_utxos_pos = alice.addr |> get_utxos_pos() |> Enum.take(2)
+      [pos_1, pos_2] = Enum.map(spent_utxos_pos, &Position.encode/1)
+
+      # Note: event data is the same for InFlightExitStarted and InFlightExitOutputWithdrawn events
+      event_data = [
+        %{log_index: 2, root_chain_txhash: <<12::256>>, call_data: %{utxo_pos: pos_2}},
```
pgebal, line 67: `<<12::256>>` is repeated many times. Maybe extract it to module attribute.

priv/cabbage/apps/itest/test/features/in_flight_exits.feature
pgebal: I think test we have in `OMG.WatcherInfo.ExitConsumerTest` do not add up to the scenario I posted.
We do not have a test where event comes before the utxo.
If this is implied from the tests we have, please put a comment explaining how that works.

pgebal: Reasoning from what I can find in the code it looks like it will work well but tests would still be good.
And I still do not get why we need spending on finalization. I think all would work well without it. Could you please explain why we need that?

apps/omg_watcher_info/test/omg_watcher_info/exit_consumer_test.exs
pnowosie: The only thing here is the process could be terminated at this point, as it’s linked to the test process. So the code makes sure it’s still alive before trying to stop it.

priv/cabbage/apps/itest/test/features/in_flight_exits.feature
pnowosie: In the scenario you posted:
1. Watcher-Info receives an event IFE-started and speds inputs
2. Watcher-Info receives an event Output piggybacked, but block including in-flight tx has not been received yet (👇 next point)
3. Watcher-Info receives an event New block - which creates an in-flight tx outputs (it does not matter for Watcher-Info that inputs are already spend)

Without additinal „sanity” spending on finalizations the outputs will persist.

pnowosie: Regarding the explicit test for the scenario I think we have it splitted into separated (smaller) tests:
Pt 1 is covered,
Pt 2 - test showing that spending non-existing output does not break,
Pt 3 - test spending output by „finalization”-like event

I’ll add a comment in the test module to describe the scenario and it’s coverage by the tests already present

pnowosie:
NOTE: I discussed the concern with @pgebal. It still might be that Watcher-Info is very late in `BlockApplicationConsumer` that both `Ife-started` & `piggybacks` events will try to spend not yet existed outputs (which will be created only when W-I catchs up with blocks).
Therefore we also handle `ife finalization` events which spends everything again.
There is still hypothetical case that W-I still be late with blocks on finalization - but I found such cases very rare to implement fix for such special case.

apps/omg_watcher_info/test/omg_watcher_info/exit_consumer_test.exs
pnowosie: Both are very common and quite simple fixtures - IMHO we'll gain nothing removing them.

PR #1505, opened 2020-05-12
fix: exclude active exiting utxos from calls to /account.get_exitable_utxos

pgebal:
Closes https://github.com/omisego/security-issues/issues/30
## Overview

Active exiting utxos are excluded from `/account.get_exitable_utxos` response.


apps/omg_watcher/test/omg_watcher/api/account_test.exs
```diff
@@ -63,5 +66,44 @@ defmodule OMG.Watcher.API.AccountTest do
 
       assert %{blknum: ^blknum, txindex: ^txindex, oindex: ^oindex} = utxo
     end
+
+    @tag fixtures: [:db_initialized]
```
pgebal, line 70: I will get rid of this fixture before making it ready to review, but review comments are welcomed.

apps/omg_watcher/lib/omg_watcher/api/status.ex
```diff
@@ -58,7 +58,7 @@ defmodule OMG.Watcher.API.Status do
     eth_syncing = syncing?()
 
     validated_child_block_number = get_validated_child_block_number()
-    # wtf is eth diagnostics?
+    # Where is eth diagnostics?
```
pnowosie, line 61: Can see any connection with the code - please remove comment.

apps/omg_watcher/test/omg_watcher/api/account_test.exs
pnowosie: I don't think we're heading to ditch all fixtures, once they already there. Such fixtures are very useful and also keeps the codebase clean. That's my opinion.

pgebal: You are right. I like them to. The problem we have is that if test uses a fixture defined in other app it can be run only in bulk, not as a single test. There might be a problem in how we designed tests. In this case removing fixture was easy, so I did it.

apps/omg_watcher/test/omg_watcher/api/account_test.exs
```diff
@@ -13,24 +13,42 @@
...
   alias OMG.TestHelper
+  alias OMG.Utxo
   alias OMG.Watcher.API.Account
 
+  require Utxo
+
   @eth OMG.Eth.zero_address()
   @payment_output_type OMG.WireFormatTypes.output_type_for(:output_payment_v1)
 
+  setup do
```
unnawut, line 27: +1 on moving fixture to setup, this makes it more explicit, plus avoids deep fixture inheritance that fixtures allow

PR #1513, opened 2020-05-15
feat: NewExits and NewInflightExits for non state updating functions

boolafish:
:clipboard: https://github.com/omisego/new-tx-type-poc/issues/13

## Overview

As the first step to make the Core state thinner, we are first removing the places that insert new state about exit events to the process memory.

Previously, for most event action it would: 1. update the state of the ExitProcessor 2. Return `db_updates` and will eventually be send to OMG.DB to update the db accordingly.

Now we are trying to remove the usage of the state of the ExitProcessor. Thus provide and use function that does not need to access state.

This PR provides the initial ones: `new_exit` and `new_in_flight_exit`.

## Changes

- Add new modules `NewExits` and `NewInflightExits` for the update_db related functions.
- Use the new module to provide `db_updates` while still calling the `Core` to keep the state getting update until we are able to remove all state reading.

## Testing

- added unit test
- compare the result (`db_updates`) of new modules with the existing Core function.


apps/omg_watcher/lib/omg_watcher/exit_processor/update_db_cores/new_exits.ex
```diff
@@ -0,0 +1,48 @@
...
+
+  @doc """
+  DB updates needed to add new exits from Ethereum events into tracked state.
+
+  The list of `exit_contract_statuses` is used to track current (as in wall-clock "now", not syncing "now") status.
+  This is to prevent spurious invalid exit events being fired during syncing for exits that were challenged/finalized
+  Still we do want to track these exits when syncing, to have them spend from `OMG.State` on their finalization
+  """
+  @spec get_db_update(list(map()), list(map)) :: {:ok, list()} | {:error, :unexpected_events}
+  def get_db_update(new_exits, exit_contract_statuses)
```
boolafish, line 30: Copy of the `new_exits` function without using `state` from `Core`: 
https://github.com/omisego/elixir-omg/blob/d118f210b41c49aa29d5f5e2e1bdea70c5ab9fa2/apps/omg_watcher/lib/omg_watcher/exit_processor/core.ex#L176-L198

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -316,7 +317,16 @@ defmodule OMG.Watcher.ExitProcessor do
         put_timestamp_and_sft(exit_event, state.min_exit_period_seconds, state.child_block_interval)
       end)
 
+    {:ok, db_updates_new} = NewExits.get_db_update(exit_maps, exit_contract_statuses)
     {new_state, db_updates} = Core.new_exits(state, exit_maps, exit_contract_statuses)
+
+    _ =
+      if db_updates_new != db_updates,
```
boolafish, line 324: showing the idea of comparing db_updates if any diverge occurs: https://github.com/omisego/new-tx-type-poc/issues/13#issuecomment-628512991

T-Dnzt: I don't think that's the place for this. I think duplicating the tests for the new module or adding tests that specifically test that these two modules generate the same updates are better approaches.

And if we don't have those tests yet, well it's time to write them before making any changes :p

boolafish: > duplicating the tests for the new module

Yes, want to give something for discussion so did not write it yet. But for sure any new module should have tests.

> adding tests that specifically test that these two modules generate the same updates are better approaches.

I don't really agree this worth the effort. This would be one time using tests and throw away immediately after refactor. Though I agree that is in practice giving more confidence especially if we consider how much traffic we have on env...

Code change wise:

I think we can either do this way (shadow + logging) or be more aggressive and return `db_update_new` in the last line instead directly. However, we would need to keep the line of `Core.new_exits` until we have everything in place as the other functions would rely on that.

That is a bit why I am proposing to shadow since we will need this anyway...

boolafish: Cool, have a quick sync with @T-Dnzt 
We will continue small refactoring (one-by-one) path that keeps the old modules still running.
Meanwhile adding new unit tests + comparing tests of the two modules. Though would be removed, it gives us better confidence. The effort is quite less anyway as the module already exists. Just add one extra line at the end of the unit tests perhaps.

apps/omg_eth/lib/omg_eth/root_chain.ex
```diff
@@ -85,6 +85,7 @@ defmodule OMG.Eth.RootChain do
     {:array, {:tuple, [:bool, {:uint, 256}, {:bytes, 32}, :address, {:uint, 256}, {:uint, 256}]}}
 
     # solidity does not return arrays of structs
+    # following is tuple representing the struct in solidity: https://git.io/JfEi5
```
pnowosie, line 88: Wow `git.io` - nice to know the tricks :)
...but as it's a comment and may stay here for some long time - are we aware of data retention of this shortener service?

apps/omg_watcher/test/omg_watcher/exit_processor/update_db_cores/new_exits_test.exs
```diff
@@ -0,0 +1,142 @@
+# Copyright 2019-2020 OmiseGO Pte Ltd
```
pnowosie, line 1: Module path is `update_db_cores` ❓ Should correspond to the app module, so `update_db`

apps/omg_watcher/test/support/exit_processor/test_helper.ex
```diff
@@ -77,8 +77,9 @@ defmodule OMG.Watcher.ExitProcessor.TestHelper do
     processor
   end
 
-  # See `OMG.Eth.RootChain.get_in_flight_exits_structs/2` for reference of where this comes from
+  # See `OMG.Eth.RootChain.get_in_flight_exit_structs/2` for reference of where this comes from
   # `nil`s are unused portions of the returns data from the contract
+  # also see the reference the solidity struct data: https://git.io/JfEi5
```
pnowosie, line 82: Same here - prefer full links maybe

apps/omg_watcher/test/omg_watcher/exit_processor/update_db_cores/new_exits_test.exs
boolafish: ouch, thanks

apps/omg_eth/lib/omg_eth/root_chain.ex
boolafish: I've updated anyway, the url is a bit long though....

I can't find the data retention policy of this service but as it is native GitHub Service I would not be too much concern though...

boolafish:
we are pausing refactoring works due to low engineer resources in the cycle C. We will prioritize exercise part further on this cycle. closing this PR to avoid introducing half way refactoring.

PR #1519, opened 2020-05-20
use fixed version of ex_abi

ayrat555:
the latest version of `ex_abi` introduced a couple of bugs. The PR uses the patched version of ex_abi which fixes those bugs


InoMurko:
Perhaps another thing we could now do - since ex_abi works - is to fetch models from plasma contracts for these two:
https://github.com/omisego/elixir-omg/blob/master/apps/omg_eth/lib/omg_eth/root_chain.ex#L61-L98
wdyt @ayrat555 ?

ayrat555:
@InoMurko Some abi encodings have failed in integration tests. I'll look into it

ayrat555:
> Perhaps another thing we could now do - since ex_abi works - is to fetch ABI.FunctionSelector from plasma contracts for these two:
https://github.com/omisego/elixir-omg/blob/master/apps/omg_eth/lib/omg_eth/root_chain.ex#L61-L98
wdyt @ayrat555 ?

@InoMurko How exactly these functions should be changed?

InoMurko:
> > Perhaps another thing we could now do - since ex_abi works - is to fetch ABI.FunctionSelector from plasma contracts for these two:
> > https://github.com/omisego/elixir-omg/blob/master/apps/omg_eth/lib/omg_eth/root_chain.ex#L61-L98
> > wdyt @ayrat555 ?
> 
> @InoMurko How exactly these functions should be changed?

We can put both `standardExits(uint160[])` and `inFlightExits(uint160[])` `ABI.FunctionSelector` definitions into `OMG.Eth.RootChain.AbiFunctionSelector`. 

When we have that, we can rewrite these two functions:
```
##
  ## these two cannot be parsed with ABI decoder!
  ##
  @doc """
  Returns standard exits data from the contract for a list of `exit_id`s. Calls contract method.
  """
  def get_standard_exit_structs(exit_ids) do
    contract = Configuration.contracts().payment_exit_game

    return_types = [
      {:array, {:tuple, [:bool, {:uint, 256}, {:bytes, 32}, :address, {:uint, 256}, {:uint, 256}]}}
    ]

    # TODO: hack around an issue with `ex_abi` https://github.com/poanetwork/ex_abi/issues/22
    #       We procure a hacky version of `OMG.Eth.Client.call_contract` which strips the offending offsets from
    #       the ABI-encoded binary and proceeds to decode the array without the offset
    #       Revert to `call_contract` when that issue is resolved
    call_contract_manual_exits(
      contract,
      "standardExits(uint160[])",
      [exit_ids],
      return_types
    )
  end

  @doc """
  Returns in flight exits of the specified ids. Calls a contract method.
  """
  def get_in_flight_exit_structs(in_flight_exit_ids) do
    contract = Configuration.contracts().payment_exit_game
    {:array, {:tuple, [:bool, {:uint, 256}, {:bytes, 32}, :address, {:uint, 256}, {:uint, 256}]}}

    # solidity does not return arrays of structs
    return_types = [
      {:array, {:tuple, [:bool, {:uint, 64}, {:uint, 256}, {:uint, 256}, :address, {:uint, 256}, {:uint, 256}]}}
    ]

    call_contract_manual_exits(
      contract,
      "inFlightExits(uint160[])",
      [in_flight_exit_ids],
      return_types
    )
  end
```

into something like `RootChain` `blocks/1` `or next_child_block/0`, `get_mined_child_block/0` so that the response data gets decoded via:
```
defp get_external_data(contract_address, signature, args) do
    {:ok, data} = Rpc.call_contract(contract_address, signature, args)
    Abi.decode_function(data, signature)
  end
```

InoMurko:
Both ABIs are defined in PaymentExitGame.json. You can run `make init-contracts` in the root of elixir-omg and search for the file.
`find . -name PaymentExitGame.json`

You can extract the abi defintion if you do this:
```
abi_path_to_payment_exit_game_json
      |> File.read!()
      |> Jason.decode!()
      |> Map.fetch!("abi")
      |> ABI.parse_specification(include_events?: true)
```

InoMurko:
OH this is amazing and it cleanups rootchain!

Issue #1477, opened 2020-04-20
UndefinedFunctionError: function Mix.env/0 is undefined (module Mix is not available)


unnawut:
## Actual behaviour

```
** (UndefinedFunctionError) function Mix.env/0 is undefined (module Mix is not available)
```

## Expected behaviour

`Mix.env()` is not used in runtime code and therefore exception not thrown.

unnawut:
Probably from https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher_info/lib/omg_watcher_info/supervisor.ex#L24-L45

unnawut:
Actually it's from the childchain:

```
:gen_event handler Ink installed in Logger terminating
** (exit) an exception was raised:
** (UndefinedFunctionError) function Mix.env/0 is undefined (module Mix is not available)
Mix.env()
(ink 1.0.3) lib/ink.ex:158: Ink.log_json/2
(ink 1.0.3) lib/ink.ex:97: Ink.handle_event/2
(stdlib 3.12) gen_event.erl:577: :gen_event.server_update/4
(stdlib 3.12) gen_event.erl:559: :gen_event.server_notify/4
(stdlib 3.12) gen_event.erl:300: :gen_event.handle_msg/6
(stdlib 3.12) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
```

unnawut:
Ugh it's from Ink itself: https://github.com/ivx/ink/blob/6f2e9dfdd104b5d8188748767977d5c7f0c06163/lib/ink.ex#L158

unnawut:
It's a cascading failure. Something needs to fail prior for it to reach this condition. Which for this case is https://github.com/omisego/elixir-omg/issues/1478

unnawut:
Pending upstream fix: https://github.com/ivx/ink/pull/57

unnawut:
Upstream updated. I'll take this in a bit.

PR #1532, opened 2020-05-25
fix: error attempting to log txhash in binary

unnawut:
Relates to #1477

## Overview

Fix the error when attempting to log the transaction hash while it's still in binary format.

In the log, this error looks like `{:error, %Jason.EncodeError{message: "invalid byte 0xDC in <<67, 111, 117, ...`

## Changes

- Convert txhash to its hex string before logging

## Testing

Each block submission should render the proper error message `Could not get gas used for txhash ... after ... retries.` instead of the encoding error.


unnawut:
bruh, back to draft

PR #1534, opened 2020-05-26
refactor: consistent log message for new events

unnawut:
## Overview

This fixes and makes consistent the event logging e.g.

```elixir
1. inspect(Elixir.OMG.Status.Metric.StatsdMonitor) got event: {:clear_alarm, {:main_supervisor_halted, ...

2. Monitor got event: {:clear_alarm, {:main_supervisor_halted, ...

3. Elixir.OMG.Eth.EthereumHeightMonitor.AlarmHandler got event: {:clear_alarm, {:main_supervisor_halted, ...
```

by converting all into the last format with the full module name

## Changes

- Convert all event ignores to log with full module names

## Testing

Check the logs. Some monitors will ignore some alarms so the log should appear within a few minutes after boot.


unnawut:
> Looks good, for a future PR, might be worth while to centralize our logging messages in some module?

True, the module name is easily helpful to be a json log attribute instead

unnawut:
Coming back reading again, this looks like a silly PR. The original issue the unparsed `inspect(Elixir.OMG.Status.Metric.StatsdMonitor)` in the log that's been bothering me when log-skimming.

Issue #1520, opened 2020-05-20
Watcher Info persists in-flight exit events as :standard_exit type


okalouti:
This came up as I was running the cabbage tests and seeing how exit events were being persisted in the `watcher_info` DB. 

The `ExitConsumer` inserts all exits using  `EthEvent.insert_exits!(data)`. See [here](https://github.com/omisego/elixir-omg/blob/master/apps/omg_watcher_info/lib/omg_watcher_info/exit_consumer.ex).

However, this method takes an `event_type` argument that defaults to `:standard_exit`. See [here](https://github.com/omisego/elixir-omg/blob/7e4a23198d6634140bde90ed18b70439f8464794/apps/omg_watcher_info/lib/omg_watcher_info/db/eth_event.ex#L111).

However, no differentiation is made in the ExitConsumer.

cc: @pnowosie

pnowosie:
Thanks @okalouti 
It slipped out despite [the tests](https://github.com/omisego/elixir-omg/blob/a72aa5102d2fe29baa3a2ebcb2887a63fb003108/apps/omg_watcher_info/test/omg_watcher_info/db/eth_event_test.exs#L280) which makes me think that we should expose `TxOutput`'s `events` from `/transaction.get` 🤔 (or smth else) to allow us to test it properly.

okalouti:
I think the most straightforward fix is to differentiate in the exit consumer?

okalouti:
Because insert_exits is behaving as it was designed to. Though we could perhaps remove the default argument.

PR #1538, opened 2020-05-27
feat: add event type when consumer is spending utxos

pnowosie:
Addresses #1520 needed for 0.4.9 😑 

## Overview

Reporting appropriate event type from Watcher-Info's exit consumer.

## Changes

Pushing appropriate event type to the `ExitConsumer` so the correct type of the spending event can be persisted in Postgres DB.

## Testing

Exit consumer tests cover the change.

### DIFF
--- a/apps/omg_watcher_info/lib/omg_watcher_info/exit_consumer.ex
+++ b/apps/omg_watcher_info/lib/omg_watcher_info/exit_consumer.ex
@@ -31,13 +31,13 @@ defmodule OMG.WatcherInfo.ExitConsumer do
 
   def init(args) do
     topic = Keyword.fetch!(args, :topic)
-    event_type = Keyword.fetch!(args, :event_type)
     bus_module = Keyword.get(args, :bus_module, @default_bus_module)
+    state = %{event_type: Keyword.fetch!(args, :event_type)}
 
     :ok = bus_module.subscribe(topic, link: true)
 
     _ = Logger.info("Started #{inspect(__MODULE__)}, listen to #{inspect(topic)}")
-    {:ok, %{event_type: event_type}}
+    {:ok, state}
   end
 
   def handle_info({:internal_event_bus, :data, data}, state) do
--- a/apps/omg_watcher_info/lib/omg_watcher_info/exit_consumer.ex
+++ b/apps/omg_watcher_info/lib/omg_watcher_info/exit_consumer.ex
@@ -31,16 +31,17 @@ defmodule OMG.WatcherInfo.ExitConsumer do
 
   def init(args) do
     topic = Keyword.fetch!(args, :topic)
+    event_type = Keyword.fetch!(args, :event_type)
     bus_module = Keyword.get(args, :bus_module, @default_bus_module)
 
     :ok = bus_module.subscribe(topic, link: true)
 
     _ = Logger.info("Started #{inspect(__MODULE__)}, listen to #{inspect(topic)}")
-    {:ok, %{}}
+    {:ok, %{event_type: event_type}}
   end
 
   def handle_info({:internal_event_bus, :data, data}, state) do
-    _ = EthEvent.insert_exits!(data)
+    _ = EthEvent.insert_exits!(data, state.event_type)
     {:noreply, state}
   end
 end
--- a/apps/omg_watcher_info/lib/omg_watcher_info/exit_consumer.ex
+++ b/apps/omg_watcher_info/lib/omg_watcher_info/exit_consumer.ex
@@ -30,17 +30,25 @@ defmodule OMG.WatcherInfo.ExitConsumer do
   use GenServer
 
   def init(args) do
-    topic = Keyword.fetch!(args, :topic)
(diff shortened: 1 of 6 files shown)


apps/omg_watcher_info/lib/omg_watcher_info/exit_consumer.ex
```diff
@@ -31,16 +31,17 @@ defmodule OMG.WatcherInfo.ExitConsumer do
...
   def init(args) do
     topic = Keyword.fetch!(args, :topic)
+    event_type = Keyword.fetch!(args, :event_type)
     bus_module = Keyword.get(args, :bus_module, @default_bus_module)
 
     :ok = bus_module.subscribe(topic, link: true)
 
     _ = Logger.info("Started #{inspect(__MODULE__)}, listen to #{inspect(topic)}")
-    {:ok, %{}}
+    {:ok, %{event_type: event_type}}
```
InoMurko, line 40: Struct?

InoMurko, line 40: If the plan is to expand this (eventually) I suggest you hold a struct. If not, keep this value as the state and not a map.

pnowosie: I'll keep it simple as I don't foresee expansion of this module. I'll introduce the state var there 👍

apps/omg_watcher_info/lib/omg_watcher_info/exit_consumer.ex
```diff
@@ -32,15 +32,16 @@ defmodule OMG.WatcherInfo.ExitConsumer do
   def init(args) do
     topic = Keyword.fetch!(args, :topic)
     bus_module = Keyword.get(args, :bus_module, @default_bus_module)
+    state = %{event_type: Keyword.fetch!(args, :event_type)}
 
     :ok = bus_module.subscribe(topic, link: true)
 
     _ = Logger.info("Started #{inspect(__MODULE__)}, listen to #{inspect(topic)}")
-    {:ok, %{}}
+    {:ok, state}
```
InoMurko, line 40: Hmm. I think we misunderstood each other. By state war I meant keeping just event_type in state. 

{:ok, Keyword.fetch!(args, :event_type)}

pnowosie: Ooch - it's clever, but it's quite uncommon in our code... I would leave it as it looks now if you not strongly against ❓

InoMurko: Then make a struct. Which is the pattern thats most common.

InoMurko: Or ignore me, no strong objections 😹

apps/omg_watcher_info/lib/omg_watcher_info/supervisor.ex
```diff
@@ -64,19 +64,19 @@ defmodule OMG.WatcherInfo.Supervisor do
...
+        {OMG.WatcherInfo.ExitConsumer, [topic: {:watcher, "InFlightExitStarted"}, event_type: :in_flight_exit]},
         id: :ife_exit_consumer
       ),
       Supervisor.child_spec(
-        {OMG.WatcherInfo.ExitConsumer, [topic: {:watcher, "InFlightTxOutputPiggybacked"}]},
+        {OMG.WatcherInfo.ExitConsumer, [topic: {:watcher, "InFlightTxOutputPiggybacked"}, event_type: :in_flight_exit]},
         id: :ife_exit_output_piggybacked_consumer
       ),
       Supervisor.child_spec(
-        {OMG.WatcherInfo.ExitConsumer, [topic: {:watcher, "InFlightExitOutputWithdrawn"}]},
```
okalouti, line 79: Could we name the last two events differently? Say `in_flight_exit_output_piggybacked` and `in_flight_exit_output_withdrawn`

From the consuming applications' perspective (mobile wallet, block explorer, web wallet) we would want to distinguish between these events in future features - i.e. tracking the progression of an exit.

So concerned that storing every event related to IFEs under the same `event_type` will cause headaches down the line

pnowosie: Simple answer - we can - this PR makes it possible (I tune the UTXO query which supports 2 event types, so we can have more without effort).

I'm reluctant to do this in this PR because we not need it and we aren't clear about the future requirements. I'd prefer to see the product requirement and consider the changes in the system that satisfy it. 

One specific objection I have is as you might know from the PR which introduced these events `output_withdrawn` is a last resort in case syncing went weird, so normally I expect to see former two event types in the live system in majority of cases.

Pongch: I missed a lot of contexts here to provide a useful comment. But as long as we leave the possibility open in the future for client application to come to an accurate state of the UTXO based on the events reported by the watcher. I think its 🆗

PR #1543, opened 2020-05-28
fix: apply retry on root chain call

boolafish:
:clipboard: fix https://github.com/omisego/elixir-omg/issues/1526

## Overview

We have seen quite some transient error when fetching contract data with `{:error, :timeout}` (or 502). This PR adds retry logic with exponential backoff to mitigate the issue by using `ElixirRetry` library.

## Changes

- introduce dependency `retry`, [ElixirRetry](https://github.com/safwank/ElixirRetry). I have checked the test code and it looks quite nice to me. I did not really reviewed the implementation though as macro quote and all elixir magic are a bit hard for me now....
- use exponential backoff with max to 30 sec strategy
- I am using annotation on private functions. It is tested [here](https://github.com/safwank/ElixirRetry/blob/ff0a1674b3467545db77a679c49afc207203e23f/test/retry/annotation_test.exs#L244-L252)
- Since all of the failure recorded by Sentry follows the pattern `{:error, _}`, the library support reacting with `:error` atom by default. So we do not need to specify the error pattern ourself.
- apply retry to `RootChain.get_external_data` and `RootChain.Rpc.get_ethereum_events` as the two functions have shown error to Sentry.

## Testing

N/A


InoMurko:
@boolafish I respect the effort, but I think this is a step in the wrong direction.

The purpose of supervision trees in elixir and erlang is to respawn on transient errors. Basically errors that go away themselves. This way we avoid writing defensive code and let the supervisor handle errors. If you have time, Joe Armstrong has amazing talks about erlang on youtube https://www.youtube.com/watch?v=YaUPdgtUYko
where he explains why they designed the language this way. Or if you prefer to read, his PhD on erlang usage in distributed systems https://erlang.org/download/armstrong_thesis_2003.pdf

oh oh, if we're concerned about sentry - sentry allows us to ignore errors.

boolafish:
thanks for the links. just watch the video and is a nice one. The paper might be a bit too long to digest atm 😅

Back to the problem here, I see two potential places for improvement with current code though:
1. A network transient failure would cause the whole process using it to fail. Which would impact some sort of large process like `BlockGetter` and `BlockQueue`. (it looks like large and complex to me not sure if everyone agrees though). Also, `BlockGetter` process is used in watcher API (`get_status`), so that means there is a small down time on the api when waiting for the process to spin up again.
1. I don't think you could do backoff strategy with retry triggered eventually by OTP supervisor. Though you can argue that we don't need such strategy yet. And doing that is more of being nice member to the network so would be nice to infura instead of our own system.

If retry on network failure is not the common path to go in elixir, what should it be? Or we just embrace that failure impacting to higher level like it is now? (Then I kind of argue this is against the philosophy of supervisor tree as I understand it to encourage you write code that would fail as smaller part with smaller impact)

InoMurko:
> thanks for the links. just watch the video and is a nice one. The paper might be a bit too long to digest atm 😅
> 
> Back to the problem here, I see two potential places for improvement with current code though:
> 
> 1. A network transient failure would cause the whole process using it to fail. Which would impact some sort of large process like `BlockGetter` and `BlockQueue`. (it looks like large and complex to me not sure if everyone agrees though). Also, `BlockGetter` process is used in watcher API (`get_status`), so that means there is a small down time on the api when waiting for the process to spin up again.
> 2. I don't think you could do backoff strategy with retry triggered eventually by OTP supervisor. Though you can argue that we don't need such strategy yet. And doing that is more of being nice member to the network so would be nice to infura instead of our own system.
> 
> If retry on network failure is not the common path to go in elixir, what should it be? Or we just embrace that failure impacting to higher level like it is now? (Then I kind of argue this is against the philosophy of supervisor tree as I understand it to encourage you write code that would fail as smaller part with smaller impact)

1. All elixir processes are really lightweight. It really takes microseconds to restart a process. A bit more if they're attached to a supervisor, and a bit more if the `init/1` does heavy lifting.
The get_status API call is sorted in https://github.com/omisego/elixir-omg/pull/1535

2. The thing is, that we have two kinds of network issues. Transient that go away because they're just glitches in the network and complete network outage. The second one is sorted by a custom monitor above SyncSupervisor https://github.com/omisego/elixir-omg/blob/v0.4.8/apps/omg_watcher/lib/omg_watcher/supervisor.ex#L37-L48

So the retry logic you added is a redundant mechanism in terms of OTP... if your processes and supervision trees are designed correctly. And I think we have a track record that proves they're OK.

image



>Instead of let it crash, you can say that Erlang has auto healing mechanisms. Healing is a good thing and accurately describes what happens in the system.
(The title is a play on words - you should, in fact, let it crash).
https://ninenines.eu/articles/dont-let-it-crash/

boolafish:
thanks for your patience explaining this 🙇‍♂️ ✨

PR #1545, opened 2020-05-28
chore: move production code debug log to info level

boolafish:

## Overview

I have found it sometimes hard to collect enough info to understand a root cause of a Sentry issue.
As a result, I would like to propose move all debug logs to info level to ensure it would show up in data dog.

## Changes

find all non test code with `Logger.debug` and move them to `Logger.info`.

## Testing

N/A

### DIFF
--- a/apps/omg/lib/omg/root_chain_coordinator.ex
+++ b/apps/omg/lib/omg/root_chain_coordinator.ex
@@ -106,7 +106,7 @@ defmodule OMG.RootChainCoordinator do
   end
 
   def handle_call({:check_in, synced_height, service_name}, {pid, _ref}, state) do
-    _ = Logger.debug("#{inspect(service_name)} checks in on height #{inspect(synced_height)}")
+    _ = Logger.info("#{inspect(service_name)} checks in on height #{inspect(synced_height)}")
 
     {:ok, state} = Core.check_in(state, pid, synced_height, service_name)
     {:reply, :ok, state}
--- a/apps/omg_child_chain/lib/child_chain.ex
+++ b/apps/omg_child_chain/lib/child_chain.ex
@@ -67,7 +67,7 @@ defmodule OMG.ChildChain do
   defp is_supported(%Transaction.Recovered{}), do: :ok
 
   defp result_with_logging(result) do
-    _ = Logger.debug(" resulted with #{inspect(result)}")
+    _ = Logger.info(" resulted with #{inspect(result)}")
     result
   end
 end
--- a/apps/omg_child_chain/lib/omg_child_chain/ethereum_event_aggregator.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/ethereum_event_aggregator.ex
@@ -14,7 +14,7 @@
 defmodule OMG.ChildChain.EthereumEventAggregator do
   @moduledoc """
   This process combines all plasma contract events we're interested in and does eth_getLogs + enriches them if needed
-  for all Ethereum Event Listener processes. 
+  for all Ethereum Event Listener processes.
   """
   use GenServer
   require Logger
@@ -258,7 +258,7 @@ defmodule OMG.ChildChain.EthereumEventAggregator do
         missing_to_block = List.last(missing_blocks)
 
         _ =
-          Logger.debug(
+          Logger.info(
             "Missing block information (#{missing_from_block}, #{missing_to_block}) in event fetcher. Additional RPC call to gather logs."
           )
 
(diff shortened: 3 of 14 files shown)


apps/omg/lib/omg/root_chain_coordinator.ex
```diff
@@ -106,7 +106,7 @@ defmodule OMG.RootChainCoordinator do
   end
 
   def handle_call({:check_in, synced_height, service_name}, {pid, _ref}, state) do
-    _ = Logger.debug("#{inspect(service_name)} checks in on height #{inspect(synced_height)}")
+    _ = Logger.info("#{inspect(service_name)} checks in on height #{inspect(synced_height)}")
```
InoMurko, line 109: I don't think you want to do this, this creates a lot of noise.

apps/omg_child_chain/lib/child_chain.ex
```diff
@@ -67,7 +67,7 @@ defmodule OMG.ChildChain do
   defp is_supported(%Transaction.Recovered{}), do: :ok
 
   defp result_with_logging(result) do
-    _ = Logger.debug(" resulted with #{inspect(result)}")
+    _ = Logger.info(" resulted with #{inspect(result)}")
```
InoMurko, line 70: is this the only occurance of the call? https://github.com/omgnetwork/elixir-omg/pull/1545/files#diff-6a1bd5749a6c9a2d4f25c09e137226a4R61

otherwise you don't really know what the result is from

apps/omg_child_chain/lib/omg_child_chain/ethereum_event_aggregator.ex
```diff
@@ -258,7 +258,7 @@ defmodule OMG.ChildChain.EthereumEventAggregator do
         missing_to_block = List.last(missing_blocks)
 
         _ =
-          Logger.debug(
+          Logger.info(
```
InoMurko, line 261: lots of noise this one too

apps/omg_eth/lib/omg_eth/ethereum_height.ex
```diff
@@ -46,7 +46,7 @@ defmodule OMG.Eth.EthereumHeight do
   end
 
   def handle_info({:internal_event_bus, :ethereum_new_height, new_height}, _state) do
-    _ = Logger.debug("Got an internal :ethereum_new_height event with height: #{new_height}.")
+    _ = Logger.info("Got an internal :ethereum_new_height event with height: #{new_height}.")
```
InoMurko, line 49: is this needed? it's every 15 seconds

apps/omg/lib/omg/root_chain_coordinator.ex
boolafish: then we delete those you think we don't need. I am actually doing this PR blindly 😛

apps/omg_utils/lib/omg_utils/http_rpc/adapter.ex
```diff
@@ -28,7 +28,7 @@ defmodule OMG.Utils.HttpRPC.Adapter do
 
     with {:ok, body} <- Jason.encode(body),
          {:ok, %HTTPoison.Response{} = response} <- HTTPoison.post(addr, body, headers) do
-      _ = Logger.debug("rpc post #{inspect(addr)} completed successfully")
+      _ = Logger.info("rpc post #{inspect(addr)} completed successfully")
```
InoMurko, line 31: ughh, lots of noise as well. every http request towards childchain goes through this

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -175,7 +175,7 @@ defmodule OMG.Watcher.BlockGetter do
 
     {state, synced_height, db_updates} = Core.apply_block(state, block_application)
 
-    _ = Logger.debug("Synced height update: #{inspect(db_updates)}")
+    _ = Logger.info("Synced height update: #{inspect(db_updates)}")
```
InoMurko, line 178: lots of noise

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -292,7 +292,7 @@ defmodule OMG.Watcher.BlockGetter do
       {blocks_to_apply, synced_height, db_updates, state} =
         Core.get_blocks_to_apply(state, submissions, next_synced_height)
 
-      _ = Logger.debug("Synced height is #{inspect(synced_height)}, got #{length(blocks_to_apply)} blocks to apply")
+      _ = Logger.info("Synced height is #{inspect(synced_height)}, got #{length(blocks_to_apply)} blocks to apply")
```
InoMurko, line 295: lots of noise

apps/omg/lib/omg/root_chain_coordinator.ex
InoMurko: if you want to remove logs from production code you can do purging of a specific level, https://hexdocs.pm/logger/Logger.html#module-application-configuration
compile_time_purge_matching 

I would suggest you enable all these logs locally, build your cluster with dockercompose and run cabbage tests :)

PR #1546, opened 2020-05-28
fix: skip byzantine event tests

T-Dnzt:
## Overview

Some of the Byzantine perf tests are randomly failing. This PR introduces a `skip` tag to skip those tests for now. They should be moved to a different place in the future.

## Changes

- New `skip` tag
- Skip `skip` tests in Perf sub app

### DIFF
--- a/apps/omg/test/test_helper.exs
+++ b/apps/omg/test/test_helper.exs
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 ExUnitFixtures.start()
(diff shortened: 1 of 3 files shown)


unnawut:
How abouttt deleting them altogether? The skip means we're still stuck with all the fixtures and setups that these tests need. If we remove them we can move quicker to re-implement those tests.

InoMurko:
I'm afraid these will never really get implemented anywhere else.

PR #1558, opened 2020-06-01
prevent race condition for status cache

InoMurko:
https://github.com/omgnetwork/elixir-omg/issues/1557

## Overview

Fixing the order of booted dependencies and made some bootup synchronus instead of async.

## Changes

- status.get endpoint has dependencies to underlying services like OMG.Watcher.BlockGetter, OMG.Watcher.ExitProcessor, OMG.State. These dependencies need to be available with the first status.get call (even if it's cache!). 

## Testing

`mix test`

### DIFF
--- a/apps/omg/lib/omg/root_chain_coordinator.ex
+++ b/apps/omg/lib/omg/root_chain_coordinator.ex
@@ -72,10 +72,6 @@ defmodule OMG.RootChainCoordinator do
   end
 
   def init({args, configs_services}) do
-    {:ok, {args, configs_services}, {:continue, :setup}}
-  end
-
-  def handle_continue(:setup, {args, configs_services}) do
     _ = Logger.info("Starting #{__MODULE__} service. #{inspect({args, configs_services})}")
     metrics_collection_interval = Keyword.fetch!(args, :metrics_collection_interval)
     coordinator_eth_height_check_interval_ms = Keyword.fetch!(args, :coordinator_eth_height_check_interval_ms)
@@ -91,7 +87,7 @@ defmodule OMG.RootChainCoordinator do
 
     _ = Logger.info("Started #{inspect(__MODULE__)}")
 
-    {:noreply, state}
+    {:ok, state}
   end
 
   def handle_info(:send_metrics, state) do
--- a/apps/omg_watcher/lib/omg_watcher/block_getter.ex
+++ b/apps/omg_watcher/lib/omg_watcher/block_getter.ex
@@ -67,17 +67,10 @@ defmodule OMG.Watcher.BlockGetter do
   end
 
   @doc """
-  Initializes the GenServer state, most work done in `handle_continue/2`.
-  """
-  def init(args) do
-    {:ok, args, {:continue, :setup}}
-  end
-
-  @doc """
   Reads the status of block getting and application from `OMG.DB`, reads the current state of the contract and root
   chain and starts the pollers that will take care of getting blocks.
   """
-  def handle_continue(:setup, args) do
+  def init(args) do
     child_block_interval = Keyword.fetch!(args, :child_block_interval)
     # how many eth blocks backward can change during an reorg
     block_getter_reorg_margin = Keyword.fetch!(args, :block_getter_reorg_margin)
@@ -131,7 +124,7 @@ defmodule OMG.Watcher.BlockGetter do
         }"
       )
 
-    {:noreply, state}
+    {:ok, state}
   end
 
   # :apply_block pipeline of steps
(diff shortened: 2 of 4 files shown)


apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -304,11 +304,7 @@ defmodule OMG.Watcher.ExitProcessor do
   - updates the `ExitProcessor`'s state
   - returns `db_updates`
   """
-  def handle_call(
-        {:new_exits, exits},
-        _from,
-        state
-      ) do
+  def handle_call({:new_exits, exits}, _from, state) do
```
InoMurko, line 307: ugly formatter, couldn't look away

apps/omg/lib/omg/root_chain_coordinator.ex
```diff
@@ -72,10 +72,6 @@ defmodule OMG.RootChainCoordinator do
   end
 
   def init({args, configs_services}) do
-    {:ok, {args, configs_services}, {:continue, :setup}}
-  end
-
-  def handle_continue(:setup, {args, configs_services}) do
```
unnawut, line 78: Curious shouldn't we move work from init to `handle_continue/2` instead of the other way around, so the supervisor can move on and start the next child sooner?

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -66,18 +66,11 @@ defmodule OMG.Watcher.BlockGetter do
...
-  def init(args) do
-    {:ok, args, {:continue, :setup}}
-  end
-
   @doc """
   Reads the status of block getting and application from `OMG.DB`, reads the current state of the contract and root
   chain and starts the pollers that will take care of getting blocks.
   """
-  def handle_continue(:setup, args) do
+  def init(args) do
```
unnawut, line 73: Same here about handle_continue, especially `ExitProcessor.check_validity`, `OMG.DB.get_single_value`, `check_in_to_coordinator ` and things that could hide behind all those functions (nothing intensive from what I can see but they jump through a lot of modules and functions).

apps/omg/lib/omg/root_chain_coordinator.ex
unnawut: Mainly the concern in this specific init is `EthereumHeight.get()`

boolafish:
from this article: http://davidsulc.com/blog/2018/09/08/otp-21-introduces-handle_continue-callback-to-genserver/

It seems like the call from mailbox would be handled only after `handle_continue` finished. And it sort of states that that is why `handle_continue` is invented for (to avoid race condition before really booting up the process).

Given this, I am not sure if this approach is correct on solving this?

apps/omg/lib/omg/root_chain_coordinator.ex
unnawut: Ah reading https://github.com/omgnetwork/elixir-omg/pull/1558#issuecomment-637282889 and re-reading the description, this is how the PR tackles the race condition, my bad 🤦 

I'm thinking through this again

InoMurko:
So there are two things going on:
- handle_continue
- init

and their relationship to the supervisor is different. How?
While you`re executing init/1, the supervisor is waiting for the process to return `{:ok, pid}`.
If you use handle_continue (with empty init/1), the supervisor immediately gets the pid of the process (`{:ok, pid}`) and continues to start other processes defined as children.

The order of children defined in the supervisor is sometimes important because of the relationships between them. 
The specific problem here was that blockgetter used `handle_continue` and in there it's booting children:
```
{:ok, _} = __MODULE__.Status.start_link()
:ok = update_status(state)
``` 

which are exposed via his API:
```
  @doc """
  Retrieves the freshest information about `OMG.Watcher.BlockGetter`'s status, as stored by the slave process `Status`.
  """
  @spec get_events() :: {:ok, Core.chain_ok_response_t()}
  def get_events(), do: __MODULE__.Status.get_events()
```


So you could have a race condition where someone was requesting data from the blockgetter api, but the blockgetter wasn't fully booted because it was still processing his `handle_continue`. That's why we're making this a blocking operation. 

cc @boolafish @unnawut

unnawut:
(thinking out loud)

The other scenario I'm thinking is how `/status.get` behaved in this case before the status cache. I think the endpoint would crash anyway so this is also not introducing a new failure mode

InoMurko:
> (thinking out loud)
> 
> The other scenario I'm thinking is how `/status.get` behaved in this case before the status cache. I think the endpoint would crash anyway so this is also not introducing a new failure mode

Yeah, if you called it too early, it would have crashed on you.

unnawut:
I think a TLDR for me would be, this makes the about-to-crash crashes earlier, and so we don't have to worry about all the possible crashing states from the async and parallel crashes later on.

PR #1565, opened 2020-06-02
restart strategy

InoMurko:
## Overview

Increasing the restart strategy.

## Changes

Screen Shot 2020-06-02 at 22 40 26

## Testing

/

### DIFF
--- a/apps/omg_child_chain/lib/omg_child_chain/sync_supervisor.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/sync_supervisor.ex
@@ -46,7 +46,7 @@ defmodule OMG.ChildChain.SyncSupervisor do
 
     # We have 9 children, roughly 7 of them have a dependency to the internetz.
     # The internetz is flaky. We account for that and allow the flakyness to pass
-    # by increasing the restart strategy. But not too much, because if the internetz is 
+    # by increasing the restart strategy. But not too much, because if the internetz is
     # really off, HeightMonitor should catch that in max 8 seconds and raise an alarm.
     max_restarts = 3
     max_seconds = 5
--- a/apps/omg_child_chain/lib/omg_child_chain/sync_supervisor.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/sync_supervisor.ex
@@ -38,7 +38,19 @@ defmodule OMG.ChildChain.SyncSupervisor do
   end
 
   def init(args) do
-    opts = [strategy: :one_for_one]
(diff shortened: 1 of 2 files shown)


boolafish:
wow...copy a explanation from slack by Ino:

---

our etheruem connection monitor checks the connection every ~8 seconds. so the connection monitor didn’t catch the network glitch and it didn’t raise a ethereum_connection_error alarm.
the processes that are dependent to a working ethereum connection started crashing, but the crash limit on the supervisor was really low (3 crashes in the 5 second time frame). so tthat means the supervisor will crash itself:
```
    # Assuming the values max_restarts and max_seconds,
    # then, if more than max_restarts restarts occur within max_seconds seconds,
    # the supervisor terminates all child processes and then itself.
    # The termination reason for the supervisor itself in that case will be shutdown.
    # max_restarts defaults to 3 and max_seconds defaults to 5.
```

apps/omg_child_chain/lib/omg_child_chain/sync_supervisor.ex
```diff
@@ -38,7 +38,19 @@ defmodule OMG.ChildChain.SyncSupervisor do
...
+    # The termination reason for the supervisor itself in that case will be shutdown.
+    # max_restarts defaults to 3 and max_seconds defaults to 5.
+
+    # We have 9 children, roughly 7 of them have a dependency to the internetz.
+    # The internetz is flaky. We account for that and allow the flakyness to pass
+    # by increasing the restart strategy. But not too much, because if the internetz is
+    # really off, HeightMonitor should catch that in max 8 seconds and raise an alarm.
+    max_restarts = 3
+    max_seconds = 5
+    opts = [strategy: :one_for_one, max_restarts: max_restarts * 5, max_seconds: max_seconds]
```
boolafish, line 53: > HeightMonitor should catch that in max 8 seconds and raise an alarm.

But you're still setting this to max 5 seconds. Is it okay to not be > 8?

InoMurko: the thing is -the timer resets itself if you don't reach `max_restarts ` in `max_seconds`. so it's a bit difficult to figure out what the correct numbers are - for example, a delay in a process `init/1` could reset the timer I'm mentioning.

boolafish: > the timer resets itself if you don't reach max_restarts in max_seconds. 

oh wow.....I see. I thought if force failure as long as it runs overtime.

PR #1573, opened 2020-06-09
multiple exit processors

boolafish:
:clipboard: ref https://github.com/omgnetwork/new-tx-type-poc/issues/10

This PR is aiming to merge to `payment_v2_poc` branch.

## Overview

TL;DR a lots of hack to have something running and make existing tests happy

## Changes

- Have `ExitProcessorDispatcher` that dispatch and forward requests to `ExitProcessors` of each exit game, and then merge the result from all exit processors.
- Add payment v2 tx type. And spin up the exit processor according to `tx_type`.
- Add `exit_games` to Configuration. However, in this PR we still keep the `payment_exit_game` under `contract_addr` so it makes the tests pass. We can refactor that out later.
- Use the snapshot with 2 payment exit games
- add `address` (the source contract address) to events. Dispatcher relies on this to know where to direct the events.

## Testing

CI testing passes


apps/omg_eth/lib/omg_eth/release_tasks/set_contract.ex
```diff
@@ -72,71 +72,76 @@ defmodule OMG.Eth.ReleaseTasks.SetContract do
...
-         child_block_interval
-       )
-       when is_binary(txhash_contract) and
-              is_binary(authority_address) and is_map(contract_addresses) and is_integer(min_exit_period_seconds) and
-              is_binary(contract_semver) and is_binary(network) do
-    contract_addresses = Enum.into(contract_addresses, %{}, fn {name, addr} -> {name, String.downcase(addr)} end)
+    {exit_games, eth_vault, erc20_vault, min_exit_period_seconds, contract_semver, child_block_interval}
+  end
+
+  defp merge_configuration(config, extra_config) do
```
boolafish, line 123: dialyzer is arguing too many args.

InoMurko: I think you mean credo probably?

InoMurko:
I don't quite understand this PR. There's lots of #TODOs and hacks. What are we looking at? What are we trying to accomplish with this PR?

boolafish:
@InoMurko so one of our team's goal in this cycle is to exercise a soft upgrade to ensure that yes, it is doable with elixir-omg code base. A nice-to-have goal is to as well refactor the code to make it ALD ready.

The goal of this PR is to handle the exit processor part on soft upgrade. However, the strategy chosen here, IMO, is sth we should avoid in long-term as it is building on top of the ExitProcessor monster 🤦‍♂️ and this strategy would probably only work for upgrading to another payment tx, not flexible enough. However, this is the simplest thing we can do now. I have tried on the path to remove the monster `state` in `ExitProcessorCore` in some of the previous PR (closed) but I re-prioritize the exercising a POC over refactoring now due to it's entering second half of the cycle and not enough engineering resource to cover full refactoring + exercising.

After adding the necessary parts, we will add cabbage test to demo multi-exit game support. I don't know exactly how yet, but with the cabbage test there it might be easier for us to refactor out to better code design afterward.

apps/omg_eth/lib/omg_eth/release_tasks/set_contract.ex
```diff
@@ -72,71 +72,76 @@ defmodule OMG.Eth.ReleaseTasks.SetContract do
...
   end
 
   defp get_external_data(plasma_framework, rpc_api) do
     min_exit_period_seconds = get_min_exit_period(plasma_framework, rpc_api)
 
-    payment_exit_game =
-      plasma_framework |> exit_game_contract_address(ExPlasma.payment_v1(), rpc_api) |> Encoding.to_hex()
+    # TODO: get the list of types from ex_plasma?
+    exit_games =
+      Enum.into(OMG.WireFormatTypes.exit_game_tx_types(), %{}, fn type ->
```
InoMurko, line 109: Not sure if this version of ex_plasma has these but in childchain v2 I call `ExPlasma.payment_v1()`

apps/omg_eth/test/fixtures.exs
```diff
@@ -64,16 +68,33 @@ defmodule OMG.Eth.Fixtures do
     token_addr
   end
 
+  # inject the exit games into :omg_eth
+  # test fixture does not rely on the release task so it would need this setup
+  defp setup_exit_games() do
```
InoMurko, line 73: this seems unnecessary.

all contracts are set in test.exs, let's keep it there not to pollute tests and global state (application config).

apps/omg_watcher/lib/omg_watcher/exit_processor_dispatcher.ex
```diff
@@ -0,0 +1,274 @@
...
+  """
+  def check_validity() do
+    do_chack_validity()
+  end
+
+  def check_validity(timeout) do
+    do_chack_validity(timeout)
+  end
+
+  defp do_chack_validity(timeout \\ 5000) do
```
InoMurko, line 37: *check* not *chack*

apps/omg_watcher/lib/omg_watcher/exit_processor_dispatcher.ex
```diff
@@ -0,0 +1,274 @@
...
+    forwarded_result = forward(:check_validity, timeout)
+
+    Enum.reduce(forwarded_result, {:ok, []}, fn {chain_validity, events}, acc ->
+      {acc_chain_validity, acc_events} = acc
+
+      case acc_chain_validity do
+        :ok ->
+          {chain_validity, acc_events ++ events}
```
InoMurko, line 45: a good practice is to avoid list concatenation this way. 
use [acc_events | events] and Enum.reverse when you're done.

apps/omg_watcher/lib/omg_watcher/exit_processor_dispatcher.ex
```diff
@@ -0,0 +1,274 @@
...
+
+  @doc """
+  Returns a map of requested in flight exits, keyed by transaction hash
+  """
+  def get_active_in_flight_exits() do
+    in_flight_exits =
+      :get_active_in_flight_exits
+      |> forward()
+      |> Enum.reduce([], fn {:ok, in_flight_exits}, acc ->
+        acc ++ in_flight_exits
```
InoMurko, line 98: same as above, [acc | in_flight_exits] and enum.reverse if order is important.

apps/omg_watcher/lib/omg_watcher/exit_processor_dispatcher.ex
```diff
@@ -0,0 +1,274 @@
...
+      end)
+
+    {:ok, in_flight_exits}
+  end
+
+  # TODO: move private funcs at the bottom of the file
+  defp forward(func) do
+    Enum.map(OMG.WireFormatTypes.exit_game_tx_types(), fn tx_type ->
+      GenServer.call(tx_type, func)
```
InoMurko, line 107: it's pretty expensive to sync call in a tight loop. might be better to decouple these.

apps/omg_watcher/lib/omg_watcher/exit_processor_dispatcher.ex
```diff
@@ -0,0 +1,274 @@
...
+    case prev_status do
+      :ok -> prev_result
+      _ -> find_first_ok_result_or_return_last(tail, current_result)
+    end
+  end
+
+  defp dispatch(event_name, events) do
+    db_updates =
+      Enum.flat_map(group_events(events), fn {transaction_type, events} ->
+        {:ok, db_updates} = GenServer.call(transaction_type, {event_name, events})
```
InoMurko, line 138: same as above.

apps/omg_eth/test/fixtures.exs
boolafish: not sure how we do this in `test.exs`. Can the config file load the module we need and make the call to Geth? (Did a bit try after seeing your comment, but it argues those module as unknown)

Alternatively is we add `payment_v2_exit_game` to `localchain_contract_addresses.env` which would need all the change from `plasma-contracts`, `docker-elixir-omg` and I doubt this is what we want, I think we probably prefer to get exit game from `plasma_framework` more?

apps/omg_watcher/lib/omg_watcher/ethereum_event_aggregator.ex
```diff
@@ -84,6 +84,10 @@ defmodule OMG.Watcher.EthereumEventAggregator do
 
   def init(opts) do
     contracts = opts |> Keyword.fetch!(:contracts) |> Map.values() |> Enum.map(&from_hex(&1))
+    # This is a pure hack....currently exit game would be double configed by
+    # Config.contracts().payment_exit_game and exit_games
+    contracts = Enum.uniq(contracts)
```
pnowosie, line 89: I don't understand why is exit games addresses configured in multiple places (?) 
Can we unify this so we can obtain exit game address for a given tx_type from single, meaningfull module?

snapshots.env
```diff
@@ -1,4 +1,4 @@
-SNAPSHOT_MIX_EXIT_PERIOD_SECONDS_20=https://storage.googleapis.com/circleci-docker-artifacts/data-elixir-omg-tester-plasma-deployer-stable-20200410-MIN_EXIT_PERIOD-20-PLASMA_CONTRACTS_SHA-a69c763f239b81c5eb46b0bbdef3459f764360dc.tar.gz
-SNAPSHOT_MIX_EXIT_PERIOD_SECONDS_120=https://storage.googleapis.com/circleci-docker-artifacts/data-elixir-omg-tester-plasma-deployer-stable-20200410-MIN_EXIT_PERIOD-120-PLASMA_CONTRACTS_SHA-a69c763f239b81c5eb46b0bbdef3459f764360dc.tar.gz
-SNAPSHOT_MIX_EXIT_PERIOD_SECONDS_240=https://storage.googleapis.com/circleci-docker-artifacts/data-elixir-omg-tester-plasma-deployer-stable-20200410-MIN_EXIT_PERIOD-240-PLASMA_CONTRACTS_SHA-a69c763f239b81c5eb46b0bbdef3459f764360dc.tar.gz
-CONTRACT_SHA=a69c763f239b81c5eb46b0bbdef3459f764360dc
+SNAPSHOT_MIX_EXIT_PERIOD_SECONDS_20=https://storage.googleapis.com/circleci-docker-artifacts/data-elixir-omg-tester-plasma-deployer-dev-d01163d-MIN_EXIT_PERIOD-20-PLASMA_CONTRACTS_SHA-1d25e66420aaa4d23cac145c2ecea64123218b9a.tar.gz
```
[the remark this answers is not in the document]
boolafish: This is the snapshot with 2 payment exit game contract deployed

apps/omg_watcher/lib/omg_watcher/exit_processor_dispatcher.ex
```diff
@@ -0,0 +1,274 @@
...
+  # TODO: move private funcs at the bottom of the file
+  defp forward(func) do
+    Enum.map(OMG.WireFormatTypes.exit_game_tx_types(), fn tx_type ->
+      GenServer.call(tx_type, func)
+    end)
+  end
+
+  defp forward(func, timeout) do
+    Enum.map(OMG.WireFormatTypes.exit_game_tx_types(), fn tx_type ->
+      GenServer.call(tx_type, func, timeout)
```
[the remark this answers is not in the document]
pnowosie: Yes, and I think it's OK for POC

apps/omg_eth/test/fixtures.exs
```diff
@@ -20,6 +20,8 @@ defmodule OMG.Eth.Fixtures do
 
   alias OMG.Eth.Configuration
   alias OMG.Eth.Encoding
+  alias OMG.Eth.RootChain.Abi, as: RootChainABI
```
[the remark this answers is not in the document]
boolafish: I am using this because in this file there is sth like `ABI.encode` where ABI is another library.
If we simply use alias, it will resulting in `ABI` and `Abi` and is pretty confusing though there is slight difference.

PR #1577, opened 2020-06-11
Kevin/load test erc20 token

kevsul:
#1093  Overview

Adds the ability to use an erc20 token when running load tests.

## Changes

- Make load tests use an erc20 token by config e.g. `test_currency: "0x942f123b3587EDe66193aa52CF2bF9264C564F87"`
- Configure `Tesla.Middleware` to _not_ retry on error. It can be useful to retry, but usually when testing you don't want to do this.


priv/perf/README.md
```diff
@@ -2,14 +2,63 @@
...
+export CONTRACT_ADDRESS_ERC20_VAULT=<address of the erc20 vault contract>
+export LOAD_TEST_FAUCET_PRIVATE_KEY=<faucet private key>
+```
+
+
+### 2. Generate the open-api client
+ ```
+make init
+```
+(You need to run this any time you change the env vars above)
```
boolafish, line 24: I don't think this is true. And if any change that causes this we should try to see if we can avoid that. The new `LoadTest.Connection.XXX` is designed to not need to `make init` with url changes.

priv/perf/README.md
```diff
@@ -2,14 +2,63 @@
...
+This makes the code simpler as it doesn't have to manage separate fee utxos. 
+However, if necessary you can configure the tests to use a different currency. e.g.
+```
+config :load_test,
+  test_currency: "0x942f123b3587EDe66193aa52CF2bF9264C564F87",
+  fee_amount: 6_000_000_000_000_000,
+```
+
+### 4. Run the test
+`MIX_ENV=dev mix test apps/load_test/test/load_tests/runner/childchain_test.exs`
```
boolafish, line 46: > `MIX_ENV=<your_env> mix test`. Or `mix test` if you want to run against local services.

I think original description is better though. this line is pretty specific to a single test. Unless that is what we need only. 🤷‍♂️

Or you can use this line as an example to run single test after the description of a more generalized test running instruction as this is probably the most runned single test so people can copy paste the instruction easier.

priv/perf/apps/load_test/lib/connection/connection_defaults.ex
```diff
@@ -34,11 +34,13 @@ defmodule LoadTest.Connection.ConnectionDefaults do
   """
   def pool_name(), do: :perf_pool
 
+  # Don't automatically retry on error
+  # It _can_ sometimes be useful to retry though, so if you need it return true here
   defp retry?() do
     fn
-      {:ok, %{status: status}} when status in 500..599 -> true
+      {:ok, %{status: status}} when status in 500..599 -> false
```
boolafish, line 41: I think instead of doing this -> false thing, better ways are:
1. restrict this to 502 if you agree we want to ignore infra failure from our load test and we really don't want our test to stop at the middle (especially if it takes some time to run). I am guessing you're avoiding retry because the previous issue that the child chain does accepts the tx but timed out and retry would just not make sense. That would be 500 if I get it right.
1. Or if you don't agree with above, then set this with config/env_var, and describe how we do this in readme instead of the comment here. We default to `false` if you have experienced that most of the time we don't want to retry.

kevsul: My thinking is that by default we don't want to ignore _any_ failures, because finding failures is the main reason for running the tests.

I get that we might want to keep the tests running regardless of errors, but if there is a known infra problem (such as the 502 error) and you want to ignore it and run the tests anyway, then I think you need to explicitly do that and understand how the error will affect the test results.

I'm not sure how to set that via config though, because you may want to retry on a status code (or a range of status codes)
```
{:ok, %{status: status}} when status = 502 -> true
```
or a particular error e.g.
```
{:error, "reason: timeout"} -> true
```

Do you have any idea?
Or should I just put a better explanation with some examples into the readme?

priv/perf/README.md
kevsul: Ah yes, you're right - this was fixed

priv/perf/apps/load_test/lib/connection/connection_defaults.ex
boolafish: >  because finding failures is the main reason for running the tests.

Out of scope of this PR I would say, but better way I can think of is collect this failure metric and let the test to continue running. Usually the test should be run for a certain period, so the cost of re-run is quite high....for an engineer to wait 😛 We just want to know whether it has failed or not but we are not fixing with the load test anyway so we can continue to test the performance. Also, I think it might be interesting to monitor how our system works after some error occurs, does the error rate explodes after the first one or it auto recovers well.

> Do you have any idea?

lol, okay you're definitely thinking sth harder then I expect haha. I was just suggesting to either allow retry on 5xx or not with env_var or config. I guess we can start by either simply adding flag to turn retry on or let the env var passing in the list of status code we retry? (so turning off we pass in empty list)

InoMurko: Failure rate! Yeah, that would be really cool!

PR #1579, opened 2020-06-13
Add reorged docker compose

ayrat555:
## Overview

Adds a step to circle ci that runs cabbage tests using 2 node ethash geth that can be used to trigger reorgs


docker-compose-reorg.yml
```diff
@@ -0,0 +1,284 @@
+version: "2.3"
+services:
+  geth-1:
+    image: ethereum/client-go:alltools-stable
+    hostname: geth-1
+    environment:
+      - account=0x6de4b3b9c28e9c3e84c2b2d3a875c947a84de68d
+      - bootnodes=enode://b655cc3e5b72ab9beb8a8536a3c3ae92fbeb79feb1ebd7f95d72be72554ca586428bd48a54eb9c2bcaae455cc674299b6dd3df3c6556a493dfd50070f1a448aa@172.25.0.103:30303
+    entrypoint:
```
InoMurko, line 9: maybe now is not a good time - or if you think it is, I've once extracted this command outside docker-compose-reorg.yml
take a look here: 
https://github.com/omgnetwork/childchain/blob/master/docker-compose.yml#L39
https://github.com/omgnetwork/childchain/blob/master/docker-compose.yml#L50
https://github.com/omgnetwork/childchain/blob/master/docker/geth/command

docker-compose-reorg.yml
```diff
@@ -0,0 +1,284 @@
+version: "2.3"
+services:
+  geth-1:
+    image: ethereum/client-go:alltools-stable
+    hostname: geth-1
+    environment:
+      - account=0x6de4b3b9c28e9c3e84c2b2d3a875c947a84de68d
```
InoMurko, line 7: maybe capitalizing env vars? same for BOOTNODES.

docker-compose-reorg.yml
```diff
@@ -0,0 +1,284 @@
...
+    healthcheck:
+      test: pg_isready -U omisego_dev
+      interval: 5s
+      timeout: 3s
+      retries: 5
+    networks:
+      chain:
+        ipv4_address: 172.25.0.110
+
+  childchain:
```
InoMurko, line 147: if you chain docker-compose files, you don't need to completely redefine them here. 
I think it's enough if you simply do:
```
childchain:
   environment:
       - ETHEREUM_RPC_URL=http://172.25.0.104:80
```
everything else is then merged with the default docker-compose.yml file.

priv/cabbage/apps/itest/lib/reorg.ex
```diff
@@ -0,0 +1,101 @@
+defmodule Itest.Reorg do
```
InoMurko, line 1: Regarding codebase style:
I think our convention is that we put genserver API, like `start_reorg/0` at the top of the module (so that it's clearly evident how one can interact with a process).
Also, not sure why this didn't trigger credo... parentheses (even for functions with 0 arity) are mandatory.

Feature request:
Are you able to detect a reorg and it's depth? perhaps using: https://docs.docker.com/engine/api/sdk/examples/#print-the-logs-of-a-specific-container

This would help observability of tests greatly!

Your question regarding where do we trigger reorgs:
I think every test needs to survive reorgs! The question is, how do we enable/disable these. Maybe with an environment variable?
I think it's important to be able to do that - because you don't want to even start `Itest.Reorg` when you're running test towards one node geth right? And I don't imagine devs running reorg tests locally often.

docker-compose-reorg.yml
```diff
@@ -0,0 +1,284 @@
+version: "2.3"
+services:
+  geth-1:
+    image: ethereum/client-go:alltools-stable
```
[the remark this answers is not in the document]
ayrat555: Done, I set it to 1.9.15. With older versions, random tests start failing

docker-compose-reorg.yml
ayrat555: I extracted this command into a file in snapshot

ayrat555:
It seems not all tests can survive reorgs

priv/cabbage/apps/itest/lib/reorg.ex
ayrat555: I addressed your comments. Logs are added as a separate background process

docker-compose-reorg.yml
ayrat555: Will it redefine all properties including network?

priv/cabbage/apps/itest/lib/account.ex
```diff
@@ -24,7 +24,7 @@ defmodule Itest.Account do
   def take_accounts(number_of_accounts) do
     1..number_of_accounts
     |> Task.async_stream(fn _ -> account() end,
-      timeout: 60_000,
+      timeout: 240_000,
```
unnawut, line 27: Just for my knowledge, why the increase?

ayrat555: the task is timing out https://app.circleci.com/pipelines/github/omgnetwork/elixir-omg/7388/workflows/229cd6c9-8ddf-4d0b-ba76-dd5827bdc467/jobs/66078/steps

docker-compose-reorg.yml
InoMurko: yeah, I think it merges and overwrites!

priv/cabbage/apps/itest/test/itest/configuration_api_test.exs
```diff
@@ -13,10 +13,14 @@
 # limitations under the License.
 defmodule ConfigurationRetrievalTests do
   use Cabbage.Feature, async: true, file: "configuration_api.feature"
+  alias Itest.Reorg
   alias Itest.Transactions.Encoding
   require Logger
 
   setup_all do
+    Reorg.finish_reorg()
```
InoMurko, line 21: if I'm not mistaking, every test in cabbage is executed in their own process. Perhaps it would be better if we tie Itest.Reorg to the test process, and make sure `Itest.Reorg` cleans up after himself in a `terminate/2` callback? If Itest.Reorg is linked with the test process, the terminate/2 will be called when the test finishes, even if abruptly!

priv/cabbage/apps/itest/test/itest/deposits_test.exs
```diff
@@ -19,11 +19,16 @@ defmodule DepositsTests do
...
   alias Itest.Client
+  alias Itest.Reorg
   alias Itest.Transactions.Currency
 
   setup do
+    Reorg.finish_reorg()
+
     [{alice_account, alice_pkey}, {bob_account, _bob_pkey}] = Account.take_accounts(2)
 
+    Reorg.start_reorg()
```
InoMurko, line 30: also, another idea. Perhaps theres a better interface and explanation to the developer what is going on.
What if we tie the whole reorg period into a anonymous fn?

this way, the developer knows which parts are going to be under stress of the chain reorg? wdyt?

this would cause more changes to the codebase, but maybe it would be more evident where a reorg is happening and when it's finished. so you could expose a particular block of code to reorg events...

priv/cabbage/apps/itest/test/itest/in_flight_exits_test.exs
```diff
@@ -62,6 +63,7 @@ defmodule InFlightExitsTests do
   @gas_process_exit_price 1_000_000_000
 
   setup do
+    Reorg.finish_reorg()
     # as we're testing IFEs, queue needs to be empty
     0 = get_next_exit_from_queue()
```
InoMurko, line 68: I think this is still a problem with eth hash nodes returning revert reasons rather then 0 right?

docker-compose-reorg.yml
ayrat555: @InoMurko I just tried overriding docker-compose file. It seems `depends_on` can be overridden.
https://github.com/docker/compose/issues/6426 It just merges values.

priv/cabbage/apps/itest/test/itest/in_flight_exits_test.exs
ayrat555: I added a hack for it https://github.com/omgnetwork/elixir-omg/pull/1579/commits/1ef4b25e4bf106cee077dc17e54a8d3b814058f4

priv/cabbage/apps/itest/test/itest/deposits_test.exs
ayrat555: I tried to make it like
```elixir
Reorg.start_reorg(fn ->
...
end)
```
But it doesn't seem to work: reorgs just doesn't have enough time to be triggered. Because the anonymous function finishes its execution faster than pausing/unpausing operations are performed. 
I think it's logical to leave it in the setup callback.

priv/cabbage/apps/itest/test/itest/configuration_api_test.exs
ayrat555: I tried to add `Reorg.finish_reorg()` to` on_exit`. Reorgs started happening less consistently. Also, more tests started failing

priv/cabbage/apps/itest/test/itest/in_flight_exits_test.exs
ayrat555: @InoMurko It seems event with this hack, in-flight tests still fail

ayrat555:
The last couple of days I’ve been investigating cabbage tests in  the reorgs setup. Some tests are very flaky. Sometimes they pass, sometimes they fail. I think it happens because  their transactions sometimes belong to blocks that were reorged. For example:
```
  3) test 1. Alice starts a Standard Exit for ERC-20 cabbage_test (StandardExitsTests)
     apps/itest/test/itest/standard_exit_test.exs:11
     Assertion with == failed
     code:  assert eth_balance == state.alice_initial_eth_balance - state.gas_used
     left:  99974425620000000000
     right: 99988425620000000000
     stacktrace:
       test/itest/standard_exit_test.exs:120: (test)
```
So I think tests can not survive reorgs consistently. 


I started skipping them. If people will want to test something specific to reorgs, reorg setup already will be there. They can even extend the logic of reorgs (Currently I start reorgs after the setup callback, but it can be started in any period of time)

priv/cabbage/apps/itest/lib/gas.ex
```diff
@@ -38,4 +38,24 @@ defmodule Itest.Gas do
...
+
+  def with_retries(func, total_time \\ 510, current_time \\ 0) do
+    case func.() do
+      {:ok, value} = result ->
+        if is_nil(value) do
+          Process.sleep(1_000)
+          with_retries(func, total_time, current_time + 1)
+        else
+          result
+        end
```
InoMurko, line 50: why not?
```suggestion
      {:ok, nil} = result ->
          Process.sleep(1_000)
          with_retries(func, total_time, current_time + 1)
      {:ok, _} = result -> result
```

priv/cabbage/apps/itest/lib/poller.ex
```diff
@@ -295,6 +291,10 @@ defmodule Itest.Poller do
         Process.sleep(@sleep_retry_sec)
         submit_typed(typed_data_signed, counter - 1)
 
+      %{"messages" => %{"code" => "operation:service_unavailable"}} ->
```
InoMurko, line 294: hmmm. why does this happen and when? this particular response means there's something wrong on the client side. for example - alarms are raised.

priv/cabbage/apps/itest/lib/reorg.ex
```diff
@@ -0,0 +1,136 @@
...
+  @node1 "geth-1"
+  @node2 "geth-2"
+  @pause_seconds 100
+
+  def finish_reorg() do
+    if Application.get_env(:cabbage, :reorg) do
+      unpause_container!(@node1)
+      unpause_container!(@node2)
+
+      Process.sleep(60 * 1000)
```
InoMurko, line 32: where does this number come from?

priv/cabbage/apps/itest/test/itest/deposits_test.exs
InoMurko: Perhaps the `start_reorg` function that executes the anonymous function can wait until reorgs start to appear, and then execute it's anonymous function that comes in as argument?

.circleci/config.yml
```diff
@@ -526,6 +529,109 @@ jobs:
...
+          command: docker-compose -f docker-compose-reorg.yml logs --follow | grep "reorg"
+          background: true
+      - run:
+          name: Run specs
+          command: |
+            cd priv/cabbage
+            make install
+            make generate_api_code
+            mix deps.get
+            mix test --exclude in_flight_exits --exclude invalid_standard_exits_tests --exclude transactions --exclude standard_exits_tests --exclude watcher_status --exclude watcher_info_api --trace
```
InoMurko, line 593: before we commit to this, I think we should figure out why things don't work here. I can take a look at this, but let's first nail the design of reorg invoking (comments bellow).

docker-compose-reorg.yml
InoMurko: what if we put nginx in docker-compose.yml?

priv/cabbage/apps/itest/lib/reorg.ex
ayrat555: After reorg finishes, nodes need time to connect to each other, it doesn't happen instantly. I set this value based on failing tests

priv/cabbage/apps/itest/test/itest/deposits_test.exs
ayrat555: @InoMurko It's not obvious to me what to wrap in this anonymous function - defgiven, defwhen, defthen?

priv/cabbage/apps/itest/lib/reorg.ex
InoMurko: Perhaps we can figure out how to make this more dynamic. Can we detect this this somehow? Parse logs?

priv/cabbage/apps/itest/test/itest/deposits_test.exs
InoMurko: aha! lemme give you an example...

docker-compose-reorg.yml
ayrat555: but it still will depend on geth. but in this docker compose geth-1 and geth-2 are used

InoMurko: yeah, so we put nginx in docker-compose.yml in front of the single -non-reorg- geth.
and nginx here points to two geths, and in docker-compose.yml to single geth

priv/cabbage/apps/itest/test/itest/deposits_test.exs
InoMurko: so for example (looking at deposits_test.exs):
```
defwhen ~r/^Alice deposits "(?<amount>[^"]+)" ETH to the root chain$/,
          %{amount: amount},
          %{alice_account: alice_account} = state do
    initial_balance = Itest.Poller.root_chain_get_balance(alice_account)
Itest.Reorg.do_it(fn-> 
    {:ok, receipt_hash} =
      amount
      |> Currency.to_wei()
      |> Client.deposit(alice_account, Itest.PlasmaFramework.vault(Currency.ether()))
end)
    gas_used = Client.get_gas_used(receipt_hash)

    {_, new_state} =
      Map.get_and_update!(state, :gas, fn current_gas ->
        {current_gas, current_gas + gas_used}
      end)

    balance_after_deposit = Itest.Poller.root_chain_get_balance(alice_account)

    state = Map.put_new(new_state, :alice_ethereum_balance, balance_after_deposit)
    {:ok, Map.put_new(state, :alice_initial_balance, initial_balance)}
  end
```

InoMurko: this way I would make sure a deposit happens on one of the blocks that gets reorged, and I prove that the childchain and watchers detect the deposit on the correct block thats also reorg protected (because the reorg is already over!)

ayrat555: @InoMurko I've done exactly that for deposits tests. Please take a look.

.circleci/config.yml
```diff
@@ -526,6 +529,109 @@ jobs:
...
+          command: docker-compose -f docker-compose-reorg.yml logs --follow | grep "reorg"
+          background: true
+      - run:
+          name: Run specs
+          command: |
+            cd priv/cabbage
+            make install
+            make generate_api_code
+            mix deps.get
+            mix test --exclude in_flight_exits --exclude invalid_standard_exits_tests --exclude transactions --exclude standard_exits_tests --exclude watcher_status --exclude watcher_info_api --exclude configuration_retrieval --exclude service_name --exclude fee_claiming --trace
```
achiurizo, line 593: why all the excludes? What is left for `mix test` to run? Could this just be something like

```suggestion
            mix test --only some_tag --trace
```

priv/cabbage/Makefile
```diff
@@ -14,6 +14,12 @@ start_daemon_services:
 	cd priv/cabbage/ && \
 	docker-compose -f ../../docker-compose.yml -f docker-compose-cabbage.yml up -d
 
+start_daemon_services_reorg:
+	cd ../../ && \
+	make init_test_reorg && \
+	cd priv/cabbage/ && \
+	docker-compose -f ../../docker-compose-reorg.yml -f docker-compose-cabbage.yml up -d
```
achiurizo, line 21: Are we using `docker-compose-reorg.yml` much outside of the `cabbage` repo/context? I've only noticed one instant where [we use it to output logs for CI](https://github.com/omgnetwork/elixir-omg/pull/1579/files#diff-1d37e48f9ceff6d8030570cd36286a61R584). Perhaps we move it into this repo?

.circleci/config.yml
ayrat555: I added `reorg` tag

priv/cabbage/Makefile
ayrat555: I move it into the `cabbage` directory

priv/cabbage/apps/itest/lib/poller.ex
ayrat555: I removed it for now. It may be caused by non-deterministic nature of reorgs. I hope the error will disappear when using Reorg.execute_in_reorg function

priv/cabbage/apps/itest/lib/reorg.ex
ayrat555: I tried decreasing this number. Logs show that reorg happened, but the node still return 502

docker-compose-reorg.yml
ayrat555: I added nginx to docker-compose.yml

docker-compose.yml
```diff
@@ -14,6 +14,9 @@ services:
       interval: 5s
       timeout: 3s
       retries: 5
+    networks:
+      chain_net:
+        ipv4_address: 172.27.0.106
```
unnawut, line 19: Is changing to ip addresses needed for postgres, childchain and watchers?

ayrat555: @unnawut Initially, I added ip for the Geth node so can define Nginx config. But after adding Nginx and Geth to the same network, they became unavailable for other containers. That's why I added all containers to the same network

unnawut: Thanks for the clarification!

priv/cabbage/apps/itest/lib/account.ex
```diff
@@ -79,9 +94,36 @@ defmodule Itest.Account do
...
+      end)
+
+      with_retries(fn ->
+        Ethereumex.HttpClient.request("personal_importRawKey", [secret, passphrase], url: "http://localhost:9001")
+      end)
+    else
+      with_retries(fn ->
+        Ethereumex.HttpClient.request("personal_importRawKey", [secret, passphrase], [])
+      end)
+    end
```
InoMurko, line 112: remind me why do we do this again? is it so that we unlock accounts on both nodes?

priv/cabbage/apps/itest/lib/reorg.ex
```diff
@@ -0,0 +1,163 @@
...
+
+      Process.sleep(@pause_seconds * 1000)
+    end
+  end
+
+  def start_reorg() do
+    if Application.get_env(:cabbage, :reorg) do
+      GenServer.cast(__MODULE__, :reorg_step1)
+    end
+  end
```
InoMurko, line 40: do we need these two?

priv/cabbage/docker-compose-reorg.yml
```diff
@@ -0,0 +1,108 @@
...
+
+  geth-2:
+    image: ethereum/client-go:v1.9.15
+    container_name: geth-2
+    depends_on:
+      - geth-1
+    environment:
+      - ACCOUNT=0xc0f780dfc35075979b0def588d999225b7ecc56f
+      - BOOTNODES=enode://4574f825d67bf570b9216e704a5b761d05d5015c458e2c9dd4b30abb2fe8c881400c2074a126df94690c4c9fb72ee046e6e3ac2bb73dede42fce66cb0a963b36@172.25.0.102:30303
+      - INIT=false
```
InoMurko, line 41: what does this do?
image

priv/cabbage/apps/itest/lib/account.ex
ayrat555: yes

InoMurko: is there an alternative way of achieving the same thing? this looks kinda brute force 🗡️

priv/cabbage/docker-compose-reorg.yml
ayrat555: the same `command` is used when contracts are deployed to the snapshot. If this variable is true, the DB is initialized using genesis file

InoMurko: ahaa, thanks for clarifying!

priv/cabbage/apps/itest/lib/reorg.ex
ayrat555: you're right I removed them. Also, I removed genserver since it's not used anymore

ayrat555:
@InoMurko I think I found a way to make it more reproducible. Deposit tests call pull_balance_until_amount which recursively checks if the current balance on the watcher is equal to the balance that the test expects. So event if the balance at some point is inconsistent because of reorg it is fixed after some time of pulling.
instead of pulling for the balance that test expects we can pull for inconsistent balance. so now I’m pulling for balance > than the balance that test expects.
what do you think?

InoMurko:
> instead of pulling for the balance that test expects we can pull for inconsistent balance. so now I’m pulling for balance > than the balance that test expects.

ahaaa. oh great catch!
what if you do this:
- all services (including watcher) exposes an API endpoint callled `request GET configuration.get` with schema:
```
deposit_finality_margin:
                              type: integer
                              format: int256
                            contract_semver:
                              type: string
                            exit_processor_sla_margin:
                              type: integer
                              format: int256
                            network:
                              type: string
```

as you can see we expose `deposit_finality_margin`. So the idea is:
- you store the ethereum height
- perform the `execute_in_reorg/1`
- compare the current ethereum height with the height in the first step
- if the difference is more then the `deposit_finality_margin` pull balance, else, wait for N additional blocks until you're pass `deposit_finality_margin`

We do something similar in `in_flight_exits_test.exs`:
```
{:ok, response} =
      WatcherSecurityCriticalAPI.Api.Configuration.configuration_get(WatcherSecurityCriticalAPI.Connection.new())

    watcher_security_critical_config =
      WatcherSecurityCriticalConfiguration.to_struct(Jason.decode!(response.body)["data"])

    finality_margin_blocks = watcher_security_critical_config.deposit_finality_margin
    to_miliseconds = 1000

    finality_margin_blocks
    |> Kernel.*(geth_block_every)
    |> Kernel.*(to_miliseconds)
    |> Kernel.round()
    |> Process.sleep()
```
wdyt?

ayrat555:
@InoMurko  I finally made test failures consistent by waiting nodes to reconnect to each by calling `net_peerCount`. https://github.com/omgnetwork/elixir-omg/pull/1579/commits/32e23948c3bda538ecf904ebde1b94323ae36ef9 
Is this last step? PR's good to merge?

priv/cabbage/apps/itest/lib/client.ex
```diff
@@ -71,18 +71,27 @@ defmodule Itest.Client do
...
+        {:ok, [sign_hash, typed_data, txbytes]}
+
+      %{"code" => "create:client_error", "messages" => %{"code" => "operation:service_unavailable"}} = result ->
+        if tries == 0 do
+          result
+        else
+          Process.sleep(1_000)
+          create_transaction(amount_in_wei, input_address, output_address, currency, tries - 1)
+        end
+    end
```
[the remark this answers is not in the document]
InoMurko: why does `operation:service_unavailable` happen?

priv/cabbage/apps/itest/lib/client.ex
```diff
@@ -157,8 +166,32 @@ defmodule Itest.Client do
     |> ExRLP.encode()
   end
 
-  def wait_until_tx_sync_to_watcher(tx_id) do
-    do_wait_until_tx_sync_to_watcher(tx_id, @default_retry_attempts)
+  def wait_until_tx_sync_to_watcher(tx_id, retry_attempts \\ nil) do
```
InoMurko, line 169: is this being used?

InoMurko: if yes, where, I can't find any reference to it. (specifically wait_until_tx_sync_to_watcher/2)

priv/cabbage/apps/itest/lib/poller.ex
InoMurko: it didn't dissapear?

priv/cabbage/apps/itest/lib/account.ex
```diff
@@ -79,9 +87,30 @@ defmodule Itest.Account do
...
+      Reorg.create_account_from_secret(secret, passphrase)
+    else
+      with_retries(fn ->
+        Ethereumex.HttpClient.request("personal_importRawKey", [secret, passphrase], [])
+      end)
+    end
   end
 
-  defp hash(message), do: ExthCrypto.Hash.hash(message, ExthCrypto.Hash.kec())
+  def with_retries(func, total_time \\ 510, current_time \\ 0) do
```
InoMurko, line 102: mixing private and public functions.
also, what is the reason we need to `retry`. Is it so that all nodes unlock an account? is there a better way to do this?

priv/cabbage/apps/itest/lib/client.ex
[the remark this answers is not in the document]
ayrat555: it's happening during reorgs. when one node is paused. I think it happens because Nginx continues to redirect to paused node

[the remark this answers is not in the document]
InoMurko: aha, so perhaps an alarm is raised (ethereum connection error?)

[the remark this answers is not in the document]
ayrat555: I can't see any alarms. but in Nginx logs I see
```
nginx           | 2020/07/06 11:36:53 [error] 29#29: *37837 upstream timed out (110: Connection timed out) while reading response header from upstream, client: 172.25.0.105, server: , request: "POST / HTTP/1.1", upstream: "http://172.25.0.103:8545/", host: "172.25.0.104:80"
```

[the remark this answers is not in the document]
InoMurko: Alarms would be raised on enpoints watchers and childchain enpoints: `/alarm.get`. Perhaps logged as well.

priv/cabbage/apps/itest/lib/account.ex
ayrat555: I removed it. It was needed when accounts were unlocked in reorg

priv/cabbage/apps/itest/lib/client.ex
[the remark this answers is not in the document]
ayrat555: I think the issue is in Nginx. during reorgs when one node is paused, Nginx still tries to access it

priv/cabbage/apps/itest/lib/account.ex
ayrat555: I didn't find any other way. It is related to Nginx timeout. Some requests can no finish in 10 seconds. I set it 10 because during reorg Nginx continues to redirect to the paused node (til the first failure?), paused node accepts a request but do not return any response for N (timeout limit) seconds

priv/cabbage/apps/itest/lib/poller.ex
ayrat555: no, I think it's related to Nginx and timeout limit to 10 seconds

.circleci/config.yml
```diff
@@ -526,6 +529,70 @@ jobs:
...
+            - ~/.asdf
+            - ~/.asdf-vm
+      - run: make install-hex-rebar
+      - run: sh .circleci/status.sh
+      - restore_cache:
+          key: v2-mix-specs-cache-{{ .Branch }}-{{ checksum "mix.lock" }}
+      - run:
+          name: Print reorg logs
+          command: docker-compose -f docker-compose.yml -f ./priv/cabbage/docker-compose-reorg.yml -f ./priv/cabbage/docker-compose-cabbage.yml logs --follow | grep "reorg"
+          background: true
```
InoMurko, line 585: can you do the same for childchain, watcher and watcher info (separately?)

priv/cabbage/apps/itest/lib/plasma_framework.ex
```diff
@@ -96,4 +100,19 @@ defmodule Itest.PlasmaFramework do
       Map.put(acc, key, value)
     end)
   end
+
+  def with_retries(func, total_time \\ 120, current_time \\ 0) do
```
InoMurko, line 104: mixing private and public functions

priv/cabbage/apps/itest/test/itest/deposits_test.exs
```diff
@@ -33,9 +35,11 @@ defmodule DepositsTests do
...
 
     {:ok, receipt_hash} =
-      amount
-      |> Currency.to_wei()
-      |> Client.deposit(alice_account, Itest.PlasmaFramework.vault(Currency.ether()))
+      Reorg.execute_in_reorg(fn ->
+        amount
+        |> Currency.to_wei()
+        |> Client.deposit(alice_account, Itest.PlasmaFramework.vault(Currency.ether()))
+      end)
```
InoMurko, line 42: after the anon. function, the `balance = Client.get_exact_balance(alice_account, expecting_amount)` would def. return the correct and LAST balance because the reorg is at this point finished, correct?

priv/cabbage/apps/itest/lib/reorg.ex
```diff
@@ -0,0 +1,121 @@
...
+      Process.sleep(30_000)
+
+      unpause_container!(@node2)
+      unpause_container!(@node1)
+
+      :ok = Poller.wait_until_peer_count(1)
+
+      Process.sleep(20_000)
+
+      response
```
InoMurko, line 56: I don't like these `sleeps`, because they're undeterministic. Is there a way for us to make this less prone to error?

ayrat555: I can request the block number from nodes and wait for 5 generated blocks. I'll do this.

priv/cabbage/apps/itest/test/itest/deposits_test.exs
ayrat555: not really, it takes some time for childchain to sync the latest data

InoMurko:
I've noticed another issue:
https://github.com/omgnetwork/elixir-omg/blob/f713eb65046eac7e86508de157106c59cdad76c0/priv/cabbage/apps/itest/test/itest/deposits_test.exs#L61-L88

`geth_block_every = 1` is basically - we assume a block definetly arrives every second (which worked fine with clique).

Perhaps a better approach, scalable at least, would be to get the curent ethereum height and compare it with `curent ethereum height + watcher_security_critical_config.deposit_finality_margin`, when the current ethereum height is above `curent ethereum height + watcher_security_critical_config.deposit_finality_margin` you're allowed to check for balance.

Or a **even better approach**,  would be to get the block number at which the deposit transaction was made and compare it with `curent ethereum height + watcher_security_critical_config.deposit_finality_margin`, when the deposit ethereum height is above `curent ethereum height + watcher_security_critical_config.deposit_finality_margin` you're allowed to check for balance.

ayrat555:
@InoMurko how does it look?

priv/cabbage/apps/itest/lib/poller.ex
InoMurko: so as per slack, this actually is because we're raising alarms on our services - "stalled sync" when reorg is in progress.

PR #1584, opened 2020-06-16
chore: minor fixes

boolafish:
## Overview

Some minors fixes.

## Changes

- better readme for cabbage test running
- update gitignore
- remove deplicated implementation on checking eth balance in transfer test

## Testing

CI

### DIFF
--- a/priv/cabbage/apps/itest/test/itest/transactions_test.exs
+++ b/priv/cabbage/apps/itest/test/itest/transactions_test.exs
@@ -73,46 +73,6 @@ defmodule TransactionsTests do
     {:ok, state}
   end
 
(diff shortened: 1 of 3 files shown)


priv/cabbage/apps/itest/test/itest/transactions_test.exs
```diff
@@ -73,46 +73,6 @@ defmodule TransactionsTests do
     {:ok, state}
   end
 
-  defthen ~r/^they should have "(?<amount>[^"]+)" ETH on the child chain$/,
```
boolafish, line 76: see line: 143. It is the same, and this is actually not used.

InoMurko: good catch, I did this! :D

PR #1585, opened 2020-06-16
docs: Swagger specs for v1.0.0

okalouti:
Deploy latest Swagger specs. Based on the specs found in [v1.0.0](https://github.com/omgnetwork/elixir-omg/tree/v1.0.0)

### DIFF
--- a/docs-ui/1.0.0/operator_api_specs.yaml
+++ b/docs-ui/1.0.0/operator_api_specs.yaml
@@ -1,0 +1,683 @@
+openapi: 3.0.0
+info:
+  version: 1.0.0
+  title: Child chain API
+  description: |
+    This is the documentation for child chain API.
+    Error codes are available in [html](https://github.com/omisego/elixir-omg/blob/master/docs/api_specs/errors.md#error-codes-description) format.
+  contact:
+    name: OmiseGO
+    email: omisego@omisego.co
+  license:
+    name: 'Apache 2.0: https://www.apache.org/licenses/LICENSE-2.0'
+    url: 'https://omisego.network/'
+tags:
+  - name: Block
+    description: Block related API.
+  - name: Transaction
+    description: Transaction related API.
+  - name: Fees
+    description: Fees related API.
+servers:
+  - url: 'http://localhost:9656/'
+paths:
+  /alarm.get:
+    get:
+      tags:
+        - Alarm
+      summary: 'Provides alarms related to system memory, cpu and storage and application specific alarms.'
+      description: |
+        **Note:** Service operator alarms.
+      operationId: alarm_get
+      responses:
+        '200':
+          description: System alarms
+          content:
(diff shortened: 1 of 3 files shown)


docs-ui/1.0.0/operator_api_specs.yaml
```diff
@@ -0,0 +1,683 @@
...
+                      data:
+                        type: object
+                    required:
+                      - version
+                      - success
+                      - data
+                      - service_name
+                    example:
+                      service_name: child_chain
+                      version: 1.0+abcdefa
```
jbunce, line 57: missing a `0` should be `1.0.0+d950ebe`

docs-ui/1.0.0/operator_api_specs.yaml
```diff
@@ -0,0 +1,683 @@
+openapi: 3.0.0
+info:
+  version: 1.0.0
+  title: Child chain API
+  description: |
+    This is the documentation for child chain API.
+    Error codes are available in [html](https://github.com/omisego/elixir-omg/blob/master/docs/api_specs/errors.md#error-codes-description) format.
+  contact:
+    name: OmiseGO
+    email: omisego@omisego.co
```
unnawut, line 10: not sure what we can update this to

docs-ui/1.0.0/operator_api_specs.yaml
```diff
@@ -0,0 +1,683 @@
...
+  license:
+    name: 'Apache 2.0: https://www.apache.org/licenses/LICENSE-2.0'
+    url: 'https://omisego.network/'
```
unnawut, line 13:
```suggestion
    url: 'https://omg.network/'
```

docs-ui/1.0.0/operator_api_specs.yaml
```diff
@@ -0,0 +1,683 @@
+openapi: 3.0.0
+info:
+  version: 1.0.0
+  title: Child chain API
+  description: |
+    This is the documentation for child chain API.
+    Error codes are available in [html](https://github.com/omisego/elixir-omg/blob/master/docs/api_specs/errors.md#error-codes-description) format.
```
unnawut, line 7:
```suggestion
    Error codes are available in [html](https://github.com/omgnetwork/elixir-omg/blob/master/docs/api_specs/errors.md#error-codes-description) format.
```

docs-ui/1.0.0/operator_api_specs.yaml
okalouti: thanks @jbunce. This just serves as an example and doesn't impact anything. It was also there in previous versions.

would change but in order to keep things consistent changing this would require a PR to `v1.0.0` too. 

This PR will get the API changes deployed, but I'd rather leave any further content to @dmitrydao now that we're all familiar with the process.

PR #1589, opened 2020-06-18
test: payment v2 with standard exit

boolafish:
:clipboard: https://github.com/omgnetwork/new-tx-type-poc/issues/5

## Overview

test to ensure payment v2 exit event is handled

## Changes

Add cabbage test. Currently is has shown that child chain and watcher-info somehow can handle 2 payment exit events.

**warning, watcher is not tested!**

Currently, child chain works and watcher can recognize the event and then pass it to watcher info. Since payment v1 and v2 has the same tx and output format, so it works to push the data to DB with different `type` info.

## Testing

CI

### DIFF
--- a/priv/cabbage/apps/itest/test/itest/payment_v2_test.exs
+++ b/priv/cabbage/apps/itest/test/itest/payment_v2_test.exs
@@ -23,111 +23,92 @@ defmodule PaymentV2Test do
 
   @payment_v2_tx_type 2
 
-  setup do
-    [alice, bob] =
-      Enum.map(
-        Account.take_accounts(2),
-        fn {account, pkey} -> %{account: account, pkey: pkey} end
-      )
-
-    %{alice: alice, bob: bob}
+  defgiven ~r/^(?<user>[\w]+) has an ethereum account$/, %{user: user}, state do
+    [{account, pkey}] = Account.take_accounts(1)
+    {:ok, Map.put(state, user, %{account: account, pkey: pkey})}
   end
 
-  defwhen ~r/^Alice deposits "(?<amount>[^"]+)" ETH to the root chain$/,
-          %{amount: amount},
-          %{alice: alice} = state do
+  defwhen ~r/^(?<user>[\w]+) deposits "(?<amount>[^"]+)" ETH to the root chain$/,
+          %{user: user, amount: amount},
+          state do
+    user = state[user]
+
     {:ok, _receipt_hash} =
       amount
       |> Currency.to_wei()
-      |> Client.deposit(alice.account, Itest.PlasmaFramework.vault(Currency.ether()))
+      |> Client.deposit(user.account, Itest.PlasmaFramework.vault(Currency.ether()))
 
     {:ok, state}
   end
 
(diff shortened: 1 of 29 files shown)


priv/cabbage/apps/itest/test/itest/payment_v2_test.exs
```diff
@@ -50,7 +51,10 @@ defmodule PaymentV2Test do
     balance = Client.get_exact_balance(alice.account, expecting_amount)
     balance = balance["amount"]
 
-    assert_equal(expecting_amount, balance, "For #{alice.account}")
+    case expecting_amount do
+      0 -> assert(balance == nil, "Alice ETH amount is not 0")
```
pnowosie, line 55: `balance == nil` is not obvious and misleading.
Can we set it to zero, kind of (line above): `balance = Map.get(balance, "amount", 0)`

priv/cabbage/apps/itest/test/itest/payment_v2_test.exs
```diff
@@ -91,13 +95,48 @@ defmodule PaymentV2Test do
           %{amount: amount},
           %{bob: bob} = state do
     expecting_amount = Currency.to_wei(amount)
+
     balance = Client.get_exact_balance(bob.account, expecting_amount)["amount"]
 
-    assert_equal(expecting_amount, balance, "For Bob: #{bob.account}")
+    case expecting_amount do
```
pnowosie, line 101: Same here

priv/cabbage/apps/itest/test/itest/payment_v2_test.exs
```diff
@@ -91,13 +95,48 @@ defmodule PaymentV2Test do
...
+    assert Itest.Poller.exitable_utxo_absent?(bob.account, state.standard_exit.utxo.utxo_pos)
+    {:ok, state}
+  end
 
+  defwhen ~r/^Alice processes the standard exit on the child chain$/, _, state do
+    StandardExitClient.wait_and_process_standard_exit(state.standard_exit)
     {:ok, state}
   end
 
+  defthen ~r/^Alice should have "(?<amount>[^"]+)" ETH on the child chain after finality margin$/,
```
pnowosie, line 127: Can we parametrize Alice - looks like it will work for any `entity`. Thank you 👍

priv/cabbage/apps/itest/test/itest/payment_v2_test.exs
boolafish: lol, TIL in elixir you can try to access a value from nil and would return nil instead of crashing.
the `Client.get_exact_balance` would actually return `nil` instead of empty map. 

I change to code to the following instead:
```
Client.get_exact_balance(bob.account, expecting_amount)["amount"] || 0
```

priv/cabbage/apps/itest/test/itest/payment_v2_test.exs
boolafish: ok, add the step `Given Alice has an ethereum account` that generates the user account and put it into state instead!

PR #1593, opened 2020-06-18
async stream + timeout

InoMurko:

## Overview

## Changes

- Change 1
- Change 2
- ...

## Testing


apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -310,9 +310,16 @@ defmodule OMG.Watcher.ExitProcessor do
...
-      Enum.map(exits, fn exit_event ->
-        put_timestamp_and_sft(exit_event, state.min_exit_period_seconds, state.child_block_interval)
-      end)
+      exits
+      |> Task.async_stream(
+        fn exit_event ->
+          put_timestamp_and_sft(exit_event, state.min_exit_period_seconds, state.child_block_interval)
+        end,
+        timeout: 50_000,
+        on_timeout: :kill_task,
```
boolafish: are we sure we don't want to exit the process?

boolafish: okay, and I refresh the page and is changed to exit 😅

InoMurko: no :) you we're 32seconds too quick

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -331,12 +338,16 @@ defmodule OMG.Watcher.ExitProcessor do
...
-        events,
+      events
+      |> Task.async_stream(
         fn %{call_data: %{in_flight_tx: txbytes}} ->
           ExPlasma.InFlightExit.txbytes_to_id(txbytes)
-        end
+        end,
+        timeout: 50_000,
+        on_timeout: :exit,
+        max_concurrency: System.schedulers_online() * 2
```
boolafish, line 348: this would not be larger then our rate limit to infura right (is there a rate limit)? (btw, I actually have no idea what both number are 😅)

InoMurko: the beam installs a scheduler per CPU core. so if you have 8 cores times 2, you'd do at most 16 exits at the same time.

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -310,9 +310,16 @@ defmodule OMG.Watcher.ExitProcessor do
...
     exit_maps =
-      Enum.map(exits, fn exit_event ->
-        put_timestamp_and_sft(exit_event, state.min_exit_period_seconds, state.child_block_interval)
-      end)
+      exits
+      |> Task.async_stream(
+        fn exit_event ->
+          put_timestamp_and_sft(exit_event, state.min_exit_period_seconds, state.child_block_interval)
+        end,
+        timeout: 50_000,
```
unnawut, line 318: Oh this `Task.async_stream`'s `:timeout` is very handy 👏

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -331,12 +338,16 @@ defmodule OMG.Watcher.ExitProcessor do
     _ = if not Enum.empty?(events), do: Logger.info("Recognized in-flight exits: #{inspect(events)}")
 
     contract_ife_ids =
-      Enum.map(
-        events,
+      events
+      |> Task.async_stream(
```
unnawut, line 342: `ExPlasma.InFlightExit.txbytes_to_id(txbytes)` is already a local computation though, no longer making RPC calls. 

Maybe at least we can remove the custom timeout and so defaults back to 5000ms which is plenty enough?

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -175,7 +175,7 @@ defmodule OMG.Watcher.ExitProcessor do
   but under unchanged conditions, it should have unchanged behavior from POV of an outside caller.
   """
   def check_validity() do
-    GenServer.call(__MODULE__, :check_validity)
+    GenServer.call(__MODULE__, :check_validity, 60_000)
```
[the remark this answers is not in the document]
InoMurko: yeah!

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
InoMurko: oh geez. I didn't even check. Now I remember you did this!

PR #1595, opened 2020-06-19
Add Transaction filter by end_datetime

jarindr:
Adding filter for finance report on all transaction endpoint on watcher info

## Overview

Add Transaction filter by end_datetime

## Changes

- Add Transaction filter by end_datetime
- Add test

## Testing

`mix test`


unnawut:
dude it doesnt even run yet

priv/cabbage/apps/itest/test/itest/watcher_info_api_test.exs
```diff
@@ -91,6 +137,17 @@ defmodule WatcherInfoApiTest do
     Itest.Poller.pull_balance_until_amount(address, amount)
   end
 
+  defp is_all_tx_behind_timestamp(transactions, timestamp) do
+    Enum.reduce(
```
boolafish, line 141: probably `Enum.any?` would be better to read

jarindr: I was looking for something like that lol thanks

unnawut: I'm late 😅  and nitpicking and pedantic 😅 😅  How about `Enum.all?` to avoid the brain fatigue from the negation.

If we're worried of full list iteration perf, it's already taken care off by the implementation:

> Iterates over the `enumerable` and invokes `fun` on each element. When an invocation of `fun` returns a falsy value (`false` or `nil`) iteration stops immediately and `false` is returned. In all other cases `true` is returned.
>
> https://hexdocs.pm/elixir/Enum.html#all?/2

jarindr: yey all is better

priv/cabbage/apps/itest/lib/client.ex
```diff
@@ -106,6 +116,27 @@ defmodule Itest.Client do
     {:ok, data}
   end
 
+  def get_transactions(params) do
+    default_paging = %{page: 1, limit: 200}
```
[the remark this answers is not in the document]
jarindr: purposely put it there so the test doesn't need to specify it every time and it is obvious when you come back to see how many it is calling by default

priv/cabbage/apps/itest/lib/client.ex
```diff
@@ -95,6 +99,12 @@ defmodule Itest.Client do
...
   end
 
+  def submit_transaction_and_wait(typed_data, sign_hash, private_keys) do
+    tx = submit_transaction(typed_data, sign_hash, private_keys)
+    :ok = wait_until_tx_sync_to_watcher(tx.txhash)
+    tx
+  end
+
   def get_utxos(params) do
     default_paging = %{page: 1, limit: 200}
```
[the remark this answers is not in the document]
jarindr: same as above

jarindr:
> unit test is missing?

added

priv/cabbage/apps/itest/test/itest/watcher_info_api_test.exs
```diff
@@ -26,9 +26,10 @@ defmodule WatcherInfoApiTest do
   @to_milliseconds 1000
```
InoMurko, line 26: can you rename this test and it's feature file?

jarindr: done

PR #1598, opened 2020-06-21
ecto APM in DD

InoMurko:


ayrat555:
@InoMurko why did you close it?
On  the weekends, I was debugging spandex_ecto on childchain. I found that with default instructions from README it doesn't work, because current_tract_id is always nil. I found a workaround. I'll update my pr today

PR #1609, opened 2020-06-25
Update request body swagger

jarindr:
## Overview

In order to keep the Swagger specs up to date with ou API, we need to update the parameters for the `transaction.all` endpoint. The new param is `end_datetime` which can be used to pull transactions between specific timeframe.


unnawut:
Can you try compile the specs? i.e. https://github.com/omgnetwork/elixir-omg/blob/master/apps/omg_watcher_rpc/priv/swagger/swagger.md

Otherwise it's a source file not picked up by all the toolings

jarindr:
> Can you try compile the specs? i.e. https://github.com/omgnetwork/elixir-omg/blob/master/apps/omg_watcher_rpc/priv/swagger/swagger.md
> 
> Otherwise it's a source file not picked up by all the toolings

recompiled

PR #1612, opened 2020-06-26
fix: prevent error log truncation

### LINKED ISSUE #1571: Missing log metadata when json-encoding the log failed
This is sent to the log collector without any metadata:

```
{:error, %Jason.EncodeError{message: "invalid byte 0x9D in <<67, ...>>"}}
```

This specific issue will be resolved by #1532, but nevertheless the metadata should still be sent in case of any error with log encoding

unnawut:
Closes #1571 

## Overview

Uses our forked `Ink` library that can display full log instead of truncated like `{:error, %Jason.EncodeError{message: "invalid byte 0x9D in <<67, ...>>"}}`.

https://github.com/omgnetwork/ink/tree/unnawut/full-log-on-error

## Changes

- Replace `:ink` with our forked repo that supports `error_inspect_limit: :infinity`
- Use `error_inspect_limit: :infinity`
- Rename `github.com/omisego` -> `github.com/omgnetwork` because I was touching the deps anyway

## Testing

Logs that have encoding error should no longer truncate its content


apps/omg_status/mix.exs
```diff
@@ -35,11 +35,11 @@ defmodule OMG.Status.Mixfile do
     do: [
       {:telemetry, "~> 0.4.1"},
       {:sentry, "~> 7.0"},
-      {:statix, git: "https://github.com/omisego/statix.git", branch: "otp-21.3.8.4-support-global-tag-patch"},
+      {:statix, git: "https://github.com/omgnetwork/statix.git", branch: "otp-21.3.8.4-support-global-tag-patch"},
       {:spandex_datadog, "~> 0.4"},
       {:decorator, "~> 1.2"},
       {:vmstats, "~> 2.3", runtime: false},
-      {:ink, "~> 1.1"},
+      {:ink, git: "https://github.com/omgnetwork/ink.git", branch: "unnawut/full-log-on-error"},
```
ayrat555, line 42: I saw this comment https://github.com/ivx/ink/pull/63#issuecomment-650132999 . Does it mean this PR is not final?

unnawut: So this is from the discussion at the last weekly sync to have the changes applied and running on our fork first. Then apply the PR upstream

InoMurko:
FYI, logs (logger) is truncating logs as part of the backpressure mechanism. if this wouldn't be there and if the amount of incoming messages with large binaries would be high, the only remaining backpressure mechanism logger has is dropping messages.

unnawut:
> FYI, logs (logger) is truncating logs as part of the backpressure mechanism. if this wouldn't be there and if the amount of incoming messages with large binaries would be high, the only remaining backpressure mechanism logger has is dropping messages.

I'll check what's the ideal length or there's another solution for `length: 50` not being enough

unnawut:
Looks like a more useful approach is to figure out a way to log the stacktrace on log encoding error. Will resubmit as a new PR.

PR #1613, opened 2020-06-30
chore: bump version in VERSION file

boolafish:
OOOPs


InoMurko:
I think your work is only getting better!

PR #1617, opened 2020-07-01
fix: handle "transaction underpriced" and other unknown server error responses

### LINKED ISSUE #1615: Handle "transaction underpriced" error response from Ethereum client
Under some high gas price conditions, the Ethereum client could return `%{"code" => -32000, "message" => "transaction underpriced"}` which is currently not handled by `BlockSubmission.process_result/3`

---

Sentry Issue: [ELIXIR-OMG-4H](https://sentry.io/organizations/omisego/issues/1589562121/?referrer=github_integration)

```
FunctionClauseError: no function clause matching in OMG.ChildChain.BlockQueue.Core.process_submit_result/3
  File "lib/omg_child_chain/block_queue/core.ex", line 247, in OMG.ChildChain.BlockQueue.Core.process_submit_result/3
  File "lib/omg_child_chain/block_queue.ex", line 212, in OMG.ChildChain.BlockQueue.submit/1
  File "lib/enum.ex", line 783, in Enum."-each/2-lists^foreach/1-0-"/2
  File "lib/enum.ex", line 783, in Enum.each/2
  File "lib/omg_child_chain/block_queue.ex", line 180, in OMG.ChildChain.BlockQueue.handle_info/2
...
(3 additional frame(s) were not displayed)

(FunctionClauseError) no function clause matching in OMG.ChildChain.BlockQueue.Core.process_submit_result/3
```

TODOs:
- [x] Handle `"transaction underpriced"` response in [`OMG.ChildChain.BlockQueue.Core`](https://github.com/omgnetwork/elixir-omg/blob/v1.0.1/apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex#L252-L341)
- [x] Catch-all [`-32000 to -32099`](https://eth.wiki/json-rpc/json-rpc-error-codes-improvement-proposal) error response. We don't really need to crash the BlockQueue just for log message mismatch
- [ ] `:block_submission_stalled` alarm

unnawut:
Closes #1615 

## Overview

Adds "transaction underpriced" and unknown server error response to block submission response handling, preventing unnecessary BlockQueue crashes.

## Changes

- Add "transaction underpriced" response handling to `BlockQueue.Core.process_submit_result/3`
- Add `code >= -32_099 and code <= -32_000` server error fallback
- Minimum code changes without refactoring because refactoring the `BlockQueue` already started as part of https://github.com/omgnetwork/elixir-omg/pull/1575

## Testing

On local environments, set [geth `--miner.gas_price`](https://github.com/omgnetwork/elixir-omg/blob/master/docker-compose.yml#L35) to a very high value e.g. 200Gwei. Then trigger a block submission. The error should be handled and logged.


apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex
```diff
@@ -340,6 +351,15 @@ defmodule OMG.ChildChain.BlockQueue.Core do
     :ok
   end
 
+  # Fallback for unknown server errors: https://eth.wiki/json-rpc/json-rpc-error-codes-improvement-proposal
+  # Only server errors are handled as they are the only set of errors that we have no control of.
+  # Returns `:ok` so that BlockQueue can continue and do a retry, similar to a low replacement price error.
+  def process_submit_result(submission, {:error, %{"code" => code} = error}, _newest_mined_blknum)
+      when code >= -32_099 and code <= -32_000 do
+    _ = Logger.error("Submission #{inspect(submission)} resulted in an unknown server error: #{inspect(error)}.")
```
boolafish, line 359: Do we alarm on all error log. If not let's make sure we add alarm on this, either via Datadog parsing the error message or do the direct alarm (?) from code?

apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex
```diff
@@ -616,6 +636,11 @@ defmodule OMG.ChildChain.BlockQueue.Core do
     :ok
   end
 
+  defp log_tx_underpriced(submission) do
+    _ = Logger.debug("Submission #{inspect(submission)} got rejected due to low gas price - ignored")
```
boolafish, line 640: I again don't think debug log brings much value (probably 0 in this case as this would probably only happen on mainnet env).

unnawut: whoops

unnawut: d7b5d12

apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex
unnawut: Good idea! I think we can do with an app alarm when a block submission is pending for more than 2 blocks (since we're aiming for next-block inclusion, but I'll make number configurable in case of too many false panics). Then Datadog can consume this app alarm similar to `:ethereum_stalled_sync` and friends.

Adding now to this PR.

unnawut:
Adding @InoMurko as reviewer as this relates to block submission

apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex
InoMurko: > Since we're losing this visibility because this PR is suppressing the exception, I'll add :block_submission_stalled alarm to this PR.


Queue observability size. Hell yeah!

apps/omg_child_chain/lib/omg_child_chain/block_queue/core.ex
```diff
@@ -281,6 +281,17 @@ defmodule OMG.ChildChain.BlockQueue.Core do
     :ok
   end
 
+  # https://github.com/ethereum/go-ethereum/blob/v1.9.15/core/tx_pool.go#L62-L64
+  def process_submit_result(
```
InoMurko, line 285: I'm not entirely sure why do we even bother "handling" results. We don't do anything with them, we're just spending effort defining all function headers and then we ... log and return :ok?
I've never understood the purpose.

unnawut: I was wondering the same. It turns about the block queue uses these :ok, {:error, ...} returns to determine if the submission was successful or if it needs to retry the submission :(

I don't like it too but it's not a small refactor. I'm kind of unbundling this in https://github.com/omgnetwork/elixir-omg/pull/1575 though

unnawut:
I'll follow up with :block_submission_stalled in a separate PR

PR #1618, opened 2020-07-03
fix: stop load test when config have float number

boolafish:
## Overview

When there is float number, the transaction can fail due to the function mathching with integer only.
eg. 15.0 should become 15 before passing into the test.

## Testing


boolafish:
ah...just saw the latest PR for load test has trunc stuff already. closing this PR

PR #1620, opened 2020-07-06
Multiple instances of key-value database (rocksdb)

pnowosie:
Solves omgnetwork/new-tx-type-poc#12

This is a continuation of #1570 but being merged to `payment_v2_poc` branch instead of `master`.

❗ Note: We're not ready to merge it to `master` because it changes the directory structure of the `rocksdb`. E.g. to have child_chain running we need to stop the service na copy files from `DB_PATH` to `DB_PATH/default`. We need to do the same for Watcher (and Watcher-Info) but also initialize database for `exit_processor` instance.

## Overview

This PR introduces multi-instance OMG.DB databases for Watcher (and Watcher-Info, hence it requires Watcher in cabbage tests).
`OMG.DB.init(path)` function was extended, so under the `path` it creates `default` folder for default database and allows for additional instances. Watcher introduces another instance for Exit processors, called `OMG.DB.Instance.ExitProcessor`.
To initialize database storage for Watcher we're using `OMG.DB.init/2` passing path and instances which in this case are:
`[OMG.DB.Instance.Default, OMG.DB.Instance.ExitProcessor]`. Under the `path` it creates 2 folders `default` and `exit_processor`.
Appliction environment variable `{:omg_db, :path}` is pointing at base folder (one containing `default`) - so in this aspect it wasn't change. However `OMG.DB` module handles paths when connecting to the particular instance with `child_spec` or `start_link`.
I refactored the existing code so it does not use `OMG.DB.RocksDB` module directly.

## Changes

- Allows for multiple storages and instances of key-value database (rocksdb).
- During initialization and process start one need to provide base path (as now in `DB_PATH` env var) and `OMG.DB` takes care of the instances' paths.

## Testing

Test were adjusted. Added new tests for paths manipulation.


pnowosie:
Some review comments to #1570 have suggested to remove manipulation of the App's environment. To do that has been suggested to remove `omg_db` application from auto-start in `Mix.Project` and then start db instances in Watcher supervisor.

I'd happy to align to this. However as I discovered auto-start from `Mix.Project` already setups db before all other apps depending on it, and this is not only the Watcher but also `omg` and `omg_eth` that Watcher depends on.

To remove App's env variable we would need to start also these apps on Watcher supervisor to remain proper start order...

As an alternative I can see removing `omg_db` application (which I think @InoMurko has suggested) and then each umbrella app that needs db should start it in its supervisor (if it's not started already)...

I'd rather focus on the above in a separate task as its not related directly to the database multi-instance.
WDYT @unnawut , @boolafish

pnowosie:
Other comments I'm planning to align here are:

- remove `OMG.DB.start_link` as it's only for tests ([comment](https://github.com/omgnetwork/elixir-omg/pull/1570#discussion_r446810975))
- make private `OMG.DB.prepare_args` ([comment](https://github.com/omgnetwork/elixir-omg/pull/1570#discussion_r446994366))
- move exit_processor instance related queries to separate module ([comment](https://github.com/omgnetwork/elixir-omg/pull/1570#discussion_r446988994))

I'm not going to make `join_path` private [as suggested in](https://github.com/omgnetwork/elixir-omg/pull/1570#discussion_r446994648)). I need it to be well test as its critical for the implemented feature.

CC: @boolafish , @unnawut

pnowosie:
Regarding 
> move exit_processor instance related queries to separate module 

I'll skip this one for now - exit processor is using just generic DB functions - unless we'll get a direct need to refactor this.

PR #1622, opened 2020-07-06
enable margin

InoMurko:

## Overview

## Changes

- Change 1
- Change 2
- ...

## Testing


boolafish:
hmmm, I think the real bug is in this line: https://github.com/omgnetwork/elixir-omg/blob/ca108e0c974170d48f9f4b9db6561b9a4f19335c/apps/omg/lib/omg/ethereum_event_listener/core.ex#L115

It should be using `min` instead of `max`? The `sync_height` would take `wait_for` into account: https://github.com/omgnetwork/elixir-omg/blob/ca108e0c974170d48f9f4b9db6561b9a4f19335c/apps/omg/lib/omg/root_chain_coordinator/core.ex#L115-L122

However, since we are using `max` here, it will use the ethereum height instead of the sync height that should be the real guidance.

InoMurko:
@boolafish you might be right, and fixing this probably makes sense. But since I looked into it, it would have a bigger impact then the approach in this PR.
This PR is actually verified - since that's how Watcher syncs.

So I suggest we go with what we have here and address/simplify Root Chain Coordinator at a later stage. WDYT

pgebal:
> hmmm, I think the real bug is in this line:
> 
> https://github.com/omgnetwork/elixir-omg/blob/ca108e0c974170d48f9f4b9db6561b9a4f19335c/apps/omg/lib/omg/ethereum_event_listener/core.ex#L115
> 
> It should be using `min` instead of `max`? The `sync_height` would take `wait_for` into account:
> 
> https://github.com/omgnetwork/elixir-omg/blob/ca108e0c974170d48f9f4b9db6561b9a4f19335c/apps/omg/lib/omg/root_chain_coordinator/core.ex#L115-L122
> 
> However, since we are using `max` here, it will use the "root_chain_height" instead of the sync height that should be the real guidance.
> 
> The PR here is fixing this issue in an abusing way of how implementation works I think. It will solve the issue because the "root_chain_height" would take the finality margin into account.
> 
> https://github.com/omgnetwork/elixir-omg/blob/ca108e0c974170d48f9f4b9db6561b9a4f19335c/apps/omg/lib/omg/root_chain_coordinator/core.ex#L124

If we want to do that this way, we could remove root_chain_height from `SyncGuide`, since sync height is never greater than root_chain_height.

InoMurko:
> The fix lgtm and we can continue the longer term fix in parallel 👍

https://github.com/omgnetwork/elixir-omg/issues/1624

PR #1636, opened 2020-07-09
use cabbage tests from a separate repo

ayrat555:

## Overview

## Changes

- Change 1
- Change 2
- ...

## Testing


Makefile
```diff
@@ -250,6 +250,9 @@ init-contracts-reorg: clean-contracts
 	PLASMA_FRAMEWORK_TX_HASH=$$PLASMA_FRAMEWORK_TX_HASH PLASMA_FRAMEWORK=$$PLASMA_FRAMEWORK \
 	PAYMENT_EIP712_LIBMOCK=$$PAYMENT_EIP712_LIBMOCK MERKLE_WRAPPER=$$MERKLE_WRAPPER ERC20_MINTABLE=$$ERC20_MINTABLE
 
+init-integration-tests:
```
InoMurko, line 253: should we git submodule perhaps?

boolafish:
🤔 a bit out topic of this PR. But do we want the `spec` to be private repo. At least the acceptance test on watcher should be public IMO.

.circleci/config.yml
```diff
@@ -581,19 +591,23 @@ jobs:
           key: v2-mix-specs-cache-{{ .Branch }}-{{ checksum "mix.lock" }}
       - run:
           name: Print watcher logs
-          command: docker-compose -f docker-compose.yml -f ./priv/cabbage/docker-compose-reorg.yml -f ./priv/cabbage/docker-compose-cabbage.yml logs --follow watcher
+          command: docker-compose -f docker-compose.yml -f ./priv/cabbage/docker-compose-reorg.yml -f ./priv/cabbage/docker-compose-2-specs.yml logs --follow watcher
```
InoMurko, line 594: how do we name this better, `docker-compose-2-specs` is confusing I think...

ayrat555: @InoMurko it means the docker-compose version 2. the childchain users version 3. what do you think about `docker-compose-version-2-specs`?

InoMurko: oh, okay... that's the reason then. perhaps there's two things:
- should we name them by functionality?
- should we wrap these into makefile recipes?

ayrat555: @InoMurko I moved to makefile. It doesn't have any functionality, it only overrides 2 env variables

achiurizo:
> 🤔 a bit out topic of this PR. But do we want the `spec` to be private repo. At least the acceptance test on watcher should be public IMO.
> 
> cc. @achiurizo @T-Dnzt @kasima

Let's figure this out later once we are comfortable with how we run the `specs` repo then we can figure out how to open up the watcher acceptance tests there.

boolafish:
Why do we even wants to move to private repo now though?

 I can *guess* why you’d like to move to separate repo but i don’t see why private yet. This change of moving it private is a degradation of what we are providing to the users now but I don’t see any discussion and justification reasoning. Would appreciate if this kinds of change can clearly list out the motivation of change in PR or have early discussion with the team 🙏

ayrat555:
@InoMurko I added a section about running integration tests to README

InoMurko:
At least a partial reason for having these tests private is the future compliance with childchain2. Tests there will expose some of the underlying functionality we probably don't want the world to see. Or maybe we do (from the top of my head):
- cluster failover
- database failover
- internal accounting (backoffice)

PR #1648, opened 2020-07-16
feat: pin elixir and erlang versions for asdf

### LINKED ISSUE #1647: Pin elixir and erlang versions for asdf
Since asdf is being used to manage erlang & elixir versions in our CI pipeline and also suggested in the installation docs, adding this version pinning to the repo helps us enforcing consistent versions across local dev machines, CI, external docs, etc.

So this is exempted from the "avoid personal configs in the repo" rule.

unnawut:
Closes #1647

## Overview

Pin elixir and erlang versions through `.tool-versions` file instead of in CI config, so we also enforce the same versions to local environments and installation docs too.

## Changes

- Add [asdf's `.tool-versions`](https://asdf-vm.com/#/core-manage-versions?id=set-current-version) and pin the versions to the ones currently used by the CI
- Remove the elixir/erlang version specs from CI config so a basic `asdf install` refers to `.tool-versions`

## Testing

CI should pass, meaning that it's able to install from `.tool-versions` correctly and build/test are successful with this approach


unnawut:
Thanks everyone!

PR #1649, opened 2020-07-16
feat: block queue metrics and stalled submission alarm

### LINKED ISSUE #1629: Monitoring & alert for pending block submissions
Spinning off from https://github.com/omgnetwork/elixir-omg/pull/1617#discussion_r448697480

> Do we alarm on all error log. If not let's make sure we add alarm on this, either via Datadog parsing the error message or do the direct alarm (?) from code?
>
> Good idea! I think we can do with an app alarm when a block submission is pending for more than 2 blocks (since we're aiming for next-block inclusion, but I'll make number configurable in case of too many false panics). Then Datadog can consume this app alarm similar to :ethereum_stalled_sync and friends.
>
> Since we're losing this visibility because this PR is suppressing the exception, I'll add :block_submission_stalled alarm to this PR.
>
> Queue observability size. Hell yeah!

#### TODOs:
- [ ] Add a monitor/metrics to track pending chch block submissions (`pending_block_submissions`)
- [ ] There's a similar naming `:pending_block_queue_length` but this is watcher_info's mechanism. Come up with better metrics names to separate the two.
- [ ] Raises an alarm when `pending_block_submissions > 2`
- [ ] Clears the alarm when `pending_block_submissions == 0` (we shouldn't be expecting the queue to be around for too long)

unnawut:
Closes #1629

## Overview

Publish various block queue metrics (see the list in Changes section) and `block_submit_stalled` alarm.

### Telemetry
- Publishes `:blknum_submitting` telemetry event with the `blknum` when a block is about to be submitted
- Publishes `:blknum_submitted` telemetry event with the `blknum` when a block is successfully submitted

These are used for metrics reporting.

### OMG.Bus
- Broadcasts `:block_submitting` event with the `blknum` when a block is about to be submitted
- Broadcasts `:block_submitted` event with the `blknum` when a block is successfully submitted

These are used by `BlockQueue.Monitor` to keep tracking of stalled blocks.

### Datadog metrics
- `block_queue_blknum_submitting`: Reports the blknum being attempted (e.g. 1000, 1000, 2000, 1000, 2000, 3000, 2000, 3000, ...)
- `block_queue_blknum_submitted`: Reports the blknum succesfully submitted (e.g. 1000, 2000, 3000, ...)
- `block_submission_attempt`: Increments by 1 when a block submission is attempted
- `block_submission_success`: Increments by 1 when a block submission is successfully submitted
- `block_queue_num_blocks_stalled`: Reports the number of stalled blocks at the moment of metric collection (e.g. 0, 0, 2, 1, 0, 1, 0, ...)
- Rename `block_submission` gauge to `block_submission_gas` for better naming accuracy

### Alarm
- Raises `:block_submit_stalled` alarm when at least one block has been submitting for the past `BLOCK_SUBMIT_STALL_THRESHOLD_BLOCKS` root chain blocks or longer
- Clears `:block_submit_stalled` alarm when all submitting blocks have been submitting for less than `BLOCK_SUBMIT_STALL_THRESHOLD_BLOCKS` root chain blocks
- Defaults `BLOCK_SUBMIT_STALL_THRESHOLD_BLOCKS` env var to 4 root chain blocks

The threshold is in number of root chain blocks because blocktime can vary greatly between networks. Even on mainnet blocktime can vary between ~ 1-60 seconds, and a block submission should not be considered stalled, for example, after 30 seconds of submitting if the blocktime is 45 seconds.

### Miscellaneous
- Change log level for block formation from debug to info

## Testing

1. Set geth's `--miner.gasprice` to a very high value, e.g. `--miner.gasprice "500000000000"` (500 Gwei)
1. Submit some transactions to the network
1. `block_queue_blknum_submitting` should increase and `block_submission_attempt` is tallied
1. Childchain block submissions stay pending, not getting included in root chain blocks
1. `block_submit_stalled` alarm should raise
1. Set geth's `--miner.gasprice` to a very low value, e.g. `--miner.gasprice "100000000"` (0.1 Gwei)
1. Childchain block submissions start getting included in root chain blocks
1. `block_queue_blknum_submitted` should increase and `block_submission_success` is tallied
1. `block_submit_stalled` alarm should be cleared


unnawut:
Tagging quite a bit more reviewers than usual since it's touching the BlockQueue, although only emitting out telemetry and bus events.

docker-compose.yml
```diff
@@ -41,7 +41,6 @@ services:
             --datadir data/ \
             --syncmode 'full' \
             --networkid 1337 \
-            --gasprice '1' \
```
unnawut, line 44: now replaced by `--miner.gasprice` above

apps/omg_child_chain/lib/omg_child_chain/block_queue/submission_monitor.ex
```diff
@@ -0,0 +1,174 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.ChildChain.BlockQueue.SubmissionMonitor do
+  @moduledoc """
+  Listens to block events and raises :block_submit_stalled alarm when a pending block
```
InoMurko, line 17: so essentially what you want to figure out is:
the difference between blocks mined minus blocks formed, correct?

If yes, why are you storing pending blocks?

apps/omg_child_chain/lib/omg_child_chain/block_queue/submission_monitor.ex
```diff
@@ -0,0 +1,174 @@
...
+    case Enum.member?(:gen_event.which_handlers(:alarm_handler), __MODULE__.AlarmHandler) do
+      true -> :ok
+      _ -> :alarm_handler.add_alarm_handler(__MODULE__.AlarmHandler)
+    end
+  end
+
+  @spec trigger_alarm(module(), boolean(), [blknum()]) :: :ok
+  defp trigger_alarm(_alarm_module, false, []), do: :ok
+
+  defp trigger_alarm(alarm_module, false, _stalled_blocks) do
```
boolafish, line 153: why `state.alarm_raised` false should set alarm while true we clear? Should it be the opposite?

apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
```diff
@@ -194,6 +194,11 @@ defmodule OMG.ChildChain.BlockQueue do
     {:noreply, state1}
   end
 
+  @doc false
+  # Ignore unrelated events
+  def handle_info({:internal_event_bus, :block_submitting, _}, state), do: {:noreply, state}
```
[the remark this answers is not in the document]
unnawut: It's sharing the same `"blocks"` topic in [`:ok = OMG.Bus.subscribe({:child_chain, "blocks"}, link: true)`](https://github.com/omgnetwork/elixir-omg/pull/1649/files/365d0297b683a9c1f1f9f2bb8182bed5da9db78e#diff-7f8f0e0490710d5a45c427d163153f16R143).

We can make the topic more granular but it opens a different can of worms, i.e. setting a new convention, to include in this PR 😢

apps/omg_child_chain/lib/omg_child_chain/block_queue/submission_monitor.ex
unnawut: The 2nd parameter is whether the alarm was already raised, and when `_stalled_blocks` is non-zero, it will be caught by this case.

But I see it's a valid confusion, updated the clauses to this instead https://github.com/omgnetwork/elixir-omg/pull/1649/commits/d676020cda4efa555e2e376f811cc063e5a78020:

```elixir
  @spec trigger_alarm(alarm :: module(), alarm_raised :: boolean(), stalled_blocks :: [blknum()]) :: :ok
  defp trigger_alarm(alarm_module, false, stalled_blocks) when length(stalled_blocks) >= 1 do
    alarm_module.set(alarm_module.block_submit_stalled(__MODULE__))
  end

  defp trigger_alarm(alarm_module, true, []) do
    alarm_module.clear(alarm_module.block_submit_stalled(__MODULE__))
  end

  defp trigger_alarm(_alarm_module, _alarm_raised, _stalled_blocks), do: :ok
```

boolafish: I see

apps/omg_child_chain/lib/omg_child_chain/block_queue/submission_monitor.ex
unnawut: Mainly to track `{blknum, root_chain_height}`. There was a lot of false alarms when I tracked just the `mined - formed` because of the large swing in block time. And it's a large swing all the time, not just once in a while, so the alarm is barely useful with an alarm every few minutes.

Some supporting info. Below is rootchain blocktime over a single day. You can see the spikes throughout the day, multiple times each hour:

blocktime

And in the diagram below, if we use the average blocktime of 12.5 - 15s as usual, only 6% (369/6401) of all blocks in a day will fall in the range.

Histogram of blocktime

So in my trial it was impossible to stick to `mined - formed`, because over a single day `mined - formed >= 4` could be from 10 seconds to 188 seconds. So tracking the chch blknum against the first root chain height it was submitted at was the most reliable way to track stalled submissions as far as I tried.

---

Source data: https://docs.google.com/spreadsheets/d/1wzniBoO2nonRo83im0VRcaSPj_pr-ZZXKV-yiOOFVOo/edit

T-Dnzt (approved):
LGTM 👍

InoMurko (approved):
IMPRESSIVE STUFF

unnawut:
## Metrics
image

## Submission stall monitor

```json
$ curl -s <dev_url>/alarm.get -w "\n" | jq

{
  "data": [
    {
      "block_submit_stalled": {
        "node": "child_chain@127.0.0.1",
        "reporter": "Elixir.OMG.ChildChain.BlockQueue.SubmissionMonitor"
      }
    }
  ],
  "service_name": "child_chain",
  "success": true,
  "version": "1.0.3+733aab2"
}
```

image

PR #1656, opened 2020-07-21
Add missing clause on witness validation check

mederic-p:
Closes https://github.com/omgnetwork/elixir-omg/issues/1158

Fixes a crash when given an array of array of RLP encoded signatures.


unnawut:
much wow!

PR #1662, opened 2020-07-22
Fix fee adapter to accept decimal value in fee rules

jarindr:
## Overview
pegged amount now can be decimal point, we add that and also enable `pegged` type

## Changes

- remove `integer` validation on `validate_optional_positive_amount`
- add default value for fee interval fetching
## Testing

- run test normally

### DIFF
--- a/apps/omg_child_chain/test/omg_child_chain/fees/json_single_spec_parser_test.exs
+++ b/apps/omg_child_chain/test/omg_child_chain/fees/json_single_spec_parser_test.exs
@@ -66,6 +66,22 @@ defmodule OMG.ChildChain.Fees.JSONSingleSpecParserTest do
               }} == JSONSingleSpecParser.parse({@eth_hex, spec})
     end
 
+    test "returns ok when given decimal to pegged_amount" do
+      spec = Map.put(@valid_spec, "pegged_amount", 0.33)
+
+      assert {:ok,
+              %{
+                token: @eth,
+                amount: 1,
+                subunit_to_unit: 1_000_000_000_000_000_000,
+                pegged_amount: 0.33,
+                pegged_currency: "USD",
+                pegged_subunit_to_unit: 100,
+                updated_at: "2019-01-01T10:10:00+00:00" |> DateTime.from_iso8601() |> elem(1),
+                type: :fixed
+              }} == JSONSingleSpecParser.parse({@eth_hex, spec})
+    end
+
     test "returns an `invalid_pegged_fields` error when given a nil pegged_amount alone" do
       spec = Map.put(@valid_spec, "pegged_amount", nil)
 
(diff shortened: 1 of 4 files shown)


unnawut:
2 points:

1. When would the fee amount be decimal point? Shouldn't it already be covered by subunit_to_unit?
2. `validate_optional_positive_amount/2` is also used to `subunit_to_unit` which shouldn't have decimal, so I think if this change is still needed you'll need to split it to a new validator function.

jarindr:
> 2 points:
> 
> 1. When would the fee amount be decimal point? Shouldn't it already be covered by subunit_to_unit?
> 2. `validate_optional_positive_amount/2` is also used to `subunit_to_unit` which shouldn't have decimal, so I think if this change is still needed you'll need to split it to a new validator function.

1. we use `pegged_amount` as a multiplier for the amount to peg to the gas price, eg. if we want 1/3 of the price we put it as 0.33

2. true, will fix now

unnawut:
> 1. we use `pegged_amount` as a multiplier for the amount to peg to the gas price, eg. if we want 1/3 of the price we put it as 0.33

Ah ok I see it only affect `"pegged_amount"` and not `"amount"` 👍

unnawut:
Btw. how did this get through without needing to update the tests? 😂  Can you add to cover these changes? 😂

jarindr:
> Btw. how did this get through without needing to update the tests? 😂 Can you add to cover these changes? 😂

added test to ensure magic doesn't happen again.

apps/omg_child_chain/test/omg_child_chain/fees/json_single_spec_parser_test.exs
```diff
@@ -66,6 +66,22 @@ defmodule OMG.ChildChain.Fees.JSONSingleSpecParserTest do
...
+      spec = Map.put(@valid_spec, "pegged_amount", 0.33)
+
+      assert {:ok,
+              %{
+                token: @eth,
+                amount: 1,
+                subunit_to_unit: 1_000_000_000_000_000_000,
+                pegged_amount: 0.33,
+                pegged_currency: "USD",
+                pegged_subunit_to_unit: 100,
```
unnawut, line 79: so `pegged_amount: 0.33` with `subunit_to_unit: 100` is 0.0033 USD? 🤔

unnawut:
As explained by @T-Dnzt. For future reference:

> `pegged_subunit_to_unit` is not for the `pegged_amount`, it's the decimals for the `pegged_currency` (so `1000000000000000000` for OMG)
>
> (amount in ETH to pay (gas_price * avg_gas_price_for_tx) * pegged_subunit_to_unit) * (Exchange rate ETH <> ERC-20 Token) * subunit_to_unit * pegged_amount

> type: pegged
> symbol: OMG
> subunit_to_unit: 1000000000000000000
> pegged_currency: eth_gas_price
> pegged_subunit_to_unit: 1000000000000000000
> pegged_amount: 0.33 (set manually to reflect how much cheaper we want to be, 1/3 of ETH)

PR #1664, opened 2020-07-23
debug: enable debug log in dev

boolafish:
Is this the right way to enable debug log?

Trying to understand what is the cause of the 502 response. This commit enables the debug log for dev env. Hopefully the Plug.Logger would log the requests.
    
ref: https://github.com/omgnetwork/devops/issues/373

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -1,5 +1,7 @@
 import Config
 
+config :logger, level: :debug
+
 config :logger,
   backends: [:console, Sentry.LoggerBackend]
 


config/dev.exs
```diff
@@ -1,5 +1,7 @@
 import Config
 
+config :logger, level: :debug
```
unnawut, line 3: A little more ideal is to combine with below

```elixir
config :logger,
  level: :debug,
  backends: [:console, Sentry.LoggerBackend]
```

boolafish:
> For your local dev environment right?

no, it is for the dev env. We want to know what is happening in the 502 thing and that need a full infra

boolafish:
wrong approach.

unnawut:
👍  For deployed environments changing `DD_LOG_LEVEL` is more ideal

PR #1666, opened 2020-07-23
feat: add metrics: balance per currency, number of unspent outputs, number of addresses

### LINKED ISSUE #1148: Calculate metrics on lazily loaded UTXO set
PR #1103 allows to load utxo subset when needed instead of loading it entirely during startup.
This meas that in-memory UTXO set contains information processed since the last service restart.

Metrics collected so far:
* balance
* unique addresses
(see: [{repo_root}/apps/omg/lib/omg/state/measurement_calculation.ex](https://github.com/omisego/elixir-omg/blob/f92cb6fbca27bdfc6ca31c518dbb867a0b5f4551/apps/omg/lib/omg/state/measurement_calculation.ex))

Where we should move metrics calculation :question:

unnawut:
Close #1148

## Overview

Add metrics for childchain's balances, number of unspent outputs and addresses with unspent outputs.

image

## Changes

- Adds `total_unspent_outputs`: The total number of unspent outputs being tracked in the childchain service
- Adds `total_unspent_addresses`: The total number of addresses holding at least 1 unspent output
- Updates `balance`: Same definition. Fixed where the balance was calculating only from the recent cache
- Removes `unique_users`: Replaced by `total_unspent_addresses`

Note that I originally wanted to avoid or optimize utxos calling but it requires lower level opimization and the current DB can handle the load for the foreseeable future (see https://github.com/omgnetwork/elixir-omg/pull/1667 for some benchmark). 

With this PR we get the utxos db observability and can monitor when the metrics become concerning so I think it's worth the extra call.

## Testing

Follow these steps:

1. Configure `DD_DISABLED: false` and `DD_API_KEY` at least on the childchain
1. Run `make docker-build`
1. Run `cd priv/cabbage && make start_daemon_services-2`
1. Make some deposits and transfers
1. The 3 metrics should be reported


apps/omg_db/lib/omg_db/monitor.ex
```diff
@@ -0,0 +1,97 @@
...
+        %{total_unspent_outputs: stats.total_unspent_outputs},
+        %{}
+      )
+
+    _ = Logger.info("#{__MODULE__} observed #{stats.total_unspent_outputs} unspent outputs.")
+    {:noreply, state}
+  end
+
+  defp collect_stats(db) do
+    {:ok, utxos} = db.utxos()
```
InoMurko, line 89: isn't this like the worst query you could make? pull millions of records into memory?

unnawut: Yup! Without revamping the data structure, we can chunk-calculate this, but that means a full table scan for each query, e.g. one to group by currency, another one to group by address, etc. and that's read-heavy on rocksdb which is even more severe than being memory-intensive.

And we need utxoset observability so 🤷‍♂️

unnawut: Actually hang on, lemme try create `OMG.DB.RocksDB.reduce()` that processes each utxo while it iterates through the full utxoset.

unnawut:
Will reopen in https://github.com/omgnetwork/omg-childchain-v1

PR #1667, opened 2020-07-23
benchmark: rocksdb concurrent reads/writes

unnawut:
Opening this PR just for reference so we can copy & paste to do more benchmarking in the future.

This is done on my local machine so the numbers are much lower than what we're seeing on deployed environments. But got 3 things:

1. Rocksdb reads and writes do interleave. But we route through [`OMG.DB.RocksDB.Server`](https://github.com/omgnetwork/elixir-omg/blob/master/apps/omg_db/lib/omg_db/rocksdb/server.ex) which handles one message at a time and so is _probably_ the reason why our read/write operations are blocking.
    - Implement read/write as async tasks into this GenServer?
    - Reads could also be turned into a stream I think.
2. Within `OMG.DB.utxos()`, the read iterator and decoder takes [most of the time](https://github.com/omgnetwork/elixir-omg/compare/unnawut/benchmark?expand=1#diff-0d4a86fbc510a6669d7b5603e2b94509R72-R93). The decoder takes [most of the memory](https://github.com/omgnetwork/elixir-omg/compare/unnawut/benchmark?expand=1#diff-0d4a86fbc510a6669d7b5603e2b94509R85-R93).
    - Can try [optimize on the iterator](https://github.com/omgnetwork/elixir-omg/issues/1660) to optimize for time. Decoding could be a stream to optimize for memory
3. Tried [rocksdb snapshots](https://github.com/facebook/rocksdb/wiki/Snapshot), with or without, doesn't significantly impact concurrent reads/writes (see [results](https://github.com/omgnetwork/elixir-omg/compare/unnawut/benchmark?expand=1#diff-6994845488110f644f6c550cde0ad57d))
    - Ignore rocksdb snapshots for now


unnawut:
Not intending to merge this but want to keep the files for the record.

Issue #1643, opened 2020-07-14
Endpoint: /block.validate


okalouti:
In the current setup, the child chain composes a block and submits to the root chain without any further validation from an external component. This poses the risk of a byzantine event due to a software error in the child chain.

The following flow is envisioned to mitigate this: 

*_The `Vault` will manage block submission to the Ethereum Network_

1. Child chain requests the Vault to publish a block (Merkle) root hash
2. Vault calls `child_chain.get_block(root_hash)`
3. Vault calls Watcher to perform stateful block validation.
4. If the validation is successful, publish. If it's not successful send an alarm on multiple channels.

okalouti:
Checklist: 
- [ ] Stateless validation: given Merkle root matches reconstructed Merkle root
- [ ] Stateless validation: all transactions are correctly formed
- [ ] Stateful validation: UTXOs in the given transactions are not spent in a child chain transaction or exit

To discuss: where/how do we draw the line between what is validated by the Watcher and what is left to the exit/challenge game?

thec00n:
Here are my requirements for the stateful validation rules. 

Payment Tx (Ref: https://github.com/omgnetwork/elixir-omg/blob/master/docs/transaction_validation.md): 
* A tx input can not be spent by another tx in the same block 
* A tx input has to be included in a previous block (including deposit blocks) and therefore has to be included in  the list of unspent utxos 
* A tx input must have a valid signature from the owner 
* The sum of tx inputs for a currency needs to be greater than the sum of tx outputs 

Fee Tx (Ref: https://github.com/omgnetwork/elixir-omg/blob/master/docs/fee_design.md):
* The sum of inputs minus the sum of outputs for a given currency across all payment tx is the amount that the operator can transfer to himself
* Match the fee tx type 
* Must have 0 inputs, 1 output
* Fee tx address should be matched against a specific value?  

Block: 
* Block has less than 2^16 tx and more than 0 tx 
* Fee tx need to come at the end of a block after all other tx 
* 1 fee tx per asset 
* An input can only be spent once (per block). This is more a reminder that the watcher needs to keep track of all spent inputs and consider the ones that are spent in the current block.

okalouti:
| Requirement  | Type  | Status  | Relevant Code |  
|---|---|---|---|
|  **Payment** |   |   |   |   
| has inputs that are not spent in a previous block |   |   |   |   
| matches the payment tx type |   |   |   |  
[...]

PR #1668, opened 2020-07-24
 /block.validate endpoint

okalouti:
#1643 
## Overview

Creates an endpoint `/block.validate` that conducts **stateless** block validation. To be expanded to a stateful validation in future work. 

The endpoint takes 3 parameters: `transactions`, `hash` and `number`. 

Stateless validation consists of: 
- Verifying that the parameters are correctly formed.  ✅ 
- Verifying that the block hash corresponds to the Merkle root hash.  ✅ 
- Verifying that the transaction bytes correspond to correctly formed transactions. ✅ 

## Changes

-  Additions to `block_constraints.ex` to validate parameters.
- New module `block_validator.ex` to validate submitted block.
- Requisite changes in controller and router file

## Testing

`mix test test/omg_watcher_rpc/web/controllers/block_test.exs`
`mix test test/omg_watcher/block_validator_test.exs`
`mix test test/omg_watcher_rpc/web/validators/block_constraints_test.exs`

The following negative cases are tested: 
- Incorrectly formed `hash` parameter
- Incorrectly formed `number` parameter
- Incorrectly formed `transactions` parameter
- Incorrectly formed elements (i.e not hexadecimal transaction bytes) in `transactions` array.  
- Invalid Merkle root 
- Incorrectly formed transactions

Props to @ripzery for the work here!


apps/omg_watcher_rpc/lib/web/validators/block_validator.ex
```diff
@@ -0,0 +1,72 @@
...
+  def verify_merkle_root(block) do
+    %{hash: reconstructed_merkle_root_hash} =
+      block.transactions
+      |> Enum.map(fn tx ->
+        {:ok, recovered_tx} = Transaction.Recovered.recover_from(tx)
+        recovered_tx
+      end)
+      |> OMG.Block.hashed_txs_at(block.number)
+
+    case block.hash == reconstructed_merkle_root_hash do
```
InoMurko, line 48: Use a pin operator here instead of comparison

apps/omg_watcher_rpc/lib/web/validators/block_validator.ex
```diff
@@ -0,0 +1,72 @@
...
+  @doc """
+  Verifies that transactions are correctly formed.
+  """
+  @spec verify_transactions(transactions :: list(Transaction.Recovered.t())) ::
+          {:ok, list(Transaction.Recovered.t())}
+          | {:error, Transaction.Recovered.recover_tx_error()}
+  def verify_transactions(transactions) do
+    Enum.reduce_while(transactions, {:ok, []}, fn tx, {:ok, already_recovered} ->
+      case Transaction.Recovered.recover_from(tx) do
+        {:ok, recovered} -> {:cont, {:ok, already_recovered ++ [recovered]}}
```
InoMurko, line 63: This is slow because you're recreating a list on each iteration. Can you refactor this?

apps/omg_watcher_rpc/lib/web/validators/block_validator.ex
```diff
@@ -0,0 +1,77 @@
...
+    |> Enum.reduce_while({:ok, []}, fn tx, {:ok, already_recovered} ->
+      case Transaction.Recovered.recover_from(tx) do
+        {:ok, recovered} ->
+          {:cont, {:ok, [recovered | already_recovered]}}
+
+        error ->
+          {:halt, error}
+      end
+    end)
+  end
```
unnawut, line 72: Can we move this to `OMG.Watcher`? This seems to be a core business validating logic and not request validation like `parse_to_validate()`

apps/omg_watcher_rpc/lib/web/controllers/block.ex
```diff
@@ -43,4 +44,22 @@ defmodule OMG.WatcherRPC.Web.Controller.Block do
...
+    end
+  end
+
+  @spec stateless_validate(Block.t()) :: any
+  defp stateless_validate(submitted_block) do
+    with {:ok, recovered_transactions} <- Validator.BlockValidator.verify_transactions(submitted_block.transactions),
+         {:ok, block} <- Validator.BlockValidator.verify_merkle_root(submitted_block, recovered_transactions) do
+      {:ok, block}
+    end
+  end
```
unnawut, line 64: I think this should be moved to `OMG.Watcher` along with my other comment in https://github.com/omgnetwork/elixir-omg/pull/1668/files#diff-fbb5622ef4f2eb45b525f9d2cb8014dc

And remaining here would be something like:

```
  def validate_block(conn, params) do
    with {:ok, block} <- Validator.BlockValidator.parse_to_validate(params),
         {:ok, _block} <- OMG.Watcher.Block.stateless_validate(block) do
      api_response(block, conn, :validate_block)
    end
  end
```

okalouti:
hey @unnawut thanks, makes sense. have move the block validation logic into `OMG.Watcher.BlockValidator` and parameter validation logic into existing `block_constraints.ex` file.

apps/omg_watcher/lib/omg_watcher/block_validator.ex
```diff
@@ -0,0 +1,71 @@
...
+      {:ok, block}
+    end
+  end
+
+  @doc """
+  Verifies that given Merkle root matches reconstructed Merkle root.
+  """
+  @spec verify_merkle_root(Block.t(), list(Transaction.Recovered.t())) ::
+          {:ok, Block.t()} | {:error, :mismatched_merkle_root}
+  def verify_merkle_root(block, transactions) do
```
InoMurko, line 40: Why is this public?

apps/omg_watcher/lib/omg_watcher/block_validator.ex
```diff
@@ -0,0 +1,71 @@
...
+    end
+  end
+
+  @doc """
+  Verifies that transactions are correctly formed.
+  """
+  @spec verify_transactions(transactions :: list(Transaction.Recovered.t())) ::
+          {:ok, list(Transaction.Recovered.t())}
+          | {:error, Transaction.Recovered.recover_tx_error()}
+  def verify_transactions(transactions) do
```
InoMurko, line 58: Why is this public?

PR #1670, opened 2020-07-27
feat: remove utxos cache from OMG.State on form_block

boolafish:
**blocked by https://github.com/omgnetwork/private-issues/issues/70**

:clipboard: closes https://github.com/omgnetwork/private-issues/issues/62

## Overview

This is to avoid potential memory leak due to increasing size of utxos.
See @pnowosie 's comment: https://github.com/omgnetwork/private-issues/issues/62#issuecomment-662907908

## Changes

- reset to empty map for `utxos` on `form_block`
- add a unit test for above behavior
- fix fee claiming unit test
- remove one OMG.State core test as it is really testing integration behavior and not really possible to test with this change introduced. However, the test should already be covered by cabbage naturally.

## Testing

CI

### DIFF
--- a/apps/omg/test/omg/state/core_test.exs
+++ b/apps/omg/test/omg/state/core_test.exs
@@ -470,18 +470,6 @@ defmodule OMG.State.CoreTest do
   end
 
   @tag fixtures: [:alice, :bob, :state_alice_deposit]
-  test "can spend after block is formed", %{alice: alice, bob: bob, state_alice_deposit: state} do
-    next_block_height = @blknum2
-    {:ok, {_, _}, state} = form_block_check(state)
-
-    state
-    |> Core.exec(create_recovered([{1, 0, 0, alice}], @eth, [{bob, 7}, {alice, 2}]), @fee)
-    |> success?
-    |> Core.exec(create_recovered([{next_block_height, 0, 0, bob}], @eth, [{bob, 6}]), @fee)
-    |> success?
-  end
-
-  @tag fixtures: [:alice, :bob, :state_alice_deposit]
   test "forming block doesn't unspend", %{alice: alice, bob: bob, state_alice_deposit: state} do
     recovered = create_recovered([{1, 0, 0, alice}], @eth, [{bob, 7}, {alice, 2}])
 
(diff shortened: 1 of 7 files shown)


apps/omg/test/omg/state/core_test.exs
```diff
@@ -469,18 +469,6 @@ defmodule OMG.State.CoreTest do
     |> success?
   end
 
-  @tag fixtures: [:alice, :bob, :state_alice_deposit]
-  test "can spend after block is formed", %{alice: alice, bob: bob, state_alice_deposit: state} do
```
boolafish, line 473: Not really possible to run this test. The test relies on utxo staying in memory.
The test description itself indicates it to be integration tests. As a result leaving this to be covered by cabbage.

apps/omg_watcher/test/omg_watcher/exit_processor/core/state_interaction_test.exs
```diff
@@ -179,55 +179,6 @@ defmodule OMG.Watcher.ExitProcessor.Core.StateInteractionTest do
     assert Utxo.position(1, 0, 0) in spends_to_get
   end
 
-  test "can work with State to exit utxos from in-flight transactions",
```
boolafish, line 182: There is IFE cabbage test covering IFE flow of startIFE, piggyback, and process Exit

boolafish:
okay, confirmed an integration test that suppose not to fail failed due to the introduction of the change.

https://github.com/omgnetwork/elixir-omg/blob/1f7a6351b3c7b8af1de144936784ca22bbc1d1c5/apps/omg_child_chain/test/omg_child_chain/integration/happy_path_test.exs#L214

boolafish:
When receiving piggyback output event, it will try to get the utxo with the following line:
https://github.com/omgnetwork/elixir-omg/blob/522fb32c493196e3fe1d32b38dd61a9949c741ba/apps/omg/lib/omg/state.ex#L211

And the Core implementation would require the `utxos` memory: 
https://github.com/omgnetwork/elixir-omg/blob/1feed637da2e5caf49b7fe2bb3a2872818baee46/apps/omg/lib/omg/state/core.ex#L339-L349

in `find_utxo_matching_piggyback`
https://github.com/omgnetwork/elixir-omg/blob/1feed637da2e5caf49b7fe2bb3a2872818baee46/apps/omg/lib/omg/state/core.ex#L502-L506

PR #1676, opened 2020-07-28
account.get_exitable_utxos is unaware of in-flight exited inputs

pnowosie:
Solves omgnetwork/private-issues#41

## Overview

Assuming standard IFE case where in-flight transaction is not submitted. This means transaction's outputs don't exist and inputs regardless transaction's canonicity cannot be standard-exited. Therefore they are being filtered from the result of the `account.get_exitable_utxos` endpoint 

## Changes

Inputs of the transaction in-flight are being removed from the results returned by the `account.get_exitable_utxos` endpoint 

## Testing

Specs test demonstrate the feature https://github.com/omgnetwork/specs/pull/4

### DIFF
--- a/apps/omg_watcher/lib/omg_watcher/api/account.ex
+++ b/apps/omg_watcher/lib/omg_watcher/api/account.ex
@@ -31,10 +31,16 @@ defmodule OMG.Watcher.API.Account do
 
     # PaymentExitInfo.all_exit_infos() takes a while.
     {:ok, standard_exits} = PaymentExitInfo.all_exit_infos()
-    active_standard_exiting_utxos = OMG.Watcher.ExitProcessor.Core.active_standard_exiting_utxos(standard_exits)
+    {:ok, in_flight_exits} = PaymentExitInfo.all_in_flight_exits_infos()
+
+    active_exiting_utxos =
+      MapSet.union(
+        OMG.Watcher.ExitProcessor.Core.active_standard_exiting_utxos(standard_exits),
+        OMG.Watcher.ExitProcessor.Core.active_in_flight_exiting_inputs(in_flight_exits)
+      )
 
     # active standard exiting utxos are excluded
-    filter_standard_exiting_utxos(standard_exitable_utxos, active_standard_exiting_utxos)
+    filter_standard_exiting_utxos(standard_exitable_utxos, active_exiting_utxos)
   end
 
   defp filter_standard_exiting_utxos(standard_exitable_utxos, active_standard_exiting_utxos) do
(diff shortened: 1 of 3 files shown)


apps/omg_watcher/lib/omg_watcher/api/account.ex
```diff
@@ -31,10 +31,16 @@ defmodule OMG.Watcher.API.Account do
...
+        OMG.Watcher.ExitProcessor.Core.active_standard_exiting_utxos(standard_exits),
+        OMG.Watcher.ExitProcessor.Core.active_in_flight_exiting_inputs(in_flight_exits)
+      )
 
     # active standard exiting utxos are excluded
-    filter_standard_exiting_utxos(standard_exitable_utxos, active_standard_exiting_utxos)
+    filter_standard_exiting_utxos(standard_exitable_utxos, active_exiting_utxos)
   end
 
   defp filter_standard_exiting_utxos(standard_exitable_utxos, active_standard_exiting_utxos) do
```
boolafish, line 46: perhaps rename the param `active_standard_exiting_utxos` here too.

apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
```diff
@@ -287,9 +288,51 @@ defmodule OMG.Watcher.ExitProcessor.CoreTest do
...
 
-      expected = MapSet.new([OMG.Utxo.position(active_blknum, active_txindex, active_txoutput)])
+      expected = MapSet.new([Utxo.position(active_blknum, active_txindex, active_txoutput)])
 
       assert expected == Core.active_standard_exiting_utxos(db_exits)
     end
   end
+
+  describe "active_in_flight_exiting_inputs" do
+    test "in-flight exit info db value is formed correctly" do
```
boolafish, line 298: this is kinda weird as test though I get you'd like to verify your test functions are acting as expected.....personally I would just give up on this (or only tested locally at the beginning)

apps/omg_watcher/lib/omg_watcher/api/account.ex
```diff
@@ -31,10 +31,16 @@ defmodule OMG.Watcher.API.Account do
 
     # PaymentExitInfo.all_exit_infos() takes a while.
     {:ok, standard_exits} = PaymentExitInfo.all_exit_infos()
-    active_standard_exiting_utxos = OMG.Watcher.ExitProcessor.Core.active_standard_exiting_utxos(standard_exits)
+    {:ok, in_flight_exits} = PaymentExitInfo.all_in_flight_exits_infos()
+
+    active_exiting_utxos =
```
boolafish, line 36: consider to left a comment to link the issue for people to understand why filtering both SE and IFE inputs: `https://github.com/omgnetwork/private-issues/issues/41`

apps/omg_watcher/test/omg_watcher/exit_processor/core_test.exs
pnowosie: Yeah, I agree that this test does not belongs here. I think I have a better place for the assertion. Thanks!

PR #1678, opened 2020-07-29
corrrectly serialize PIDs in alarms.get

ayrat555:
Fixes https://github.com/omgnetwork/elixir-omg/issues/1619

## Overview

Add Jason.Encoder for PID.


InoMurko:
I would actually go down the route of setting the treshold to 1.0 so that the memory alarm never gets triggered (cc @unnawut). We report memory to Datadog and we can set our monitoring there. Wdyt @unnawut ?

unnawut:
Hmm true I think this alarm is not needed for external consumption

InoMurko:
cool. @ayrat555 if you could adapt this based on the comments it would be wonderful/better. thanks!

ayrat555:
@InoMurko @unnawut done

unnawut:
Amazing thanks 😂  Can you also set this alarm up in our monitoring tool as we merge this so we don't lose the alarm?

ayrat555:
@unnawut what do you mean? I think this change will remove the alarm

unnawut:
> @unnawut what do you mean? I think this change will remove the alarm

Ah Ino said it better than me prior 😂  "We report memory to Datadog and we can set our monitoring there."

ayrat555:
@InoMurko @unnawut I don't think Datadog has the ability to monitor erlang processed

unnawut:
> @InoMurko @unnawut I don't think Datadog has the ability to monitor erlang processes

Maybe https://github.com/ferd/vmstats that we have now? It taps into what erlang's memory module provides and already reporting to Datadog. There's `vmstats.memory.proc_used` and etc. but will need some digging into how to calculate for the alarm.

(I think this doesn't block merging of this PR though)

PR #1683, opened 2020-08-05
feat: transaction.create optimisation

### LINKED ISSUE #1645: Optimise usage of UTXOs in transaction.create
In line with the objective of reducing the amount of UTXOs on the network, implement logic in  `/transaction.create` to select the maximum number of UTXOs as transaction inputs.

okalouti:
Closes #1645 

## Overview

This PR optimizes utxos from the `transaction.create` usage by selecting more utxos to inputs if there're rooms available.

Eligibility of utxos to be selected:
- has the same currency as either payment or fee.

Priority to select the utxos are:
- selects currency which has the largest number of utxos
- selects the smallest amount utxo

Condition to stop selecting utxos:
- reaches the maximum number of inputs
- have no more utxos to be selected

## Changes

- Adds utxos to the original inputs which sufficient to satisfy payment and fee.
- Returns error instead of auto-merging utxos.

## Testing

`mix test test/omg_watcher_info/utxo_selection_test.exs`
`mix test test/omg_watcher_rpc/web/controllers/transaction_test.exs`


apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -23,11 +23,44 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
+  require Transaction.Payment
 
   @default_transactions_limit 200
   @empty_metadata <<0::256>>
 
+  @type create_t() ::
+          {:ok,
+           %{
+             result: :complete | :intermediate,
+             transactions: nonempty_list(UtxoSelection.transaction_t())
```
okalouti, line 38: move `UtxoSelection.transaction_t()` to this module?

okalouti, line 38: Should this be from the UtxoSelection module?

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -77,10 +110,41 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
+        result
+        |> create_transaction(order)
+        |> respond(:complete)
+    end
+  end
+
+  # Given order finds spender's inputs sufficient to perform a payment.
+  # If also provided with receiver's address, creates and encodes a transaction.
+  @spec create_inputs(utxos_map_t(), order_t()) :: inputs_t()
+  defp create_inputs(utxos, %{payments: payments, fee: fee}) do
```
okalouti, line 131: I would call this `select_inpust.ex`

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -77,19 +109,50 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
+         |> select_inputs(order) do
+      {:error, reason} ->
+        {:error, reason}
+
+      result ->
+        result
+        |> create_transaction(order)
+        |> respond()
+    end
+  end
```
okalouti, line 125:
```suggestion
    with {:ok, inputs} <- order.owner |> DB.TxOutput.get_sorted_grouped_utxos() |> select_inputs() do 
          {:ok, create_transaction(inputs)}
    end
```

okalouti: I don't think you need `respond()` anymore?

okalouti: May be wrong about respond() but above is how I'd refactor the rest

apps/omg_watcher_info/lib/omg_watcher_info/utxo_selection.ex
```diff
@@ -19,93 +19,115 @@ defmodule OMG.WatcherInfo.UtxoSelection do
...
-  @type advice_t() ::
-          {:ok,
-           %{
-             result: :complete | :intermediate,
-             transactions: nonempty_list(transaction_t())
-           }}
-          | {:error, {:insufficient_funds, list(map())}}
-          | {:error, :too_many_outputs}
-          | {:error, :empty_transaction}
+  @spec filter_unselected(utxo_list_t(), %{currency_t() => boolean()}) :: utxo_list_t()
```
okalouti, line 72: think we will need to make this private?

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -1313,6 +1204,114 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
...
+
+      assert %{
+               "transactions" => [
+                 %{
+                   "fee" => %{
+                     "amount" => 5,
+                     "currency" => @eth_hex
+                   },
+                   "inputs" => [
+                     %{"amount" => 20},
```
okalouti, line 1305: Test for the right currencies in the inputs as well?

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -77,19 +107,53 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
   """
-  @spec create(UtxoSelection.order_t()) :: UtxoSelection.advice_t()
+  @spec create(order_t()) :: create_t()
   def create(order) do
-    utxos = DB.TxOutput.get_sorted_grouped_utxos(order.owner)
-    UtxoSelection.create_advice(utxos, order)
+    with {:ok, inputs} <-
+           order.owner
+           |> DB.TxOutput.get_sorted_grouped_utxos()
+           |> select_inputs(order) do
```
InoMurko, line 115: this is a single with statement which means a `case` is prefered. `with` - `Used to combine matching clauses.`

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -77,19 +107,53 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
+  # Given an `order`, finds spender's inputs sufficient to perform a payment.
+  # If also provided with receiver's address, creates and encodes a transaction.
+  @spec select_inputs(utxos_map_t(), order_t()) :: inputs_t()
+  defp select_inputs(utxos, %{payments: payments, fee: fee}) do
+    token_utxo_selection =
+      payments
+      |> UtxoSelection.needed_funds(fee)
+      |> UtxoSelection.select_utxo(utxos)
+
+    with {:ok, funds} <- UtxoSelection.funds_sufficient?(token_utxo_selection) do
```
InoMurko, line 129: this is a flaw, a questionmark function returns boolean values only.

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -77,19 +107,53 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
+          {:error, :too_many_inputs}
+      end
+    end
   end
 
   @spec include_typed_data(UtxoSelection.advice_t()) :: UtxoSelection.advice_t()
   def include_typed_data({:error, _} = err), do: err
 
-  def include_typed_data({:ok, %{transactions: txs} = advice}),
+  def include_typed_data({:ok, txs}),
```
InoMurko, line 153: you're mixing public and private functions here. it also seems the function needs to be written in multiline

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -126,6 +188,78 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
+    |> (fn output -> Enum.zip([:output0, :output1, :output2, :output3], output) end).()
+  end
+
+  defp create_transaction(utxos_per_token, %{
+         owner: owner,
+         payments: payments,
+         metadata: metadata,
+         fee: fee
+       }) do
+    rests =
```
InoMurko, line 200: please don't break down arguments in function headers in there are no multiple function headers that you need to pattern match.

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -126,6 +188,78 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
+      utxos_per_token
+      |> Enum.map(fn {_, utxos} -> utxos end)
+      |> List.flatten()
+
+    cond do
+      Enum.count(outputs) > Transaction.Payment.max_outputs() ->
+        {:error, :too_many_outputs}
+
+      Enum.empty?(inputs) ->
+        {:error, :empty_transaction}
```
InoMurko, line 226: this whole thing really looks like it's a subset of some larger functionality it would make sense to extract into a separate module perhaps

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -126,6 +188,78 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
+
+  defp create_raw_transaction(inputs, outputs, metadata) do
+    if Enum.any?(outputs, fn %{owner: owner} -> owner == nil end),
+      do: nil,
+      else:
+        Transaction.Payment.new(
+          Enum.map(inputs, fn input -> {input.blknum, input.txindex, input.oindex} end),
+          Enum.map(outputs, fn output -> {output.owner, output.currency, output.amount} end),
+          metadata || @empty_metadata
+        )
```
InoMurko, line 253: use short if/else notation only if you don't have multiline data structures or flows.

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -126,6 +188,78 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
+             metadata: metadata,
+             txbytes: create_txbytes(raw_tx),
+             sign_hash: compute_sign_hash(raw_tx)
+           }
+         ]}
+    end
+  end
+
+  defp create_raw_transaction(inputs, outputs, metadata) do
+    if Enum.any?(outputs, fn %{owner: owner} -> owner == nil end),
```
InoMurko, line 246: what is this check doing? don't we have validation modules for that?

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -126,6 +188,78 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
+          Enum.map(inputs, fn input -> {input.blknum, input.txindex, input.oindex} end),
+          Enum.map(outputs, fn output -> {output.owner, output.currency, output.amount} end),
+          metadata || @empty_metadata
+        )
+  end
+
+  defp create_txbytes(tx) do
+    with tx when not is_nil(tx) <- tx,
+         do: Transaction.raw_txbytes(tx)
+  end
```
InoMurko, line 259: hm... this seems kinda randoom, why do you need the is nil check?
also, a single with is a `case`.

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -126,6 +188,78 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
+
+  defp compute_sign_hash(tx) do
```
InoMurko, line 261: why the defensive coding again? what are you supposed to do if tx is nil coming from `create_raw_transaction`?

apps/omg_watcher_info/lib/omg_watcher_info/utxo_selection.ex
```diff
@@ -19,93 +19,116 @@ defmodule OMG.WatcherInfo.UtxoSelection do
...
-      utxo_count =
-        funds
-        |> Stream.map(fn {_, utxos} -> length(utxos) end)
-        |> Enum.sum()
-
-      if utxo_count <= Transaction.Payment.max_inputs(),
-        do: create_transaction(funds, order) |> respond(:complete),
-        else: create_merge(owner, funds) |> respond(:intermediate)
+  @spec add_utxos_for_stealth_merge(utxo_list_t(), utxos_map_t()) :: utxos_map_t()
+  def add_utxos_for_stealth_merge(available_utxos, selected_utxos) do
```
InoMurko, line 91: mixing public and private functions again

apps/omg_watcher_info/lib/omg_watcher_info/utxo_selection.ex
```diff
@@ -19,93 +19,116 @@ defmodule OMG.WatcherInfo.UtxoSelection do
...
+    cond do
+      get_number_of_utxos(selected_utxos) == Transaction.Payment.max_inputs() ->
+        selected_utxos
+
+      Enum.empty?(available_utxos) ->
+        selected_utxos
```
InoMurko, line 97: you can remove this clause if you add a function header above the current one
```
def add_utxos_for_stealth_merge([], selected_utxos), do: selected_utxos
```

apps/omg_watcher_info/lib/omg_watcher_info/utxo_selection.ex
```diff
@@ -19,93 +19,116 @@ defmodule OMG.WatcherInfo.UtxoSelection do
...
-  defp select_utxo(utxos, needed_funds) do
+  @doc """
+  Given the available set of UTXOs and the needed amount by currency, tries to find a UTXO that satisfies the payment with no change.
+  If this fails, starts to collect UTXOs (starting from the largest amount) until the payment is covered.
+  Returns {currency, { variance, [utxos] }}. A `variance` greater than zero means insufficient funds.
+  The ordering of UTXOs in descending order of amount is implicitly assumed for this algorithm to work deterministically.
+  """
+  @spec select_utxo(%{currency_t() => pos_integer()}, utxos_map_t()) ::
+          list({currency_t(), {integer, utxo_list_t()}})
+  def select_utxo(needed_funds, utxos) do
```
InoMurko, line 119: mixing public and private functions

apps/omg_watcher_info/lib/omg_watcher_info/utxo_selection.ex
```diff
@@ -114,118 +137,46 @@ defmodule OMG.WatcherInfo.UtxoSelection do
...
-  end
-
-  defp compute_sign_hash(tx) do
-    with tx when not is_nil(tx) <- tx,
-         do: TypedDataHash.hash_struct(tx)
+         utxo_selection
+         |> Enum.reduce(%{}, fn {token, {_missing_amount, utxos}}, acc ->
+           Map.put(acc, token, utxos)
+         end)},
+      else: {:error, {:insufficient_funds, missing_funds}}
```
InoMurko, line 180: don't use single line if/else notation if the statements are multiline

InoMurko (changes requested):

apps/omg_watcher_info/lib/omg_watcher_info/utxo_selection.ex
```diff
@@ -19,93 +19,116 @@ defmodule OMG.WatcherInfo.UtxoSelection do
...
+      get_number_of_utxos(selected_utxos) == Transaction.Payment.max_inputs() ->
+        selected_utxos
+
+      Enum.empty?(available_utxos) ->
+        selected_utxos
+
+      true ->
+        [priority_utxo | remaining_available_utxos] = available_utxos
+
+        stealth_merged_utxos =
```
okalouti, line 102: suggest `merge` instead of `merged`

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -1359,6 +1268,348 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
...
+                     %{"amount" => 10, "currency" => @eth_hex},
+                     %{"amount" => 5, "currency" => @eth_hex},
+                     %{"amount" => 20, "currency" => @eth_hex},
+                     %{"amount" => 30, "currency" => @eth_hex}
+                   ]
+                 }
+               ]
+             } = WatcherHelper.success?("transaction.create", params)
+
+      # Assert when payment amount exactly matched with input amounts
```
okalouti, line 1484: comment here should be _doesn't_ match.

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -1359,6 +1268,348 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
...
+          %{"amount" => 20, "currency" => @eth_hex, "owner" => Encoding.to_hex(bob.addr)}
+        ],
+        "fee" => %{"currency" => @eth_hex}
+      }
+
+      assert %{
+               "transactions" => [
+                 %{
+                   "fee" => %{
+                     "amount" => 5,
```
okalouti, line 1599: Question for all the tests. Anywhere you can access the fee amount from this server in a variable? So that we don't hard-code it?

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -1359,6 +1268,348 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
...
+                     "currency" => @eth_hex
+                   },
```
okalouti, line 1601: Another thought: at the endpoint level I think we should also include expected outputs in these tests, to ensure that  the transactions are "what we expect" in a more comprehensive sense. No new tests, just adding outputs into the assertions.

okalouti:
For when you're done with the "code part" ... should probably add a note about this functionality in the Swagger docs.

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
ripzery: As you said, we already had that check in the validator very early from the controller.

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
ripzery: It won't be needed now since `owner` from `create_raw_transaction` (conversation above) can't be `nil` at this point.

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -112,12 +173,10 @@ defmodule OMG.WatcherInfo.API.Transaction do
   end
 
   defp create_inputs(inputs) do
```
okalouti, line 175: Do `create_inputs` and `create_outputs` belong better in the transaction creator module?

apps/omg_watcher_info/lib/omg_watcher_info/transaction.ex
```diff
@@ -0,0 +1,127 @@
...
+  @doc """
+  Given selected utxos and order, create inputs and outputs, then returns a transaction.
+
+  Returns an error when any of the following conditions is met:
+
+  1. A number of outputs overs maximum.
+  2. A number of inputs overs maximum.
+  3. An inputs are empty.
+  """
+  @spec create(utxos_map_t(), order_t()) :: create_t()
```
okalouti, line 64: could we add some positive and negative tests for this function?

apps/omg_watcher_info/lib/omg_watcher_info/transaction.ex
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+  @type order_t() :: %{
+          owner: Crypto.address_t(),
+          payments: nonempty_list(UtxoSelection.payment_t()),
+          metadata: binary() | nil,
+          fee: UtxoSelection.fee_t()
+        }
+
+  @type utxos_map_t() :: %{UtxoSelection.currency_t() => UtxoSelection.utxo_list_t()}
```
okalouti, line 53: Can we just use the type defined in `utxo_selection.ex` instead of redefining it ?

apps/omg_watcher_info/lib/omg_watcher_info/utxo_selection.ex
```diff
@@ -17,95 +17,96 @@ defmodule OMG.WatcherInfo.UtxoSelection do
...
-        do: create_transaction(funds, order) |> respond(:complete),
-        else: create_merge(owner, funds) |> respond(:intermediate)
+  @spec prioritize_merge_utxos(utxos_map_t(), list({currency_t(), utxo_list_t()})) :: utxo_list_t()
+  def prioritize_merge_utxos(utxos, selected_utxos) do
+    selected_utxo_hashes =
+      selected_utxos
+      |> Enum.flat_map(fn {_ccy, utxos} -> utxos end)
+      |> Enum.reduce(%{}, fn utxo, acc -> Map.put(acc, utxo.child_chain_utxohash, true) end)
+
+    case selected_utxo_hashes |> Map.keys() |> length() do
```
mederic-p, line 46: Can't you just check if it's equal to `%{}`?

apps/omg_watcher_info/lib/omg_watcher_info/utxo_selection.ex
```diff
@@ -114,118 +115,60 @@ defmodule OMG.WatcherInfo.UtxoSelection do
...
-  defp needed_funds(payments, %{currency: fee_currency, amount: fee_amount}) do
+  @doc """
+  Sums up payable amount by token, including the fee.
+  """
+  @spec needed_funds(list(TransactionCreator.payment_t()), %{amount: pos_integer(), currency: currency_t()}) ::
+          %{currency_t() => pos_integer()}
+  def needed_funds(payments, %{currency: fee_currency, amount: fee_amount}) do
     needed_funds =
       payments
-      |> Enum.group_by(& &1.currency)
```
mederic-p, line 121: I have no problem with the `&` short notation

mederic-p (commented):
I have the feeling that the overall process could be simplified, it seems that we are doing a loootttt of Enum.Map / Reduce / Stream ...
On the other side, this is not a critical endpoint and performance shouldn't be too much of a problem, but it's always nice to have at least a minimum of optimization.

apps/omg_watcher_info/lib/omg_watcher_info/transaction.ex
```diff
@@ -0,0 +1,228 @@
...
+          transaction
+      end
+    end)
+  end
+
+  defp create_merge(inputs) do
+    %{currency: currency, owner: owner} = List.first(inputs)
+
+    create([{currency, inputs}], %{
+      fee: %{amount: 0, currency: currency},
```
mederic-p, line 156: Can we move the `0` here to a `@merge_fee` at the top?

apps/omg_watcher_info/lib/omg_watcher_info/utxo_selection.ex
```diff
@@ -17,95 +17,98 @@ defmodule OMG.WatcherInfo.UtxoSelection do
...
-      utxo_count =
-        funds
-        |> Stream.map(fn {_, utxos} -> length(utxos) end)
-        |> Enum.sum()
-
-      if utxo_count <= Transaction.Payment.max_inputs(),
-        do: create_transaction(funds, order) |> respond(:complete),
-        else: create_merge(owner, funds) |> respond(:intermediate)
+  @spec prioritize_merge_utxos(utxos_map_t(), utxos_map_t()) :: utxo_list_t()
+  def prioritize_merge_utxos(utxos, selected_utxos) do
```
mederic-p, line 40: Should be able to write
```
def prioritize_merge_utxos(utxos, selected_utxos) do
  selected_utxos
  |> Enum.flat_map(fn {_ccy, utxos} -> utxos end)
  |> Enum.reduce(%{}, fn utxo, acc -> Map.put(acc, utxo.child_chain_utxohash, true) end)
  |> case do
    %{} ->
      []

    hashes_map ->
      selected_utxos
      |> Enum.map(&prioritize_utxos_by_currency(&1, utxos, hashes_map))
      |> Enum.sort_by(&length/1, :desc)
      |> Enum.map(fn ccy_group -> Enum.slice(ccy_group, 0, 3) end)
      |> List.flatten()
  end
end
```

apps/omg_watcher_info/lib/omg_watcher_info/utxo_selection.ex
```diff
@@ -17,95 +17,98 @@ defmodule OMG.WatcherInfo.UtxoSelection do
...
     Enum.map(needed_funds, fn {token, need} ->
       token_utxos = Map.get(utxos, token, [])
 
       {token,
        case Enum.find(token_utxos, fn %DB.TxOutput{amount: amount} -> amount == need end) do
          nil ->
            Enum.reduce_while(token_utxos, {need, []}, fn
-             _, {need, acc} when need <= 0 -> {:halt, {need, acc}}
-             %DB.TxOutput{amount: amount} = utxo, {need, acc} -> {:cont, {need - amount, [utxo | acc]}}
+             _, {need, acc} when need <= 0 ->
```
mederic-p, line 107: This is a reduce_while in a case in a map, I know it's not from you but if you can simply it a bit (split in functions?) it'd be great

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -77,55 +90,45 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
+    |> List.flatten()
+    |> Enum.count()
   end
 
-  defp create_inputs(inputs) do
-    empty_gen = fn -> %{blknum: 0, txindex: 0, oindex: 0} end
+  defp create_transaction(utxos_count, inputs, _order) when utxos_count > Transaction.Payment.max_inputs() do
+    transactions =
+      inputs
+      |> Enum.map(fn {_currency, inputs} ->
```
mederic-p, line 118: Maybe use a different variable name here `inputs` is already the one you Map on.

apps/omg_watcher_info/lib/omg_watcher_info/transaction.ex
```diff
@@ -0,0 +1,252 @@
...
+  @type inputs_t() :: {:ok, utxos_map_t()} | {:error, {:insufficient_funds, list(map())}}
+
+  @doc """
+  Given an `order`, finds spender's inputs sufficient to perform a payment.
+  If also provided with receiver's address, creates and encodes a transaction.
+  """
+  @spec select_inputs(utxos_map_t(), order_t()) :: inputs_t()
+  def select_inputs(utxos, %{payments: payments, fee: fee}) do
+    payments
+    |> UtxoSelection.needed_funds(fee)
```
mederic-p, line 88: I think we could rename these functions to be more explicit on what they do. It's a bit confusing.
(I'm talking about `needed_funds`, `select_utxo`, `funds_sufficient`).
Or/and you could also add more comments in the code to tell exactly what it does.

apps/omg_watcher_info/lib/omg_watcher_info/transaction.ex
```diff
@@ -0,0 +1,252 @@
...
+        [metadata: metadata || @empty_metadata]
+      ]
+      |> Enum.concat()
+      |> Map.new()
+
+    %{
+      domain: TypedDataHash.Config.domain_data_from_config(),
+      message: message
+    }
+    |> Map.merge(TypedDataHash.Types.eip712_types_specification())
```
mederic-p, line 234: Don't use pipes (`|>`) when you only have 2 parts.

apps/omg_watcher_info/lib/omg_watcher_info/transaction.ex
```diff
@@ -85,9 +85,9 @@ defmodule OMG.WatcherInfo.Transaction do
   @spec select_inputs(utxos_map_t(), order_t()) :: inputs_t()
   def select_inputs(utxos, %{payments: payments, fee: fee}) do
     payments
-    |> UtxoSelection.needed_funds(fee)
-    |> UtxoSelection.select_utxo(utxos)
-    |> UtxoSelection.funds_sufficient()
+    |> UtxoSelection.calculate_net_amount(fee) # calculate total amount to satisfy payments and fee.
+    |> UtxoSelection.select_utxos(utxos) # tries to find utxos that satisfies payments and fee.
+    |> UtxoSelection.review_selected_utxos() # review if utxos can satisfy payments and fee
```
mederic-p: 👍

apps/omg_watcher_info/lib/omg_watcher_info/api/transaction.ex
```diff
@@ -77,55 +88,45 @@ defmodule OMG.WatcherInfo.API.Transaction do
...
   If also provided with receiver's address, creates and encodes a transaction.
   """
-  @spec create(UtxoSelection.order_t()) :: UtxoSelection.advice_t()
+  @spec create(TransactionCreator.order_t()) :: create_t()
   def create(order) do
-    utxos = DB.TxOutput.get_sorted_grouped_utxos(order.owner)
-    UtxoSelection.create_advice(utxos, order)
+    case order.owner
+         |> DB.TxOutput.get_sorted_grouped_utxos()
+         |> TransactionCreator.select_inputs(order) do
```
InoMurko, line 95:
```suggestion
owner_inputs = order.owner |> DB.TxOutput.get_sorted_grouped_utxos() |> TransactionCreator.select_inputs(order)
    case owner_inputs do
```

InoMurko (changes requested):

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -19,7 +19,8 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
   use OMG.WatcherInfo.Fixtures
   use OMG.Watcher.Fixtures
 
-  import OMG.WatcherInfo.Factory, only: [build: 1, with_deposit: 1, insert: 1, with_inputs: 2, with_outputs: 2]
+  # only: [build: 1, with_deposit: 1, insert: 1, with_inputs: 2, with_outputs: 2]
+  import OMG.WatcherInfo.Factory
```
InoMurko, line 23: you need to specify what you're importing or not use import at all

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -32,6 +33,9 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
   alias Support.WatcherHelper
 
   require OMG.State.Transaction.Payment
+  import OMG.WatcherInfo.Factory
```
InoMurko, line 36: above ^^

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -781,7 +785,6 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
       creating_txhash = inserted_txs |> Enum.at(0) |> elem(2) |> Encoding.to_hex()
 
       assert %{
-               "result" => "complete",
```
InoMurko, line 784: this PR seems to remove things from the response?

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -1357,6 +1267,419 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
       |> TestServer.make_response()
       |> TestServer.with_response(context, "/fees.all")
     end
+
+    defp max_amount_spendable_in_single_tx(address, token) do
```
InoMurko, line 1271: tests are mixed with private functions

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
```diff
@@ -1357,6 +1267,419 @@ defmodule OMG.WatcherRPC.Web.Controller.TransactionTest do
...
+
+  describe "/transaction.create stealth add inputs" do
+    setup tags do
+      context = TestServer.start()
+      on_exit(fn -> TestServer.stop(context) end)
+      Map.put(tags, :test_server, context)
+    end
+
+    @tag fixtures: [:phoenix_ecto_sandbox]
+    test "does not stealth add inputs where the number of inputs reached maximum", %{
```
InoMurko, line 1292: can you elaborate what this fraze means? `does not stealth add inputs`?

apps/omg_watcher_info/lib/omg_watcher_info/utxo_selection.ex
```diff
@@ -17,215 +17,163 @@ defmodule OMG.WatcherInfo.UtxoSelection do
...
-         nil ->
-           Enum.reduce_while(token_utxos, {need, []}, fn
-             _, {need, acc} when need <= 0 -> {:halt, {need, acc}}
-             %DB.TxOutput{amount: amount} = utxo, {need, acc} -> {:cont, {need - amount, [utxo | acc]}}
-           end)
-
-         utxo ->
-           {0, [utxo]}
-       end}
+  defp prioritize_utxos_by_currency({currency, _utxos}, utxos, selected_utxo_hashes) do
```
InoMurko, line 56: I have now repeated this sooo many times :)))
don't mix private and public functions

apps/omg_watcher_info/lib/omg_watcher_info/transaction.ex
```diff
@@ -0,0 +1,257 @@
...
+          utxos
+          |> UtxoSelection.prioritize_merge_utxos(funds)
+          |> UtxoSelection.add_utxos_for_stealth_merge(funds)
+
+        {:ok, stealth_merge_utxos}
+
+      err ->
+        err
+    end
+  end
```
InoMurko, line 106: break this down, no need to pipe into a case

InoMurko (changes requested):

apps/omg_watcher_rpc/test/omg_watcher_rpc/web/controllers/transaction_test.exs
ripzery: we use `stealth` word when we're going to select extra inputs just for merging (optimising) it.

apps/omg_watcher_info/lib/omg_watcher_info/transaction.ex
```diff
@@ -0,0 +1,258 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.WatcherInfo.Transaction do
+  @moduledoc """
+  Module create transaction from selected utxos and order.
```
InoMurko, line 17:
```suggestion
  Module creates transaction from selected utxos and order.
```

mederic-p (approved):

InoMurko (approved):

PR #1687, opened 2020-08-07
feat: use fallback calldata provider

pgebal:
References https://github.com/omgnetwork/private-issues/issues/64.
Uses a fallback call data provider - Tenderly in case there are problems parsing function call data.


InoMurko:
Before I review this ... is the plan that everyone that runs a watcher or watcher info has a tenderly account setup?

pgebal:
> Before I review this ... is the plan that everyone that runs a watcher or watcher info has a tenderly account setup?
> 
> Because if we don't, I don't see how this is solving a problem. All watchers can come to a halt if they don't know how to get calldata similarly to our childchain.

No. The plan is that we use tenderly as a fallback, so our child chain and watcher doesn't halt. It's an interim solution until we deploy new contracts, where all the necessary data will be provided either in events or via functions.

InoMurko:
> > Before I review this ... is the plan that everyone that runs a watcher or watcher info has a tenderly account setup?
> > Because if we don't, I don't see how this is solving a problem. All watchers can come to a halt if they don't know how to get calldata similarly to our childchain.
> 
> No. The plan is that we use tenderly as a fallback, so our child chain and watcher doesn't halt. It's an interim solution until we deploy new contracts, where all the necessary data will be provided either in events or via functions.

Does Bitfinex need a tenderly account if their watcher doesn't want to halt?

pgebal:
> > > Before I review this ... is the plan that everyone that runs a watcher or watcher info has a tenderly account setup?
> > > Because if we don't, I don't see how this is solving a problem. All watchers can come to a halt if they don't know how to get calldata similarly to our childchain.
> > 
> > 
> > No. The plan is that we use tenderly as a fallback, so our child chain and watcher doesn't halt. It's an interim solution until we deploy new contracts, where all the necessary data will be provided either in events or via functions.
> 
> Does Bitfinex need a tenderly account if their watcher doesn't want to halt?
Yes.

InoMurko:
Is that acceptable from the product? @achiurizo @T-Dnzt ?

boolafish:
Probably a possible  solution for watcher is to fallback to a new event api from child chain. We solve ChildChain by tenderly and solves watcher by ChildChain

pgebal:
> Probably a possible solution for watcher is to fallback to a new event api from child chain. We solve ChildChain by tenderly and solves watcher by ChildChain

Very nice. Is this API merged?

boolafish:
> Very nice. Is this API merged?

lol this is a theoretical solution so I don't think anything is worked on that yet....😅
I have not even verified this is an okay solution 😬

pgebal:
> > Very nice. Is this API merged?
> 
> lol this is a theoretical solution so I don't think anything is worked on that yet....sweat_smile
> I have not even verified this is a truly okay without issue solution grimacing

Ok. I thought I missed something and we have it already. On the other hand, I don't thing watcher should rely on contracts event data to be provided by child-chain. It can be an attack vector if child-chain is malicious.

boolafish:
For the patching of the issue itself, given the specific data we are trying to fetch: `utxoPos`, `rlpOutputTx`, `outputTxInclusionProof` watcher can do sth like the following to not need to trust childchain on the fetched event data:
- verify the utxoPos + rlpOutputTx -> same exit Id from the event
- verify the output tx inclusion proof is indeed the proof of that tx with the data from watcher itself

(should be enough right?)

However, I do not know if there is a way to generalize the solution instead of patching for this specific use case.

pgebal:
@InoMurko   could you please do a CR? So we have this PR ready just in case we have to deploy it to rescue our instances of child-chain and watcher.

InoMurko:
> @InoMurko could you please do a CR? So we have this PR ready just in case we have to deploy it to rescue our instances of child-chain and watcher.

I don't see how this is a usable solution in any way. How do we roll this out? Do we give free Tenderly accounts to everyone? What would the plan look like?

pgebal:
> > @InoMurko could you please do a CR? So we have this PR ready just in case we have to deploy it to rescue our instances of child-chain and watcher.
> 
> I don't see how this is a usable solution in any way. How do we roll this out? Do we give free Tenderly accounts to everyone? What would the plan look like?

It might be useful as an emergency for us if it happens that we don't have access to ABI of a contract that made the call.

InoMurko:
> > > @InoMurko could you please do a CR? So we have this PR ready just in case we have to deploy it to rescue our instances of child-chain and watcher.
> > 
> > 
> > I don't see how this is a usable solution in any way. How do we roll this out? Do we give free Tenderly accounts to everyone? What would the plan look like?
> 
> It might be useful as an emergency for us if it happens that we don't have access to ABI of a contract that made the call.

I prefer if we extract this into a tool that we can use that would help us create a fix of childchain and both watcher, rather then keeping this in master.

pgebal:
> > > > @InoMurko could you please do a CR? So we have this PR ready just in case we have to deploy it to rescue our instances of child-chain and watcher.
> > > 
> > > 
> > > I don't see how this is a usable solution in any way. How do we roll this out? Do we give free Tenderly accounts to everyone? What would the plan look like?
> > 
> > 
> > It might be useful as an emergency for us if it happens that we don't have access to ABI of a contract that made the call.
> 
> I prefer if we extract this into a tool that we can use that would help us create a fix of childchain and both watchers, rather then keeping this in master.

That's a good idea.

pgebal:
Closing until we decide if we want to fix this issue using full archive node geth.

PR #1690, opened 2020-08-10
increase linger timeouts

ayrat555:


config/config.exs
```diff
@@ -210,7 +214,11 @@ config :omg_watcher_rpc,
 config :omg_watcher_rpc, OMG.WatcherRPC.Web.Endpoint,
   render_errors: [view: OMG.WatcherRPC.Web.Views.Error, accepts: ~w(json)],
   enable_cors: true,
-  http: [:inet6, port: 7434, protocol_options: [max_request_line_length: 8192, max_header_value_length: 8192]],
+  http: [
+    :inet6,
+    port: 7434,
+    protocol_options: [max_request_line_length: 8192, max_header_value_length: 8192, linger_timeout: 5_000]
```
InoMurko, line 220: I think we agreed that we don't have to change `linger_timeout`. The default is:
```
linger_timeout (1000)
Time in ms that Cowboy will wait when closing the connection. This is necessary to avoid the TCP reset problem as described in the section 6.6 of RFC7230.

```
which should solve the problem. SO I guess we can close this PR. cc @arthurk @jbunce

InoMurko: the confusion was that I thought the cowboy update PR hasn't been merged. :))) sry

arthurk: Yes 5 seconds is too much and we can keep the default of 1 second. I've checked a few tcp streams and the delayed packages arrived around 0.000404000 seconds (no idea what this is, nanoseconds?) after the socket was closed

PR #1691, opened 2020-08-10
move omg_performance json rpc tests to perf project

ayrat555:
This PR moves `omg_perfomance` json rpc tests from `omg_performance` project to `perf`. Unfortunately, not all tests can be moved because some of them test internal module functions

Fixes https://github.com/omgnetwork/elixir-omg/issues/1600


priv/perf/Makefile
```diff
@@ -16,7 +16,7 @@ init:
 	mix deps.get
 
 test: # runs test against child chain and geth provided in test docker containers
-	CONTRACT_ADDRESS_ETH_VAULT=0x4e3aeff70f022a6d4cc5947423887e7152826cf7 LOAD_TEST_FAUCET_PRIVATE_KEY=0xd885a307e35738f773d8c9c63c7a3f3977819274638d04aaf934a1e1158513ce CONTRACT_ADDRESS_PLASMA_FRAMEWORK=0xc673e4ffcb8464faff908a6804fe0e635af0ea2f mix test
+	CONTRACT_ADDRESS_ETH_VAULT=0x4e3aeff70f022a6d4cc5947423887e7152826cf7 LOAD_TEST_FAUCET_PRIVATE_KEY=0xd885a307e35738f773d8c9c63c7a3f3977819274638d04aaf934a1e1158513ce CONTRACT_ADDRESS_PLASMA_FRAMEWORK=0xc673e4ffcb8464faff908a6804fe0e635af0ea2f CONTRACT_ADDRESS_PAYMENT_EXIT_GAME=0x89afce326e7da55647d22e24336c6a2816c99f6b mix test --trace
```
InoMurko, line 19: why is this data hardcoded?

priv/perf/apps/load_test/lib/common/byzantine_events.ex
```diff
@@ -216,20 +246,23 @@ defmodule OMG.Performance.ByzantineEvents do
...
-  defp watcher_synchronized?(root_chain_height, watcher_url) do
-    {:ok, status} = WatcherClient.get_status(watcher_url)
+  defp watcher_synchronized?(root_chain_height) do
+    {:ok, status_response} =
+      WatcherSecurityCriticalAPI.Api.Status.status_get(LoadTest.Connection.WatcherSecurity.client())
+
+    status = Jason.decode!(status_response.body)["data"]
 
-    with true <- watcher_synchronized_to_mined_block?(status),
+    with :ok <- watcher_synchronized_to_mined_block?(status),
```
InoMurko, line 265: hmmm this should be a true/false function

priv/perf/apps/load_test/lib/common/extended_perftest.ex
```diff
@@ -0,0 +1,91 @@
...
+
+    defaults = [destdir: "."]
+
+    opts = Keyword.merge(defaults, opts)
+
+    utxos = create_deposits(spenders, ntx_to_send)
+
+    result = LoadTest.Common.Runner.run(ntx_to_send, utxos, opts, false)
+
+    Process.sleep(30_000)
```
InoMurko, line 68: is there a way to make this poll for results?

priv/perf/apps/load_test/lib/common/generators.ex
```diff
@@ -0,0 +1,128 @@
...
+  Creates addresses with private keys and funds them with given `initial_funds_wei` on geth.
+
+  Options:
+    - :faucet - the address to send the test ETH from, assumed to be unlocked and have the necessary funds
+    - :initial_funds_wei - the amount of test ETH that will be granted to every generated user
+  """
+  @spec generate_users(non_neg_integer) :: [OMG.TestHelper.entity()]
+  def generate_users(size) do
+    1..size
+    |> Task.async_stream(fn _ -> generate_user() end, timeout: @generate_user_timeout)
```
InoMurko, line 37: the problem with this test is that they only run towards geth or ganache in ci or locally, and because of that, it makes little sense to call them performance tests.  they cannot give any reliable information of the performance of our system because they're not being run in isolation, on the same hardware, with vacuum setup.
so the question is, what value do these tests give us? I think it's very limited if not harmful because they take engineering hours and CPU cycles (running them, maintaining them).

@achiurizo

priv/perf/apps/load_test/lib/common/sender_server.ex
```diff
@@ -0,0 +1,236 @@
...
+
+  # Handles result of successful Tx submission or retry request into new state and sends :do message
+  @spec update_state_with_tx_submission(
+          tx_submit_result :: {:ok, map} | :retry | {:error, any},
+          state :: __MODULE__.state()
+        ) :: __MODULE__.state()
+  defp update_state_with_tx_submission(
+         tx_submit_result,
+         %__MODULE__{seqnum: seqnum, last_tx: last_tx} = state
+       ) do
```
InoMurko, line 169:
```suggestion
  defp update_state_with_tx_submission(tx_submit_result, state) do
```
and then use `.` access modifier (state.last_tx)

priv/perf/apps/load_test/lib/common/sender_server.ex
```diff
@@ -0,0 +1,236 @@
...
+
+        if newblknum > last_tx.blknum,
+          do:
+            LoadTest.Common.SenderManager.sender_stats(%{
+              seqnum: seqnum,
+              blknum: last_tx.blknum,
+              txindex: last_tx.txindex
+            })
+
+        state |> next_state(newblknum, newtxindex, newvalue)
```
InoMurko, line 182:
```suggestion
        next_state(state, newblknum, newtxindex, newvalue)
```

boolafish:
Just saw Ino’s comment on running against geth locally. I second his opinion for those tests that can only run against local geth we should not put into perf. The perf is designed to run agains an env though we put docker in CI to really just sanity check the perf test code is runable.

priv/perf/apps/load_test/lib/common/generators.ex
ayrat555: I think most of the tests in perf (if not all) are running against geth

InoMurko: yeah, that's probably true. if we want to make them useful in any way, it makes sense to review them against the existing perf tests and adapt them into the same "form", so that we can use infura or geth to stress test!

priv/perf/Makefile
ayrat555: should I use `localchain_contract_addresses.env` file?

.circleci/config.yml
```diff
@@ -709,11 +725,11 @@ jobs:
             ps axww | grep watcher_info
             ps axww | grep child_chain
       - run: sh .circleci/status.sh
-      - run:
-          name: Run specs
-          command: |
-            cd priv/cabbage
-            mix test
+      # - run:
```
ayrat555, line 728: I commented cabbage test for faster load test runs. I'll uncomment them

priv/perf/Makefile
InoMurko: yeah, that's like the location for all contract addresses coming from the snapshot, everyone should be using this, so that when we upgrade to v2 contracts, we change only the snapshot URL.

.circleci/config.yml
InoMurko: what if we separate cabbage and load test runs into two jobs?

priv/perf/apps/load_test/lib/common/sender_server.ex
```diff
@@ -0,0 +1,237 @@
...
+      ntx_to_send: ntx_to_send,
+      spender: spender,
+      last_tx: %LastTx{
+        # initial state takes deposited value, put there on :init
+        blknum: blknum,
+        txindex: txindex,
+        oindex: oindex,
+        amount: amount
+      },
+      child_chain_url: Application.fetch_env!(:load_test, :child_chain_url),
```
InoMurko, line 207: where are you actually setting this?

priv/perf/apps/load_test/lib/common/sender_server.ex
```diff
@@ -0,0 +1,237 @@
...
+          {:noreply, new_state :: __MODULE__.state()} | {:stop, :normal, __MODULE__.state()}
+  def handle_info(
+        :do,
+        %__MODULE__{ntx_to_send: 0, seqnum: seqnum, last_tx: %LastTx{blknum: blknum, txindex: txindex}} = state
+      ) do
+    _ = Logger.info("[#{inspect(seqnum)}] Stoping...")
+
+    LoadTest.Common.SenderManager.sender_stats(%{seqnum: seqnum, blknum: blknum, txindex: txindex})
+    {:stop, :normal, state}
+  end
```
InoMurko, line 111:
```suggestion
  def handle_info(:do, %__MODULE__{ntx_to_send: 0} = state) do
    _ = Logger.info("[#{inspect(seqnum)}] Stoping...")

    LoadTest.Common.SenderManager.sender_stats(%{seqnum: state.seqnum, blknum: state.last_tx.blknum, txindex: state.last_tx.txindex})
    {:stop, :normal, state}
  end
```

priv/perf/apps/load_test/lib/common/generators.ex
```diff
@@ -0,0 +1,128 @@
...
+  defp poll_get_block(block_hash, child_chain_url) do
+    poll_get_block(block_hash, child_chain_url, 50)
+  end
+
+  defp poll_get_block(block_hash, child_chain_url, retry) do
+    case ChildChainAPI.Api.Block.block_get(
+           LoadTest.Connection.ChildChain.client(),
+           %ChildChainAPI.Model.GetBlockBodySchema{hash: Encoding.to_hex(block_hash)}
+         ) do
+      {:ok, block_response} ->
```
InoMurko, line 120:
```suggestion
client = LoadTest.Connection.ChildChain.client()
body = %ChildChainAPI.Model.GetBlockBodySchema{hash: Encoding.to_hex(block_hash)}
block = ChildChainAPI.Api.Block.block_get(client, body)
    case block do
      {:ok, block_response} ->
```

priv/perf/apps/load_test/lib/common/byzantine_events.ex
```diff
@@ -216,44 +249,70 @@ defmodule OMG.Performance.ByzantineEvents do
...
     else
       _ -> :repeat
     end
   end
 
-  defp root_chain_synced?(nil, _), do: true
+  defp root_chain_synced?(nil, _, _), do: true
+
+  defp root_chain_synced?(root_chain_height, status, nil) do
+    IO.inspect(
```
InoMurko, line 279: inspect!

InoMurko (changes requested):
There's a lot of things that was copied (like the ABI decoind modules). Is that used? I didn't find a lot of things actually being used...

.circleci/config.yml
ayrat555: I created a separate ci step for load tests

priv/perf/apps/load_test/lib/common/sender_server.ex
ayrat555: removed

priv/perf/apps/load_test/lib/child_chain/transaction.ex
```diff
@@ -103,6 +110,100 @@ defmodule LoadTest.ChildChain.Transaction do
     end)
   end
 
+  def recover(encoded_signed_tx) do
```
InoMurko, line 113: mixing private and public functions here

priv/perf/apps/load_test/lib/performance.ex
```diff
@@ -0,0 +1,62 @@
...
+
+      :ok
+    end
+  end
+
+  def init() do
+    {:ok, _} = Application.ensure_all_started(:briefly)
+    {:ok, _} = Application.ensure_all_started(:ethereumex)
+    {:ok, _} = Application.ensure_all_started(:hackney)
+    {:ok, _} = Application.ensure_all_started(:cowboy)
```
InoMurko, line 40: hmm... why cowboy?

InoMurko (approved):

priv/perf/apps/load_test/lib/performance.ex
ayrat555: it seems this function is not needed

priv/perf/apps/load_test/lib/common/byzantine_events.ex
```diff
@@ -191,9 +216,12 @@ defmodule OMG.Performance.ByzantineEvents do
...
   def get_byzantine_events() do
-    watcher_url = Application.fetch_env!(:omg_performance, :watcher_url)
-    {:ok, status_response} = WatcherClient.get_status(watcher_url)
-    status_response[:byzantine_events]
+    {:ok, status_response} =
+      WatcherSecurityCriticalAPI.Api.Status.status_get(LoadTest.Connection.WatcherSecurity.client())
+
+    status = Jason.decode!(status_response.body)["data"]
+
+    status["byzantine_events"]
```
achiurizo, line 224: seems brittle here. Should we be more defensive here? Also if you really want to go this route, may as well condense:

```suggestion
    status = Jason.decode!(status_response.body)["data"]["byzantine_events"]
```

priv/perf/apps/load_test/lib/common/byzantine_events.ex
```diff
@@ -216,44 +244,63 @@ defmodule OMG.Performance.ByzantineEvents do
...
-  defp root_chain_synced?(root_chain_height, status) do
+  defp root_chain_synced?(root_chain_height, status, nil) do
     status
-    |> Access.get(:services_synced_heights)
+    |> Map.get("services_synced_heights")
+    |> Enum.reject(fn height ->
+      service = height["service"]
+      # these service heights are stuck on circle ci, but they work fine locally
+      # I think ci machin is not powerful enough
+      service == "block_getter" || service == "exit_finalizer" || service == "ife_exit_finalizer"
```
achiurizo, line 280: Open an issue for this? :pray:

priv/perf/apps/load_test/lib/common/sender_server.ex
```diff
@@ -109,114 +105,92 @@ defmodule OMG.Performance.SenderServer do
...
+    change_output =
+      if new_amount > 0, do: [%ExPlasma.Utxo{owner: state.spender.addr, currency: @eth, amount: new_amount}], else: []
+
+    [%{blknum: blknum, txindex: txindex, amount: amount} | _] =
+      Transaction.submit_tx(
+        [%ExPlasma.Utxo{blknum: state.last_tx.blknum, txindex: state.last_tx.txindex, oindex: state.last_tx.oindex}],
+        change_output ++ recipient_output,
+        [
+          state.spender
+        ],
```
achiurizo, line 152:
```suggestion
        [state.spender],
```

priv/perf/apps/load_test/lib/common/byzantine_events.ex
ayrat555: https://github.com/omgnetwork/elixir-omg/issues/1711

PR #1693, opened 2020-08-14
Block Validation: New Features

okalouti:
…on in block validation

## Overview

## Changes

- Change 1
- Change 2
- ...

## Testing

### DIFF
--- a/apps/omg_watcher/lib/omg_watcher/block_validator.ex
+++ b/apps/omg_watcher/lib/omg_watcher/block_validator.ex
@@ -20,15 +20,22 @@ defmodule OMG.Watcher.BlockValidator do
   alias OMG.Block
   alias OMG.Merkle
   alias OMG.State.Transaction
+  alias OMG.Utxo
+
+  @transaction_upper_limit 2 |> :math.pow(16) |> Kernel.trunc()
 
   @doc """
   Executes stateless validation of a submitted block:
+  - Verifies that the number of transactions falls within the accepted range.
   - Verifies that transactions are correctly formed.
+  - Verifies that there are no duplicate inputs at the block level.
   - Verifies that given Merkle root matches reconstructed Merkle root.
   """
   @spec stateless_validate(Block.t()) :: {:ok, boolean()} | {:error, atom()}
   def stateless_validate(submitted_block) do
-    with {:ok, recovered_transactions} <- verify_transactions(submitted_block.transactions),
+    with :ok <- number_of_transactions_within_limit(submitted_block.transactions),
+         {:ok, recovered_transactions} <- verify_transactions(submitted_block.transactions),
+         {:ok, _inputs} <- verify_no_duplicate_inputs(recovered_transactions),
          {:ok, _block} <- verify_merkle_root(submitted_block, recovered_transactions) do
       {:ok, true}
     end
@@ -48,7 +55,7 @@ defmodule OMG.Watcher.BlockValidator do
     end
   end
 
-  @spec verify_transactions(transactions :: list(Transaction.Recovered.t())) ::
+  @spec verify_transactions(transactions :: list(Transaction.Signed.txbytes())) ::
           {:ok, list(Transaction.Recovered.t())}
           | {:error, Transaction.Recovered.recover_tx_error()}
   defp verify_transactions(transactions) do
@@ -62,6 +69,43 @@ defmodule OMG.Watcher.BlockValidator do
         error ->
           {:halt, error}
       end
+    end)
+  end
+
+  @spec number_of_transactions_within_limit([Transaction.Signed.tx_bytes()]) ::
+          :ok | {:error, atom()}
+  defp number_of_transactions_within_limit(transactions) do
+    case length(transactions) do
+      # A block should at least have two transactions: a payment transaction and a fee transaction.
+      n when n < 2 ->
+        {:error, :transactions_below_block_minimum}
+
+      n when n > @transaction_upper_limit ->
+        {:error, :transactions_exceed_block_limit}
+
+      _ ->
+        :ok
+    end
+  end
+
+  @spec verify_no_duplicate_inputs([Transaction.Recovered.t()]) :: {:ok, map()}
+  defp(verify_no_duplicate_inputs(transactions)) do
+    transactions
+    |> Enum.reduce_while({:ok, %{}}, fn tx, {:ok, input_set} ->
+      inputs =
+        tx
+        |> Map.get(:signed_tx)
+        |> Map.get(:raw_tx)
+        |> Map.get(:inputs)
+
+      Enum.reduce_while(inputs, {:cont, {:ok, input_set}}, fn input, {:cont, {:ok, input_set}} ->
+        input_position = Utxo.Position.encode(input)
+
+        case Map.has_key?(input_set, input_position) do
+          true -> {:halt, {:halt, {:error, :block_duplicate_inputs}}}
+          false -> {:cont, {:cont, {:ok, Map.put(input_set, input_position, true)}}}
+        end
+      end)
     end)
   end
 end
--- a/apps/omg_watcher/test/omg_watcher/block_validator_test.exs
+++ b/apps/omg_watcher/test/omg_watcher/block_validator_test.exs
@@ -25,6 +25,7 @@ defmodule OMG.WatcherRPC.Web.Validator.BlockValidatorTest do
   @bob TestHelper.generate_entity()
   @eth OMG.Eth.zero_address()
   @payment_tx_type OMG.WireFormatTypes.tx_type_for(:tx_payment_v1)
+  @transaction_upper_limit 2 |> :math.pow(16) |> Kernel.trunc()
 
   describe "stateless_validate/1" do
     test "returns error if a transaction is not correctly formed (e.g. duplicate inputs)" do
@@ -119,6 +120,46 @@ defmodule OMG.WatcherRPC.Web.Validator.BlockValidatorTest do
       }
 
       assert {:ok, true} = BlockValidator.stateless_validate(block)
+    end
+
+    test "rejects a block with fewer than two transactions or more transactions than the defined limit" do
+      oversize_block = %{
+        hash: "0x0",
+        number: 1000,
+        transactions: List.duplicate("0x0", @transaction_upper_limit + 1)
+      }
+
+      undersize_block = %{
+        hash: "0x0",
+        number: 1000,
+        transactions: ["0x0"]
+      }
+
+      assert {:error, :transactions_exceed_block_limit} = BlockValidator.stateless_validate(oversize_block)
+      assert {:error, :transactions_below_block_minimum} = BlockValidator.stateless_validate(undersize_block)
+    end
+
+    test "rejects a block that uses the same input in different transactions" do
+      duplicate_input = {1, 0, 0, @alice}
+
+      recovered_tx_1 = TestHelper.create_recovered([duplicate_input], @eth, [{@bob, 10}])
+      recovered_tx_2 = TestHelper.create_recovered([duplicate_input], @eth, [{@bob, 10}])
+
+      signed_txbytes_1 = recovered_tx_1.signed_tx_bytes
+      signed_txbytes_2 = recovered_tx_2.signed_tx_bytes
+
+      merkle_root =
+        [recovered_tx_1, recovered_tx_2]
+        |> Enum.map(&Transaction.raw_txbytes/1)
+        |> Merkle.hash()
+
+      block = %{
+        hash: merkle_root,
+        number: 1000,
+        transactions: [signed_txbytes_1, signed_txbytes_2]
+      }
+
+      assert {:error, :block_duplicate_inputs} == BlockValidator.stateless_validate(block)
     end
   end
 end


apps/omg_watcher/test/omg_watcher/block_validator_test.exs
```diff
@@ -23,9 +23,14 @@ defmodule OMG.WatcherRPC.Web.Validator.BlockValidatorTest do
...
+
   @eth OMG.Eth.zero_address()
+
   @payment_tx_type OMG.WireFormatTypes.tx_type_for(:tx_payment_v1)
 
+  @fee_claimer <<27::160>>
+  @transaction_upper_limit 2 |> :math.pow(16) |> Kernel.trunc()
+
   describe "stateless_validate/1" do
     test "returns error if a transaction is not correctly formed (e.g. duplicate inputs)" do
```
thec00n, line 35:
```suggestion
    test "returns an error if a transaction within a block is not correctly formed (e.g. duplicate inputs)" do
```

apps/omg_watcher/test/omg_watcher/block_validator_test.exs
```diff
@@ -120,5 +118,116 @@ defmodule OMG.WatcherRPC.Web.Validator.BlockValidatorTest do
 
       assert {:ok, true} = BlockValidator.stateless_validate(block)
     end
+
+    test "rejects a block with fewer than two transactions or more transactions than the defined limit" do
```
thec00n, line 122:
```suggestion
    test "rejects a block with fewer than one transaction or more transactions than the defined limit" do
```

apps/omg_watcher/test/omg_watcher/block_validator_test.exs
```diff
@@ -120,5 +118,116 @@ defmodule OMG.WatcherRPC.Web.Validator.BlockValidatorTest do
...
+      block = %{
+        hash: derive_merkle_root(invalid_ordered_transactions),
+        number: 1000,
+        transactions: signed_txbytes
+      }
+
+      assert {:error, :unexpected_transaction_type_at_fee_index} = BlockValidator.stateless_validate(block)
+    end
+
+    test "rejects a block if fee transactions are not at the tail of the transactions' list (two fee currencies)" do
```
thec00n, line 200: Can we also add a test case in case there is two fee tx of the same output currency?

apps/omg_watcher/lib/omg_watcher/block_validator.ex
```diff
@@ -20,15 +20,24 @@ defmodule OMG.Watcher.BlockValidator do
...
+
+  @transaction_upper_limit 2 |> :math.pow(16) |> Kernel.trunc()
 
   @doc """
   Executes stateless validation of a submitted block:
+  - Verifies that the number of transactions falls within the accepted range.
   - Verifies that transactions are correctly formed.
+  - Verifies that there are no duplicate inputs at the block level.
   - Verifies that given Merkle root matches reconstructed Merkle root.
+  - Verifies that fee transactions are correctly placed, calculated and unique per currency.
```
thec00n, line 33:
```suggestion
  - Verifies that fee transactions are correctly placed, well formed and unique per currency.
```

thec00n (changes requested):

apps/omg_watcher/lib/omg_watcher/block_validator.ex
```diff
@@ -64,4 +73,93 @@ defmodule OMG.Watcher.BlockValidator do
       end
     end)
   end
+
+  @spec number_of_transactions_within_limit([Transaction.Signed.tx_bytes()]) ::
+          :ok | {:error, atom()}
+  defp number_of_transactions_within_limit(transactions) do
```
mederic-p, line 79: I personally prefer this to be written in a pattern matching guard style:
```
# A block with a merge transaction only would have no fee transaction.
  defp number_of_transactions_within_limit(tx) when length(tx) == 0, do: {:error, :empty_block}
  defp number_of_transactions_within_limit(tx) when length(tx) > @transaction_upper_limit do 
    {:error, :transactions_exceed_block_limit}
  end
  defp number_of_transactions_within_limit(tx), do: :ok
```

apps/omg_watcher/lib/omg_watcher/block_validator.ex
```diff
@@ -64,4 +73,93 @@ defmodule OMG.Watcher.BlockValidator do
...
+
+      _ ->
+        :ok
+    end
+  end
+
+  @spec verify_no_duplicate_inputs([Transaction.Recovered.t()]) :: {:ok, map()}
+  defp verify_no_duplicate_inputs(transactions) do
+    Enum.reduce_while(transactions, {:ok, %{}}, fn tx, {:ok, input_set} ->
+      tx
```
mederic-p, line 96: You should be able to call `Transaction.get_inputs(tx)`

apps/omg_watcher/lib/omg_watcher/block_validator.ex
```diff
@@ -64,4 +73,93 @@ defmodule OMG.Watcher.BlockValidator do
...
+      |> Map.get(:inputs, [])
+      # Setting an empty array as default because fee transactions will not have an `input` key.
+      |> scan_for_duplicates({:cont, {:ok, input_set}})
+    end)
+  end
+
+  # Nested reducer executing duplicate verification logic for `verify_no_duplicate_inputs/1`
+  @spec scan_for_duplicates(Transaction.Recovered.t(), {:cont, {:ok, map()}}) :: {:cont, {:ok, map()}}
+  defp scan_for_duplicates(tx_input_set, {:cont, {:ok, input_set}}) do
+    Enum.reduce_while(tx_input_set, {:cont, {:ok, input_set}}, fn input, {:cont, {:ok, input_set}} ->
```
mederic-p, line 108: [Enum.uniq_by](https://hexdocs.pm/elixir/Enum.html#uniq_by/2) might be faster, not sure.

apps/omg_watcher/lib/omg_watcher/block_validator.ex
```diff
@@ -64,4 +73,93 @@ defmodule OMG.Watcher.BlockValidator do
...
+        {:error, :transactions_exceed_block_limit}
+
+      _ ->
+        :ok
+    end
+  end
+
+  @spec verify_no_duplicate_inputs([Transaction.Recovered.t()]) :: {:ok, map()}
+  defp verify_no_duplicate_inputs(transactions) do
+    Enum.reduce_while(transactions, {:ok, %{}}, fn tx, {:ok, input_set} ->
```
InoMurko, line 95: this looks like a reduce abuse + scan_for_duplicates. A simple recursive function will be more readable.

InoMurko (changes requested):

apps/omg_watcher/lib/omg_watcher/block_validator.ex
```diff
@@ -48,7 +57,7 @@ defmodule OMG.Watcher.BlockValidator do
     end
   end
 
-  @spec verify_transactions(transactions :: list(Transaction.Recovered.t())) ::
+  @spec verify_transactions(transactions :: list(Transaction.Signed.tx_bytes())) ::
           {:ok, list(Transaction.Recovered.t())}
           | {:error, Transaction.Recovered.recover_tx_error()}
   defp verify_transactions(transactions) do
```
mederic-p, line 63: Didn't check all the logic, but are you sure the txbytes that you receive here are signed (included signatures in their RLP array)?

mederic-p: Ahh, seems that you have them signed yeah: https://github.com/omgnetwork/elixir-omg/blob/master/apps/omg/lib/omg/block.ex#L48
So the hash is calculated using unsigned bytes but you have the signed bytes in the block object as well

apps/omg_watcher/lib/omg_watcher/block_validator.ex
InoMurko: or not, but this doesn't look too clean!

apps/omg_watcher/lib/omg_watcher/block_validator.ex
mederic-p: So the question is, do you need the signed bytes or not? Maybe we can save a few Mb per block using unsigned tx bytes

apps/omg_watcher/lib/omg_watcher/block_validator.ex
```diff
@@ -64,4 +73,93 @@ defmodule OMG.Watcher.BlockValidator do
...
+      _ -> {:error, :duplicate_fee_transaction_for_ccy}
+    end
+  end
+
+  defp is_fee(%Transaction.Recovered{signed_tx: %Transaction.Signed{raw_tx: %Transaction.Fee{}}}),
+    do: true
+
+  defp is_fee(_), do: false
+
+  defp get_fee_output(fee_transaction) do
```
mederic-p, line 154: You should be able to use `fee_transaction |> Transaction.get_outputs() |> Enum.at(0)`

apps/omg_watcher/lib/omg_watcher/block_validator.ex
mederic-p: Ino hates `reduce_while` :trollface:

InoMurko: yeah when I see things like `{:halt, {:halt, {:error, :block_duplicate_inputs}}}` 🗡️ 🗡️ 🗡️  :)))))

mederic-p: Oh I missed the double `:halt`, needs some refactor indeed!

apps/omg_watcher/lib/omg_watcher/block_validator.ex
```diff
@@ -64,4 +73,71 @@ defmodule OMG.Watcher.BlockValidator do
...
+  defp number_of_transactions_within_limit(transactions) when length(transactions) == 0, do: {:error, :empty_block}
+
+  defp number_of_transactions_within_limit(transactions) when length(transactions) > @transaction_upper_limit do
+    {:error, :transactions_exceed_block_limit}
+  end
+
+  defp number_of_transactions_within_limit(_transactions), do: :ok
+
+  @spec verify_no_duplicate_inputs([Transaction.Recovered.t()], map()) :: {:ok, map()}
+  defp verify_no_duplicate_inputs([], input_set), do: {:ok, input_set}
```
mederic-p, line 87: can this be simplified by something like 
```
defp verify_no_duplicate_inputs(transactions) do
          all_inputs = Enum.flat_map(transactions, &Transaction.get_inputs/1)
          uniq_inputs = Enum.uniq_by(all_inputs, &Position.encode/1)
          
          case length(all_inputs) == length(uniq_inputs) do
            true -> :ok
            false -> {:error, :block_duplicate_inputs}
          end    
        end
```
Not sure if it's faster

apps/omg_watcher/test/omg_watcher/block_validator_test.exs
okalouti: hey Gerhard, what difference do you have in mind from the test on line 160?

apps/omg_watcher/lib/omg_watcher/block_validator.ex
okalouti: hey @mederic-p. Thanks. 

The Merkle root will be constructed from the raw transaction bytes, but the transaction bytes in the list will be signed. 

As we would like to validate that transactions are correctly signed, we need the signed bytes. 

(... for now validation is stateless, so checking that signature is correctly formed and a signature exists for each input)

apps/omg_watcher/lib/omg_watcher/block_validator.ex
okalouti: Haha point taken. I was going for performance but it got a bit ugly. Refactored this.

apps/omg_watcher/lib/omg_watcher/block_validator.ex
okalouti: Comparing both versions: 

t = number of transactions
i = number of inputs 

** Proposed version **

map to inputs: O(t)
flatten: O(t)
encode position: O(i)
`uniq_by`: O(i) [best case]

perf: O(2t + 2i) 

** Current version ** 

get_inputs: O(t)
endode: O(i)
check if in accumulator map: O(i)

perf: O(t + 2i)

pasting my version for ease
```  
defp verify_no_duplicate_inputs([], input_set), do: {:ok, input_set}

  defp verify_no_duplicate_inputs([transaction | rest], input_set) do
    current_input_positions = transaction |> Transaction.get_inputs() |> Enum.map(&Position.encode/1)

    current_input_positions
    |> Enum.any?(fn input_position -> Map.has_key?(input_set, input_position) end)
    |> case do
      true ->
        {:error, :block_duplicate_inputs}

      false ->
        new_input_set = current_input_positions |> Map.new(fn pos -> {pos, true} end) |> Map.merge(input_set)
        verify_no_duplicate_inputs(rest, new_input_set)
    end
```

@InoMurko would appreciate your insights on this one! @mederic-p 's definitely looks better though.

InoMurko: pick the one that's easier to read and understand. we can make it fast if we need to.

okalouti:
Updated the documentation to reflect new checks.

thec00n (approved):

PR #1696, opened 2020-08-20
Verity: New deployment-time configuration properties

pnowosie:

Closes https://github.com/omgnetwork/private-issues/issues/76

## Overview

See mentioned issue 👆 

## Changes

Allows following configuration properties to be set at deployment time
- SUBMIT_WHEN_N_TXS - submit block if there is `C`=10k txs already (or wait n-Eth block otherwise),
- SUBMIT_EVERY_NTH - submit a block every nth ethereum blocks,

## Testing

Added release_tasks tests.


pnowosie:
Moved to https://github.com/omgnetwork/elixir-omg/pull/1706

PR #1701, opened 2020-08-25
feat: handle in-flight exits deletions

pgebal:
Closes https://github.com/omgnetwork/private-issues/issues/35

Follows solution proposed by @pnowosie:
spends inputs of in-flight transaction (child-chain does the same when ife starts) and deletes the in-flight exit from ExitProcessor.


apps/omg_watcher/lib/omg_watcher/exit_processor/core.ex
```diff
@@ -641,4 +641,32 @@ defmodule OMG.Watcher.ExitProcessor.Core do
 
   defp sla_margin_safe?(exit_processor_sla_margin, min_exit_period_seconds, ethereum_block_time_seconds),
     do: exit_processor_sla_margin * ethereum_block_time_seconds < min_exit_period_seconds
+
+  @doc """
+  Deletes in-flight exits from state and returns deleted exits
+  """
+  @spec delete_in_flight_exits(__MODULE__.t(), list(map)) :: {__MODULE__.t(), list(InFlightExitInfo.t()), list(any())}
+  def delete_in_flight_exits(%__MODULE__{in_flight_exits: ifes} = state, deletions) do
```
InoMurko, line 649:
```suggestion
  def delete_in_flight_exits(state, deletions) do
```
and use `state.ifes`

apps/omg_watcher/lib/omg_watcher/exit_processor/in_flight_exit_info.ex
```diff
@@ -417,6 +427,12 @@ defmodule OMG.Watcher.ExitProcessor.InFlightExitInfo do
       }),
       do: is_older?(seen_in_pos, oldest_competitor_pos)
 
+  @doc """
+  Converts integer to in-flight exit contract id
+  """
+  @spec to_contract_id(non_neg_integer) :: <<_::192>>
+  def to_contract_id(id), do: <<id::192>>
```
InoMurko, line 434: you're mixing private and public functions

InoMurko (approved):

apps/omg_watcher/test/omg_watcher/exit_processor/core/state_interaction_test.exs
```diff
@@ -289,6 +289,23 @@ defmodule OMG.Watcher.ExitProcessor.Core.StateInteractionTest do
...
+
+    state = TestHelper.do_deposit(state, alice, %{amount: 10, currency: @eth, blknum: 1})
+    {:ok, _, state} = State.Core.form_block(state)
+
+    {_processor, deleted_utxos, _db_updates} =
+      processor
+      |> start_ife_from(ife_exit_tx, exit_id: ife_id)
+      |> Core.delete_in_flight_exits([%{exit_id: ife_id}])
+
+    assert {:ok, {[{:delete, :utxo, _}], {[{:utxo_position, 1, 0, 0}], []}}, _} =
```
boolafish, line 305: slightly curious, what will be in the `_` part and why we don't match it to detail?

boolafish (approved):
LGTM

apps/omg_watcher/test/omg_watcher/exit_processor/core/state_interaction_test.exs
pgebal: @boolafish  Key for utxo to delete from the db. We don't match it, because it's not a unit test that tests `State.exit_utxos` (it's tested in detail in other file). We just wan to know if what `delete_in_flight_exit` returns fits into `State.exit_utxos` and do just basic sanity check, hence matching the returned utxo.

PR #1704, opened 2020-08-26
Trigger block submit when enough pending txs

pnowosie:

Fixes https://github.com/omgnetwork/private-issues/issues/77

## Overview

Allows to shorten the time for next block submission (which is not based on mined Ethereum blocks count) when there is enough pending transaction collected already.
Pending transaction count which triggers blocks submission can be configured by `BLOCK_SUBMIT_WHEN_N_TXS` environment variable.
When not set it defaults to _infinity_ which effectively switch off the feature.

## Changes

Added 2 more Childchain's configuration properties and pass them down to the `BlockQueue` module.
- submit_block_when_n_txs
- submit_block_when_n_txs

When `submit_block_when_n_txs` is set to a number it will set a timer checking pending transactions count in the interval of `submit_block_when_n_txs` milliseconds. Timer will trigger block forming and submission to the rootchain according to the transactions count in the block and `submit_block_when_n_txs` property.

- ...

## Testing


pnowosie:
Moved to https://github.com/omgnetwork/elixir-omg/pull/1707

PR #1706, opened 2020-08-26
Verity: exposing more configs via envvar

pnowosie:

Closes https://github.com/omgnetwork/private-issues/issues/76

## Overview

See mentioned issue 👆 

## Changes

Allows following configuration properties to be set at deployment time
- SUBMIT_WHEN_N_TXS - submit block if there is `C`=10k txs already (or wait n-Eth block otherwise),
- SUBMIT_EVERY_NTH - submit a block every nth ethereum blocks,

## Testing

Added release_tasks tests.


docs/details.md
```diff
@@ -203,6 +203,8 @@ Affects how quick the services reading Ethereum events realize there's a new blo
 
 * **`block_submit_every_nth`** - how many new Ethereum blocks must be mined, since previous submission **attempt**, before another block is going to be formed and submitted.
 
+* **`block_submit_when_n_txs`** - pending transaction count that will trigger block submission faster than in required Ethereum's blocks (see: `block_submi_every_nth`). When not set defaults to infinity, effecively switching this feature off.
```
pgebal, line 206: 1. typo, `block_submi_every_nth`
2. how about we name it `block_submit_after_n_txs`?
3. I think it could be explained more clearly: `maximum number of transactions in a block. If block reaches this number of transactions it is submitted to plasma contracts` or something similar.

pgebal (approved):

PR #1707, opened 2020-08-26
Verity: trigger block submit when enough pending txs

pnowosie:

Fixes https://github.com/omgnetwork/private-issues/issues/77

## Overview

Allows to shorten the time for next block submission (which is not based on mined Ethereum blocks count) when there is enough pending transaction collected already.
Pending transaction count which triggers blocks submission can be configured by `BLOCK_SUBMIT_WHEN_N_TXS` environment variable.
When not set it defaults to _infinity_ which effectively switch off the feature.

## Changes

Added 2 more Childchain's configuration properties and pass them down to the `BlockQueue` module.
- submit_block_when_n_txs
- block_fulfillment_check_interval_ms

When `submit_block_when_n_txs` is set to a number it will set a timer checking pending transactions count in the interval of `submit_block_when_n_txs` milliseconds. Timer will trigger block forming and submission to the rootchain according to the transactions count in the block and `submit_block_when_n_txs` property.

- ...

## Testing

### DIFF
--- a/apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
+++ b/apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
@@ -104,7 +104,7 @@ defmodule OMG.ChildChain.BlockQueue do
     _ = Logger.info("Starting BlockQueue, top_mined_hash: #{inspect(Encoding.to_hex(top_mined_hash))}")
 
     block_submit_every_nth = Keyword.fetch!(args, :block_submit_every_nth)
-    block_submit_when_n_txs = Keyword.fetch!(args, :block_submit_when_n_txs)
+    block_submit_after_n_txs = Keyword.fetch!(args, :block_submit_after_n_txs)
     block_submit_max_gas_price = Keyword.fetch!(args, :block_submit_max_gas_price)
     gas_price_adj_params = %GasPriceAdjustment{max_gas_price: block_submit_max_gas_price}
 
@@ -116,7 +116,7 @@ defmodule OMG.ChildChain.BlockQueue do
         parent_height: parent_height,
         child_block_interval: child_block_interval,
         block_submit_every_nth: block_submit_every_nth,
-        block_submit_when_n_txs: block_submit_when_n_txs,
+        block_submit_after_n_txs: block_submit_after_n_txs,
         finality_threshold: finality_threshold,
         gas_price_adj_params: gas_price_adj_params
       )
@@ -142,7 +142,7 @@ defmodule OMG.ChildChain.BlockQueue do
     {:ok, _} = :timer.send_interval(interval, self(), :check_ethereum_status)
 
     _ =
-      if is_number(block_submit_when_n_txs) do
+      if is_number(block_submit_after_n_txs) do
         block_fulfillment_check_interval_ms = Keyword.fetch!(args, :block_fulfillment_check_interval_ms)
         {:ok, _} = :timer.send_interval(block_fulfillment_check_interval_ms, self(), :check_block_fulfillment)
       end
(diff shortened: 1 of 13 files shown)


kevsul:
> Allows to shorten the time for next block submission (which is not based on mined Ethereum blocks count) when there is enough pending transaction collected already.

Is this really what's required? My understanding was that we want to _lengthen_ the time between blocks to that they're submitted _less_ frequently. 

Or can this be achieved this by setting e.g.
```
block_submit_every_nth: 10,
block_submit_when_n_txs: 10_000
```

and then a block will be submitted either after 10 rootchain blocks or 10,000 childchain txs, whichever comes first?

kevsul:
```
Added 2 more Childchain's configuration properties and pass them down to the BlockQueue module.
- submit_block_when_n_txs
- submit_block_when_n_txs
```

Typo in the description - I think you meant `block_fulfillment_check_interval_ms`?

pgebal (commented):
> > Allows to shorten the time for next block submission (which is not based on mined Ethereum blocks count) when there is enough pending transaction collected already.
> 
> Is this really what's required? My understanding was that we want to _lengthen_ the time between blocks to that they're submitted _less_ frequently.

Effect of this PR is that blocks will be submitted more frequently.

pgebal:
What we need that PR for? Just to know if it works? Was it checked that it really works?

pnowosie:
> then a block will be submitted either after 10 rootchain blocks or 10,000 childchain txs, whichever comes first?

This is correct. 10k txs block submission comes out of lack of visibility of pending transactions.

apps/omg_child_chain/lib/omg_child_chain/block_queue.ex
```diff
@@ -163,12 +171,34 @@ defmodule OMG.ChildChain.BlockQueue do
...
 
     mined_blknum = RootChain.get_mined_child_block()
-    {_, is_empty_block} = OMG.State.get_status()
+    {_, _, txs_count} = OMG.State.get_status()
 
     _ = Logger.debug("Ethereum at \#'#{inspect(ethereum_height)}', mined child at \#'#{inspect(mined_blknum)}'")
 
     state1 =
-      case Core.set_ethereum_status(state, ethereum_height, mined_blknum, is_empty_block) do
+      case Core.set_ethereum_status(state, ethereum_height, mined_blknum, txs_count) do
```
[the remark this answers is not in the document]
pnowosie: I agree. Having in mind this not going to master and on-going effort to depreciate the `grandpa` :joy:  I'll leave it as is.

apps/omg_child_chain/test/omg_child_chain/block_queue/core_test.exs
```diff
@@ -237,130 +237,186 @@ defmodule OMG.ChildChain.BlockQueue.CoreTest do
...
                |> elem(1)
                |> Core.enqueue_block("1", 1000, 0)
                |> Core.enqueue_block("2", 2000, 1)
-               |> Core.set_ethereum_status(1, 2000, false)
+               |> Core.set_ethereum_status(1, 2000, 1)
+    end
+  end
+
+  describe "check_block_fulfilled/2" do
+    test "does not form block by default", %{empty: empty} do
```
pgebal, line 368: what's the purpose of this test? Shouldn't we reject txs if we get to `max_block_size`?

pnowosie: By default decision to form a block is to align with Ethereum blocks (`block_submit_every_nth`) and block size isn't considered. I don't change default behavior, so here we test old expected behavior. 

And yes - new transaction above max block size are rejected - but its different flow.

pgebal (approved):

PR #1724, opened 2020-09-03
fix: set :last_ife_exit_deleted_eth_height

pgebal:
That's an interim solution to fix dev env.
I think better solution is proposed here: https://github.com/omgnetwork/elixir-omg/pull/1720, but I can't make it run yet.


pgebal:
That's not an elegant solution.

PR #1726, opened 2020-09-04
Allow to run docker-compose without feefeed docker

boolafish:
:clipboard: fixes: https://github.com/omgnetwork/elixir-omg/issues/1723

requires: https://github.com/omgnetwork/specs/pull/10

## Overview

Provide a new docker-compose file: `docker-compose.feefeed.yaml` that could override the docker-compose with using feefeed instead of file. If not combining the docker-compose files, it will run `docker-compose` without the need of feefeed.

## Changes

- add `docker-compose.feefeed.yml` file and port out the feefeed from `docker-compose.yml`
- in `make` targets, use the composed feefeed service
- [x] TODO: switch cabbage back to master once: https://github.com/omgnetwork/specs/pull/10 merged

## Testing

CI


priv/perf/Makefile
```diff
@@ -2,14 +2,20 @@
 list:
 	@$(MAKE) -pRrq -f $(lastword $(MAKEFILE_LIST)) : 2>/dev/null | awk -v RS= -F: '/^# File/,/^# Finished Make data base/ {if ($$1 !~ "^[#.]") {print $$1}}' | sort | egrep -v -e '^[^[:alnum:]]' -e '^$@$$'
 
+clean:
+	docker-compose -f ../../docker-compose.yml -f ../../docker-compose.feefeed.yml down && docker volume prune --force
```
unnawut, line 6: Wdyt if we move the config to a constant so we can define the intention, i.e. starting full set of services? E.g.

```shell
COMPOSE_FULL_SERVICES=-f ../../docker-compose.yml -f ../../docker-compose.feefeed.yml

docker-compose $(COMPOSE_FULL_SERVICES) down && docker volume prune --force
```

unnawut (approved):
LGTM while we still figuring out the service boundaries

achiurizo (approved):

PR #1732, opened 2020-09-09
Kevsul/standard exit perf test

### LINKED ISSUE #1600: Clean up perf to be able to run against env

kevsul:
Fixes #1600 

## Overview

Port the byzantine tests to the Chaperon framework

## Changes

- Added new Chaperon scenarios to test valid Standard Exits
- Removed the code that only worked with local geth. Apologies again @ayrat555 for not being clear about this in the original issue.

Note: This PR does not include tests for invalid standard exits and challenges. That can be done as a separate task.


priv/perf/apps/load_test/lib/scenario/start_standard_exit.ex
```diff
@@ -0,0 +1,41 @@
...
+  - `exiter` the account that's starting the exit
+  - `utxo` the utxo to exit
+  """
+
+  use Chaperon.Scenario
+
+  alias Chaperon.Session
+  alias LoadTest.ChildChain.Exit
+
+  def run(session) do
```
boolafish, line 29: I am slightly concerned that we are over moduling on the scenario level where we wants really are some helper client functions to wrap a bit of complexity to make the call. Relying on session to pass around vars are not that great in my opinion as it is harder to follow at certain point, what var is in the session when we need to cross read multiple Scenario files in a single runner.

I would bias over to use sth like in the place that runs this scenario
```elixir
tx_hash = root_chain.start_standard_exit()
session = Session.assign(session, tx_hash: tx_hash)
```
if necessary. I think most of the cases we don't even need the part of assign it to session.

Same for the watcher info scenario

priv/perf/apps/load_test/lib/runner/standard_exits.ex
```diff
@@ -0,0 +1,60 @@
...
+    test_currency = Application.fetch_env!(:load_test, :test_currency)
+    gas_price = Application.fetch_env!(:load_test, :gas_price)
+    config = default_config()
+
+    # Use the faucet account to add the token's exit queue if necessary
+    {:ok, faucet} = Faucet.get_faucet()
+    _ = Exit.add_exit_queue(1, Encoding.to_binary(test_currency), faucet, gas_price)
+
+    [
+      {{config.concurrent_sessions, [LoadTest.Scenario.ManyStandardExits, LoadTest.Scenario.WatcherStatus]},
```
boolafish, line 53: just curious, what is the purpose of watcher status call at the end?

boolafish (commented):
🚀🚀🚀🚀

priv/perf/config/test.exs
```diff
@@ -9,7 +9,7 @@ config :load_test,
   watcher_info_url: "http://localhost:7534",
   faucet_deposit_amount: trunc(:math.pow(10, 18) * 10),
   # fee testing setup: https://github.com/omgnetwork/fee-rules-public/blob/master/fee_rules.json
-  fee_amount: 75,
+  fee_amount: 1,
```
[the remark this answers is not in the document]
kevsul: Oops, yeah I couldn't get feefeed working locally, so I backed out 😂

priv/perf/apps/load_test/lib/runner/standard_exits.ex
kevsul: That was in the original test - start many standard exits and measure the response of watcher status.get

priv/perf/apps/load_test/lib/scenario/start_standard_exit.ex
kevsul: I made it a scenario for 2 reasons:

1. Chaperon automatically adds metrics to scenarios, so you don't have to add them explicitly (as in your comment above)
2. Chaperon allows you compose scenarios, which can make it easier to create complex scenarios. e.g. to measure watcher status.get you just add `LoadTest.Scenario.WatcherStatus` to the list (see comment below).

But TBH I'm kind of in two minds about this. 
As you pointed out, the 'composability' is a bit weird - I think it can reduce code reuse in simple cases like `WatcherStatus`, but for more complex scenarios storing everything in the session map can be a bit confusing. 

The automatic metrics are nice to have, but can be added explicitly where necessary.

I'm fine with either way really...

priv/perf/apps/load_test/lib/scenario/start_standard_exit.ex
```diff
@@ -0,0 +1,41 @@
...
+    exiter = config(session, [:exiter])
+    utxo = config(session, [:utxo])
+    gas_price = config(session, [:gas_price])
+
+    tx_hash =
```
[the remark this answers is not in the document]
kevsul: metrics are added automatically because it's a scenario, see comment below

priv/perf/apps/load_test/lib/runner/standard_exits.ex
boolafish: I guess if we cannot understand that or justify that ..... we can probably throw it away?

priv/perf/apps/load_test/lib/scenario/start_standard_exit.ex
boolafish: hmmm I am okay to continue like this. Will leave this to your decision.

Though I do believe client code would be more reusable and useful compare to this. But yeah, free metric is a selling point. 

Composability wise I am on the fence, lower level compatibility seems a bit more useful, at least I am not really buyed with the compatibility in this PR. But the previous example of composing scenario for create UTXO and spend UTXO was great. Probably to me it is more about the complexity your wrapping on 🤔 But again I am not strongly against this yet it is more of my intuition so far.

priv/perf/apps/load_test/lib/runner/standard_exits.ex
kevsul: I don't think it's a bad idea to have it. 
Starting exits happens mainly on the root chain (apart from getting the exit data), so we want to have some indication of how the watcher / childchain is behaving as a result.

Can you think of a better way of measuring that other than status.get?

boolafish: You mean....testing watcher is still alive?

kevsul: Well, testing that it's still alive is a start and we can probably do that manually when testing the limits of the system.

But long term, I mean testing how long it takes to respond to status.get and tracking if any code changes affect that.
Otherwise we're just executing root chain transactions with no visibility on how that affects our services.

boolafish: sure. I was more like just wondering what is that single call to watcher status is doing. If we know currently it help sanity checks watcher is still alive after the load, I am okay to keep it as long as we know what that is doing now.

priv/perf/apps/load_test/lib/scenario/many_standard_exits.ex
```diff
@@ -0,0 +1,79 @@
...
+  def run(session) do
+    exits_per_session = config(session, [:exits_per_session])
+    gas_price = config(session, [:gas_price])
+
+    # Create a new exiter account
+    {:ok, exiter} = Account.new()
+    amount = (@gas_start_exit * gas_price + @standard_exit_bond) * exits_per_session
+
+    # Fund the exiter with some root chain eth
+    session = run_scenario(session, LoadTest.Scenario.FundAccountRootchain, %{account: exiter.addr, amount: amount})
```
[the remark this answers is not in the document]
kevsul: Yeah, I'll remove this scenario - it's not something we're going to want metrics for

boolafish (approved):

priv/perf/apps/load_test/lib/child_chain/exit.ex
```diff
@@ -6,80 +20,135 @@ defmodule LoadTest.ChildChain.Exit do
...
+  def wait_for_exit_data(utxo_pos, counter) do
+    data = get_exit_data(utxo_pos)
+
+    if data.proof do
+      data
+    else
+      _ = Logger.info("Waiting for exit data")
+      Process.sleep(@poll_interval)
+      wait_for_exit_data(utxo_pos, counter - 1)
+    end
```
InoMurko, line 78: `case` statements are preferred to avoid ambiguity, use if/else for trailing question mark function only

InoMurko (approved):
nitpick over if/else usage

PR #1737, opened 2020-09-14
Inomurko/remove child chain

InoMurko:
This PR removes all childchain code and makes ready to OS grandfather's repo.


.circleci/config.yml
```diff
@@ -348,166 +351,97 @@ jobs:
...
-    machine:
-      image: ubuntu-1604:201903-01
-    environment:
-      MIX_ENV: test
-    steps:
-      - checkout
       - setup_childchain
       - run:
-          name: Integration Tests & Coveralls Part Watcher
+          name: invalid_exit_2_test.exs
```
boolafish, line 381: lol nice to move those back in same workflow/job 😅

priv/perf/scripts/generate_api_client.sh
```diff
@@ -23,10 +23,9 @@ echo "Generating childchain clients..."
 echo "------------------------------------------------------"
 docker run --rm \
     -v ${PWD}/apps:/apps \
-    -v ${PWD}/../../apps/omg_child_chain_rpc/priv/swagger/:/swagger \
     --user $(id -u):$(id -g) \
     openapitools/openapi-generator-cli generate \
-    -i /swagger/operator_api_specs.yaml \
+    -i https://raw.githubusercontent.com/omgnetwork/grandfather/master/apps/omg_child_chain_rpc/priv/swagger/operator_api_specs.yaml?token=AATWQG6U2GJE3VHVMVDEDWK7NCGYE \
```
boolafish, line 28: I guess grandfather got rebranded 😛.
Also.....the token if that is sth should be confidential you might want to delete that token. (I guess it is GitHub API token?)

achiurizo, line 28: is this token okay?

boolafish (commented):
hi, though I am not the next release manager, but what is the plan for next release (1.0.5?). Assuming we are going to merge this soon:

1. Is `1.0.5` to be without childchain already? (so we only release watcher, watcher-info with `1.0.5`)
1. If so, have we checked that there is no changes on childchain after 1.0.4? Or everything is ported to `omg-childchain-v1` already?

If we are not releasing chldchain even for next release 1.0.5, I can remove the CI steps for release that triggers update for childchain on the release tag flow. 🙏

(Update, or probably you can just update in this PR? So they got removed together)

achiurizo (approved):
🍰

InoMurko:
Childchain repo: https://github.com/omgnetwork/omg-childchain-v1

priv/perf/scripts/generate_api_client.sh
InoMurko: it's just their tracking token, anyhow, its open now so I;ll switch to omg-childchain-v1!

InoMurko: yes, this pr happened before the `omg-childchain-v1 `

ayrat555 (approved):
Looks good.

But I saw you made some changes to the `specs` submodule. I think you should push your changes to the `specs` repo first and only then update it here

PR #1743, opened 2020-09-18
Chore: use exexec from upstream

ayrat555:
fixes https://github.com/omgnetwork/elixir-omg/issues/1616


apps/omg_eth/test/support/dev_geth.ex
```diff
@@ -36,14 +36,13 @@ defmodule OMG.Eth.DevGeth do
     geth = ~s(geth --miner.gastarget 7500000 \
             --nodiscover \
             --maxpeers 0 \
-            --miner.gasprice \"10\" \
-            --datadir /data/ \
```
ayrat555, line 40: the second datadir is on line 50

apps/omg_eth/test/support/dev_geth.ex
```diff
@@ -36,27 +38,75 @@ defmodule OMG.Eth.DevGeth do
...
+    GenServer.call(pid, :ready?)
+  end
+
+  def stop_geth(pid) do
+    GenServer.cast(pid, :stop)
+  end
+
+  def start_link(cmd) do
+    GenServer.start_link(__MODULE__, cmd)
+  end
```
InoMurko, line 71: these could be private?

ayrat555: @InoMurko I made them private and fixed tests

ayrat555:
@InoMurko @pgebal does it look good?

apps/omg_eth/test/support/dev_geth.ex
```diff
@@ -36,54 +38,88 @@ defmodule OMG.Eth.DevGeth do
...
-      end
+    pid
+  end
+
+  defp wait_for_rpc(pid) do
+    if ready?(pid) do
+      :ok
+    else
+      Process.sleep(1_000)
+      ready?(pid)
```
pgebal, line 114: shouldn't it be `wait_for_rpc`?

pgebal (changes requested):

apps/omg_eth/test/support/dev_geth.ex
```diff
@@ -36,54 +38,88 @@ defmodule OMG.Eth.DevGeth do
...
+
+  # PRIVATE
+
   defp launch(cmd) do
-    _ = Logger.debug("Starting geth")
+    {:ok, pid} = start_link(cmd)
 
-    {:ok, geth_proc, _ref, [{:stream, geth_out, _stream_server}]} =
-      Exexec.run(cmd, stdout: :stream, kill_command: "pkill -9 geth")
+    waiting_task_function = fn ->
```
pgebal, line 98: I think `function` part in var name in unnecessary. We can take a look on the right side of the assignment and we know it's a function.

pgebal (changes requested):

apps/omg_eth/test/support/dev_geth.ex
ayrat555: you're right

apps/omg_eth/test/support/dev_geth.ex
```diff
@@ -36,54 +38,88 @@ defmodule OMG.Eth.DevGeth do
...
+  @impl true
+  def handle_info({:stdout, pid, stdout}, %{os_proc: pid} = state) do
+    new_state =
+      if String.contains?(stdout, "IPC endpoint opened") do
+        Map.put(state, :ready?, true)
+      else
+        state
+      end
+
+    if Application.get_env(:omg_eth, :node_logging_in_debug) do
```
InoMurko, line 81: Not your doing, but this should be a case.

InoMurko (approved):
I guess same goes into omg childchain v1!

pgebal (approved):

PR #1744, opened 2020-09-21
Chore: parallelize tests by tags

ayrat555:
resolves https://github.com/omgnetwork/elixir-omg/issues/1740

## Overview

This PR parallelize feature tags using module tags with circle ci parallelism (https://circleci.com/docs/2.0/parallelism-faster-jobs/)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -629,7 +629,7 @@ jobs:
             make install
             make generate_api_code
             mix deps.get
-            echo &("configuration deposit fee in_flight_exit invalid_standard_exits service_name standard_exits transactions watcher_info_api watcher_status" | circleci tests split)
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -628,7 +629,9 @@ jobs:
             make install
             make generate_api_code
             mix deps.get
-            mix test
+            echo $(circleci tests split ./tags.txt)
```
InoMurko, line 632: I think that with a little bit of `grep` , `awk` and `tr` you could go without this file?

grep -r "@moduletag" | awk '{print $3}' | tr -d ':'

InoMurko (approved):

mederic-p (approved):

PR #1745, opened 2020-09-23
deposits performance tests bot

ayrat555:
This PR implements `deposits` test that can be run with `mix run`:

```
 mix run -e "LoadTest.TestRunner.run()" -- "deposits" "10" "50"
```
Parameters are:
- "deposits" - test name
- "10" - rate in tests per second
- "50" - period in seconds


priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -108,6 +125,242 @@ defmodule LoadTest.Ethereum do
...
+  defp process_transaction_result(
+         %{"code" => "create:client_error", "messages" => %{"code" => "operation:service_unavailable"}} = result,
+         _amount_in_wei,
+         _input_address,
+         _output_address,
+         _currency,
+         0
+       ) do
+    result
+  end
```
InoMurko, line 253: this looks horrible, sorry :)

ayrat555:
@InoMurko I didn't implement api and environment selection for now. I think this should pass review first. For now, I implemented test runs which periodically are being dumped to the postgres. The only supported test is deposits test.

boolafish:
Hi @ayrat555 can you describe how do you expect user to interact with the code change here? If we still plan to trigger the test via Kubernate job, which is more or less like trigger a command with docker. In this case, if we later package this to docker, we will end up talk to a command that trigger the `Front.Runner` and pass all the required params, right? 

But how do we use the UI dashboard and Postgres here?

priv/perf/apps/load_test/lib/child_chain/deposit.ex
```diff
@@ -32,21 +32,30 @@ defmodule LoadTest.ChildChain.Deposit do
...
     {:ok, deposit} = Deposit.new(deposit_utxo)
-    {:ok, {deposit_blknum, eth_blknum}} = send_deposit(deposit, depositor, amount, currency, gas_price)
+    {:ok, {deposit_blknum, eth_blknum, eth_blkhash}} = send_deposit(deposit, depositor, amount, currency, gas_price)
+
     :ok = wait_deposit_finality(eth_blknum, deposit_finality_margin)
-    Utxo.new(%{blknum: deposit_blknum, txindex: 0, oindex: 0, amount: amount})
+
+    case return do
+      :utxo -> Utxo.new(%{blknum: deposit_blknum, txindex: 0, oindex: 0, amount: amount})
+      _ -> eth_blkhash
```
ayrat555, line 57: now it can also return block hash so we can use it to fetch used gas

priv/perf/apps/load_test/lib/test_runner.ex
```diff
@@ -0,0 +1,49 @@
+defmodule LoadTest.TestRunner do
+  alias ExPlasma.Encoding
+
+  @tests %{
+    "deposits" => {LoadTest.Runner.Deposits, LoadTest.Scenario.Deposits}
+  }
+
+  @configs %{
```
ayrat555, line 8: currently, params are hardcoded. should I make them configurable?

priv/perf/apps/load_test/test/load_tests/runner/deposits_test.exs
```diff
@@ -0,0 +1,44 @@
...
+# limitations under the License.
+
+defmodule LoadTest.Runner.DepositsTest do
+  use ExUnit.Case
+
+  @moduletag :deposits
+
+  alias ExPlasma.Encoding
+
+  test "deposits test" do
```
ayrat555, line 22: this is just a happy path test for deposits test the can be run using `mix run` with LoadTest.TestRunner.run

priv/perf/apps/load_test/lib/test_runner.ex
```diff
@@ -0,0 +1,49 @@
...
+    "deposits" => %{
+      token: Encoding.to_binary("0x0000000000000000000000000000000000000000"),
+      initial_amount: 500_000_000_000_000_000,
+      deposited_amount: 200_000_000_000_000_000,
+      transferred_amount: 100_000_000_000_000_000
+    }
+  }
+
+  def run() do
+    [test, rate, period] = System.argv()
```
ayrat555, line 18: @boolafish Can you point out what params should have to be configured aside from rate and period? (watcher_url, ethereumex_ url etc)

priv/perf/apps/load_test/lib/scenario/deposits.ex
```diff
@@ -0,0 +1,129 @@
...
+    txhash =
+      Deposit.deposit_from(from_address, deposited_amount, token,
+        return: :txhash,
+        deposit_finality_margin: 10,
+        gas_price: 0
+      )
+
+    gas_used = Ethereum.get_gas_used(txhash)
+
+    with :ok <-
```
ayrat555, line 60: 1. check childchain balance after deposit for the depositor
2. check rootchain balance after the deposit for the depositor
3. send value to another account
4. check childchain value for the receiver

priv/perf/apps/load_test/lib/test_runner.ex
InoMurko: @ayrat555 it's entirely up to your discretion, this *entire story* is part of work that **you're owning** and are **responsible**. if you make a change - it needs to make sense based on the requirements **you** provided and we **all did at the beginning**. 

I don't know what this `configs` does, so you'll need to clarify if you're unsure.

ayrat555: these values are used in test:
token - token address used for deposits
initial_amount - the amount that the depositor gets from the faucet
deposited_amount - the amount that the depositor deposits to the childchain
transferred_amount - the amount that the depositor send to another account on the childchain

priv/perf/apps/load_test/lib/test_runner.ex
ayrat555: ok. it's not an  issue  since it supports the same env vars that old perf tests support

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -108,6 +125,229 @@ defmodule LoadTest.Ethereum do
...
+
+  defp process_transaction_result(
+         result,
+         amount_in_wei,
+         input_address,
+         output_address,
+         currency,
+         tries
+       ) do
```
InoMurko, line 231: what happened here?

InoMurko:
```suggestion
  defp process_transaction_result(result, amount_in_wei, input_address, output_address, currency, tries) do
```

ayrat555: I addressed your comment

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -49,6 +49,23 @@ defmodule LoadTest.Ethereum do
...
+      {Ethereumex.HttpClient.eth_get_transaction_receipt(receipt_hash),
+       Ethereumex.HttpClient.eth_get_transaction_by_hash(receipt_hash)}
+
+    case result do
+      {{:ok, %{"gasUsed" => gas_used}}, {:ok, %{"gasPrice" => gas_price}}} ->
+        {gas_price_value, ""} = gas_price |> String.replace_prefix("0x", "") |> Integer.parse(16)
+        {gas_used_value, ""} = gas_used |> String.replace_prefix("0x", "") |> Integer.parse(16)
+        gas_price_value * gas_used_value
+
+      {{:ok, nil}, {:ok, nil}} ->
```
pgebal, line 64: Doesn't it break something when we return that transaction used 0 gas?
In what cases receipt and hash are nil?

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -108,6 +125,222 @@ defmodule LoadTest.Ethereum do
...
+    end
+  end
+
+  defp execute_submit_typed(typed_data_signed) do
+    WatcherInfoAPI.Api.Transaction.submit_typed(LoadTest.Connection.WatcherInfo.client(), typed_data_signed)
+  end
+
+  defp process_transaction_result(result, amount_in_wei, input_address, output_address, currency, tries) do
+    case {result, tries} do
+      {%{"code" => "create:client_error", "messages" => %{"code" => "operation:service_unavailable"}}, 0} ->
```
pgebal, line 226: Is it possible to get a different error response?
If it is, then we will loop here.
If `tries` is 0, we should always stop retrying.

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -108,6 +125,222 @@ defmodule LoadTest.Ethereum do
     Sync.repeat_until_success(f, timeout)
   end
 
+  def fetch_balance(address, amount, currency \\ <<0::160>>) do
```
pgebal, line 128: Please add specs to all new public functions.

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -108,6 +125,222 @@ defmodule LoadTest.Ethereum do
...
+       }, _} ->
+        {:ok, [sign_hash, typed_data, txbytes]}
+
+      _ ->
+        Process.sleep(1_000)
+        create_transaction(amount_in_wei, input_address, output_address, currency, tries - 1)
+    end
+  end
+
+  defp root_chain_get_eth_balance(address, 0) do
```
pgebal, line 247: Just return error here. Calls are done in the second definition. We can get the same by just increasing counter value by 1 when calling `root_chain_get_eth_balance`.

priv/perf/apps/load_test/lib/service/faucet.ex
```diff
@@ -200,7 +200,8 @@ defmodule LoadTest.Service.Faucet do
   defp deposit(faucet_account, amount, currency, deposit_finality_margin, gas_price) do
     Logger.debug("Not enough funds in the faucet, depositing #{amount} from the root chain")
 
-    {:ok, utxo} = Deposit.deposit_from(faucet_account, amount, currency, deposit_finality_margin, gas_price)
+    options = [gas_price: gas_price, deposit_finality_margin: deposit_finality_margin]
+    {:ok, utxo} = Deposit.deposit_from(faucet_account, amount, currency, options)
```
pgebal, line 204: That's the only call to `deposit_from` we have and we do not provide `return` option. Let's get rid off `:return` in `deposit_from` definition.

priv/perf/apps/load_test/test/load_tests/runner/deposits_test.exs
```diff
@@ -0,0 +1,44 @@
...
+        transferred_amount: transferred_amount
+      },
+      run_config: %{
+        tps: 1,
+        period_in_seconds: 20
+      },
+      timeout: :infinity
+    }
+
+    Chaperon.run_load_test(LoadTest.Runner.Deposits, print_results: true, config: config)
```
pgebal, line 42: Do we want to print results here? Isn't just checking if run did not fail enough? I have no strong opinion on that, just thinking.

pgebal (changes requested):

priv/perf/apps/load_test/lib/service/faucet.ex
ayrat555: > That's the only call to deposit_from we have and we do not provide return option

Because by default utxo is returned.

> Let's get rid off :return in deposit_from definition.

For some cases we need tx hash.

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
ayrat555: For now, I removed `case`, this function is copied from cabbage tests

priv/perf/apps/load_test/test/load_tests/runner/deposits_test.exs
ayrat555: The rest of the performance tests print results, I think to be consistent this tests should also print results

ayrat555:
@pgebal @InoMurko I addressed your comments. Can you please take another look?

priv/perf/apps/load_test/lib/service/faucet.ex
pgebal: Ok. I couldn't find those cases in the code.

priv/perf/apps/load_test/test/load_tests/runner/deposits_test.exs
boolafish, line 22: Probably we can do the same for your test runner, but instead of exit we assert the metric result here.

priv/perf/config/config.exs
```diff
@@ -9,15 +9,17 @@ ethereum_client_timeout_ms = 20_000
 
 config :ethereumex,
   http_options: [recv_timeout: ethereum_client_timeout_ms],
-  url: System.get_env("ETHEREUM_RPC_URL")
+  url: System.get_env("ETHEREUM_RPC_URL") || "http://localhost:8545"
```
boolafish, line 12: Not sure what is best practice of config in elixir now. Previous idea was to run `MIX_ENV=<env> mix blabla` for each env so we don't need to re-config the common things. However, seems like elixir is moving away from config system?

Probably we can start to have some `local_docker.env` file for local docker testings and cleanup the other configs?

Anyway, not a blocker of this PR. But I am curious to learn from elixir expert what should be a good solution 😛

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -108,6 +120,232 @@ defmodule LoadTest.Ethereum do
     Sync.repeat_until_success(f, timeout)
   end
 
+  @spec fetch_balance(Account.addr_t(), non_neg_integer(), Account.addr_t()) :: non_neg_integer() | :error | nil | map()
+  def fetch_balance(address, amount, currency \\ <<0::160>>) do
+    fetch_balance(Encoding.to_hex(address), amount, Encoding.to_hex(currency), 60)
```
boolafish, line 125: can we name the magic numbers?

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -108,6 +120,232 @@ defmodule LoadTest.Ethereum do
...
+  end
+
+  @spec fetch_rootchain_balance(Account.addr_t(), Account.addr_t()) :: non_neg_integer() | no_return()
+  def fetch_rootchain_balance(address, <<0::160>>) do
+    root_chain_get_eth_balance(Encoding.to_hex(address), 10)
+  end
+
+  def fetch_rootchain_balance(address, currency) do
+    root_chain_get_erc20_balance(Encoding.to_hex(address), Encoding.to_hex(currency), 10)
```
boolafish, line 134: same for magic number

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
InoMurko: it seems like this conversation was marked as resolved, but not addressed?

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -108,6 +120,232 @@ defmodule LoadTest.Ethereum do
...
+             "sign_hash" => sign_hash,
+             "typed_data" => typed_data,
+             "txbytes" => txbytes
+           }
+         ]
+       }, _} ->
+        {:ok, [sign_hash, typed_data, txbytes]}
+
+      _ ->
+        Process.sleep(1_000)
```
boolafish, line 249: do we want to have same sleep value for all networks? 1 seconds might work for docker geth but I would doubt useful for real network?

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -108,6 +120,232 @@ defmodule LoadTest.Ethereum do
...
+  defp root_chain_get_eth_balance(address, counter) do
+    response = eth_account_get_balance(address)
+
+    case response do
+      {:ok, initial_balance} ->
+        {initial_balance, ""} = initial_balance |> String.replace_prefix("0x", "") |> Integer.parse(16)
+        initial_balance
+
+      _ ->
+        Process.sleep(1_000)
```
boolafish, line 267: Please consider to add log for all retries in this file.

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -108,6 +120,232 @@ defmodule LoadTest.Ethereum do
...
+  defp fetch_balance(address, amount, currency, counter) do
+    response =
+      case account_get_balances(address) do
+        {:ok, response} ->
+          decoded_response = Jason.decode!(response.body)
+          Enum.find(decoded_response["data"], fn data -> data["currency"] == currency end)
+
+        _ ->
+          # socket closed etc.
+          :error
```
InoMurko, line 318: logging the error might make sense for visibility

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
boolafish: Also, please consider to add log for all retries in this file

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -108,6 +120,232 @@ defmodule LoadTest.Ethereum do
...
+      end
+
+    case response do
+      # empty response is considered no account balance!
+      nil when amount == 0 ->
+        nil
+
+      %{"amount" => ^amount} = balance ->
```
InoMurko, line 326: you're pinning the amount twice. Once in the calling function `fetch_childchain_balance` and the second time here. is this necessary?  what are you trying to acomplish?

I see you're also mixing approaches. 
https://github.com/omgnetwork/elixir-omg/pull/1745/files#diff-3e3ab486b14889eab1712a81574bd2c6a194527da3c65eec103b399e86a6a896R125

perhaps this should be `do_fetch_balance `. I've seen you used this naming convention before.

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
ayrat555: It was trying to fetch balance. Now I'm raising an exception https://github.com/omgnetwork/elixir-omg/pull/1745/files#diff-3e3ab486b14889eab1712a81574bd2c6a194527da3c65eec103b399e86a6a896R255
Should I return {:error, reason} ?

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -108,6 +120,232 @@ defmodule LoadTest.Ethereum do
...
+  end
+
+  defp account_get_balances(address) do
+    WatcherInfoAPI.Api.Account.account_get_balance(
+      LoadTest.Connection.WatcherInfo.client(),
+      %{
+        address: address
+      }
+    )
+  end
```
InoMurko, line 347:
```suggestion
  defp account_get_balances(address) do
    client = LoadTest.Connection.WatcherInfo.client()
    WatcherInfoAPI.Api.Account.account_get_balance(client, %{ address: address})
  end
```

priv/perf/apps/load_test/lib/scenario/deposits.ex
```diff
@@ -0,0 +1,159 @@
...
+    with {:ok, from, to} <- create_accounts(session),
+         :ok <- create_deposit(from, session),
+         :ok <- send_value_to_receiver(from, to, session) do
+      Session.add_metric(session, "success_rate", 1)
+    else
+      error ->
+        log_error(session, "#{__MODULE__} failed with #{inspect(error)}")
+
+        session
+        |> Session.add_metric("success_rate", 0)
```
InoMurko, line 63: should this be `error_rate`?

priv/perf/apps/load_test/lib/scenario/deposits.ex
InoMurko: leave this as a comment! it's good!

priv/perf/apps/load_test/lib/scenario/deposits.ex
```diff
@@ -0,0 +1,159 @@
...
+  defp send_amount_on_childchain(from, to, token, amount) do
+    {:ok, [sign_hash, typed_data, _txbytes]} =
+      Ethereum.create_transaction(
+        amount,
+        from.addr,
+        to.addr,
+        token
+      )
+
+    _ = Ethereum.submit_transaction(typed_data, sign_hash, [from.priv])
```
InoMurko, line 139: this doesn't do anything
```suggestion
   Ethereum.submit_transaction(typed_data, sign_hash, [from.priv])
```

priv/perf/apps/load_test/lib/test_runner.ex
```diff
@@ -0,0 +1,42 @@
...
+
+  It fetches all configuration params from env vars.
+  """
+  alias LoadTest.TestRunner.Config
+
+  def run() do
+    {runner_module, config} = Config.parse()
+    result = Chaperon.run_load_test(runner_module, print_results: true, config: config)
+
+    case result.metrics["success_rate"][:mean] do
```
boolafish, line 34: slightly concern how scaleable is this approach. I do think we might assert on different conditions for different perf tests. For instance, in watcher info perf tests, I might be interested to fail when `p99 of API_X latency > Y` or when an `API error rate > Y%`. 

Probably need sth like a `verify_result_module` that we can parse out from Config too? (just like the runner_module)

priv/perf/apps/load_test/lib/test_runner.ex
```diff
@@ -0,0 +1,42 @@
...
+      1.0 ->
+        System.halt(0)
+
+      _ ->
+        System.halt(1)
+    end
```
InoMurko, line 40: what is this?

boolafish (commented):
mostly minor comments

priv/perf/apps/load_test/lib/test_runner/config.ex
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+  def parse() do
+    [test, rate, period] = System.argv()
+
+    config = config(test, rate, period)
+
+    runner_module = Map.fetch!(@tests, test)
+
+    {runner_module, config}
```
InoMurko, line 40: should we start a document in describing how to actually run this?

also, I think there was a concern chaperon would blow up if huge parameters would be used? should we validate inputs?

InoMurko (changes requested):

InoMurko (commented):
https://github.com/omgnetwork/elixir-omg/pull/1745/files#diff-fb22ea18bf1eff6706b99ba0780d4f3d5c148f9fade3b73bbf62527bf292a2deR86

https://github.com/omgnetwork/elixir-omg/pull/1745/files#diff-eacc35f8be0345d703df84cd915b5daccd12e8f28ad95a71ecdef5bbaa9b08f7R31


what is the relationship between these two params ... they have the same name but different values.

priv/perf/apps/load_test/lib/test_runner.ex
ayrat555: it's Unix exit codes:
- 0 - success
- 1 - failure

InoMurko: https://tenor.com/vzXt.gif

InoMurko: my question was... what are you measuring with `result.metrics["success_rate"][:mean]`? Why `mean`? how are you making sure you won't get too strict tests in performance tests?

like, you might be simply having network issues in a 72h test and you hit a 90% success rate.

ayrat555: Ok. I can make this configurable
I think @boolafish  has the same comment
https://github.com/omgnetwork/elixir-omg/pull/1745/files#r503739580

ayrat555:
> https://github.com/omgnetwork/elixir-omg/pull/1745/files#diff-fb22ea18bf1eff6706b99ba0780d4f3d5c148f9fade3b73bbf62527bf292a2deR86
> 
> https://github.com/omgnetwork/elixir-omg/pull/1745/files#diff-eacc35f8be0345d703df84cd915b5daccd12e8f28ad95a71ecdef5bbaa9b08f7R31
> 
> what is the relationship between these two params ... they have the same name but different values.

the module attribute value (1) is the default value for `deposit_finality_margin`. In test, I use value different value - 10.

InoMurko:
> > https://github.com/omgnetwork/elixir-omg/pull/1745/files#diff-fb22ea18bf1eff6706b99ba0780d4f3d5c148f9fade3b73bbf62527bf292a2deR86
> > https://github.com/omgnetwork/elixir-omg/pull/1745/files#diff-eacc35f8be0345d703df84cd915b5daccd12e8f28ad95a71ecdef5bbaa9b08f7R31
> > what is the relationship between these two params ... they have the same name but different values.
> 
> the module attribute value (1) is the default value for `deposit_finality_margin`. In test, I use value different value - 10.

Let's drop default values and make the API usage strict.

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
ayrat555: I made a sleep period configurable. Also, I added logging to all sleeps even outside of my pr in the perf project

priv/perf/apps/load_test/lib/test_runner.ex
ayrat555: I added percentile as an optional fourth parameter https://github.com/omgnetwork/elixir-omg/pull/1745/commits/3071f826c1dc9c2d5c5e55b87efe823032373f85

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
ayrat555: now I'm using `Sync` module for every retry

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
ayrat555: Now I use `Sync` module for every retry in the perf project

priv/perf/apps/load_test/lib/scenario/deposits.ex
ayrat555: I renamed it

priv/perf/apps/load_test/lib/scenario/deposits.ex
ayrat555: I added it to the moduledoc

priv/perf/apps/load_test/lib/test_runner.ex
ayrat555: I added percentile as an optional fourth parameter https://github.com/omgnetwork/elixir-omg/pull/1745

priv/perf/apps/load_test/lib/test_runner/config.ex
ayrat555: > should we start a document in describing how to actually run this?

I added moduledoc to the `TestRunner` module

> also, I think there was a concern chaperon would blow up if huge parameters would be used? should we validate inputs?

I added validation. if it spawns more than 200_000 processes, an error will be raised

priv/perf/apps/load_test/lib/child_chain/exit.ex
```diff
@@ -63,21 +63,18 @@ defmodule LoadTest.ChildChain.Exit do
...
   @spec wait_for_exit_data(Utxo.t(), pos_integer()) :: any()
-  def wait_for_exit_data(utxo_pos, counter \\ 10)
-  def wait_for_exit_data(_, 0), do: :error
+  def wait_for_exit_data(utxo_pos, timeout \\ 100_000) do
+    func = fn ->
+      data = get_exit_data(utxo_pos)
 
-  def wait_for_exit_data(utxo_pos, counter) do
-    data = get_exit_data(utxo_pos)
+      if not is_nil(data.proof) do
```
pgebal, line 70: Should data.proof ever be `nil`?

priv/perf/apps/load_test/lib/ethereum/ethereum.ex
```diff
@@ -105,7 +118,222 @@ defmodule LoadTest.Ethereum do
...
+
+  @spec fetch_balance(Account.addr_t(), non_neg_integer(), Account.addr_t()) :: non_neg_integer() | :error | nil | map()
+  def fetch_balance(address, amount, currency \\ <<0::160>>) do
+    {:ok, result} =
+      Sync.repeat_until_success(
+        fn ->
+          do_fetch_balance(Encoding.to_hex(address), amount, Encoding.to_hex(currency))
+        end,
+        @poll_timeout,
+        "Failed to fetch childcahin balance"
```
pgebal, line 132:
```suggestion
        "Failed to fetch childchain balance"
```

priv/perf/apps/load_test/lib/ethereum/sync.ex
```diff
@@ -16,32 +16,41 @@ defmodule LoadTest.Ethereum.Sync do
...
-        :ok = return -> return
-        {:ok, _} = return -> return
-        _ -> do_repeat_until_success(f)
-      end
-    catch
-      _ -> do_repeat_until_success(f)
-      :error, {:badmatch, _} -> do_repeat_until_success(f)
+  defp do_repeat_until_success(f, message) do
+    case f.() do
+      :ok = return ->
```
pgebal, line 38: `:ok -> :ok`
no point for `return` var

priv/perf/apps/load_test/lib/service/sleeper.ex
```diff
@@ -0,0 +1,31 @@
...
+defmodule LoadTest.Service.Sleeper do
+  @moduledoc """
+    Sleeps the current process for the given timeout and print the given warning.
+  """
+
+  require Logger
+
+  @spec sleep(String.t()) :: Ok
+  def sleep(message) do
+    _ = Logger.warn(message)
```
pgebal, line 24: Isn't `warn` level too much?

priv/perf/apps/load_test/lib/service/sleeper.ex
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+
+defmodule LoadTest.Service.Sleeper do
+  @moduledoc """
+    Sleeps the current process for the given timeout and print the given warning.
+  """
+
+  require Logger
+
+  @spec sleep(String.t()) :: Ok
```
pgebal, line 22: Shouldn't return type be `:ok`?

priv/perf/apps/load_test/lib/scenario/deposits.ex
```diff
@@ -0,0 +1,154 @@
...
+           ),
+         :ok <-
+           fetch_rootchain_balance(
+             from_address,
+             amount: initial_amount - deposited_amount - gas_used,
+             token: token,
+             error: :wrong_rootchain_balance_after_deposit
+           ) do
+      :ok
+    else
```
pgebal, line 101: I don't think we need `else` branch here if we don't do anything with the error.

priv/perf/apps/load_test/lib/scenario/deposits.ex
```diff
@@ -0,0 +1,154 @@
...
+    with _ <- send_amount_on_childchain(from_address, to_address, token, transferred_amount),
+         :ok <-
+           fetch_childchain_balance(
+             to_address,
+             amount: transferred_amount,
+             token: token,
+             error: :wrong_childchain_to_balance_after_sending_deposit
+           ) do
+      :ok
+    else
```
pgebal, line 120: Same here.

priv/perf/apps/load_test/lib/scenario/deposits.ex
```diff
@@ -0,0 +1,154 @@
...
+      error -> error
+    end
+  end
+
+  defp send_amount_on_childchain(from, to, token, amount) do
+    {:ok, [sign_hash, typed_data, _txbytes]} =
+      Ethereum.create_transaction(
```
pgebal, line 127: I think this function, and couple others, should not be in module named `Ethereum`? It has nothing to do with Ethereum.

priv/perf/apps/load_test/lib/test_runner.ex
```diff
@@ -0,0 +1,48 @@
...
+  It fetches all configuration params from env vars.
+  """
+  alias LoadTest.TestRunner.Config
+
+  def run() do
+    {runner_module, config, property} = Config.parse()
+    result = Chaperon.run_load_test(runner_module, print_results: true, config: config)
+
+    case result.metrics["error_rate"][property] do
+      0.0 ->
```
pgebal, line 38: Why it's sometimes 0.0 and other times 0? Can't we merge it into a single branch?

pgebal (changes requested):

priv/perf/apps/load_test/lib/test_runner.ex
boolafish: haha my concern is slightly different. I am just thinking not all test want to assert on same stuff. Probably some does not even care about error_rate. But I think we can iterate on this as you're going to continue working with perf anyway. I guess you'll figure the solution out when adding more tests with different conditions to assert.

priv/perf/apps/load_test/lib/child_chain/exit.ex
ayrat555: It can be nil. Because its key is set directly from response data without checking the result

priv/perf/apps/load_test/lib/scenario/deposits.ex
ayrat555: I moved unrelated to Ethereum API functions to separate modules

ayrat555:
@pgebal @InoMurko @boolafish I addressed your comments. Can you please take another look?

priv/perf/apps/load_test/lib/service/sync.ex
```diff
@@ -12,36 +12,45 @@
...
 
   @doc """
   Repeats f until f returns {:ok, ...}, :ok OR exception is raised (see :erlang.exit, :erlang.error) OR timeout
   after `timeout` milliseconds specified
 
   Simple throws and :badmatch are treated as signals to repeat
   """
-  def repeat_until_success(f, timeout) do
-    fn -> do_repeat_until_success(f) end
+  def repeat_until_success(f, timeout, message) do
```
boolafish, line 30: 🤔 I know this is old code logic, but do we not want a circuit breaker after certain times of retries??

priv/perf/apps/load_test/lib/service/sync.ex
```diff
@@ -12,36 +12,45 @@
...
-defmodule LoadTest.Ethereum.Sync do
+defmodule LoadTest.Service.Sync do
   @moduledoc """
   Provides a function for repeating a function call until a given criteria is met.
   """
-  @sleep_interval_ms 100
+
+  require Logger
+
+  import LoadTest.Service.Sleeper
```
boolafish, line 22: I thought `import` is not preferred ?

priv/perf/apps/load_test/lib/service/sync.ex
ayrat555: Most of the usages of this function are related for waiting for some condition/even in the chain. 

For cases where an external service is not available, I think we can implement some kind of circuit breaker that will fail immediately.

boolafish:
@ayrat555 do you mind to split the docker and k-job part to different PR? I think the deposit code part has enough sizing and is closed to merge? 

Another reason asking for another PR is the docker should be pushed to a docker hub or google container registry from CI. So the pipeline tooling can just use the image directly. The dockerizing and k-job would have certain size of code as well.

boolafish (approved):
Giving approval, but please move docker and k-job to a separate PR for review

ayrat555:
@boolafish I removed dockerfile and kubernetes job from this PR. For now I'm trying to make it work in my local cluster. I'll add publishing to docker hub after I make it work

priv/perf/apps/load_test/lib/child_chain/deposit.ex
```diff
@@ -16,37 +16,57 @@ defmodule LoadTest.ChildChain.Deposit do
...
-          pos_integer(),
-          LoadTest.Ethereum.Account.t(),
-          pos_integer(),
-          pos_integer()
-        ) :: Utxo.t()
-  def deposit_from(%Account{} = depositor, amount, currency, deposit_finality_margin, gas_price) do
+  @spec deposit_from(Account.t(), pos_integer(), Account.t(), non_neg_integer(), non_neg_integer, atom()) ::
+          Utxo.t() | binary()
+  def deposit_from(
+        %Account{} = depositor,
```
InoMurko, line 52: why pattern match?

priv/perf/apps/load_test/lib/test_runner.ex
```diff
@@ -0,0 +1,46 @@
...
+
+  mix run -e "LoadTest.TestRunner.run()" -- "deposits" "1" "5"
+
+  It accepts three arguments:
+  - test name
+  - transactions per seconds
+  - period in seconds
+
+  Optional paramters:
+  - Percentile. You can pass percentile as the fourth parameter. By default `mean` value is used.
```
InoMurko, line 27: where are these used???

InoMurko: whats the purpose?

priv/perf/apps/load_test/lib/test_runner.ex
```diff
@@ -0,0 +1,46 @@
...
+  It fetches all configuration params from env vars.
+  """
+  alias LoadTest.TestRunner.Config
+
+  def run() do
+    {runner_module, config, property} = Config.parse()
+    result = Chaperon.run_load_test(runner_module, print_results: true, config: config)
+
+    # / 1 is to convert result to float (mean is float, percentiles are integers)[O
+    case result.metrics["error_rate"][property] / 1 do
```
InoMurko, line 38: how are you interpreting the results?

priv/perf/apps/load_test/lib/test_runner/config.ex
```diff
@@ -0,0 +1,80 @@
...
+
+    config = config(test, rate_int, period_int)
+
+    runner_module = Map.fetch!(@tests, test)
+
+    {runner_module, config, property}
+  end
+
+  defp config(test, rate_int, period_int) do
+    if rate_int * period_int > 200_000, do: raise("too many processes")
```
InoMurko, line 55: someone not familiar with the test framework is not going to know what this means.

priv/perf/apps/load_test/lib/test_runner/config.ex
```diff
@@ -0,0 +1,80 @@
...
+      chain_config: chain_config,
+      timeout: :infinity
+    }
+  end
+
+  defp parse_percentile(percentile) do
+    percentile_int = String.to_integer(percentile)
+
+    unless rem(percentile_int, 10) == 0 and div(percentile_int, 10) < 10 do
+      raise("Wrong percentile")
```
InoMurko, line 75: why is it wrong?

priv/perf/apps/load_test/lib/scenario/deposits.ex
```diff
@@ -0,0 +1,151 @@
...
+
+    {:ok, from_address, to_address}
+  end
+
+  defp create_deposit(from_address, session) do
+    token = config(session, [:chain_config, :token])
+    deposited_amount = config(session, [:chain_config, :deposited_amount])
+    initial_amount = config(session, [:chain_config, :initial_amount])
+
+    txhash = Deposit.deposit_from(from_address, deposited_amount, token, 10, 0, :txhash)
```
InoMurko, line 85: why is `gas_price` 0? are you sure this is going to work against ropsten or rinkeby?

InoMurko (changes requested):

priv/perf/apps/load_test/lib/test_runner/config.ex
ayrat555: I added a comment

priv/perf/apps/load_test/lib/scenario/deposits.ex
ayrat555: I set it to `2_000_000_000`. Also, now every test value is overridable through a file that can be passed in the env variable.

priv/perf/apps/load_test/lib/test_runner.ex
ayrat555: It's used when checking error_rate. By default mean value is used. if percentile is set, error_rate percentile is used

InoMurko (commented):
are we able to run these tests towards docker-compose setup *and* rinkeby, mainnet?

ayrat555:
@InoMurko yes, by setting env variables

InoMurko (approved):

pgebal (approved):

PR #1747, opened 2020-10-06
chore: changelog/version bump for v1.0.5

achiurizo:
## Overview

Release for `v1.0.5-pre.0`

[v1.0.5](https://github.com/omgnetwork/elixir-omg/tree/v1.0.5) is up-to date with [master](https://github.com/omgnetwork/elixir-omg/tree/master)

## Changes

`CHANGELOG` and `VERSION` bump.

## Testing

CI? CI.


boolafish:
some changes looks bit off to me, can you double check @unnawut is this expected? also cc @jarindr as you have experience last time

unnawut:
Yeah we had similar issues with v1.0.4 that the generator went back and change older version's content. Weird that it also affects the versions that we have not touched for months and not just the recent ones.

Couldn't find the root cause so far, so we just revert those changes :(

boolafish:
what if we don't revert and take in that change. Would the next generation still have such issue?

unnawut:
Yup can try! Those changes don't seem too critical

boolafish (approved):

boolafish (commented):
ah I think we should change `Unreleased` to 1.0.5

achiurizo:
I've merged master into this and also moved all the Unreleased into this version as well `v1.0.5-pre.0` cc @unnawut @boolafish

CHANGELOG.md
```diff
@@ -1,25 +1,41 @@
 # Changelog
 
-## [Unreleased](https://github.com/omgnetwork/elixir-omg/tree/HEAD)
+## [v1.0.5](https://github.com/omgnetwork/elixir-omg/tree/v1.0.5) (2020-10-01)
 
-[Full Changelog](https://github.com/omgnetwork/elixir-omg/compare/v1.0.4...HEAD)
+[Full Changelog](https://github.com/omgnetwork/elixir-omg/compare/v1.0.4-pre.2...v1.0.5)
```
T-Dnzt, line 5: Shouldn't we compare with the last released release? `1.0.3` instead of `v1.0.4`?

T-Dnzt (approved):

CHANGELOG.md
boolafish: I don't think that is a good idea. We have changelog for 1.0.4 already and 1.0.4 is a stable release (as it passed sandbox) just it did not reach production. Compare to 1.0.3 might cause more confusion IMO.

boolafish (approved):

PR #1750, opened 2020-10-16
add docker file for perf

ayrat555:
This pr adds docker file and k8s job for the perf project

### DIFF
--- a/priv/perf/docker/Dockerfile
+++ b/priv/perf/docker/Dockerfile
@@ -1,0 +1,41 @@
+FROM elixir:1.10.4-alpine
+
+RUN apk add git curl bash maven jq \
+        autoconf \
+        automake \
+        gmp \
+        gmp-dev \
+        libtool \
+        gcc \
+        cmake \
+        gnupg \
+        alpine-sdk
+
+RUN git clone https://github.com/omgnetwork/elixir-omg
+
+WORKDIR ./elixir-omg
+
+RUN mkdir -p priv/openapitools \
+        && curl https://raw.githubusercontent.com/OpenAPITools/openapi-generator/master/bin/utils/openapi-generator-cli.sh > priv/openapitools/openapi-generator-cli \
+        && chmod u+x priv/openapitools/openapi-generator-cli
+
+RUN priv/openapitools/openapi-generator-cli generate \
+        -i https://raw.githubusercontent.com/omgnetwork/omg-childchain-v1/master/apps/omg_child_chain_rpc/priv/swagger/operator_api_specs.yaml \
+        -g elixir \
+        -o priv/perf/apps/child_chain_api/
+
+RUN priv/openapitools/openapi-generator-cli generate \
+        -i apps/omg_watcher_rpc/priv/swagger/security_critical_api_specs.yaml \
+        -g elixir \
+        -o priv/perf/apps/watcher_security_critical_api/
+
+RUN priv/openapitools/openapi-generator-cli generate \
+        -i apps/omg_watcher_rpc/priv/swagger/info_api_specs.yaml \
+        -g elixir \
+        -o priv/perf/apps/watcher_info_api/
+
+RUN mix local.hex --force && mix local.rebar --force
+
+WORKDIR ./priv/perf
+
+RUN mix deps.get && mix compile
--- a/priv/perf/docker/Dockerfile
+++ b/priv/perf/docker/Dockerfile
@@ -16,7 +16,7 @@ RUN git clone https://github.com/omgnetwork/elixir-omg
 WORKDIR ./elixir-omg
 
 RUN mkdir -p priv/openapitools \
-        && curl https://raw.githubusercontent.com/OpenAPITools/openapi-generator/master/bin/utils/openapi-generator-cli.sh > priv/openapitools/openapi-generator-cli \
+        && curl https://raw.githubusercontent.com/OpenAPITools/openapi-generator/v4.3.1/bin/utils/openapi-generator-cli.sh > priv/openapitools/openapi-generator-cli \
         && chmod u+x priv/openapitools/openapi-generator-cli
 
 RUN priv/openapitools/openapi-generator-cli generate \
--- a/priv/perf/docker/k8s_job.yaml
+++ b/priv/perf/docker/k8s_job.yaml
@@ -1,0 +1,85 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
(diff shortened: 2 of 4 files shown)


priv/perf/docker/k8s_job.yaml
```diff
@@ -0,0 +1,44 @@
...
+  name: perf-job
+spec:
+  template:
+    spec:
+      containers:
+      - name: perf
+        image: perf:lo
+        imagePullPolicy: Never
+        command: ["mix",  "run", "-e", "LoadTest.TestRunner.run()", "--", "deposits", "1", "5"]
+        env:
```
ayrat555, line 13: I'm not sure how should I set them. should I hardcode them and create different files for different env. or should I use something like https://kubernetes.io/docs/concepts/configuration/configmap/

boolafish: Later in the stage it makes sense to port those env non-changing or quite likely wouldn't change values into config file so we can just have them in the cluster that runs the job. But I think starting with file per test is good for now.

priv/perf/docker/running_k8s_job_locally.md
```diff
@@ -0,0 +1,71 @@
...
+eval $(minikube docker-env)
+```
+
+2. Build docker image:
+
+```
+docker build -f Dockerfile -t perf:latest .
+```
+
+#### 3. Add host alias of your machine to the k8s job
```
boolafish, line 31: This is coooool 🤯

boolafish (commented):
It is really cool to see it is possible to run k-job against local docker-composed services.

However, I am not really following the plan here. In practice, we will only need job running against sandbox env (or stress). And I doubt for development, people would bother to run perf in k-job mode. People would probably only use that when trying to develop the k-job config itself. But then they don't need to develop a k-job config running with local docker services.

Do you have plan on how this is going to be used??

priv/perf/docker/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+        gcc \
+        cmake \
+        gnupg \
+        alpine-sdk
+
+RUN git clone https://github.com/omgnetwork/elixir-omg
+
+WORKDIR ./elixir-omg
+
+RUN mkdir -p priv/openapitools \
```
boolafish, line 18: 1. If this is the preferred way, probably consider to replace this script too: https://github.com/omgnetwork/elixir-omg/blob/master/priv/perf/scripts/generate_api_client.sh
And just run the script inside so we get consistent setup when developing perf and using this docker.

1. probably we should pin to release version tag instead of `master`?

ayrat555:
> People would probably only use that when trying to develop the k-job config itself. But then they don't need to develop a k-job config running with local docker services.

Yes, that's the use case I created the instruction for running it locally. Otherwise, changes will have to be tested directly on spinnaker

For other environments, I think separate k jobs can be created as you suggested

priv/perf/docker/Dockerfile
ayrat555: > If this is the preferred way, probably consider to replace this script too: https://github.com/omgnetwork/elixir-omg/blob/master/priv/perf/scripts/generate_api_client.sh
And just run the script inside so we get consistent setup when developing perf and using this docker.

I don't think it's the most convenient way because it requires installing java, maven, jq. 

2. You're right. I'll do this now

boolafish:
> therwise, changes will have to be tested directly on spinnaker

Not really though. For instance, you can set the k-job configs to sandbox and then run it in your Minikube, right?
And those is really what people would want to develop. People don't really care about a k-job that runs with docker-composes?

ayrat555:
I'm not familiar with sandbox env. is it a separate env for testing? How frequently is it updated?

Don't you think it's safer and more convenient testing it locally first?
1. You already know all env vars
2. Bug in your code or too many test runs can drain faucet account. (Sandbox env will have to be re-deployed?)
3. You don't depend on sandbox env

boolafish:
All env vars you need for development should be here: https://github.com/omgnetwork/devops/wiki/DevOps-Wiki
Faucet is on your choice and you can use your own address on testet. (But we can announce some shared faucet address and add alarm on those)

The problem is the k-job config that will be useful and we want to maintain is really the one that hits sandbox (or other real env). So I am not sure how people would use your k-job file for docker-compose as the correctness of that local k-job is not that interested to anyone.

InoMurko:
I think the scope of this repository is to ensure that a container image can be built and that the container image that was built can be used.

It's up to infra (marely talking scope here) to determine how that image is going to be used. Kubertnetes or docker-compose, or whatever. You just need to make sure you can configure the application in a way that can be used (env vars).

I don't see much value in commiting k8 infra files here, because running minikube locally is too much - also, if you insist in running minikube locally for this solution, you need to setup cirleci as well.

boolafish:
> I don't see much value in commiting k8 infra files here

yeah agree actually. I think publishing docker would be enough. We can develop k-job later. But for you to get a sense of it, you can run it your way or just set to real env values

ayrat555:
@InoMurko @boolafish 
I think we decided to not use a separate image

```
Ino Murko:speech_balloon:  2 days ago
ah... I don't think you need another image, then! :slightly_smiling_face:

Ino Murko:speech_balloon:  2 days ago
one less step.

Ayrat Badykov  2 days ago
but it takes a couple of minutes to generate API clients, fetch and compile deps

Ino Murko:speech_balloon:  2 days ago
well, since these tests are gonna run N hours, I don't think a couple of minutes is a lot, by contrast

Ayrat Badykov  2 days ago
Ok. I'll try moving these commands to k job
```
So I moved all commands to k8s job https://github.com/omgnetwork/elixir-omg/pull/1750/commits/98ef8d081dbc7a0e1e1e036b563e2f547dac5b73

boolafish:
oh I am a bit surprised that it works for you to just write those commands as script inside that block! Was trying it before somehow failed for me.

lol but can I challenge that decision again, not really a fan of coupling with childchain docker and publish public docker should not be that hard right? And your [adding a lots of lines](https://github.com/omgnetwork/elixir-omg/blob/98ef8d081dbc7a0e1e1e036b563e2f547dac5b73/priv/perf/docker/k8s_job.yaml#L13-L53) to whoever is going to write kubernate job. Not to mention we will have lots of k-jobs for each tests. (eg. deposit, childchain utxo....)

ayrat555:
closed in favour of https://github.com/omgnetwork/elixir-omg/pull/1753

PR #1753, opened 2020-10-22
Publish the perf project

ayrat555:
This PR adds publishing of the perf project docker image to docker hub during circle ci builds so the perf project can be used as a standalone application outside of elixir-omg.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1111,15 +1111,16 @@ workflows:
           filters: *master_and_version_branches_and_all_tags
 
       - publish_perf:
-          filters: *all_branches_and_tags
-          #   branches:
-          #     only:
-          #       - master
-          #       # vMAJOR.MINOR (e.g. v0.1, v0.2, v1.0, v2.1, etc.)
-          #       - /^v[0-9]+\.[0-9]+/
-          #   tags:
-          #     only:
-          #       - /.+/
+          requires: [test_docker_compose_performance]
+          filters:
+            branches:
+              only:
+                - master
+                # vMAJOR.MINOR (e.g. v0.1, v0.2, v1.0, v2.1, etc.)
+                - /^v[0-9]+\.[0-9]+/
+            tags:
+              only:
+                - /.+/
       # Increase chart version for master, this will end up trigger deployment on dev
       - increase_chart_version_watcher_master:
           requires: [publish_watcher, publish_watcher_info]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1109,15 +1109,17 @@ workflows:
               audit_deps
             ]
           filters: *master_and_version_branches_and_all_tags
+
       - publish_perf:
-          branches:
-            only:
-              - master
-              # vMAJOR.MINOR (e.g. v0.1, v0.2, v1.0, v2.1, etc.)
-              - /^v[0-9]+\.[0-9]+/
-          tags:
-            only:
-              - /.+/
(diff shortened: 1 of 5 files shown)


priv/perf/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+        automake \
+        gmp \
+        gmp-dev \
+        libtool \
+        gcc \
+        cmake \
+        gnupg \
+        alpine-sdk
+
+RUN git clone https://github.com/omgnetwork/elixir-omg
```
boolafish, line 14: I think you should be directly copy from the repo itself instead though. like: https://github.com/omgnetwork/elixir-omg/blob/c6cb81f063e2c4ffa0ef65da4f00406f9fa7fd2a/Dockerfile.watcher#L39

So you are not publishing current master but *this* current perf?

.circleci/config.yml
```diff
@@ -1099,6 +1109,17 @@ workflows:
               audit_deps
             ]
           filters: *master_and_version_branches_and_all_tags
+
+      - publish_perf:
```
boolafish, line 1113: Probably like publish_watcher, only publish after some required tests have been done? eg the local docker perf test one so we can make sure this version of perf is runnable.

priv/perf/Dockerfile
ayrat555: instead of cloning the repo in the Dockerfile, I now copy the current elixir-omg directory https://github.com/omgnetwork/elixir-omg/pull/1753/files#diff-4a6ddb0f679098f7358db41ea45a306bc35e49328fc46afcd5a1e7bec2bf0130R14

.circleci/config.yml
ayrat555: I added  `requires: [test_docker_compose_performance]`

.circleci/config.yml
```diff
@@ -870,6 +870,16 @@ jobs:
           path: current_release/
       - run: IMAGE_NAME=$WATCHER_IMAGE_NAME sh .circleci/ci_publish.sh
 
+  publish_perf:
+    machine:
+      image: ubuntu-1604:201903-01
+    environment:
+      PERF_IMAGE_NAME: "omisego/perf"
```
boolafish, line 877: just curious do we want to start migrate the naming to omgnetwork?

apps/omg_eth/test/support/dev_geth.ex
```diff
@@ -103,7 +103,7 @@ defmodule OMG.Eth.DevGeth do
 
     waiting_task
     |> Task.async()
-    |> Task.await(30_000)
+    |> Task.await(90_000)
```
boolafish, line 106: just curious what is leading to this change?

boolafish (approved):

apps/omg_eth/test/support/dev_geth.ex
ayrat555: dev geth is timing out which causes ci failures. I increased it so dev geth has more time to prepare itself

.circleci/config.yml
ayrat555: I'm not sure. All dockers images have the same prefix
cc @achiurizo @InoMurko

InoMurko: I'm not aware of a way to migrate a docker hub account to a new namespace! (which is what `omisego` is)

InoMurko (approved):

PR #1756, opened 2020-10-26
Utxo perf test

ayrat555:
This PR adds utxo perf test that can be run from the command line using `mix run`:

A single iteration of this test consists of the following steps:

1.1 Two accounts are created - the sender and the receiver

2.1 The sender account is funded with `initial_amount`
2.2 The balance on the childchain of the sender is validated using `WatcherInfoAPI.Api.Account.account_get_balance` API.
2.3 Utxos of the sender are validated using `WatcherInfoAPI.Api.Account.account_get_utxos` API

3.1 The sender sends al his value to the receiver
3.2 The balance on the childchain of the sender is validated
3.3 The balance on the childchain of the receiver is validated
3.4 Utxos of the sender are validated
3.5 Utxos of the receiver are validated

You can run it using:
```sh
mix run -e "LoadTest.TestRunner.run()" -- "utxos" 2 5
```
Parameters:
1. test name - utxos
2. tps - 2 t/s
3. period - 5 seconds

So the command above means "Run utxos test for 5 seconds with 2 tests / second". Keep in mind that using 100 t/s causes elixir-omg to start returning internal server error.

You can modify the following test paramers:
      - token -address of the token. default value is "0x0000000000000000000000000000000000000000",
      - initial_amount - this value used when funding the receiver. default value is 760
      - fee - fee used when transferring value from the sender to the receiver

  You can also modify values for the by providing `TEST_CONFIG_PATH` env variable, it should contain
  the path to json file. For example:
```sh
    TEST_CONFIG_PATH=./my_file mix run -e "LoadTest.TestRunner.run()" -- "utxos" 1 5
```


priv/perf/apps/load_test/lib/watcher_info/utxo.ex
```diff
@@ -0,0 +1,101 @@
...
+           }
+         ) do
+      {:ok, result} ->
+        result.body
+        |> Jason.decode!()
+        |> find_utxo(utxo)
+
+      other ->
+        other
+    end
```
InoMurko, line 53: why not spend a little bit more time on the code design? who likes to read this, honestly?

```suggestion
client = WatcherInfo.client()
body = %WatcherInfoAPI.Model.AddressBodySchema1{address: Encoding.to_hex(sender.addr}
account_utxos = WatcherInfoAPI.Api.Account.account_get_utxos(client, body) 
    case account_utxos do
      {:ok, result} ->
        result.body
        |> Jason.decode!()
        |> find_utxo(utxo)

      other ->
        other
    end
```

priv/perf/apps/load_test/lib/scenario/utxos.ex
```diff
@@ -0,0 +1,141 @@
...
+      :ok
+    end
+  end
+
+  defp fetch_childchain_balance(account, amount: amount, token: token, error: error) do
+    childchain_balance = Balance.fetch_balance(account.addr, amount, token)
+
+    case childchain_balance do
+      nil when amount == 0 -> :ok
+      %{"amount" => ^amount} -> :ok
```
InoMurko, line 137: are you sure this two are the same?

priv/perf/apps/load_test/lib/watcher_info/utxo.ex
```diff
@@ -0,0 +1,101 @@
...
+      :empty ->
+        case decoded_response["data"] do
+          [] -> {:ok, []}
+          other -> {:error, other}
+        end
+
+      _data ->
+        do_find_utxo(decoded_response, utxo)
+    end
+  end
```
InoMurko, line 70:
```suggestion
  defp find_utxo(decoded_response, nil) do
  {:ok, decoded_response}
  end
  defp find_utxo(decoded_response, :empty) do
 case decoded_response["data"] do
          [] -> {:ok, []}
          other -> {:error, other}
        end
  end
  defp find_utxo(decoded_response, _utxo) do
do_find_utxo(decoded_response, utxo)
  end
```

InoMurko:
Perhaps a bit more effort on the PR description and the test description. `utxos` is not a very good descriptive name of a test. 
the parameters 1 and 1, also don't ring a bell what they mean, could we paramatrice these so that we know what we're setting?
--num_senders
--num_receivers

do the senders and receivers need to be the same? what happens if they're not? if they need to be the same, why are there two separate paramters?

priv/perf/apps/load_test/lib/scenario/utxos.ex
ayrat555: `nil when amount == 0` means we're expecting amount to be empty in the childchain for the specified currency. if it's not found in the json returned from watcherinfo, it means amount is zero.

ayrat555:
@InoMurko I extended PR description

InoMurko (approved):

priv/perf/apps/load_test/test/load_tests/runner/utxos_test.exs
```diff
@@ -0,0 +1,42 @@
...
+      chain_config: %{
+        token: token,
+        initial_amount: initial_amount,
+        fee: fee
+      },
+      run_config: %{
+        tps: 1,
+        period_in_seconds: 20
+      },
+      timeout: :infinity
```
boolafish, line 35: why infinity??

boolafish (approved):
> Keep in mind that using 100 t/s causes elixir-omg to start returning internal server error.

how do you get this?? do you run with docker or existing env? Cause I am quite sure we run TPS 100 several times on our sandbox

boolafish:
Can you link this PR to the issue?? so it can be easier tracked. @ayrat555 

I am also curious on what your approach is. Are you removing existing stuffs like `LoadTest.Runner.UtxosLoad` afterward??

priv/perf/apps/load_test/test/load_tests/runner/utxos_test.exs
ayrat555: the default timeout is 10_000 ms for task execution inside of chaperon

```
2020-10-29 13:24:13.437 [error] ⋅Task #PID<0.500.0> started from #PID<0.501.0> terminating
** (stop) exited in: Task.await(%Task{owner: #PID<0.501.0>, pid: #PID<0.879.0>, ref: #Reference<0.3543989965.2161639427.128884>}, 10000)
    ** (EXIT) time out
    (elixir 1.10.2) lib/task.ex:607: Task.await/2
    (chaperon 0.3.1) lib/chaperon/session.ex:366: Chaperon.Session.await/3
    (elixir 1.10.2) lib/enum.ex:2111: Enum."-reduce/3-lists^foldl/2-0-"/3
    (chaperon 0.3.1) lib/chaperon/scenario.ex:198: Chaperon.Scenario.with_scenario/3
    (chaperon 0.3.1) lib/chaperon/scenario.ex:172: Chaperon.Scenario.run/2
    (chaperon 0.3.1) lib/chaperon/scenario.ex:122: Chaperon.Scenario.execute/2
    (elixir 1.10.2) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2
    (elixir 1.10.2) lib/task/supervised.ex:35: Task.Supervised.reply/5
    (stdlib 3.12) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
```

ayrat555:
> how do you get this?? do you run with docker or existing env? Cause I am quite sure we run TPS 100 several times on our sandbox

`WatcherInfoAPI.Api.Account.account_get_utxos(client, body)` starts returning `server:internal_server_error`

ayrat555:
Also, I see errors in watch_info logs
```
Server: localhost:7534 (http)
watcher_info    | Request: POST /account.get_utxos
watcher_info    | ** (exit) an exception was raised:
watcher_info    |     ** (DBConnection.ConnectionError) connection not available and request was dropped from queue after 171ms. This means requests are coming in and your connection pool cannot serve them fast enough. You can address this by:
watcher_info    | 
watcher_info    |   1. By tracking down slow queries and making sure they are running fast enough
watcher_info    |   2. Increasing the pool_size (albeit it increases resource consumption)
watcher_info    |   3. Allow requests to wait longer by increasing :queue_target and :queue_interval
watcher_info    | 
watcher_info    | See DBConnection.start_link/2 for more information

```

ayrat555:
> Are you removing existing stuff like LoadTest.Runner.UtxosLoad afterwards??

I didn't plan to do it yet. as @kevsul mentioned, existing tests have a different purpose?

PR #1762, opened 2020-11-03
add perf cli smoke test

ayrat555:
This PR adds ci step that runs `transactions` perf test

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1040,137 +1040,137 @@ workflows:
               only: /.+/
             tags:
               only: /.+/
-      # - test_barebone_release:
-      #     filters: *all_branches_and_tags
-      # - notify_services:
-      #     requires:
-      #       - increase_chart_version_watcher_master
-      #       - increase_chart_version_watcher_info_master
-      #     filters:
-      #       branches:
-      #         only:
-      #           - master
-      # - coveralls_report:
-      #     requires:
-      #       - watcher_coveralls_and_integration_tests
-      #       - watcher_info_coveralls_and_integration_tests
-      #       - common_coveralls_and_integration_tests
-      #       - test
-      # - watcher_coveralls_and_integration_tests:
-      #     requires: [build]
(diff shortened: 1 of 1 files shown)


ayrat555:
ci fails because `feefeed` is broken

.circleci/config.yml
```diff
@@ -880,6 +880,31 @@ jobs:
       - run: make docker-perf IMAGE_NAME=$PERF_IMAGE_NAME
       - run: IMAGE_NAME=$PERF_IMAGE_NAME sh .circleci/ci_publish.sh
 
+  run_perf_cli_smoke_test:
+    machine:
+      image: ubuntu-1604:201903-01
+    environment:
+      PERF_IMAGE_NAME: "omisego/perf:ci"
```
boolafish, line 887: where is generating the `:ci` tagged image?? Also when will this tag get updated?

ayrat555: it's a name for the temporary image used only in circle ci.
Generation is here - https://github.com/omgnetwork/elixir-omg/pull/1762/files/4ee5857447db0e44a4b6c70663d41ab015d0d902#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R901

boolafish: ah my bad....it is just local image where I somehow thought it was from remote.

boolafish (approved):
The step itself looks good to me. Do you think there is potential to just merge `test_docker_compose_performance` with this (?) They seems overlapping a bit on the purpose of it.

.circleci/config.yml
InoMurko: what is true is the convention is to name the `current` image - `latest` :)

InoMurko:
> The step itself looks good to me. Do you think there is potential to just merge `test_docker_compose_performance` with this (?) They seems overlapping a bit on the purpose of it.

I would second that!

InoMurko (approved):

PR #1763, opened 2020-11-06
feat: add new metadata to Datadog traces

macaptain:
## Overview

Because API responses are always 200, even if there's an error message in the body of the response, observability is impacted and it's hard to understand the health of services. This PR adds extra fields to the spans of tags passing through the Watcher and Watcher Info Phoenix based APIs, which flags responses with errors in the trace in Datadog.

While Datadog generates metrics from traces, unfortunately it isn't possible to group the traces by error_type in the metrics count (see https://docs.datadoghq.com/tracing/guide/metrics_namespace/#errors), so we will need to add more instrumentation, counting errors hit with custom metrics. This'll be a different PR.

## Changes

What you see in Datadog traces after this change:
* The service name (default 'web' for Phoenix applications) is the nameof the OMG application.
* Traces are tagged with application version number
* For API responses which error, the trace is flagged red as an error, the error code is marked on the trace and the error message is found in the tags.

image

This is implemented by using `customize_metadata` on the SpandexPhoenix tracer: https://hexdocs.pm/spandex_phoenix/SpandexPhoenix.html, parsing the JSON response body with Jason (hopefully a relatively cheap operation), and adding the fields to the default metadata.

## Testing

```
DD_API_KEY=<some-api-key> docker-compose -f docker-compose-watcher.yml -f docker-compose.dev.yml up
```

Make API calls to the endpoints, and check the local-development environment APM in Datadog. Look for services "watcher" and "watcher_info", and check the traces in the DataDog GUI.

## P.S.

This is my first Elixir PR. Please let me know how my code can be more idiomatic!

### DIFF
--- a/apps/omg_watcher_rpc/lib/tracer.ex
+++ b/apps/omg_watcher_rpc/lib/tracer.ex
@@ -18,4 +18,69 @@ defmodule OMG.WatcherRPC.Tracer do
   """
 
   use Spandex.Tracer, otp_app: :omg_watcher_rpc
+
+  @doc """
+  Given conn and parsed response body, return a span with extra metadata.
+
+  The service medata (default ':web') is overridden to be the name of the OMG service.
+
+  Metadata added to the span is not arbitrary and is validated by Spandex against a schema:
+
+  service(:atom) Required: The default service name to use for spans declared without a service
+  env(:string): A name used to identify the environment name, e.g prod or development
+  services([{:keyword, :atom}, :keyword]): A mapping of service name to the default span types. - Default: []
+  completion_time(:integer)
+  error(:keyword)
+  http(:keyword)
+  id(:any)
+  name(:string)
+  parent_id(:any)
+  private(:keyword) - Default: []
+  resource([:atom, :string])
+  sql_query(:keyword)
+  start(:integer)
+  tags(:keyword) - Default: []
+  trace_id(:any)
+  type(:atom)
+  """
+  defp add_metadata_from_response_body(conn, json_resp_body) do
+    service = String.to_atom(json_resp_body["service_name"])
+    error = if !json_resp_body["success"], do: Keyword.new([{:error, true}])
+    error_type = if !!error and json_resp_body["data"], do: json_resp_body["data"]["code"]
+    error_msg = if !!error and json_resp_body["data"], do: json_resp_body["data"]["description"]
+
+    tags = [
+      {:version, json_resp_body["version"]}
+    ]
+
+    _ = if error_type, do: tags ++ {String.to_atom("error.type"), error_type}
+    _ = if error_msg, do: tags ++ {String.to_atom("error.msg"), error_msg}
+
+    trace_data =
+      conn
+      |> SpandexPhoenix.default_metadata()
+      |> Keyword.put(:service, service)
+      |> Keyword.put(:tags, tags)
+
+    if error,
+      do:
+        trace_data
+        |> Keyword.put(:error, error),
+      else: trace_data
+  end
+
+  @doc """
+  Adds metadata from the response body to the Spandex span.
+
+  The conn is inspected just before sending back the API response.
+
+  Handles failure to decode the response body gracefully.
+  """
+
+  def add_trace_metadata(conn) do
+    case Jason.decode(conn.resp_body) do
+      {:ok, json_resp_body} -> add_metadata_from_response_body(conn, json_resp_body)
+      {:error, _} -> SpandexPhoenix.default_metadata(conn)
+    end
+  end
 end
--- a/apps/omg_watcher_rpc/test/omg_watcher_rpc/tracer_test.exs
+++ b/apps/omg_watcher_rpc/test/omg_watcher_rpc/tracer_test.exs
@@ -1,0 +1,98 @@
+# Copyright 2019-2020 OmiseGO Pte Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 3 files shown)


apps/omg_watcher_rpc/lib/tracer.ex
```diff
@@ -18,4 +18,69 @@ defmodule OMG.WatcherRPC.Tracer do
...
+
+  Handles failure to decode the response body gracefully.
+  """
+
+  def add_trace_metadata(conn) do
+    case Jason.decode(conn.resp_body) do
+      {:ok, json_resp_body} -> add_metadata_from_response_body(conn, json_resp_body)
+      {:error, _} -> SpandexPhoenix.default_metadata(conn)
+    end
+  end
```
InoMurko, line 85: it seems to me that we could find a way that would avoid this serialization and deserialization.

if you look at `OMG.WatcherRPC.Web.api_response/3` this gets called when success == true
if you look at `OMG.WatcherRPC.Web.Controller.Fallback.call/2` this gets called when success == false

and you could attach metadata at this point in time with assigns https://elixircasts.io/exploring-phoenix-assigns

try it, let me know :)))

apps/omg_watcher_rpc/lib/tracer.ex
```diff
@@ -18,4 +18,69 @@ defmodule OMG.WatcherRPC.Tracer do
...
+  private(:keyword) - Default: []
+  resource([:atom, :string])
+  sql_query(:keyword)
+  start(:integer)
+  tags(:keyword) - Default: []
+  trace_id(:any)
+  type(:atom)
+  """
+  defp add_metadata_from_response_body(conn, json_resp_body) do
+    service = String.to_atom(json_resp_body["service_name"])
```
InoMurko, line 47: this seems somewhat dangerous that could peraps blow up the VM unintentionally. try to choose the atom, for example:
```
case (json_resp_body["service_name"] do
"watcher"->:watcher
"watcher_info"-> :watcher_info
end
```
the reason being, atoms are not garbage collected and turning strings to atoms is something we should avoid.

apps/omg_watcher_rpc/lib/tracer.ex
```diff
@@ -18,4 +18,69 @@ defmodule OMG.WatcherRPC.Tracer do
...
+    error = if !json_resp_body["success"], do: Keyword.new([{:error, true}])
+    error_type = if !!error and json_resp_body["data"], do: json_resp_body["data"]["code"]
+    error_msg = if !!error and json_resp_body["data"], do: json_resp_body["data"]["description"]
```
InoMurko, line 50: if clauses are to be used only with functions with questionmarks. in elixir, we prefer patter matching with `case` clauses

apps/omg_watcher_rpc/lib/tracer.ex
```diff
@@ -18,4 +18,69 @@ defmodule OMG.WatcherRPC.Tracer do
...
+
+    tags = [
+      {:version, json_resp_body["version"]}
+    ]
+
+    _ = if error_type, do: tags ++ {String.to_atom("error.type"), error_type}
+    _ = if error_msg, do: tags ++ {String.to_atom("error.msg"), error_msg}
```
InoMurko, line 57: as said, `String.to_atom` is to be avoided whenever possible.

apps/omg_watcher_rpc/lib/tracer.ex
```diff
@@ -18,4 +18,69 @@ defmodule OMG.WatcherRPC.Tracer do
...
+      conn
+      |> SpandexPhoenix.default_metadata()
+      |> Keyword.put(:service, service)
+      |> Keyword.put(:tags, tags)
+
+    if error,
+      do:
+        trace_data
+        |> Keyword.put(:error, error),
+      else: trace_data
```
InoMurko, line 69: this is a short if notation for a multiline statement... anyhow, use case clauses as they're prefered.

apps/omg_watcher_rpc/lib/tracer.ex
```diff
@@ -18,4 +18,69 @@ defmodule OMG.WatcherRPC.Tracer do
...
+
+  @doc """
+  Adds metadata from the response body to the Spandex span.
+
+  The conn is inspected just before sending back the API response.
+
+  Handles failure to decode the response body gracefully.
+  """
+
+  def add_trace_metadata(conn) do
```
InoMurko, line 80: public functions on top, private at the bottom

apps/omg_watcher_rpc/test/omg_watcher_rpc/tracer_test.exs
```diff
@@ -0,0 +1,98 @@
...
+      "data": [],
+      "service_name": "watcher",
+      "success": true,
+      "version": "1.0.4+33c3300"
+    }
+    """
+
+    conn =
+      Phoenix.ConnTest.build_conn(:get, "/alerts.get")
+      |> Plug.Conn.resp(200, resp_body)
```
InoMurko, line 31: piping should be used for multiple statements piped.
yes:
```
:get
|> Phoenix.ConnTest.build_conn("/alerts.get")
|> Plug.Conn.resp(200, resp_body)
```

InoMurko (changes requested):

apps/omg_watcher_rpc/lib/tracer.ex
macaptain: I agree that it would be great to avoid decoding the response body back to JSON. I'll take a look at using Phoenix assigns instead. Thanks for the suggestion!

macaptain:
Thanks for the review @InoMurko.

I'll close this PR and open a new one. There's too much to fix here in one pass. I'll re-use `customize_metadata`, but take a different approach with `assigns` for adding metadata to the `conn`.

PR #1765, opened 2020-11-09
Collect metrics through datadog for the perf project

ayrat555:
This PR adds the following:

- sending of request metrics and test results to Datadog
- making assertions based on datadog monitor events. Monitor events are emitted on:

1. If the value of the metric is above a threshold. I set thresholds based on average time during my testing
2. If the test failed at least once during the current test run.

I created two monitors for both tests:

1. Deposits test dashboard https://app.datadoghq.com/dashboard/rpx-xu2-b2g/deposits-perf-tests?from_ts=1605343939055&live=true&to_ts=1605430339055&tpl_var_env=perf_circleci 
2. Transaction test dashboard https://app.datadoghq.com/dashboard/7kh-xx4-9qu/transactions-perf-tests

Created monitors:
- Watcher info monitors - https://app.datadoghq.com/monitors/manage?q=watcherinfo
- Childchain monitors - https://app.datadoghq.com/monitors/manage?q=childchain.submit
- test monitors - https://app.datadoghq.com/monitors/manage?q=test%20failed

Notes:
- Monitors and dashboards are bound to env tag set for `Statix` so they can be distinguished by env.
- For circleci,  additional env tag is prepended so a failure in one test run won't cause failures in all test runs


docker-compose.datadog.yml
```diff
@@ -0,0 +1,20 @@
+version: "2.3"
+services:
+  datadog:
+    image: datadog/agent:latest
```
ayrat555, line 4: datadog agent is added to send metrics during circle ci test runs.
example of failure - https://app.circleci.com/pipelines/github/omgnetwork/elixir-omg/8939/workflows/bcebdc24-e205-4f32-b009-64455735beb2/jobs/87535 - I added sleep(100) to simulate it

priv/perf/apps/load_test/lib/test_runner.ex
```diff
@@ -23,35 +23,59 @@ defmodule LoadTest.TestRunner do
...
 
-      _ ->
+      {:error, errors} ->
+        # credo:disable-for-next-line
+        IO.inspect("errors: #{inspect(errors)}")
         System.halt(1)
     end
   end
+
+  defp maybe_add_custom_tag(start_date) do
```
ayrat555, line 66: custom tags are only added during circle ci runs so test runs can be distinguished

InoMurko:
Feedback:

The dasboards are great, but:
- we lack information about what "Average test count" means. Average of what? What is being counter? This needs to be somehow reflected and evident in the dasboard (title, description... whatever is available).

- there's one instance of the dashboard per test type (transaction tests, deposit tests). Correct?
This means, that we're only able to run one instance of the test at the same time? What if we want to run one of these tests towards development and another one towards sandbox?

priv/perf/apps/load_test/lib/test_runner.ex
InoMurko: This goes well with my comment: https://github.com/omgnetwork/elixir-omg/pull/1765#issuecomment-729644723

Are dashboards and monitors able to distinguish by this parameter?

docker-compose.datadog.yml
InoMurko: FYI: datadog is also defined in `docker-compose.dev.yml` 
maybe they could be merged?

priv/perf/apps/load_test/lib/application.ex
```diff
@@ -4,24 +4,32 @@ defmodule LoadTest.Application do
...
+  alias LoadTest.Ethereum.NonceTracker
+  alias LoadTest.Service.Datadog
+  alias LoadTest.Service.Faucet
+
   def start(_type, _args) do
-    pool_size = Application.fetch_env!(:load_test, :pool_size)
-    max_connections = Application.fetch_env!(:load_test, :max_connection)
+    :ok = start_hackney_pool()
+
+    NonceTracker.init()
```
InoMurko, line 15: because of the nonce tracking, you can run only one instance of a particula test with the same private key/ethereum account, right? 
let's make sure we're all aware of that.

priv/perf/apps/load_test/lib/service/datadog/api.ex
```diff
@@ -0,0 +1,166 @@
...
+  @datadog_events_api_path "api/v1/events"
+  @datadog_monitor_resolve_path "monitor/bulk_resolve"
+  @datadog_app_url "https://app.datadoghq.com"
+
+  def assert_metrics(environment, start_datetime, end_datetime) do
+    start_unix = DateTime.to_unix(start_datetime)
+    end_unix = DateTime.to_unix(end_datetime)
+
+    # wait for events to be emitted
+    Process.sleep(30_000)
```
InoMurko, line 31: why do we need to wait? where is 30seconds coming from?

priv/perf/apps/load_test/lib/service/datadog/api.ex
```diff
@@ -0,0 +1,166 @@
...
+        :ok
+
+      {:ok, %{body: body}} ->
+        Logger.error("failed to resolve monitors #{inspect(body)}")
+
+        if retries == 0 do
+          do_resolve_monitors(params, retries - 1)
+        else
+          {:error, body}
+        end
```
InoMurko, line 124: same rules apply, if/else only for functions with questionmarks
```suggestion
        case retries do
        0 ->
          do_resolve_monitors(params, retries - 1)
        _ ->
          {:error, body}
        end
```

priv/perf/apps/load_test/lib/service/datadog/api.ex
```diff
@@ -0,0 +1,166 @@
...
+
+      {:error, error} = other ->
+        if retries == 0 do
```
InoMurko, line 127: same as above

priv/perf/config/config.exs
```diff
@@ -45,4 +46,14 @@ config :logger, :console,
   discard_threshold: 2000,
   metadata: [:module, :function, :request_id, :trace_id, :span_id]
 
+config :statix,
+  host: "localhost",
+  port: 8125,
+  tags: [System.get_env("STATIX_TAG", "env:perf_test")]
```
InoMurko, line 52: maybe it makes sense to yield error if this isn't set, otherwise we might overwrite data in datadog?

InoMurko (commented):
Lots of work here! Kudos! 

Left a few comments I would like to see answers to and will approve the PR once that gets resolved.

priv/perf/apps/load_test/lib/test_runner.ex
ayrat555: yes, tests are distinguished by env tag. dashboards show data for a specific tag. To show a different env, you change the env template parameter:


For circle ci, custom tags are added which contain current timestamp so two circle ci runs won't intersect

InoMurko: Awesome!

InoMurko (approved):

priv/perf/apps/load_test/lib/service/datadog/api.ex
ayrat555: monitor events are emitted based on 1 minute time intervals. If events are fetched right after test execution, events may not be published yet

InoMurko: Would it be better to poll for this data? Maybe 30s is too much, or too little and we would randomly fail.

ayrat555:
@boolafish can you please review this pr

boolafish (approved):
Amazing PR 💯💯💯💯

Probably to be different PR (or even not need a PR)
But I strongly recommend to start documenting:

1. How to setup the perf tests here.  What kind of pre-setup is required?
  - Do we need to first manually add those monitors in datadog?
  - what service dependency is required to live in the cluster? (eg. datadog agent?)
1. Update the existing doc of how to run perf test

---

one side note: I am not sure how much do we need to chain the tests now (I know it was previously brought up for really long running tests like for days). One note/idea after looking at the PR is that we can actually separate traffic trigger and result checking into 2 spinnaker stages. This would make it flexible on having multiple previous stages that triggers the traffic and the one single last stage that check the metrics. 🚀

ayrat555:
@boolafish 
> one side note: I am not sure how much do we need to chain the tests now (I know it was previously brought up for really long running tests like for days). One note/idea after looking at the PR is that we can actually separate traffic trigger and result checking into 2 spinnaker stages. This would make it flexible on having multiple previous stages that triggers the traffic and the one single last stage that check the metrics

I added the ability to run tests and make assertions separately.

From docs:
```
  Running tests without assertions.
  You can run tests without assertions by passing `false` as the last parameter:

    STATIX_TAG="env:perf_circleci" mix run -e "LoadTest.TestRunner.run()" -- "transactions" 1 80 false
  
  To just check if there are any events in the given period of time run passing start time and end time:

  STATIX_TAG="env:perf_circleci" mix run -e "LoadTest.TestRunner.run()" -- "make_assertions" 1605775276 1605785276

```

PR #1768, opened 2020-11-12
Chore: fix feefeed container healthcheck

ayrat555:
Check if feefeed fetched fees instead of checking if it's up


InoMurko:
Perhaps it would also be valuable that we add `docker-compose logs -f`  where feefeed is running. This would give us more visibility. 
```
- run:
          name: Log daemon services
          command: |
            cd priv/cabbage
            docker-compose logs -f
```
right after
```
- run:
          name: Start daemon services
          command: |
            cd priv/cabbage
            make start_daemon_services-2 || (START_RESULT=$?; docker-compose logs; exit $START_RESULT;)
```

InoMurko (approved):
good stuff, I like that you're bing pro-active!

macaptain (approved):

PR #1769, opened 2020-11-13
Watcher_Info events race cause spent output presented unspent

### LINKED ISSUE #1760: Utxo not found when attempting transaction with some utxos
Some utxos returned from `account.get_utxos` cause `utxo_not_found` error when creating transactions with them. observed on mainnet.

`account.get_utxos` returns these 2 utxos that i want to merge:
```
amount: 10000000000000000
blknum: 83000
creating_txhash: "0x447ec5ec92a647b1d6947c2faf9a575486e5811e4c0576f3d4015722124573b2"
currency: "0xd26114cd6ee289accf82350c8d8487fedb8a0c07"
inserted_at: "2020-06-18T09:40:06Z"
oindex: 1
otype: 1
owner: "0x0339b1187101bce651ee6215c9c9b34b560873c5"
spending_txhash: null
txindex: 0
updated_at: "2020-06-18T09:40:06Z"
utxo_pos: 83000000000001
```

```
amount: 3830922229089327600
blknum: 3692000
creating_txhash: "0xb0bcd3164ef339c63fadbf5e0fbd457470ba8c02ef980e2d5f32617b569cbcaf"
currency: "0xd26114cd6ee289accf82350c8d8487fedb8a0c07"
inserted_at: "2020-10-29T08:06:19Z"
oindex: 0
otype: 1
owner: "0x0339b1187101bce651ee6215c9c9b34b560873c5"
spending_txhash: null
txindex: 0
updated_at: "2020-10-29T08:06:19Z"
utxo_pos: 3692000000000000
```

i create a merge transaction using these 2 utxos and form this typed data:
```
txType:1
input0:
  blknum:83000
  txindex:0
  oindex:1
input1:
  blknum:3692000
  txindex:0
  oindex:0
input2:
  blknum:0
  txindex:0
  oindex:0
input3:
  blknum:0
  txindex:0
  oindex:0
output0:
  outputType:1
  outputGuard:0x0339B1187101bCE651EE6215C9c9b34B560873c5
  currency:0xd26114cd6ee289accf82350c8d8487fedb8a0c07
  amount:3840922229089327700
output1:
  outputType:0
  outputGuard:0x0000000000000000000000000000000000000000
  currency:0x0000000000000000000000000000000000000000
  amount:0
output2:
  outputType:0
  outputGuard:0x0000000000000000000000000000000000000000
  currency:0x0000000000000000000000000000000000000000
  amount: 0
output3:
  outputType: 0
  outputGuard: 0x0000000000000000000000000000000000000000
  currency: 0x0000000000000000000000000000000000000000
  amount: 0
  txData: 0
metadata: 0x0000000000000000000000000000000000000000004d65726765205554584f73
```

I sign and submit the transaction calling `transaction.submit` and get back:
```
code: "submit:utxo_not_found"
description: null
object: "error
```

I suspect the first utxo is the erroneous one. for some reason I cannot spend this utxo.

pnowosie:
Closes #1760
Related to https://github.com/omgnetwork/private-issues/issues/59 

## Overview

As described in the issue we will drop WatcherInfo's event cosumption based on internal `OMG.Bus`
and couple it with Watcher's event listeners which impose event processing order.

We're allowing to couple WatcherInfo to SecurityWatcher (as Security Watcher's behavior stays unchanged) because it quite quick and solid remedy. Also keep in mind that [Watcher Info](https://omisego.atlassian.net/wiki/spaces/PM/pages/622198952/DACI+-+RFC+Watcher+Data+Service+Offerings) is planned for decommissioning.

## Changes

- Removes `ExitConsumer` module
- Removes `DepositConsumer` module
- Removes `BlockApplicationConsumer` module
- Adds direct call to `insert_exits` from Watcher's ExitProcessor when WatcherInfo's code is loaded
- Adds direct call to `insert_deposit` when WatcherInfo's code is loaded
- Directly insert pending block in WatcherInfo from `BlockGetter` when WatcherInfo's code is loaded

Regarding 'keeping two databases in sync' as pointed out in this [comment](https://github.com/omgnetwork/elixir-omg/issues/1760#issuecomment-722574344). __Fortunately we don't need this.__

It is implemented as follows:
- First WatcherInfo code is executed (on deposit, new block or any exit)
- It either: 
  - succeed and proceed with Watcher module's state modification
  - dies causing Watcher's process to restart
- RocksDB data is persisted.

So we only persist Watcher's state in RocksDB if WatcherInfo's related code succeed. And also events are immutable.
So what is sufficient is to ensure that WatcherInfo operation called from Watcher are idempotent (do not break when repeated) in the case that process died before write its state into RocksDB. 

## Testing

- [Writes of deposits and exits are idempotent](https://github.com/omgnetwork/elixir-omg/blob/972c70db1b9b1dcfd521d51dfcbde7631eb0a2ff/apps/omg_watcher_info/test/omg_watcher_info/db/eth_event_test.exs#L276)
- [Writes of pending block are indempotent](https://github.com/omgnetwork/elixir-omg/blob/972c70db1b9b1dcfd521d51dfcbde7631eb0a2ff/apps/omg_watcher_info/test/omg_watcher_info/block_applicator_test.exs#L54)


InoMurko:
I think you took a wrong turn on doing this.

All events that are produced from the sync supervisors children, ethereum event listeners do not obey any kind of order of events being pushed to the bus, and that's where you should tackle the problem. The callbacks for applying the events need to *block* events being consumed. Look at my comment: https://github.com/omgnetwork/elixir-omg/issues/1760#issuecomment-722571378
```
get_events_callback: fn - 
case Code.ensure_loaded?(OMG.WatcherInfo.ExitConsumer) do
false ->&Watcher.ExitProcessor.new_exits/1
true ->
{{OMG.WatcherInfo.ExitConsumer, :exit_started}, &Watcher.ExitProcessor.new_exits/1}
end
```

pnowosie:
> I think you took a wrong turn on doing this.

So EthereumEventListener accepts [two callbacks](https://github.com/omgnetwork/elixir-omg/blob/972c70db1b9b1dcfd521d51dfcbde7631eb0a2ff/apps/omg_watcher/lib/omg_watcher/sync_supervisor.ex#L133-L134) `get_events_callback` and `process_events_callback`.
I've chosen the later one because on processing exits I have access to the ExitProcessor's data. Basically I send the same data that was transferred by the Bus but in direct manner (as you described in the comment).
Processing order is maintained by EventListener which state is updated after both callbacks succeds.

InoMurko:
> > I think you took a wrong turn on doing this.
> 
> So EthereumEventListener accepts [two callbacks](https://github.com/omgnetwork/elixir-omg/blob/972c70db1b9b1dcfd521d51dfcbde7631eb0a2ff/apps/omg_watcher/lib/omg_watcher/sync_supervisor.ex#L133-L134) `get_events_callback` and `process_events_callback`.
> I've chosen the later one because on processing exits I have access to the ExitProcessor's data. Basically I send the same data that was transferred by the Bus but in direct manner (as you described in the comment).
> Processing order is maintained by EventListener which state is updated after both callbacks succeds.

Why is the only change apps/omg_watcher/lib/omg_watcher/sync_supervisor.ex for deposits? **What about the exits?**
https://github.com/omgnetwork/elixir-omg/pull/1769/files#diff-883a96009d28dbd262906c31ba5307923a9c83219c34b9f53cbee34cb399fc69R338-R339
I see you did not go the callback route on the exits, which meant this isn't consistent as much as we would like?



I now see `.../lib/omg_watcher_info/deposit_consumer.ex → .../omg_watcher/lib/omg_watcher/depositor.ex ` you've actually moved the file. Would have been easier to review if you deleted the old one.

pnowosie:
> Would have been easier to review if you deleted the old one.

Oh, it's just git is fooling us - I deleted `DepositConsumer` in WatcherInfo and made a small module that wraps the callback  for deposit... [it was a direct call to `State`](https://github.com/omgnetwork/elixir-omg/blob/8c01e81ee651bba64b7a54becfd4d93b155f5e06/apps/omg_watcher/lib/omg_watcher/sync_supervisor.ex#L134) which we definitely want to mess around.

> which meant this isn't consistent as much as we would like?

I regret this - but it was the easiest way to not change the Watcher ExitProcessor's logic ... for exits it basically replace events emitting with direct call to WatcherInfo...

apps/omg_watcher/lib/omg_watcher/block_getter.ex
```diff
@@ -136,10 +136,8 @@ defmodule OMG.Watcher.BlockGetter do
 
     case Core.validate_executions(tx_exec_results, block_application, state) do
       {:ok, state} ->
-        :ok =
-          {:child_chain, "block.get"}
-          |> OMG.Bus.Event.new(:block_received, block_application)
-          |> OMG.Bus.direct_local_broadcast()
+        if Code.ensure_loaded?(OMG.WatcherInfo.DB.Block),
+          do: Kernel.apply(OMG.WatcherInfo.BlockApplicator, :insert_block!, [block_application])
```
InoMurko, line 140: I need you to rethink this.

Above this line theres:
```
tx_exec_results = for(tx <- block_application.transactions, do: OMG.State.exec(tx, :ignore_fees))
```

Which applies the block to the Watcher ledger. After this, you're sending the block to postgres.
What happens if PG can't swallow this block?


In `OMG.Watcher.Depositor` you're first applying the deposits to PG and then on RocksDB. You don't have a consistent approach in solving this ticket. You need to consider failure! How are you going to keep two databases in sync?

pnowosie: Here the approach is similar to the mentioned deposit. 
The code you mentioned does the stateful validation of the block's transactions but does not persist the legder and block's data into RocksDB yet. It is done in the [next step](https://github.com/omgnetwork/elixir-omg/blob/972c70db1b9b1dcfd521d51dfcbde7631eb0a2ff/apps/omg_watcher/lib/omg_watcher/block_getter.ex#L170)

So let consider the scenarios here:

1) __PG write fails__
The in-memory `State` already contains new block transactions applied but RocksDB is one block behind. ([RocksDB update](https://github.com/omgnetwork/elixir-omg/blob/972c70db1b9b1dcfd521d51dfcbde7631eb0a2ff/apps/omg_watcher/lib/omg_watcher/block_getter.ex#L170) is in the following `handle_continue`)

Failure of `WatcherInfo` call will result of  `BlockGetter` process death which also brings down the `State` process (see: [OMG.Watcher.BlockGetter.Supervisor](https://github.com/omgnetwork/elixir-omg/blob/972c70db1b9b1dcfd521d51dfcbde7631eb0a2ff/apps/omg_watcher/lib/omg_watcher/block_getter/supervisor.ex)).
Both processes will be restarted by the :point_up: supervisor - so `State` gets RocksDB version of the ledger.


2) __PG write succeeds and RocksDB update fails__
`BlockGetter` and `State` are being restarted and gets the state from the previous block. PG already has the current block's data. Processing the block again WatcherInfo code gets executed again but fortunately [its idempotent]() - so PG will stay unchanged.

InoMurko: Okay, it makes sense!
```
Processing the block again WatcherInfo code gets executed again but fortunately its idempotent - so PG will stay unchanged.
```

are you sure about this tho?

pnowosie: Fortunately it's tested already (I added the test for block insert indempothency)
- [Exits & deposits writes](https://github.com/omgnetwork/elixir-omg/blob/972c70db1b9b1dcfd521d51dfcbde7631eb0a2ff/apps/omg_watcher_info/test/omg_watcher_info/db/eth_event_test.exs#L276)
- [Block applicator test](https://github.com/omgnetwork/elixir-omg/blob/972c70db1b9b1dcfd521d51dfcbde7631eb0a2ff/apps/omg_watcher_info/test/omg_watcher_info/block_applicator_test.exs#L54)

apps/omg_watcher_info/lib/omg_watcher_info/block_applicator.ex
```diff
@@ -12,39 +12,29 @@
...
+  def insert_block!(block) do
+    block
+    |> to_pending_block()
+    |> DB.PendingBlock.insert()
+    |> case do
+      {:ok, _} ->
+        :ok
+
+      {:error, changeset} ->
+        [{:blknum, {_msg, [constraint: :unique, constraint_name: _name]}}] = changeset.errors()
```
InoMurko, line 35: can you explain why this error is ok? leave a comment

apps/omg_watcher_info/lib/omg_watcher_info/db/eth_event.ex
```diff
@@ -237,23 +240,25 @@ defmodule OMG.WatcherInfo.DB.EthEvent do
           event
       end
 
-    tx_output = resolve_tx_output(output_pointer)
+    case resolve_tx_output(output_pointer) do
+      %DB.TxOutput{} = tx_output ->
+        :ok = insert_exit_if_not_exist(ethevent, tx_output)
 
-    insert_exit_if_not_exist(ethevent, tx_output)
+      # it dies when output is expected to exist (most of the cases, see: `elixir-omg/issues/1760`)
```
InoMurko, line 247: what is `most of the cases`?

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -418,10 +425,11 @@ defmodule OMG.Watcher.ExitProcessor do
...
 
-    :ok =
-      exits
-      |> Tools.to_bus_events_data()
-      |> publish_internal_bus_events("InFlightTxOutputPiggybacked")
+    events = Tools.to_bus_events_data(exits)
+    :ok = publish_internal_bus_events(events, "InFlightTxOutputPiggybacked")
+
+    if Code.ensure_loaded?(OMG.WatcherInfo.DB.EthEvent),
+      do: Kernel.apply(OMG.WatcherInfo.DB.EthEvent, :insert_exits!, [events, :in_flight_exit, true])
```
InoMurko, line 432: all are calls towards `:insert_exits!` idempotent?

apps/omg_watcher_info/lib/omg_watcher_info/db/eth_event.ex
pnowosie: I've added better description with a link to the original issue.

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pnowosie: Yes, exit_processor also persist its state in rocksDB similarly to event listeners.

pgebal: Why do we insert in-flights on piggyback?

If this functions does something more than inserting exits let's call it `insert_exit_events!`.

It also applies to line 530.

apps/omg_watcher_info/lib/omg_watcher_info/block_applicator.ex
```diff
@@ -12,39 +12,29 @@
...
-      |> DB.PendingBlock.insert()
-
-    {:noreply, state}
+  @doc """
+  Inserts a block to pending blocks, does not break when block already exists.
+  """
+  @spec insert_block!(OMG.Watcher.BlockGetter.BlockApplication.t()) :: :ok
+  def insert_block!(block) do
+    block
+    |> to_pending_block()
```
pgebal, line 28: No need to pattern match in `to_pending_block` header.

apps/omg_watcher_info/test/omg_watcher_info/block_applicator_test.exs
```diff
@@ -0,0 +1,67 @@
...
+          blkhash: block_application.hash,
+          timestamp: block_application.timestamp,
+          transactions: block_application.transactions
+        })
+
+      assert [%DB.PendingBlock{blknum: 1_000, data: ^expected_data}] = DB.Repo.all(DB.PendingBlock)
+    end
+
+    @tag fixtures: [:phoenix_ecto_sandbox]
+    test "inserts the the same block does not break", %{block_application: block_application} do
```
pgebal, line 54:
```suggestion
    test "inserts the the same block does not break", %{block_application: block_application} do
```
```suggestion
    test "inserting bock is idempotent", %{block_application: block_application} do
```

pgebal, line 54: Let's check if after the two inserts there is only one block in the db.

apps/omg_watcher_info/test/omg_watcher_info/db/eth_event_test.exs
```diff
@@ -418,6 +419,119 @@ defmodule OMG.WatcherInfo.DB.EthEventTest do
...
+    assert %DB.TxOutput{ethevents: events} = DB.TxOutput.get_by_position(Utxo.position(1, 0, 0))
+
+    assert [
+             %DB.EthEvent{event_type: :deposit},
+             %DB.EthEvent{event_type: :in_flight_exit}
+           ] = Enum.sort(events, &(&1.eth_height < &2.eth_height))
+  end
+
+  @tag fixtures: [:initial_blocks]
+  test "only one exit type can be asssotiated to output which is retrievable by output_id",
```
pgebal, line 496:
```suggestion
  test "only one exit type can be asssotiated to output which is retrievable by output_id",
```
```suggestion
  test "only one exit type can be associated to output which is retrievable by output_id",
```

pgebal, line 496: why only one?

pgebal (changes requested):

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
pnowosie: Important thing here is to spend tx's output that was piggybacked during IFE.
This is done by associate `in_flight_exit` record of EthEvent table with it. It's not piggyback event recorded here.

apps/omg_watcher_info/test/omg_watcher_info/db/eth_event_test.exs
pnowosie: Because one exit event is enough to spend UTXO which is WI mostly awared of (and also only one effectively exits).

apps/omg_watcher_info/lib/omg_watcher_info/block_applicator.ex
pnowosie: Done

pgebal (approved):

apps/omg_watcher_info/lib/omg_watcher_info/supervisor.ex
```diff
@@ -63,26 +63,8 @@ defmodule OMG.WatcherInfo.Supervisor do
...
-      ),
-      Supervisor.child_spec(
-        {OMG.WatcherInfo.ExitConsumer, [topic: {:watcher, "InFlightTxOutputPiggybacked"}, event_type: :in_flight_exit]},
-        id: :ife_exit_output_piggybacked_consumer
-      ),
-      Supervisor.child_spec(
-        {OMG.WatcherInfo.ExitConsumer, [topic: {:watcher, "InFlightExitOutputWithdrawn"}, event_type: :in_flight_exit]},
-        id: :ife_exit_processed_consumer
-      )
+      {OMG.WatcherInfo.PendingBlockQueueLengthChecker, [check_interval: Configuration.block_queue_check_interval()]}
```
InoMurko, line 67: is there a need for this anymore? I don't think there is right? you're blocking block getter until you process the block?

apps/omg_watcher_info/test/omg_watcher_info/block_applicator_test.exs
```diff
@@ -0,0 +1,71 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule OMG.WatcherInfo.BlockApplicatorTest do
+  use ExUnitFixtures
+  use ExUnit.Case, async: true
```
InoMurko, line 17:
```suggestion
  use ExUnit.Case, async: false
```
needs to be false when using ecto. cc @pgebal

apps/omg_watcher_info/lib/omg_watcher_info/supervisor.ex
pnowosie: It _is_ needed. I haven't changed block consumption process, as it might fail for many reason so I want this goes with pending block. But it indeed blocks until block will be fully applied. (Besides it's easier/quicker to ensure idempotency of writes with pending block)

apps/omg_watcher_info/test/omg_watcher_info/block_applicator_test.exs
pnowosie: https://github.com/omgnetwork/elixir-omg/pull/1769/commits/028b4bea38c5ae0d8258bf1e850173cea6ebc7c5

InoMurko (commented):
Pls ping me when I can look at this!

apps/omg_watcher_info/lib/omg_watcher_info/db/eth_event.ex
```diff
@@ -209,16 +210,18 @@ defmodule OMG.WatcherInfo.DB.EthEvent do
...
+        ) :: :ok | :noop
   defp insert_exit!(
          %{
            root_chain_txhash: root_chain_txhash,
            log_index: log_index,
            eth_height: eth_height,
            output_pointer: output_pointer
          },
-         event_type
+         event_type,
```
InoMurko, line 223: maybe at this point it also make sense to refactor the function header by not pattern match in it

apps/omg_watcher_info/lib/omg_watcher_info/db/eth_event.ex
```diff
@@ -237,23 +240,26 @@ defmodule OMG.WatcherInfo.DB.EthEvent do
           event
       end
 
-    tx_output = resolve_tx_output(output_pointer)
+    case resolve_tx_output(output_pointer) do
+      %DB.TxOutput{} = tx_output ->
+        :ok = insert_exit_if_not_exist(ethevent, tx_output)
```
InoMurko, line 245: why is this a case is it's supposed to exit?
wouldn't this be better?
```
tx_output = DB.TxOutput{}  = resolve_tx_output(output_pointer) 
:ok = insert_exit_if_not_exist(ethevent, tx_output)
```

apps/omg_watcher_info/lib/omg_watcher_info/db/eth_event.ex
```diff
@@ -267,11 +273,9 @@ defmodule OMG.WatcherInfo.DB.EthEvent do
 
     tx_output
     |> txoutput_changeset(%{child_chain_utxohash: generate_child_chain_utxohash(decoded_utxo_position)}, ethevent)
-    |> DB.Repo.update()
+    |> DB.Repo.update!()
```
InoMurko, line 276: nice!

apps/omg_watcher_info/lib/omg_watcher_info/supervisor.ex
InoMurko: the problem is that block consumption process, as it's currently, is async. the first insert is in a queue. how are you then blocking block getter to continue?

apps/omg_watcher_info/lib/omg_watcher_info/db/eth_event.ex
pnowosie: Unfortunately for one branch (IFE piggybacked outputs) we need to allow that outputs might not be present in the database. So the other `case` match checks for it.

InoMurko: okay, so let's be very specific about that. let the code branch out if the exit involves `IFE piggybacked outputs`. this way the code will be more assertive about what is going on.

apps/omg_watcher_info/lib/omg_watcher_info/db/eth_event.ex
pnowosie: I think extracting the event values is justified as we need all of them few lines below in `%__MODULE__{...}`

InoMurko: yeah, but we're trying to get away from the code style where pattern matching is being done in function headers

```
defp insert_exit!(event, event_type, ensure_output) do
%{root_chain_txhash: root_chain_txhash, log_index: log_index, eth_height: eth_height, output_pointer: output_pointer} = event
root_chain_txhash_event = generate_root_chain_txhash_event(root_chain_txhash, log_index)
...
```

the module becomes much more readable this way.

InoMurko:
Have you tried perhaps to build watcher and watcher info and point it to production and sync mainnet? that would be good to see if there are any problems.

pnowosie:
> Have you tried perhaps to build watcher and watcher info and point it to production and sync mainnet? that would be good to see if there are any problems.

That would be great exercise!

apps/omg_watcher_info/lib/omg_watcher_info/db/eth_event.ex
pnowosie: Considering formatter "help" I have a mixed feeling about readability... but at least it's closer to the convention

apps/omg_watcher_info/lib/omg_watcher_info/db/eth_event.ex
pnowosie: https://github.com/omgnetwork/elixir-omg/pull/1769/commits/78f5ba4e99fd444b2d38a847c413a0afb3629db5

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -444,10 +451,16 @@ defmodule OMG.Watcher.ExitProcessor do
...
-      |> publish_internal_bus_events("InFlightTxOutputPiggybacked")
+    events = Tools.to_bus_events_data(exits)
+    :ok = publish_internal_bus_events(events, "InFlightTxOutputPiggybacked")
+
+    if Code.ensure_loaded?(OMG.WatcherInfo.DB.EthEvent),
+      do:
+        Kernel.apply(
+          OMG.WatcherInfo.DB.EthEvent,
+          :insert_exits!,
+          [events, :in_flight_exit, "InFlightTxOutputPiggybacked"]
```
InoMurko, line 462: these need to be atoms if you're using them as identifiers

InoMurko, line 462:
```suggestion
          [events, :in_flight_exit, :InFlightTxOutputPiggybacked]
```

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -501,10 +514,16 @@ defmodule OMG.Watcher.ExitProcessor do
 
     {:ok, state3, db_updates} = Core.finalize_in_flight_exits(state2, finalizations, invalidities)
 
-    :ok =
-      events_with_utxos
-      |> Tools.to_bus_events_data()
-      |> publish_internal_bus_events("InFlightExitOutputWithdrawn")
+    events = Tools.to_bus_events_data(events_with_utxos)
+    :ok = publish_internal_bus_events(events, "InFlightExitOutputWithdrawn")
```
InoMurko, line 518:
```suggestion
    :ok = publish_internal_bus_events(events, :InFlightExitOutputWithdrawn)
```

apps/omg_watcher/lib/omg_watcher/exit_processor.ex
```diff
@@ -501,10 +514,16 @@ defmodule OMG.Watcher.ExitProcessor do
...
+
+    if Code.ensure_loaded?(OMG.WatcherInfo.DB.EthEvent),
+      do:
+        Kernel.apply(
+          OMG.WatcherInfo.DB.EthEvent,
+          :insert_exits!,
+          [events, :in_flight_exit, "InFlightExitOutputWithdrawn"]
```
InoMurko, line 525:
```suggestion
          [events, :in_flight_exit, :InFlightExitOutputWithdrawn]
```

InoMurko (approved):
turn those string binary identifiers to atoms and you should be good

how did syncing production go?

apps/omg_watcher/lib/omg_watcher/depositor.ex
```diff
@@ -12,21 +12,21 @@
...
-  @spec get_next_pending_block() :: nil | %PendingBlock{}
-  def get_next_pending_block() do
-    PendingBlock.get_next_to_process()
-  end
+  @doc """
+  Processes a deposit event, introducing a UTXO into the ledger.
+  Note: it has to return whatever `State.deposits` returns to process deposits by the Watcher.
+  """
+  @spec new_deposits([OMG.State.Core.deposit()]) :: {:ok, list(State.Core.db_update())}
+  def new_deposits(deposits) do
```
InoMurko, line 26: is there a thing as an *old* deposit?

apps/omg_watcher/lib/omg_watcher/sync_supervisor.ex
```diff
@@ -131,7 +131,7 @@ defmodule OMG.Watcher.SyncSupervisor do
         service_name: :depositor,
         synced_height_update_key: :last_depositor_eth_height,
         get_events_callback: &EthereumEventAggregator.deposit_created/2,
-        process_events_callback: &OMG.State.deposit/1
+        process_events_callback: &Watcher.Depositor.new_deposits/1
```
InoMurko, line 134: I don't see a particular reason why this `Watcher.Depositor` a new module, where all other events are being handled in place (where they were before).

we should put 
```
if Code.ensure_loaded?(OMG.WatcherInfo.DB.EthEvent),
      do: Kernel.apply(OMG.WatcherInfo.DB.EthEvent, :insert_deposits!, [deposits])
```
into `OMG.State.deposit/1` for consistency with everything else

InoMurko (approved):

PR #1774, opened 2020-11-19
retry datadog errors

ayrat555:
in https://github.com/omgnetwork/elixir-omg/commit/dcc2d3d05383531ececef79020c266d112fb1438 datadog metrics were introduced.

Datadog api is timing out during ci runs sometimes

https://app.circleci.com/pipelines/github/omgnetwork/elixir-omg/8969/workflows/bdf890f7-2cd8-435e-81d7-8c4519e153d4/jobs/88002

### DIFF
--- a/priv/perf/apps/load_test/lib/service/datadog/api.ex
+++ b/priv/perf/apps/load_test/lib/service/datadog/api.ex
@@ -36,17 +36,15 @@ defmodule LoadTest.Service.Datadog.API do
     assert_metrics(environment, start_unix, end_unix)
   end
 
-  def do_assert_metrics(environment, start_unix, end_unix, poll_count \\ 60) do
+  defp do_assert_metrics(environment, start_unix, end_unix, poll_count \\ 60)
+
+  defp do_assert_metrics(_environment, _start_unix, _end_unix, 0), do: :ok
+
+  defp do_assert_metrics(environment, start_unix, end_unix, poll_count) do
     case fetch_events(start_unix, end_unix, environment) do
       {:ok, []} ->
-        case poll_count do
-          0 ->
-            :ok
-
-          _ ->
-            Process.sleep(1_000)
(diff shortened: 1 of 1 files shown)


priv/perf/apps/load_test/lib/service/datadog/api.ex
```diff
@@ -80,10 +80,16 @@ defmodule LoadTest.Service.Datadog.API do
...
+          0 -> {:error, body}
+          _ -> fetch_events(start_time, end_time, environment, retries - 1)
+        end
 
       {:error, error} ->
-        {:error, error}
+        case retries do
+          0 -> {:error, error}
+          _ -> fetch_events(start_time, end_time, environment, retries - 1)
+        end
```
InoMurko, line 92: why not provide a header function?
```
def fetch_events(start_time, end_time, environment, 0) do
##return error
```

just extract the query and you won't need to have a case inside a case

InoMurko (changes requested):

InoMurko (approved):

achiurizo (approved):

priv/perf/apps/load_test/lib/service/datadog/api.ex
```diff
@@ -131,26 +145,16 @@ defmodule LoadTest.Service.Datadog.API do
...
 
-          _ ->
-            Process.sleep(1_000)
-            do_resolve_monitors(params, retries - 1)
-        end
+        do_resolve_monitors(params, retries - 1)
 
-      {:error, error} = other ->
+      {:error, error} ->
         Logger.error("failed to resolve monitors #{inspect(error)}")
```
boolafish, line 153: nitpick: I now is not part of this PR change. But might consider using warn level and add some words to indicate it is going to retry as it is still fighting to get out of the error.

boolafish (approved):

PR #1781, opened 2020-11-22
extract geth command, remove childchain only code paths

InoMurko:
Cleanup of Watcher code by removing some of  the childchain only code and tests + simplifying PG DB configuration.


InoMurko:
@pnowosie I would kindly ask you to look at the tests I deleted/modified. I think they're childchain specific. If they're not and I deleted too much, let me know.

config/releases.exs
```diff
@@ -5,7 +5,21 @@ import Config
 #
 # See https://hexdocs.pm/mix/1.9.0/Mix.Tasks.Release.html#module-runtime-configuration
 
+mandatory = fn
+  env, exception, true, true ->
```
pnowosie, line 9:
```suggestion
  env_var_name, error_msg, true, true ->
```
Also please add a comment what these parameters are.

pnowosie (approved):
:+1:

docker-compose.yml
```diff
@@ -21,34 +21,9 @@ services:
...
-            --keystore=./data/geth/keystore/ \
-            --password /tmp/geth-blank-password \
-            --unlock "0,1" \
-            --rpc --rpcapi personal,web3,eth,net --rpcaddr 0.0.0.0 --rpcvhosts=* --rpcport=8545 \
-            --ws --wsaddr 0.0.0.0 --wsorigins='*' \
-            --mine \
-            --allow-insecure-unlock
+    entrypoint: /bin/sh -c ". data/command"
+    environment:
+      RPC_PORT: 8545
```
ayrat555, line 26: Does it need to be configurable? 8545 is the default value. And I can find places where another port is used

ayrat555 (approved):

docker-compose.yml
InoMurko: not really, I took it from childchain and ... I prefer consistency!

PR #1782, opened 2020-11-23
fix: limit matching for ethereumex responses

souradeep-das:
closes https://github.com/omgnetwork/elixir-omg/issues/1587

### Overview

Handle responses from Ethereumex http client only when geth is running


apps/omg_eth/lib/omg_eth/client.ex
```diff
@@ -35,7 +35,7 @@ defmodule OMG.Eth.Client do
     case client.eth_syncing() do
       {:ok, false} -> :ok
       {:ok, _} -> {:error, :geth_still_syncing}
-      {:error, :econnrefused} -> {:error, :geth_not_listening}
+      {:error, _} -> {:error, :geth_not_listening}
```
InoMurko, line 38: hmmm, how confident are you that you're not hidding what the error "actually" is?

souradeep-das: yeahh, I think singling out geth errors mostly econnrefused from request errors would be useful. I'll change it

InoMurko: a log would to it as well!

souradeep-das: Ino, how about this?

apps/omg_eth/lib/omg_eth/client.ex
```diff
@@ -29,13 +29,16 @@ defmodule OMG.Eth.Client do
...
+  @spec node_ready() :: :ok | Ethereumex.Client.Behaviour.error()
+  @spec node_ready(module()) :: :ok | Ethereumex.Client.Behaviour.error()
   def node_ready(client \\ Ethereumex.HttpClient) do
     case client.eth_syncing() do
       {:ok, false} -> :ok
       {:ok, _} -> {:error, :geth_still_syncing}
       {:error, :econnrefused} -> {:error, :geth_not_listening}
+
+      other ->
+        other
```
InoMurko, line 41: you can remove this, since the case clauses will produce an error. and we want to fail as early as possible!

InoMurko, line 41:
```suggestion
```

souradeep-das: oh this is for matching the other error responses for ex {:error, :closed}

InoMurko: okay, then include that :)

InoMurko, line 41:
```suggestion
      {:error, :closed} = other->
        other
```

souradeep-das: Do we want to strictly match it for {:error, :closed} or should we be open to all the other errors

InoMurko: let's match on **what we know happens occasionally** and **leave** to _fail_ for responses we don't know

InoMurko (approved):
LGTM

apps/omg_eth/lib/omg_eth/client.ex
```diff
@@ -29,13 +29,21 @@ defmodule OMG.Eth.Client do
     end
   end
 
-  @spec node_ready() :: :ok | {:error, :geth_still_syncing | :geth_not_listening}
-  @spec node_ready(module()) :: :ok | {:error, :geth_still_syncing | :geth_not_listening}
+  @spec node_ready() :: :ok | {:error, :geth_still_syncing | :geth_not_listening | :closed}
+  @spec node_ready(module()) :: :ok | {:error, :geth_still_syncing | :geth_not_listening | :closed}
   def node_ready(client \\ Ethereumex.HttpClient) do
     case client.eth_syncing() do
```
ayrat555, line 35: I think hackney can return  additional types of errors
https://github.com/benoitc/hackney/blob/b4a48ed5977c337ce0f761ad59e32ec634ebcbc8/src/hackney_stream.erl#L34
what do you think about logging unknown errors?

```elixir
{:error, _} = unknow_error -> Logger.error(...
```

InoMurko: why would we log them manually? a crash log produces an error log...

ayrat555: this function is expected to crash?

the additional clause will catch an unmatched error

InoMurko: well the question is - what can you do with a response you're not able to handle on the caller? the OTP way is to let it explode and let the process (the caller) restart and try again.

ayrat555: Some errors are handled, some errors are ok to crash

I think unknown errors can also be handled

ayrat555 (approved):

apps/omg_eth/lib/omg_eth/client.ex
souradeep-das: so we going with the some error to crash way then?

InoMurko:
If there are known errors in sentry coming from this, you could add them in the `SentryFilter` module!

InoMurko (approved):

ayrat555 (approved):

apps/omg_status/lib/omg_status/sentry_filter.ex
```diff
@@ -24,6 +24,12 @@ defmodule OMG.Status.SentryFilter do
     true
   end
 
+  # this might happen when the connection to the node is unavailable
+  # comes from OMG.Eth.Client
```
InoMurko, line 28:
```suggestion
```

PR #1783, opened 2020-11-23
Ignoring 406 status code invalid header exception

DanielaIvanova:
:clipboard: #1588 
## Overview

The status code `406` exception which is being caused by receiving invalid request header "Accept" is now ignored by Sentry. This approach does not solve the problem, it only "suppresses" exception throws, only in regards of invalid accept headers set from user side.

## Changes

- Added new `exclude_exception?` clause to Sentry Filter


InoMurko (approved):

ayrat555 (approved):

apps/omg_status/lib/omg_status/sentry_filter.ex
```diff
@@ -20,11 +20,10 @@ defmodule OMG.Status.SentryFilter do
...
-  def exclude_exception?(%MatchError{term: {:error, :nxdomain}}, _) do
-    true
-  end
+  def exclude_exception?(%MatchError{term: {:error, :nxdomain}}, _), do: true
 
-  def exclude_exception?(_, _) do
-    false
-  end
+  # Ignoring 406 status code invalid headers exception
+  def exclude_exception?(%Phoenix.NotAcceptableError{plug_status: 406}, _), do: true
```
ayrat555, line 26: is it possible to add a test?

DanielaIvanova: Currently exception handling is working locally, and should be also working on the server. I guess that this is more likely something is not right with the deployed version. Sure, I will add tests that will show that the given exception is excluded from notifying Sentry.

achiurizo (approved):

apps/omg_status/test/sentry_filter_test.exs
```diff
@@ -0,0 +1,35 @@
...
+              "Could not render \"406.json\" for OMG.WatcherRPC.Web.Views.Error, please define a matching clause for render/2 or define a template at \"lib/omg_watcher_rpc_web/templates/views/error/*\". No templates were compiled for this module."
+      end
+    )
+
+    assert :excluded =
+             Sentry.capture_exception(
+               %Phoenix.NotAcceptableError{plug_status: 406},
+               event_source: :plug,
+               result: :sync
+             )
```
InoMurko, line 33:
```suggestion
    assert Sentry.capture_exception(
               %Phoenix.NotAcceptableError{plug_status: 406},
               event_source: :plug,
               result: :sync
             ) = :excluded 
```
Put the expression being tested on the left-hand side of the operator and the value you're testing against on the right-hand side.
https://omisego.atlassian.net/wiki/spaces/ENG/pages/105414682/Elixir+Style+Guide

InoMurko (approved):

PR #1786, opened 2020-11-25
fix: set service name as a tag in metrics

macaptain:
## Overview

Currently the service name is part of the prefix for metrics, but it
would be far better if it were a tag, so that services could be compared
side by side in Datadog.

Rename the prefix to 'elixir'.

## Changes

- Move the service name from being part of the prefix in the name of a metric to a tag on the metric.

## Testing

```
docker-compose -f docker-compose-watcher.yml -f docker-compose.dev.yml up
﻿make docker-update-watcher
make docker-update-watcher_info
```

Check that metrics appear as omg.phoenix. ... with the service as a tag in Datadog. And this has been tested.

### DIFF
--- a/apps/omg_status/lib/omg_status/metric/telemetry.ex
+++ b/apps/omg_status/lib/omg_status/metric/telemetry.ex
@@ -35,11 +35,12 @@ defmodule OMG.Status.Metric.Telemetry do
         TelemetryMetricsStatsd,
         metrics: metrics(),
         global_tags: [
+          service: service,
           version: version
         ],
         host: dd_host,
         port: dd_port,
-        prefix: "omg.#{service}",
+        prefix: "elixir",
         formatter: :datadog
       }
     ]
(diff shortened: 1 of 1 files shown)


InoMurko (approved):

apps/omg_status/lib/omg_status/metric/telemetry.ex
```diff
@@ -35,11 +35,12 @@ defmodule OMG.Status.Metric.Telemetry do
         TelemetryMetricsStatsd,
         metrics: metrics(),
         global_tags: [
+          service: service,
           version: version
         ],
         host: dd_host,
         port: dd_port,
-        prefix: "omg.#{service}",
+        prefix: "omg",
```
macaptain, line 43: I may change this to "elixir" rather than "omg". I think the important part is that these metrics come from an "elixir" app.

pnowosie (approved):

PR #1790, opened 2020-12-04
chore: make sure V1.0.5 commit is in the master history

boolafish:

## Overview

Need the tagged commit in mater for semantic-release to take over counting version from there.

https://semantic-release.gitbook.io/semantic-release/usage/configuration#existing-version-tags

## Changes

As it has been merged previously (unfortunately squashed) so there should be no change at all aside from commit history merging

## Testing

N/A


boolafish:
merging as it is no change at all

PR #1794, opened 2020-12-07
Master v2  (new events)

ayrat555:
## Overview

This PRs adds new events from contracts v2


InoMurko (commented):
there is now code in the Aggregator that is not used anymore, since all `enrich` events are set to false.
for example:
```
case Keyword.fetch!(event, :enrich) do
        true ->
          {:ok, enriched_data} = rpc.get_call_data(decoded_log.root_chain_txhash)
          enriched_data_decoded = enriched_data |> from_hex |> Abi.decode_function()
          Map.put(decoded_log, :call_data, enriched_data_decoded)

        _ ->
          decoded_log
      end
```

PR #1800, opened 2020-12-10
childchain v2 compatibility with watchers

InoMurko:
Giant piece of work that makes the new childchain work with old watchers.
- updated specs in cabbage
- perf tests
- api compatibility
- ex plasma bump to 030
- cirleci optimization for mix based childchain tests
- new plasma contracts bindings fixes (bonds, ABIs)
- code cleanup (references to childchain code)
- fee feed mock docker image


.circleci/config.yml
```diff
@@ -152,6 +158,19 @@ commands:
       - restore_cache:
           key: v2-mix-specs-cache-{{ .Branch }}-{{ checksum "mix.lock" }}
 
+  run_test_in_docker:
+    description: "Quick test runs"
+    parameters:
+      test_url:
```
ayrat555, line 164: what do you think about calling this parameter `test_command`?

apps/omg/test/omg/state_test.exs
```diff
@@ -27,7 +27,7 @@ defmodule OMG.StateTest do
 
   require Utxo
 
-  @eth OMG.Eth.zero_address()
+  @eth <<0::160>>
```
ayrat555, line 30: why don't you want to use a common type?

apps/omg_conformance/test/omg_conformance/conformance/merkle_proof_test.exs
```diff
@@ -32,12 +32,16 @@ defmodule OMG.Conformance.MerkleProofTest do
   @max_block_size trunc(:math.pow(2, @proof_length))
 
   setup_all do
+    :ok = Application.put_env(:ex_plasma, :exit_id_size, 160)
```
ayrat555, line 35: why these tests don't work with `160` size?

ayrat555 (approved):

apps/omg/test/omg/state_test.exs
InoMurko: I don't think it adds anything but it increases coupledness with modules and decreases readability, I like tests self contained.

apps/omg_conformance/test/omg_conformance/conformance/merkle_proof_test.exs
InoMurko: some actually do, but I didn't want to dive into this in this PR, but in a follow up!

PR #1808, opened 2021-01-19
Merge unspent output bugfix to master v2

ayrat555:
merges https://github.com/omgnetwork/elixir-omg/pull/1769 into master-v2


apps/omg/mix.exs
```diff
@@ -40,7 +40,7 @@ defmodule OMG.MixProject do
 
   defp deps() do
     [
-      {:ex_plasma, git: "https://github.com/omgnetwork/ex_plasma.git", ref: "867fee97dd930b47d22b5b32c4ef1751424c4206"},
+      {:ex_plasma, git: "https://github.com/omgnetwork/ex_plasma.git", ref: "8d51a76a106511b014caf51fdf70814684af9b9d"},
```
InoMurko, line 43: what hash is this?

apps/omg_watcher_info/test/fixtures.exs
```diff
@@ -142,25 +147,21 @@ defmodule OMG.WatcherInfo.Fixtures do
   end
 
   defp prepare_one_block({blknum, recovered_txs}) do
-    mined_block = %{
+    block_application = %{
       transactions: recovered_txs,
-      blknum: blknum,
-      blkhash: "##{blknum}",
+      number: blknum,
+      hash: "##{blknum}",
```
InoMurko, line 153: is this ok?

apps/omg/mix.exs
ayrat555: it's hash of the master-v2. I'm not sure why it's changed. it can not find the old hash

apps/omg_watcher_info/test/fixtures.exs
ayrat555: it seems it always was that way.
by git blaming, I found this pr  - https://github.com/omgnetwork/elixir-omg/pull/1123

.circleci/config.yml
```diff
@@ -632,6 +648,7 @@ jobs:
             [ -d data ] || mkdir data && chmod 777 data
       - docker_login
       - make_docker_images
+      - install_and_setup_gcloud
```
InoMurko, line 651: gcloud is already installed, what problem have you noticed?

apps/omg/mix.exs
InoMurko: okay, we should revisit this but I think this works with `{:ex_plasma, "~> 0.2.0"},` too

apps/omg/test/omg/state/persistence_test.exs
```diff
@@ -226,7 +241,8 @@ defmodule OMG.State.PersistenceTest do
         owner: owner.addr,
         currency: currency,
         amount: amount,
-        blknum: blknum
+        blknum: blknum,
+        eth_height: 1
```
InoMurko, line 245: why do we need this?

InoMurko (approved):
lemme know about the things I commented on on slack, otherwise lgtm

apps/omg/test/omg/state/persistence_test.exs
ayrat555: it's required for `EthEvent.insert_deposit!/1`:
```elixir
    %{
      root_chain_txhash: root_chain_txhash,
      log_index: log_index,
      blknum: blknum,
      owner: owner,
      eth_height: eth_height,
      currency: currency,
      amount: amount
    } = params
```

.circleci/config.yml
ayrat555: It fails with openssl error. I saw in your childchain pr you resolved it by updating docker-compose, I resolved it this way

InoMurko: okay, so lets resolve it "the same way" :)
so that we don't wonder what it is or what the reason for it is.
consistency.

InoMurko:
I think this looks good, but I'll merge this after https://github.com/omgnetwork/elixir-omg/pull/1800.
Because it adds a set of tests that we don't have here yet (cabbage, perf, mix based childchain) and it'll allow us to check if it "really" works.

